Issue #100, opened 2016-01-06
sendMail example request is confusing (perhaps incorrect)


unpluggedk:
The example request seems to indicate that there is already an attachment to the message by having the following within the request body.

   "attachments": [
      {
        "lastModifiedDateTime": "datetime-value",
        "name": "name-value",
        "contentType": "contentType-value",
        "size": 99,
        "isInline": true,
        "id": "id-value"
      }

However, per document, attachment is only possible to an existing message with the following APIs :
POST /me/messages/<id>/attachments
POST /users/<id | userPrincipalName>/messages/<id>/attachments

unpluggedk:
It is also missing 
"@odata.type": "#Microsoft.OutlookServices.FileAttachment",

and doesn't specify what is required and what is optional.

unpluggedk:
just a note with contentBytes,
"lastModifiedDateTime", "contentType", "size", "id" is not needed.

martellaj:
@angelgolfer-ms, you're more familiar with the Outlook API. Can you please take a look at this sample request/response when you get a chance? Thanks!

angelgolfer-ms:
@unpluggedk , thank you for your patience. I am checking with the product team about this issue. I'll post back after verifying.

angelgolfer-ms:
@unpluggedk , sorry this took a while. I verified that sendmail does support including a file attachment. I clarified the support in the related content.

PR #102, opened 2016-01-06
add contentBytes to attachments

unpluggedk:
Addresses issues #100 #101


martellaj:
Thanks, @unpluggedk. Just checking with @VenkatAyyadevara-MSFT that this holds up before merging.

PR #104, opened 2016-01-07
added Authorization Header to the subscription examples

weinandt:


martellaj:
Thanks for the contribution, @weinandt. However, we intentionally do not include authorization headers in the examples because they are required for every call. You can see that the `Authorization` header is required in the section above the sample request.

PR #181, opened 2016-03-24
Docs updates for tasks/plans/buckets (no schema change)

xiangli91:


xiangli91:
cla should not be required since this is an update from MSFT internal (Planner team). Thanks.

PR #268, opened 2016-05-15
Settings update and group settings support

dkershaw10:
Updates for changes to settings and information on settings on the group entity type


dkershaw10:
Based on internal review, the "classification"  property on groups should be changed to say:
+|classification|String|Describes a classification for the group (such as low, medium or high business impact). Valid values for this property are defined by creating a ClassificationList [setting](directorySetting) value, based on the [template definition] (directorySettingTemplate.md).|

PR #313, opened 2016-06-11
Fixing example for deleting group group members

gsacavdm:


rgregg:
@sumurthy Is updating master sufficient, or does @gsacavdm need to update the beta and v1.0 branches instead?

sumurthy:
We need both v1.0 and/or beta (depending on where it is available) updated. Otherwise, master changes are lost when we copy over from beta/v1.0. This should get fixed with changes being made to the workflow.. till then, this is a painful step to follow.

sumurthy:
Merged; updated beta branch with same change

PR #321, opened 2016-06-13
Added a note that each of autoSubscribeNewMembers and allowExternalSe…

angelgolfer-ms:
…nders can be set only by a PATCH to the group, but not the initial POST for the group.


angelgolfer-ms:
My fork contained some broken links that had been checked into the officedev repo earlier. Will clean up and submit PR again. Closing this one.

PR #340, opened 2016-06-17
Fixed URL in example request to be consistent with syntax; completed …

angelgolfer-ms:
…fix for #313 and #336.


sumurthy:
Need to update the beta branch first as otherwise we won't know if there are any issues with the files...

PR #379, opened 2016-07-04
CHANGE: G$; Version 1.0 does not support direct updates of assignedLi…

GregSpyra:
To whom it may concern.

CHANGE: G$; Version **1.0** does not support direct updates of _assignedLicenses_, _assignedPlans_ attributes through User entity PATCH request. Such updates are only possible using _assignLicense_ method. Note that currently **Beta** does not support such request either.
CHANGE: G$; _CompanyName_ attribute cannot be updated neither with version **1.0** nor **Beta**. Given examples make initial evaluation of Graph functionality hard for a new customer

Would suggest giving other attribute updates as a usage example, i.e. _immutableId_ for non-federated identity or simply _surname_.

### DIFF
--- a/api-reference/api/user_update.md
+++ b/api-reference/api/user_update.md
@@ -70,25 +70,10 @@ Content-length: 491
 
 {
   "accountEnabled": true,
-  "assignedLicenses": [
-    {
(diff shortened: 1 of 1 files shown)


GregSpyra:
DELETE: G$; up-to-date

PR #386, opened 2016-07-19
Mention prohibition on anonymous links for folders

eob:
Just got off the phone with MS Support while trying to debug errors I was receiving and was informed that MS recently disabled anonymous link creation for folders because too many users were mistakenly over-sharing content (creating inadvertent security leaks).

Makes sense to include that here.


rgregg:
I'm fixing this via #676 instead of accepting this pull request. I wanted to take the time to explain more of the situation than just saying it wasn't available on folders.

PR #387, opened 2016-07-19
Correct Response Status in Example

optionset:
Was "200 OK", Should Be "201 Created"


marych:
Thanks for the fix @optionset !

PR #394, opened 2016-07-28
Patch 3

MarianaStepp:
Updated properties for /people


sumurthy:
Not a valid request. Too many changes.

PR #479, opened 2016-09-21
updates to user and group and new action

dkershaw10:
Added theme and preferredLanguage properties to group.md
Added refreshTokensValidFromDateTime and showInAddressList properties, and a new action to user.md
Added new file for invalidateAllRefreshTokens action


api-reference/beta/api/user_invalidateallrefreshtokens.md
```diff
@@ -0,0 +1,59 @@
+# user: invalidateAllRefreshTokens
+Invalidates all of the user's refresh tokens issued to applications, by resetting the **refreshTokensValidFromDateTime** user property to the current date-time. Typically, this operation is performed (by the user or an administrator) if the user has a lost or stolen device.  This operation would prevent access to any of the organization's data accessed through applications on the device without the user first being required to sign in again. In fact, this operation would force the user to sign in again for all applications that they have previously consented to, independent of device.
```
dstrockis: Doesn't this thing invalidate more than just refresh tokens? Doesn't it invalidate other credential types as well, like session tokens? It also invalidates access tokens sent to SharePoint I'm pretty sure.

dstrockis: If so, should we call it something more generic, not so specific to protocol terms? Like `signOutEverywhere` or `endAllSessions` or something?

dkershaw10: This went through schema review, but your suggestions are valid.  I think we have an opportunity to change this for /v1.0 if it makes sense.  At schema review this was explained as only for refresh tokens if I remember correctly.  Will check with evo folks.

PR #481, opened 2016-09-21
Get contact: Remove unsupported http request

Codeklopper:
The following request is not supported anymore

``` http
GET /users/<id | userPrincipalName>/contactFolders/<id>/contacts/<id>
```

This request will throw a NavigationNotSupported exception with the message: "Recursive navigation is not allowed after property 'Contacts' according to the entity schema.".

Only the following request works for getting a contact:

``` http
GET /users/<id | userPrincipalName>/contacts/<id>
```

### DIFF
--- a/api-reference/beta/api/contact_get.md
+++ b/api-reference/beta/api/contact_get.md
@@ -11,17 +11,6 @@ A [contact](../resources/contact.md) from user's default [contactFolder](../reso
 GET /me/contacts/<id>
 GET /users/<id | userPrincipalName>/contacts/<id>
 ```
-A [contact](../resources/contact.md) from a user's top level [contactFolder](../resources/contactfolder.md).
-```http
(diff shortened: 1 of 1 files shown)


dkershaw10:
Can you please resubmit once you've resolved the conflicts.

angelgolfer-ms:
Hi @Codeklopper , thank you very much for bringing this behavior to our attention. FYI this behavior is possibly an unintended bug and we're verifying this. 

Meanwhile, I have noted this issue in the [Known issues ](https://graph.microsoft.io/en-us/docs/overview/release_notes#contacts) topic to let other customers know about the behavior. 
Please use the v1.0 version of GET /contactfolders/{Id}/contacts/{id}. Or in beta, simply call GET /contacts{id} which would look in the user's mailbox (including all the contact folders) for the specified contact.

PR #483, opened 2016-09-21
Update list_device with correct permission scopes v1

viv-liu:


mattgeim:
@viv-liu can you fix up the conflicts and I will merge? Thanks!

PR #523, opened 2016-09-29
updating the deploy.cmd to point files to platform

DevArvindsa:


DevArvindsa:
discarding this pull request, we need to point to get-started instead platform

PR #570, opened 2016-10-18
Make sure appveyor script errors when errors occur

rgregg:


rgregg:
This fixes the appveyor issue, but we still need to fix the actual errors in the documentation.

PR #602, opened 2016-11-08
Fix various issues in user api articles

PatAltimore:


api-reference/beta/api/user_delete.md
```diff
@@ -30,7 +30,7 @@ Here is an example of the request.
   "name": "delete_user"
 }-->
 ```http
-DELETE https://graph.microsoft.com/beta/me
+DELETE https://graph.microsoft.com/beta/ba9a3254-9f18-4209-aeb3-9e42a35b5be4
```
dkershaw10: Yes - good - you shouldn't be able to delete yourself, so this is a good change.  But you've forgotten the "users" entity type in the URL

PatAltimore: Thanks. Making the change.

api-reference/beta/api/user_post_users.md
```diff
@@ -79,7 +79,7 @@ Content-type: application/json
     "officeLocation": null,
     "preferredLanguage": null,
     "surname": null,
-    "userPrincipalName": "upn-value@tenant-value@onmicrosoft.com"
+    "userPrincipalName": "upn-value@tenant-value.onmicrosoft.com"
```
dkershaw10: I think this is OK - but be better to show an example with the required fields

PR #606, opened 2016-11-10
Removed the "description" field

bharath063:
Removed the description field as OneDrive for Business doesn't support this field.


rgregg:
Description is a supported field for OneDrive personal drives though. Perhaps the example should be more general purpose since the property isn't available in all cases.

PR #641, opened 2016-11-28
Make it more clear which parameters are supported on delta

rgregg:


Lauragra:
Ryan, do you need to make the same change to the file in the beta folder? Thanks.

PR #751, opened 2017-01-06
Fixed indenting and numbering

mtillman:


mtillman:
@Lauragra The formatting on this tiny change I made got messed up on the live site, so I'm trying to fix. If there's a way I can see it on stage before it goes live, I could make sure it looks right. Thanks.

Issue #762, opened 2017-01-11
allowExternalSenders throws an error on group update


arnas47:
I'm trying to update allowExternalSenders group property using PATCH request, but it always returns me an error, that says: "The remote server returned an error: (404) Not Found." The permissions for MS Graph in Azure is set to Read all groups, Read and write all groups.

Here are my group properties:
{	
	"@odata.context":"https://graph.microsoft.com/v1.0/$metadata#groups/$entity",
	"id":"***",
	"classification":null,
	"createdDateTime":"2017-01-10T09:38:44Z",
	"description":"test apl",
	"displayName":"aptest33",
	"groupTypes":["Unified"],
	"mail":"***",
	"mailEnabled":true,
	"mailNickname":"aptest3",
	"onPremisesLastSyncDateTime":null,
	"onPremisesSecurityIdentifier":null,
	"onPremisesSyncEnabled":null,
	"proxyAddresses":["SMTP:***"],
	"renewedDateTime":"2017-01-10T09:38:44Z",
	"securityEnabled":false,
	"visibility":"Public"
}

marych:
Hi @arnas47. The allowExternalSenders property is not currently settable via the API. I created https://github.com/microsoftgraph/microsoft-graph-docs/issues/778 to update the documentation accordingly. That said, a 404 response would indicate another issue with the request. Could you please let me know the request-id and timestamp of a sample 404 response?

arnas47:
Hi @marych,

Request id is "fe10fb96-9dbb-4d38-8971-ce86d34b47c0" and date is "Wed, 18 Jan 2017 07:15:45 GMT"

LagerstedtMarcus:
Could this be prioritized? 
Would be nice if this function work, this is something that only can be changed through PowerShell and not in the Gui and therefore this is hard to change for new groups automatically.

marych:
@LagerstedtMarcus - the support for updating the allowExternalSenders group property was implemented a year ago. I just checked that it works fine - I'm able to update a group to set a different value for allowExternalSenders. The reason why 404 was returned for Arnas' request from Jan 2017 was that it was made with app-only token (obtained via the client credentials flow) and the service attempted and failed to resolve the user context.

LagerstedtMarcus:
@marych If I want to do this in an app without any user credentials, it's not possible? 
I got an app that I want to update this value for a group.
What is the reason to not make this able to change with an application token? 

[...]

PR #778, opened 2017-01-17
Update group.md

marych:
Removing a note that allowExternalSenders can be set with a PATCH request for the group as it's not currently supported.
https://github.com/microsoftgraph/microsoft-graph-docs/issues/762


Lauragra:
Hi Marek, can you update the file in the beta folder with this change as well? Thanks!

marych:
Sure, updated file in the beta folder as well.

PR #799, opened 2017-01-24
Update deployments.md

schmidtaa:


dkershaw10:
We'll be revamping this topic with more national cloud information.  We'll make this change with that overall larger change

PR #807, opened 2017-01-26
Fixes to subscribedsku and suspiciousipriskevent_list

gsacavdm:
General updates to all /subscribedsku pages
Fixing formatting in suspiciousipriskevent_list


api-reference/v1.0/api/subscribedsku_get.md
```diff
@@ -1,8 +1,9 @@
 # Get subscribedSku
-
-Retrieve the properties and relationships of subscribedsku object.
+Retrieve the properties and relationships of a service SKU that a company is subscribed to.
 ## Prerequisites
-One of the following **scopes** is required to execute this API: 
+One of the following **scopes** is required to execute this API:
+
+*User.Read*; *Directory.Read.All*; *Directory.AccessAsUser*
```
dkershaw10: It *may* work for user.read, but without checking I would err on the side of caution and remove that permission from the list.

gsacavdm:
#sign-off

api-reference/v1.0/api/subscribedsku_get.md
gsacavdm: Some else fixed the scopes and ensured User.Read wasn't added. So my PR is no limited to fixing the missing line break and the sample request/responses

PR #808, opened 2017-01-26
Updated calendar_list_calendarview.md example

MIchaelMainer:
The example is missing the startDateTime and endDateTime query parameter. The existing example will fail. Added the missing query parameters.


MIchaelMainer:
@angelgolfer-ms I added theupdate for the beta endpoint docs :-) thank you for the reminder to hit the docs for both endpoints.

PR #812, opened 2017-01-26
Update event.md

MIchaelMainer:
Added information about a time zone value in the case that Outlook desktop was used to set a custom time zone. I'll add this information to the beta endpoint as well.


Lauragra:
@MIchaelMainer , can you resolve the merge conflicts and resubmit please? Thanks.

PR #953, opened 2017-03-09
/drives path is not valid for a group

joeuy:
/drives path is not valid for a group. it should be /drive which returns the default document library.

Issuing a request against /drives will result in 400
https://graph.microsoft.com/v1.0/groups/{groupId}/drives
client-request-id: 1f5353f3-a0fe-4713-924d-ef247e4cbc8d
content-type: application/json
cache-control: private
request-id: 1f5353f3-a0fe-4713-924d-ef247e4cbc8d
Status Code: 400
```
{
    "error": {
        "code": "BadRequest",
        "message": "Unsupported segment type. ODataQuery: groups/adf3cd07-b2d5-4427-b859-eab8492e7e77/drives",
        "innerError": {
            "request-id": "1f5353f3-a0fe-4713-924d-ef247e4cbc8d",
            "date": "2017-03-06T22:35:32"
        }
    }
}
```


rgregg:
The right fix here is actually to remove the line that was changed in this fix. Since the topic is about listing drives, we should just remove the example of listing drives on a group. Can you update this change to just remove that line? Thanks.

joeuy:
@rgregg why are you suggesting to remove the line? The examples are indeed about listing drives, it seems useful to know how to list the drives for a group.

rgregg:
I've updated the topics mentioned here with a change to remove the /drives collection from a group and add sites to the examples. Thanks for the pull request and the suggestion on how to improve the content!

PR #972, opened 2017-03-17
Fix delta link; change target date.

PatAltimore:


api-reference/beta/api/group_delta.md
```diff
@@ -8,7 +8,7 @@
...
+Tracking changes to relationships on Users, Groups, Organizational Contacts, and Administrative Units is only supported within the specific resource class for which changes are being tracked. For example, if a client is tracking changes on *groups* and has selected the *members* relationship, the client will only receive membership updates in the delta query response if those members are also *groups*. In other words, tracking group membership for users is not yet supported. The Microsoft Graph team understands that this is a high priority scenario and an update is targeted to be delivered soon.
```
dkershaw10: Remove Organizational Contacts and AdministrativeUnits - they are not supported.  Check with Jeff on whether we'll support this both in /beta and AUs in /beta (and v1.0 when AUs are there).

PR #997, opened 2017-03-27
Update the link to emailAddress resource type in attendee.md

febilyt:


angelgolfer-ms:
@febilyt , thank you for your pull requests ( #997 and #999). You brought our attention to an internal bug. We have since fixed that bug and now the file name segment in a link is no longer case-sensitive. The fix has been deployed and the links you mentioned are now working and are case-insensitive.

PR #999, opened 2017-03-27
Update link to physicalAddress resource type in location.md

febilyt:
Update link to physicalAddress resource type


angelgolfer-ms:
@febilyt , thank you for your pull requests ( #997 and #999). You brought our attention to an internal bug. We have since fixed that bug and now the file name segment in a link is no longer case-sensitive. Therefore I am closing this PR. Thank you again and we look forward to your continued support.

PR #1017, opened 2017-03-30
Update user_get with permissions known issues

NBajanca:


dkershaw10:
This change is no longer required.  user.readwrite.all and user.read.all application permissions are now fully supported.

PR #1020, opened 2017-03-31
Fix the files permissions listing

rgregg:


rgregg:
Oops, need to make this against staging.

PR #1101, opened 2017-04-18
Update item_list_children.md

pred25:
Adding /groups/<key>/drive/root/children to list of HTTP requests to list children


viv-liu:
FYI this is in v1.0

PR #1188, opened 2017-05-04
User fixes for imAddresses and businessPhones

dkershaw10:


api-reference/beta/resources/user.md
```diff
@@ -62,13 +62,15 @@ This resource supports [delta query](../../../concepts/delta_query_overview.md)
 |assignedLicenses|[assignedLicense](assignedlicense.md) collection|The licenses that are assigned to the user. Not nullable.            |
 |assignedPlans|[assignedPlan](assignedplan.md) collection|The plans that are assigned to the user. Read-only. Not nullable. |
 |birthday|DateTimeOffset|The birthday of the user. The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`|
+|businessPhones|String collection|The telephone numbers for the user. NOTE: Although this is a string collection, only one number can be set for this property.|
```
PatAltimore: Slightly confusing. Since it can only be one number, should the description be "The telephone number for the user"? Maybe not since the property name is plural. Do we know what happens if you provide more than one number?  Does it use the first one in the collection or cause an error?

api-reference/v1.0/resources/user.md
```diff
@@ -57,13 +57,15 @@ This resource lets you add your own data to custom properties using [extensions]
 |assignedLicenses|[assignedLicense](assignedlicense.md) collection|The licenses that are assigned to the user. Not nullable.            |
 |assignedPlans|[assignedPlan](assignedplan.md) collection|The plans that are assigned to the user. Read-only. Not nullable. |
 |birthday|DateTimeOffset|The birthday of the user. The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`|
+|businessPhones|String collection|The telephone numbers for the user. NOTE: Although this is a string collection, only one number can be set for this property.|
```
PatAltimore: See comment for beta.

api-reference/beta/resources/user.md
dkershaw10: We created this as a collection, even though the store can only store one number.  This was to be more forward looking.  Yes it errors if you try to set more than one.

api-reference/v1.0/resources/user.md
PatAltimore: Looks good to me.

api-reference/v1.0/api/directoryobject_getbyids.md
```diff
@@ -0,0 +1,102 @@
...
+Returns the directory objects specified in a list of ids.  NOTE: The directory objects returned are the full objects containing **all** their properties. The `$select` query option is not available for this operation.
+
+Some common uses for this function are to:
+
+* Resolve ids returned by functions (that return collections of ids) such as [getMemberObjects](directoryobject_getmemberobjects.md) or [getMemberGroups](directoryobject_getmembergroups.md)  to their backing directory objects.
+* Resolve ids persisted in an external store by the application to their backing directory objects.
+
+## Prerequisites
+
+One of the following **scopes** are required to execute this API: _Directory.Read.All_
```
PatAltimore: One of the following **scopes** is required.

PatAltimore: One of the following scopes is required.

PR #1190, opened 2017-05-04
Fix spelling mistakes

john1506:
Spelling mistakes corrected from 'assocaited' to 'associated'


rgregg:
Thanks for the contribution John!

PR #1220, opened 2017-05-08
correct the example

Hongbo-Miao:
The old example is using object for "index" and "values" which actually should give number and array.


Lauragra:
Hi Hongbo, can you apply this fix to the file in the /beta folder as well? Thanks!

Hongbo-Miao:
Done! https://github.com/microsoftgraph/microsoft-graph-docs/pull/1375

PR #1293, opened 2017-05-11
Add description property to table

AdamDunn:
A minor change to add the "description" property to the table, in case the reader is not paying attention to the JSON representation.


rgregg:
Thank you!

PR #1322, opened 2017-05-18
Incorrect url for Azure Portal

jwidmer:
Updated with correct url for Azure Portal


Lauragra:
Thanks for submitting this fix!

PR #1325, opened 2017-05-22
Update angular.md

HardusLom:
Line 268 is incorrect. I've proposed a change. This is consistent with the sample code of the complete project.


jamescro:
Thanks for the correction!

PR #1328, opened 2017-05-22
Fix typo

arecaps:
Changed "cam" to "can".


Lauragra:
Good catch - thank you!

PR #1353, opened 2017-05-26
Update group_get.md

panjkov:
Typo spotted and corrected: Commas in API URL version (v1,0) corrected with dots(v1.0)


Lauragra:
Good catch - thanks!

PR #1361, opened 2017-05-26
Update site.md

panjkov:
Added an entry to "Tasks" table with an example (and link to site_get.md) for getting site for a group


rgregg:
Thank you!

PR #1367, opened 2017-05-30
Fix link to "Get access without a user" link

BrianTJackett:
"Get access without a user" link incorrectly pointed to the "...with a user" link.  Updating to proper link.


Lauragra:
Thanks!

PR #1387, opened 2017-06-06
Fix issues in device api articles

PatAltimore:


api-reference/beta/api/device_post_registeredowners.md
```diff
@@ -1,14 +1,15 @@
 # Create registeredOwner
 
 Add a user as a registered owner of the device.
```
dkershaw10: What additional benefits are "bestowed" on a user that is made an owner of a device?  We should doc that.  Can you find out from Shantaram please.
(On a separate note, can you check what we doc on groups, when setting a user as an owner?)

api-reference/beta/api/device_post_registeredowners.md
```diff
@@ -1,14 +1,15 @@
...
+
 ## Prerequisites
```
dkershaw10: Shouldn't this be in the new permissions format?

api-reference/beta/api/device_post_registeredowners.md
```diff
@@ -1,14 +1,15 @@
...
 One of the following **scopes** is required to execute this API:  
 *Directory.ReadWrite.All* or *Directory.AccessAsUser.All*
```
dkershaw10: I think there's also an app-only devices permission.  Did you check that ReadWrite.All actually allows this operation?

api-reference/beta/api/device_post_registeredowners.md
```diff
@@ -17,47 +18,33 @@ POST /devices/{id}/registeredOwners
 | Authorization  | string  | Bearer <token>. Required. |
```
dkershaw10: `<token>` is not appearing in the rendered topic.  How does this appear in other topics?

api-reference/beta/api/device_post_registeredowners.md
```diff
@@ -17,47 +18,33 @@ POST /devices/{id}/registeredOwners
...
+In the request body, supply a reference to the [user](../resources/user.md) to be added.
 
 ## Response
-If successful, this method returns `201, Created` response code and [directoryObject](../resources/directoryobject.md) object in the response body.
+If successful, this method returns `204, No Content` response code. It does not return anything in the response body.
 
 ## Example
 ##### Request
-Here is an example of the request.
+Here is an example of the request. In the request body, supply a reference to the [user](../resources/user.md) to be added.
```
dkershaw10: Check with Shantaram whether applications (service principals) can also be owners...

api-reference/beta/api/device_post_registeredowners.md
jasonjoh: Yep, that's because Markdown supports HTML markup, and that looks like an HTML tag. My suggestion would be to use `{token}` instead. I see a number of topics have been updated to use that convention.

PR #1392, opened 2017-06-07
Added link; minor edits.

Lauragra:


concepts/known_issues.md
```diff
@@ -1,6 +1,6 @@
 # Known issues with Microsoft Graph
 
-This article describes known issues with the Microsoft Graph. For information about the latest updates, see the [Microsoft Graph Changelog](changelog.md).
+This article describes known issues with Microsoft Graph. For information about the latest updates, see the [Microsoft Graph Changelog](changelog.md).
```
angelgolfer-ms: I think "Changelog" should be lower case 'c' here.

concepts/known_issues.md
```diff
@@ -58,15 +58,15 @@ Examples of group features that support only delegated permissions:
 
 ### Teams in Microsoft Teams (preview)
 
-Microsoft Teams are built upon Office 365 groups.  All group APIs can also be used with teams, with the exception that 'Create group' does not currently allow you to create a team.  Future API releases will support this.
+Microsoft Teams are built on Office 365 groups. All group APIs can also be used with teams, with the exception that the Create group API does not currently allow you to create a team.  Future API releases will support this.
```
angelgolfer-ms: The teams overview topic has removed this concept of "built on Office 365 groups", and we will clarify the relationship when the create team story is clearer. For now, we can say "Microsoft Teams and Office 365 groups share similar functionality. All group API can be used with teams, with the exception..."

PR #1395, opened 2017-06-07
Fixed formatting

mlafleur:
There were some markdown rendering issues that made it unclear which
properties were selected by default and which properties could not be
selected. I've moved this information under the Optional query
parameters section and cleaned up the rendering issues.


api-reference/beta/api/user_list.md
```diff
@@ -2,30 +2,41 @@
...
 <!-- { "blockType": "ignored" } -->
 ```http
 GET /users
 ```
+
 ## Optional query parameters
+
 This method supports the [OData Query Parameters](http://developer.microsoft.com/en-us/graph/docs/overview/query_parameters) to help customize the response.
+
+By default, only a limited set of properties are returned (*businessPhones, displayName, givenName, id, jobTitle, mail, mobilePhone, officeLocation, preferredLanguage, surname, userPrincipalName*). Use `$select` to customize which properties and relationships are returned for each  [user](../resources/user.md).
```
dkershaw10: Actually - for /beta this isn't even true.  In /beta we return ALL properties.
The note (below in v1.0) about certain properties only available in GET vs LIST is actually true for BOTH /beta and /v1.0.

We're planning some changes to this topic - I think the changes are good in general, but we explicitly wanted to tell devs up front that in v1.0/users we have a default select behavior.  This comes up ALL the time as developers complaining that properties in /beta are not available in /v1.0.

api-reference/v1.0/api/user_list.md
```diff
@@ -2,41 +2,44 @@
...
 ```http
 GET /users
 ```
+
 ## Optional query parameters
+
 This method supports the [OData Query Parameters](http://developer.microsoft.com/en-us/graph/docs/overview/query_parameters) to help customize the response.
+
+By default, only a limited set of properties are returned (*businessPhones, displayName, givenName, id, jobTitle, mail, mobilePhone, officeLocation, preferredLanguage, surname, userPrincipalName*). Use `$select` to customize which properties and relationships are returned for each  [user](../resources/user.md).
+
```
JimacoMS2: I think  you should be more explicit that "customize" -- more like "Use `$select` to return properties and relationships that are not in the default set."

api-reference/beta/api/user_list.md
JimacoMS2: Same comment wrt "customize" as below.

mlafleur:
I've removed the $select language from the beta endpoint as this endpoint does not use a default $select query. 

I've also cleaned up the language in the v1.0 endpoint and provided an example $select statement per previous suggestions.

PR #1420, opened 2017-06-14
Updated with another permission

viv-liu:
User.ReadWrite.All


api-reference/v1.0/api/profilephoto_update.md
```diff
@@ -13,6 +13,7 @@ You can use either PATCH or PUT for this operation in version 1.0.
 One of the following **scopes** is required to execute this API for:
 
 - Profile photo of signed-in **user** - *User.ReadWrite*
+- Profile photo of all **users** in the organization - *User.ReadWrite.All*
```
dkershaw10: Unfortunately it's a bit more screwed up than that.  For delegated flow, this permission only permits update of the signed in user's photo.  For app-only flow, it allows update of any user's photo.

viv-liu: This is something that can be addressed in a table. Thanks for calling this out 👍

PR #1443, opened 2017-06-21
Clean up default $select description

mlafleur:
Reformatted and reworded how we describe the default $select statment applied to user_list. 

I also removed this information from the beta endpoint as within beta we return the entire user object rather than a default property set.

### DIFF
--- a/api-reference/beta/api/user_list.md
+++ b/api-reference/beta/api/user_list.md
@@ -2,30 +2,39 @@
 
 Retrieve a list of user objects.
 
-> Note: Listing users returns a default set of properties only (*businessPhones, displayName, givenName, id, jobTitle, mail, mobilePhone, officeLocation, preferredLanguage, surname, userPrincipalName*). Use `$select` to get the other properties and relationships for the [user](../resources/user.md) object.
-
 ## Prerequisites
+
 One of the following **scopes** is required to execute this API:
 *User.ReadBasic.All; User.Read.All; User.ReadWrite.All; Directory.Read.All; Directory.ReadWrite.All; Directory.AccessAsUser.All*
+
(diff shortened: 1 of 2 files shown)


api-reference/beta/api/user_list.md
```diff
@@ -2,30 +2,39 @@
...
 This method supports the [OData Query Parameters](http://developer.microsoft.com/en-us/graph/docs/overview/query_parameters) to help customize the response.
+
 ## Request headers
-| Header       | Value|
-|:-----------|:------|
-| Authorization  | Bearer {token}. Required.  |
-| Content-Type   | application/json | 
+| Header        | Value                      |
+|:--------------|:---------------------------|
+| Authorization | Bearer <token> (required)  |
```
PatAltimore: @mlafleur This won't render correctly.  The publishing system thinks it's an HTML tag and drops it. They did a sweep of the content to fix this.  Change to Bearer {token}. Required.

api-reference/v1.0/api/user_list.md
```diff
@@ -2,41 +2,46 @@
...
+
 ## Request headers
-| Header       | Value|
-|:-----------|:------|
-| Authorization  | Bearer {token}. Required.  |
-| Content-Type   | application/json | 
+
+| Header        | Value                      |
+|:--------------|:---------------------------|
+| Authorization | Bearer <token> (required)  |
```
PatAltimore: Same comment. change to {token}.

api-reference/beta/api/user_list.md
mlafleur: DOH! Great catch, thank you. I've checked in a fix for this.

PR #1486, opened 2017-07-14
Edits to the Azure AD overview topic.

Lauragra:


api-reference/v1.0/resources/azure_ad_overview.md
```diff
@@ -1,42 +1,42 @@
...
+To call the Microsoft Graph APIs on Azure AD resources, your app will need the appropriate permissions. Many of the APIs exposed on Azure AD resources require one of the [_Directory_ permissions](../../../concepts/permissions_reference.md#directory-permissions). Directory permissions are highly privileged and always require administrator consent. 
 
-There are two kinds of permissions: delegated and application. If your app is calling an API on behalf of a user, it will need delegated permissions. If it is calling an API as itself without a signed-in user, it will need application permissions. This latter scenario is generally the case with back-end services or daemons. For more information about delegated and application permissions, see [Permissions](../../../concepts/permissions_reference.md). 
+If your app is acting on behalf of a user (application permissions), that user will likely need to be a member of an appropriate [administrator role](https://docs.microsoft.com/azure/active-directory/active-directory-assign-admin-roles) for your app to successfully call many of the Azure AD APIs.
```
JimacoMS2: IMPORTANT! change "(application permissions)" to "(delegated permissions)"

api-reference/v1.0/resources/azure_ad_overview.md
```diff
@@ -1,42 +1,42 @@
...
+With Microsoft Graph, you can access [Azure Active Directory (Azure AD)](https://docs.microsoft.com/azure/active-directory/active-directory-whatis) resources to enable scenarios like including managing administrator (directory) roles, inviting external users to an organization, and, if you are [Cloud Solution Provider (CSP)](https://partner.microsoft.com/cloud-solution-provider), managing your customer's data. Microsoft Graph also provides methods that apps can use to, for example, discover information about users' transitive group and role memberships. 
 
-> **Note**: Some Azure AD resources are documented in other sections of the API reference. For more information, see [users](users.md) and [groups](group.md).
+> **Note**: Some Azure AD resources are documented in other sections of the API reference. For more information, see [Users](users.md) and [Groups](group.md).
```
JimacoMS2: It's up to you, but the resources are actually lower case.

angelgolfer-ms: I think we should be clearer with the second sentence, by saying, "For more information, see the user and group resources.", or, "For more information, see user and group.", with hot links to the user and group resource topics.

The resource names are singular, lower (camel) case, always.

api-reference/v1.0/resources/azure_ad_overview.md
```diff
@@ -1,42 +1,42 @@
...
 
-There are two kinds of permissions: delegated and application. If your app is calling an API on behalf of a user, it will need delegated permissions. If it is calling an API as itself without a signed-in user, it will need application permissions. This latter scenario is generally the case with back-end services or daemons. For more information about delegated and application permissions, see [Permissions](../../../concepts/permissions_reference.md). 
+If your app is acting on behalf of a user (application permissions), that user will likely need to be a member of an appropriate [administrator role](https://docs.microsoft.com/azure/active-directory/active-directory-assign-admin-roles) for your app to successfully call many of the Azure AD APIs.
 
-Finally, if your app is acting on behalf of a user, that user will likely need to be a member of an appropriate [administrator role](https://docs.microsoft.com/azure/active-directory/active-directory-assign-admin-roles) for your app to successfully call many of the Azure AD APIs.
+For more information about permissions, including delegated and application permissions, see [Permissions](../../../concepts/permissions_reference.md).
```
angelgolfer-ms: In general, I would be happy to add a See-also link to the permissions content in the overview topic(s) for the Outlook API, but to bubble the permissions discussion up in the body of their respective overview topic and make it informative enough to make a point (or a few), I would be repeating content that's fully explained in the permissions content.
As a matter of fact, because these overview topics are in the API reference section, it's just as fine that we focus on the capabilities of the API in these overview topics. When the customer gets to a point to start implementing these API calls, the corresponding API reference topic will have a link to the permissions topic, is that correct?

api-reference/v1.0/resources/azure_ad_overview.md
```diff
@@ -1,42 +1,42 @@
...
 
-With Microsoft Graph, you can access [Azure Active Directory (Azure AD)](https://docs.microsoft.com/azure/active-directory/active-directory-whatis) resources that you can use to enable several different scenarios. This includes managing administrator (directory) roles, inviting external users to an organization, and, if you are [Cloud Solution Provider (CSP)](https://partner.microsoft.com/cloud-solution-provider), managing your customer's data. Microsoft Graph also provides several methods that can be used by many kinds of apps; for example, to discover information about transitive group and role memberships of users. 
+With Microsoft Graph, you can access [Azure Active Directory (Azure AD)](https://docs.microsoft.com/azure/active-directory/active-directory-whatis) resources to enable scenarios like including managing administrator (directory) roles, inviting external users to an organization, and, if you are [Cloud Solution Provider (CSP)](https://partner.microsoft.com/cloud-solution-provider), managing your customer's data. Microsoft Graph also provides methods that apps can use to, for example, discover information about users' transitive group and role memberships.
```
angelgolfer-ms: Under the Use case column, can we please start with a summary sentence for each use case?

Ideally, describing these use cases (how they are supported) should actually go into the corresponding resource topics (after the resource description in the resource topic), or in a separate lower level overview topic, and this overview topic would link to them in the second or third column. 

However, if this is currently not the case with Azure content, at least, can we separate the summary use case sentence from the rest of the use case description, by starting a new paragraph in that same cell of the table?

api-reference/v1.0/resources/azure_ad_overview.md
```diff
@@ -1,42 +1,42 @@
...
 
 ## Common use cases 
 
-The following table lists some common scenarios that Azure AD resources can be used for.
+The following table lists some common use cases for Azure AD resources.
 
 | Use cases		   | REST resources	| See also |
 |:---------------|:--------|:----------|
 | **Directory object and methods** | | |
 | `directoryObject` is the base class that many directory resources, like users and groups, inherit from. Microsoft Graph exposes several methods that you can use to discover information about users, groups, and other directory objects. For example, you can check for transitive membership in a list of groups, return all the groups and directory roles that a directory object is a transitive member of, or get all the resources of a specified type (like user or group) from a list of generic resource IDs. | [directoryObject](../resources/directoryobject.md) | N/A |
```
angelgolfer-ms: Instead of saying "Directory object and methods", can we rephrase and be more direct in describing what actions the API support?

And in general, for this table:
Under the Use case column, can we please start with a summary sentence for each use case? (This is analogous to documenting a method, in any platform, using the Microsoft dev div style. We always start with a verb and a single-sentence summarizing what that method does.)

Ideally, describing these use cases (how they are supported) should actually go into the corresponding resource topics (after the resource description in the resource topic), or in a separate lower level overview topic, and this overview topic would link to them in the second or third column. 

However, if this is currently not the case with Azure content, at least, can we separate the summary use case sentence from the rest of the use case description, by starting a new paragraph in that same cell of the table?

api-reference/v1.0/resources/azure_ad_overview.md
```diff
@@ -1,42 +1,42 @@
...
 
-> **Note**: Some Azure AD resources are documented in other sections of the API reference. For more information, see [users](users.md) and [groups](group.md).
+> **Note**: Some Azure AD resources are documented in other sections of the API reference. For more information, see [Users](users.md) and [Groups](group.md).
 
 
 ## Authorization
  
-To call the Microsoft Graph APIs on Azure AD resources, your app will need the appropriate permissions.  Many of the APIs exposed on Azure AD resources require one of the [_Directory_ permissions](../../../concepts/permissions_reference.md#directory-permissions).  _Directory_ permissions are highly privileged and always require administrator consent. 
+To call the Microsoft Graph APIs on Azure AD resources, your app will need the appropriate permissions. Many of the APIs exposed on Azure AD resources require one of the [_Directory_ permissions](../../../concepts/permissions_reference.md#directory-permissions). Directory permissions are highly privileged and always require administrator consent.
```
angelgolfer-ms: STrange. I entered a comment about permissions earlier and now it's gone.
Anyway, I'd be happy to add a See-also link to the permission content at the bottom of the overview topic. To bubble up the permissions subject in the body of the overview topic, and to make it say something useful, I would be repeating some of the content in the auth concept topics. I would instead like to dedicate the API overview topic to highlighting the capabilities of the API set. 
Look, our API capabilities and interconnection is where we can compete with Goggle. Our auth design is more complicated than Google's, our auth content is heavy, I don't want to divert customers' attention in this overview topic to link over to the auth content and lose the momentum. Put it another way, auth is not our "good side" to highlight. Let's just lure the customer to want to use our platform first, and then when they have to implement the app, they'll go through and figure out the auth stuff. :-)

api-reference/v1.0/resources/azure_ad_overview.md
JimacoMS2: Actually, Laura had a point. I think the pertinent question is whether to link to the overview topics or the resources themselves.  If the point is to link to resources then lower case is probably best. If to the overview -- upper case, although, honestly, I don't think that the user is going to really care and I could see making an equally valid argument for lower case.

My gut tells me we should probably link to the overview topics in this instance.

api-reference/v1.0/resources/azure_ad_overview.md
JimacoMS2: I think that we last agreed that Doug was going to appoint someone to be point person on this and reach a decision. Until that happens, I'd prefer to leave the auth section in the topic. I could see relocating it to the bottom, but would prefer to have a format settled first rather than iterating on it. 

Personally, I think that which permissions you use to work with an area is just as primary as the API capabilities and interconnections -- this is an area that has continued to plague customers. The permissions reference is a necessarily long topic that treats every permission exposed in Microsoft Graph. I don't really see an issue with having a small part of the overview topics that narrows the focus to the subset of permissions used by a specific workload and then treating the more granular cases in each API topic. In fact, that seems like a logical intermediate step to me.  Anyway -- my 2 cents.  For now, I'd like to leave it in these topics.

Lauragra: Doug appointed me to be the decision maker. ;) I left the Authorization section in the article because I agree that it is helpful to call this out in the overview topics. I made it shorter to address the issue with duplicate content, so instead we primarily link to the Permissions topic for the conceptual information. After we merge this PR, I'll generate a template based on this topic that we'll add to the CDK site.

api-reference/v1.0/resources/azure_ad_overview.md
JimacoMS2: Hey Angela -- I think that we do start with a summary sentence with everything except DO. The primary thing about DO is that it is the base class for a large subset of the AAD classes. It's also important to understand that we aren't really documenting methods here for AAD, beyond the ones mentioned in DO, but resources. This is because, unlike, other workloads, the AAD surface spans more than just 2-3 resources and they aren't all targeting a single set of functionality like say mail or calendars. I could split the DO row into one covering the base class and one covering the utility functions tho ... but again, I think this is kind of a one-off situation wrt the other rows in the table.

PR #1488, opened 2017-07-14
'joined Teams' API currently only works for current user

richardtaylorrt:


concepts/known_issues.md
```diff
@@ -55,19 +55,23 @@ Examples of group features that support only delegated permissions:
 * Group conversations, events, photo
 * External senders, accepted or rejected senders, group subscription
 * User favorites and unseen count
+* Microsoft Teams channels and chats.
 
-### Teams in Microsoft Teams (preview)
+### Teams in Microsoft Teams (beta)
```
angelgolfer-ms: Hi Richard, according to our style guide, we do use "(preview)" to identify a feature being in the beta endpoint only. This is because the concept of "preview" entails additional implications that a developer needs to be aware of, which include "Expect breaking changes to the beta version from time to time. Do not take a production dependency on /beta APIs."  (as documented [here](https://developer.microsoft.com/en-us/graph/docs/concepts/versioning_and_support)). This concept of "preview" is used not only in Microsoft Graph docs, but is also common inside and outside of Microsoft, and we're just banking on the familiarity of this concept.

On the other hand, "beta" simply denotes the version of an endpoint or a feature.

concepts/known_issues.md
```diff
@@ -55,19 +55,23 @@ Examples of group features that support only delegated permissions:
...
 Currently, you can read and create channels, but you cannot update or delete them.  Future API releases will support this.
 
-### Microsoft Teams chat threads and chat messages (preview)
+### Microsoft Teams chat threads and chat messages (beta)
 
 Currently, you can create chat threads in channels, but you cannot read existing chat threads or add replies to them. Also, you cannot read or write direct chats between users that are outside the scope of a team or channel.  Future API releases will add additional capabilities in this area.
 
+### Microsoft Teams user's list of joined teams (beta)
+
+Currrently, [listing the teams a user has joined](../api-reference/beta/resources/user_list_joinedteams.md) only works for the 'me' user for which the caller has [delegated permissions](../../../concepts/permissions_reference.md).  Future releases will support this operation for any specified user ID.
```
Lauragra: This link has an error - the user_list_joinedteams file is in the api folder (not resources).

PR #1497, opened 2017-07-19
Update channel_post_chatthreads.md

Unlockedluca:
Updated URL for POST request from threads to chatthreads


paddymccallan:
Can this page be updated? [Create thread](https://docs.microsoft.com/en-us/graph/api/conversation-post-threads?view=graph-rest-1.0&tabs=http ).
I Spent sometime finding this post. I was getting a 201 - Created with the old URL but nothing was appearing in the channel.

PR #1522, opened 2017-07-25
Changed sample request payload and configuration endpoint for deviceConfigurationAssignment documentation

trjast:
- Sample request should not include the "id" property since that's the unique id for the deviceConfigurationAssignment which is generated on created
- Endpoint for assigning the deviceConfiguration was missing a much needed "/" between the {deviceConfigurationId} and assign


lleonard-msft:
#hold-off

This version of the page has been superseded by a later one created by the page generator.  (The missing path separator was due to a bug that was later fixed.)  

After review, the original changes were/are correct, based on the current capabilities of the tooling.  

To clarify the page at this level, incorporate changes higher in the toolchain (otherwise, they'll be overwritten in the next drop of the generated content).

PR #1542, opened 2017-08-02
Update user.md

marmoy:
Documenting that the relation directReports does not support the odata query parameters $expand and $count. MS Graph explorer itself states that $count is not supported, whereas $expand is just ignored for the directReports relation by MS Graph


dkershaw10:
@marmoy We could take this change, but it would need to be replicated in **many** places and not just for directReports. We are about to make some major updates to the [Query Parameters/Customize queries](https://developer.microsoft.com/en-us/graph/docs/concepts/query_parameters) topic that should clarify what is and isn't supported.  Let us know what you think.

jasonjoh:
Closing per @dkershaw10's comment.

PR #1597, opened 2017-08-16
Initial update for 1708 changes.

lleonard-msft:


Lauragra:
Hey Lance,
Is there a reason why we’re changing relative links to hard-coded URLs in the Intune content? There was a separate thread about this, the hard-coded URLs cause problems for loc because they include the language identifiers (“en-us”). As a best practice, we should have relative links. There’s a longer story that I can loop you in on….

Thanks,
Laura

lleonard-msft:
@Lauragra - The links have been updated in the new commit.  Please verify.

lleonard-msft:
#sign-off

PR #1600, opened 2017-08-16
Group overview

viv-liu:


api-reference/v1.0/resources/group_overview.md
```diff
@@ -0,0 +1,126 @@
...
+|-------------------|----------|-----------|--------------|------------------|--------------------------------|
+| Office 365 groups | Facilitating user collaboration with shared Microsoft online resources. | ["Unified"] | true | false | yes | [user](user.md) | 
+| Security groups | Controlling user access to in-app resources. | [] | false | true | yes |
+| Mail-enabled security groups | Controlling user access to in-app resources, with a shared group mailbox. | [] | true | true | no |
+| Distribution groups | Distributing mail to the members of the group. It is recommended to use Office 365 groups due to the richer set of resources it provides. | [] | true | false | yes |
+
+## Office 365 groups in Outlook
+The power of Office 365 groups is in its collaborative nature, perfect for people who work together on a project or a team. They are created with resources that members of the group share including:
+
+- Outlook conversations
```
angelgolfer-ms: Suggest to call this "Conversations (displayed in Outlook)". "Outlook conversations" is a different feature from group conversations. Every Outlook user can experience Outlook conversations. Group conversations on the other hand are not displayed or accessible in Outlook for users whose mailboxes are on-premises (like me, for the longest time). Programmatically, "Outlook conversations" are supported only in the object model, PIA, etc., but not in the Outlook REST API.

angelgolfer-ms: Ok, about the last sentence, just to be clear, there is separate API in the object model, PIA... to manage conversations. In the REST API, sure, we can access group conversations as conversation, conversationThread and post resources. And that's limited to group conversations. Non-group-based, Outlook conversations have no such support in the REST API.

api-reference/v1.0/resources/group_overview.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Office 365 groups in Outlook
+Office 365 groups enable powerful collaboration, and are designed for people who work together on a project or a team. Office 365 groups include resources that members of the group share, including:
+
+- Outlook conversations
+- Outlook calendar
+- SharePoint files
+- OneNote notebook
+- SharePoint team site
+- Planner
```
Lauragra: Should we call it a "Planner Hub"?

api-reference/v1.0/resources/group_overview.md
```diff
@@ -0,0 +1,131 @@
...
+## Types of groups
+
+The following table lists the types of groups that you can access via Microsoft Graph APIs.
+
+| Type              | Used to | groupType | Mail-enabled | Security-enabled | Can be created via APIs? |
+|-------------------|----------|-----------|--------------|------------------|--------------------------------|
+| Office 365 groups | Facilitate user collaboration with shared Microsoft online resources. | ["Unified"] | True | False | Yes | [user](user.md) | 
+| Security groups | Control user access to in-app resources. | [] | False | True | Yes |
+| Mail-enabled security groups | Control user access to in-app resources, with a shared group mailbox. | [] | True | True | No |
+| Distribution groups | Distribute mail to the members of the group. We recommend that you use Office 365 groups instead of distribution groups because they provide a richer set of resources. | [] | True | False | Yes |
```
dkershaw10: I don't think distribution groups can be created through the API.  Please change the last Yes to No.
Also "...because they provide a richer set of resources." sounds weird - I prefer "...because they provide a richer set of features."

api-reference/v1.0/resources/group_overview.md
dkershaw10: What's it's official name?  It's still called "Planner" in the ToC.

api-reference/v1.0/resources/group_overview.md
```diff
@@ -0,0 +1,131 @@
...
+    "proxyAddresses": [],
+    "securityEnabled": true
+}
+```
+### Other types of groups
+
+Office 365 groups in Yammer are used to facilitate user collaboration through Yammer posts. This type of group can be returned through a read request, but their posts can't be accessed through the API. To learn more, see [Yammer developer API docs](https://developer.yammer.com/docs).
+
+## Dynamic membership 
+All types of groups can have dynamic membership rules, which automatically add or remove members from the group based on user properties. For example, a "Marketing employees" group would include every user with the department property set to "Marketing", so that new marketing employees are automatically added and employees who leave the department are automatically removed from the group. This rule can be specified in a "membershipRule" field during group creation as ```"membershipRule": 'user.department -eq "Marketing"'``` GroupType must also include ```"DynamicMembership"```. The following request creates a new Office 365 group for marketing employees.
```
dkershaw10: Something is messed up with your expressions - I think you have too many single quotes, and I don't think the syntax is correct.
`"membershipRule": "user.department -eq \"Marketing\""`

api-reference/v1.0/resources/group_overview.md
```diff
@@ -0,0 +1,131 @@
...
+    "description": "Marketing department folks",
+    "displayName": "Marketing department",
+    "groupTypes": [
+        "Unified",
+        "DynamicMembership"
+    ],
+    "mailEnabled": true,
+    "mailNickname": "marketing",
+    "securityEnabled": false,
+    "membershipRule": 'user.department -eq "Marketing"',
```
dkershaw10: This is also wrong.  I think it should be:
"membershipRule": "user.department -eq \"Marketing\""

PatAltimore:
Closing.  Creating new PR to finish changes.

PR #1621, opened 2017-08-21
Fix/1708 perms etc

lleonard-msft:


lleonard-msft:
#sign-off  - @Lauragra

PR #1623, opened 2017-08-21
Fixed GET request type that prevents users get proper result from token request.

haojunsui:


dkershaw10:
@haojunsui Can you explain why you want to make this change please?  Token acquisition for the `client_credentials` flow uses POST so that various properties including client secret or credentials are transmitted in the body of the request.

PatAltimore:
Couldn't get clarification from @haojunsui.  Closing for now.

PR #1626, opened 2017-08-22
Add missing endpoints in pattern 

shuson:
Some pages have missing endpoint in pattern /drives/{drive-id}/items/{item-id}

I tried these missing endpoint, they do work from Graph API explorer.


jthake:
@JeremyKelley do you mind reviewing this and if correct merging into the docs?

PR #1682, opened 2017-09-13
Update the syntax of the schemaExtension

Copowell:
Update the syntax of the schemaExtension creation request to include an underscore ( _ ) as the service returns an error if you only include the name of the schemaExtension (for the service to complete the **id** assignment).


api-reference/v1.0/api/schemaextension_post_schemaextensions.md
```diff
@@ -39,7 +39,7 @@ The following table shows the properties that are required when you create a sch
...
+|id|String|The unique identifier for the schema extension definition. <br>You can assign a value in one of two ways: <ul><li>Concatenate the name of one of your verified domains with a name for the schema extension to form a unique string in this format, \{_&#65279;domainName_\}\_\{_&#65279;schemaName_\}. As an example, `contoso_mySchema`. </li><li>Provide a schema name with an underscore before it, and let Microsoft Graph use that schema name to complete the **id** assignment in this format: ext\{_&#65279;8-random-alphanumeric-chars_\}\_\{_&#65279;schema-name_\}. For example: _mySchema. The created object example would be `extkvbmkofy_mySchema`.</li></ul>This property cannot be changed after creation. |
```
angelgolfer-ms: Hi @Copowell, thanks for your PR. Using the example of extkvbmkofy_mySchema - did you find that Microsoft Graph did not complete the id assignment by including an underscore between extkvbmkofy and mySchema? That is, did you end up getting a string like extkvbmkofymySchema (without the underscore) assigned to id?
Thanks.

Copowell: Sorry, this wasn't clear. When my customer tried to create the schemaExtension object, they tried with the name of "mySchema" (just as an example). The service responded with a failure saying we needed to include the domain name to which this applies (but the customer wanted to let the service determine the **id** before the underscore). When we resent the request with the name of "_mySchema" the service responded with a 200 and 'filled in' the **id** before the _mySchema. I was requesting this to be more clearly documented here since the customer was confused by the documentation since it seemed wrong.

Thanks!

jasonjoh:
@Copowell thanks for contributing. I realize it's been quite some time since you opened this, and it's possible things have changed in the Graph. I just tried creating a schema extension without an underscore and it worked fine. Given that, I'm going to close this PR. If you run into this behavior again, please open an issue so we can investigate.

Copowell:
Sounds good. Thanks Jason!

Thank you,
Colin Powell

Support Engineer
Developer - Microsoft/Azure Graph API Support

Office: +1 (469) 775-4161<tel:+14697754161>
Colin.Powell@microsoft.com<mailto:colin.powell@microsoft.com>
Working hours: M-F; 11am to 8pm CT

PR #1711, opened 2017-09-19
Revert "Schema extension updates"

PatAltimore:
Reverts microsoftgraph/microsoft-graph-docs#1706


api-reference/v1.0/resources/schemaextension.md
```diff
@@ -29,8 +29,8 @@ See the [schema extension example](../../../concepts/extensibility_schema_groups
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
 |description|String|Description for the schema extension.|
-|id|String|The unique identifier for the schema extension definition. The value must be a concatenation of one of your verified domains (e.g. contoso.com) and a name for the schema extension - e.g. *contoso_mySchema*. |
-|owner|String|The `appId` of the application that is the owner of the schema extension. This property can be supplied on creation, to set the owner.  If not supplied, then the calling application's `appId` will be set as the owner. In either case, the signed-in user must be the owner of the application. Once set, this property is read-only and cannot be changed.|
```
angelgolfer-ms: I must have overlooked this appId mention before. This should be "application ID". In general, if we refer to a property name, we need to make it clear the parent object of the property (especially the object in context, in this case that's schemaExtension, is not the parent object), and provide a link to that parent object on first mention.

api-reference/beta/resources/schemaextension.md
```diff
@@ -32,8 +32,8 @@ See the [schema extension example](../../../concepts/extensibility_schema_groups
...
-|owner|String|The `appId` of the application that is the owner of the schema extension. This property can be supplied on creation, to set the owner.  If not supplied, then the calling application's `appId` will be set as the owner. In either case, the signed-in user must be the owner of the application. Once set, this property is read-only and cannot be changed.|
+|id|String|The unique identifier for the schema extension definition. <br>You can assign a value in one of two ways: <ul><li>Concatenate the name of one of your verified domains with a name for the schema extension to form a unique string in this format, \{_&#65279;domainName_\}\_\{_&#65279;schemaName_\}. As an example, `contoso_mySchema`. </li><li>Provide a schema name, and let Microsoft Graph use that schema name to complete the **id** assignment in this format: ext\{_&#65279;8-random-alphanumeric-chars_\}\_\{_&#65279;schema-name_\}. An example would be `extkvbmkofy_mySchema`.</li></ul>This property cannot be changed after creation. |
+|owner|String|The appId of the application that created the schema extension. Read-only.|
```
angelgolfer-ms: I must have overlooked this appId mention before. This should be "application ID". In general, if we refer to a property name, we need to make it clear the parent object of the property (especially the object in context, in this case that's schemaExtension, is not the parent object), and provide a link to that parent object on first mention.

PR #1749, opened 2017-09-25
[DO NOT MERGE] Office 365 usage reports JSON version to Beta

kszb:


api-reference/beta/api/reportroot_skypeforbusinessactivityusercounts.md
```diff
@@ -0,0 +1,89 @@
...
+{
+  "@odata.context": "https://graph.microsoft.com/beta/$metadata#Collection(microsoft.graph.skypeForBusinessActivityUserCounts)", 
+  "value": [
+    {
+      "peerToPeer": 413, 
+      "organized": 30, 
+      "participated": 91, 
+      "reportRefreshDate": "2017-09-01", 
+      "reportDate": "2017-09-01", 
+      "reportPeriod": "7"
```
ElizabethSamuel-MSFT: I'm curious why reportPeriod is a string

api-reference/beta/api/reportroot_skypeforbusinessactivityuserdetail.md
```diff
@@ -0,0 +1,157 @@
...
+      "organizedConferenceIMCount": 0, 
+      "organizedConferenceAudioVideoCount": 0, 
+      "organizedConferenceAudioVideoMinutes": 0, 
+      "organizedConferenceAppSharingCount": 0, 
+      "organizedConferenceWebCount": 0, 
+      "organizedConferenceDialInOut3rdPartyCount": 0, 
+      "organizedConferenceCloudDialInOutMicrosoftCount": 0, 
+      "organizedConferenceCloudDialInMicrosoftMinutes": 0, 
+      "organizedConferenceCloudDialOutMicrosoftMinutes": 0, 
+      "participantedConferenceIMCount": 0,
```
ElizabethSamuel-MSFT: I don't understand what this represents: participantedConferenceIMCount. Do you mean, participatedInConferenceIMCount?

api-reference/beta/api/reportroot_skypeforbusinessorganizeractivityusercounts.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Request headers
+
+| Name          | Description               |
+| :------------ | :------------------------ |
+| Authorization | Bearer {token}. Required. |
+
+## Response
+
+If successful, this method returns a `200 OK` response code and an **skypeForBusinessOrganizerActivityUserCounts** object in the response body.
```
ElizabethSamuel-MSFT: Please ensure that "an" is used before vowels but use "a" otherwise. In this case, it should be "and a" instead of "and an"

api-reference/beta/api/reportroot_skypeforbusinessactivityusercounts.md
kszb: The period parameter is string type.

api-reference/beta/api/reportroot_skypeforbusinessactivityuserdetail.md
kszb: Yes. I've already reviewed the field names before. This means the conference participated in.

ElizabethSamuel-MSFT: That's a concern because participanted is not a valid word and may confuse developers.

api-reference/beta/api/reportroot_office365activeusercounts.md
```diff
@@ -21,7 +21,7 @@ One of the following permissions is required to call this API. To learn more, in
 <!-- { "blockType": "ignored" } -->
 
 ```http
-GET /reports/Office365ActiveUserCounts(period='{period_value}')?$format=application/json
```
ElizabethSamuel-MSFT: per our meeting yesterday: you should keep the $format here as it's required at this point. When the function supports both csv and json then you can remove it from the request and update the instructions accordingly.

api-reference/beta/api/reportroot_office365activeuserdetail.md
```diff
@@ -36,7 +36,7 @@ In the request URL, provide the chosen query parameter with a valid value.
 
 > **Note:** You need to set either period or date in the URL.
 
-This method supports the `$top` and `$skipToken` [OData query parameters](../../../concepts/query_parameters.md) to customize the response.
+This method supports the `$format`, `$top` and `$skipToken` [OData query parameters](../../../concepts/query_parameters.md) to customize the response. `$format` must be set application/json value.
```
ElizabethSamuel-MSFT: Query parameters should be in its own section.

ElizabethSamuel-MSFT: To keep with the pattern of the rest of the content, you should insert a "Query parameters" section with this information after the HTTP request section.

api-reference/beta/resources/office365activeuserdetail.md
```diff
@@ -0,0 +1,16 @@
+# office365ActiveUserDetail resource type
```
ElizabethSamuel-MSFT: Please include a brief description of the resource

api-reference/beta/resources/office365activeuserdetail.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Properties
+
+| Property          | Type   |
```
ElizabethSamuel-MSFT: Known pending item: Add description for each property

api-reference/beta/resources/office365activeuserdetail.md
```diff
@@ -0,0 +1,16 @@
...
+| reportRefreshDate | Date   |
+| office365         | Int64  |
+| exchange          | Int64  |
+| oneDrive          | Int64  |
+| sharePoint        | Int64  |
+| skypeForBusiness  | Int64  |
+| yammer            | Int64  |
+| teams             | Int64  |
+| reportDate        | Date   |
+| reportPeriod      | String |
```
ElizabethSamuel-MSFT: Should add a JSON representation section for the resource. Please include the comment block before the json code block to enable validation.

api-reference/beta/resources/office365activeuserdetail.md
kszb: As said before, we always have one resource type for one function, so the description of resource type is almost the same with function, I don't know whether we need to add that. For some cases, which the same resource is used by several different functions, the meaning of resource would be different when they apply in different functions.

ElizabethSamuel-MSFT: But you can't assume that the user is coming from the function context, especially if you're including it in the TOC as that means the user may come upon this documentation but not the function with the description. You should still be able to give the gist of the resource's purpose in the brief description. Also, if the difference is that important between the different functions, then it might be worth calling out here.

api-reference/beta/resources/office365activeuserdetail.md
```diff
@@ -14,3 +14,43 @@
...
+
+<!-- {
+  "blockType": "response",
+  "truncated": true,
+  "@odata.type": "microsoft.graph.office365ActiveUserDetail"
+} -->
+
+```http
+{
+  "reportRefreshDate": "2017-09-01",
```
ElizabethSamuel-MSFT: The JSON representation should in general have the type of the property instead of a sample value. Please see here for a sample: https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/resources/timezoneinformation

api-reference/beta/api/reportroot_getoffice365activeuserdetail.md
```diff
@@ -46,7 +46,7 @@ This method supports the `$format`, `$top` and `$skipToken` [OData query paramet
 
 ## Response
 
-If successful, this method returns a `200 OK` response code and an **[office365ActiveUserDetail](../resources/office365activeuserdetail)** object in the response body.
+If successful, this method returns a `200 OK` response code and an **[office365ActiveUserDetail](../resources/office365activeuserdetail.md)** object in the response body.
```
ElizabethSamuel-MSFT: verify that referenced files are correct, particularly where you updated names

PR #1755, opened 2017-09-25
Fixed Links

mlafleur:
The change log had a sizable number of invalid links. It also had a lot of full URI links rather than references to the local .md file.


mlafleur:
Outdated

PR #1804, opened 2017-10-12
Corrected DELETE rejectedSenders example

mlafleur:
The example provided was incomplete and lead to developers getting confused over the proper syntax.


madhura-sahasrabudhe:
Can we also make the same change for the DELETE acceptedSenders API? Thanks!

mlafleur:
I've corrected the DELETE acceptedSenders as well.

PR #1822, opened 2017-10-13
Correcting User companyName property

mlafleur:
Documentation for the `User` object was inconstant around `companyName`. In some areas the property was not mentioned, in others it was mentioned by not noted as `read-only`. It was also incorrectly shown in the example payload for updating a `User`.


PatAltimore:
@mlafleur Thanks for the update.  Looks good.

PR #1830, opened 2017-10-15
Update user_list_owneddevices.md [Permission for native apps]

magickaichen:
Added the permission requirement for native apps, this api will not work until Directory.AccessAsUser.All is granted.


magickaichen:
Are you guys gonna review this simple PR?

magickaichen:
Thanks!

magickaichen:
So it's already February and no one wants to look at it....

PatAltimore:
@magickaichen Sorry for the delay in reviewing your contribution. I checked the permissions. For delegated, there is a mistake in the list of permissions.  From least to most privileged it would be Directory.Read.All, Directory.ReadWrite.All, Directory.AccessAsUser.All.

Directory.AccessAsUser.All isn't always required. You can use Directory.Read.All, or Directory.ReadWrite.All.

Therefore, the note wouldn't be accurate.  If you want to remove User.Read.All, User.ReadWrite.All from delegated and application permissions, I can accept that contribution.

Or you can close the PR and I can make the change if you like.

Thanks again,
Pat

magickaichen:
I removed incorrect permissions and also removed the note since it is not accurate anymore, let me know if I need to change anything else, thanks!

PR #1839, opened 2017-10-17
Clarify Event Type and Recurrence

mlafleur:
Noted that Event `type` is Read-only
Removed some enum duplication (same values listed twice)
Corrected spelling 
Added `recurrence: null` to the example payload for updating events. It wasn't clear this was an option, leading developers to mistakenly attempted to update the read-only `type` property.


angelgolfer-ms:
Thanks for your contribution, @mlafleur. Related to your comment about specifying explicitly the type property of event as read-only, I wonder if the table of properties in the PATCH event topic, which lists those properties that can be updated, is not helping customers?

PR #1840, opened 2017-10-17
People - Update examples for Get Person and List People

ElizabethSamuel-MSFT:


api-reference/v1.0/api/person_get.md
```diff
@@ -18,11 +18,13 @@ One of the following permissions is required to call this API. To learn more, in
 <!-- { "blockType": "ignored" } -->
 
 ```http
-GET /me/people/?$search='{property_value}'
-GET /me/people/?$filter={person_property} eq '{property_value}'
+GET /me/people/{id}
+GET /users/{id}/people/{id}
```
angelgolfer-ms: Missed this part in my original request:
In MS Graph, you can index into a user collection by the UPN as well.
So please change GET /users/{id}/people/{id} to GET /users/{id | userPrincipalName}/people/{id}, for both the v1 topic and beta topic.

api-reference/v1.0/api/person_get.md
```diff
@@ -18,11 +18,13 @@ One of the following permissions is required to call this API. To learn more, in
...
 ```
 
 ## Optional query parameters
+This method supports the [OData Query Parameters](../../../concepts/query_parameters.md) to help customize the response.
+
```
angelgolfer-ms: Per item (c) in my original request, please add an intro sentence before the optional query parameters table to point customers to the concept topic for examples using the different query parameters.

api-reference/v1.0/api/person_get.md
```diff
@@ -48,6 +44,42 @@ Do not supply a request body for this method.
 ## Response
 If successful, this method returns a `200 OK` response code and a [person](../resources/person.md) object in the response body. The response can contain one person instance or a collection of person instances. 
 ## Examples
+### Get a person given the ID
```
angelgolfer-ms: This comment is about the use of H3's in the Examples section. Please remove the H3s, make sure the example request introductory sentence describes what the example does, and follow the same sub-headings in other reference topics that have multiple examples. 
I think the use of H3 with an action-based section title gives a misleading impression that the section is conceptual. In a reference topic, the example section should be tighter than a conceptual how-to. The lead-in sentence should be concise and should not bear the responsibility of explaining concepts. If there are concepts to explain, then that should go into a how-to conceptual topic.

api-reference/v1.0/api/person_get.md
```diff
@@ -48,6 +44,42 @@ Do not supply a request body for this method.
 ## Response
 If successful, this method returns a `200 OK` response code and a [person](../resources/person.md) object in the response body. The response can contain one person instance or a collection of person instances.
```
angelgolfer-ms: Remove the second sentence "The response can contain one person instance..."
We fixed this topic to be about GET /me/people/{id}, so it always returns a single instance of the person type.

api-reference/v1.0/api/person_get.md
```diff
@@ -48,6 +44,42 @@ Do not supply a request body for this method.
...
 ## Examples
+### Get a person given the ID
+The following is an example of the request that gets the person who has this ID in the user's organization. 
+
+<!-- {
+  "blockType": "request",
+  "name": "get_person_by_id"
+}-->
+```http
+GET https://graph.microsoft.com/beta/me/people/id-value
```
angelgolfer-ms: Can you provide a real GET request and response without all that pseudo data? You can try in the Graph Explorer or get a demo tenant and try it in Fiddler.
The pseudo examples are there when contributors use the topic generation tool and don't further customize the examples. While the pseudo examples are better than none, it's much more helpful for customers to see examples with real values.

api-reference/v1.0/api/person_get.md
```diff
@@ -18,11 +18,13 @@ One of the following permissions is required to call this API. To learn more, in
...
 
 ```http
-GET /me/people/?$search='{property_value}'
-GET /me/people/?$filter={person_property} eq '{property_value}'
+GET /me/people/{id}
+GET /users/{id}/people/{id}
 ```
 
 ## Optional query parameters
+This method supports the [OData Query Parameters](../../../concepts/query_parameters.md) to help customize the response.
```
angelgolfer-ms: Now that we have corrected the appropriate syntax for the GET /people/{id} operation, actually, only $select in that table of query parameters makes sense. Even if we include those in the "OData Query Parameters" topic, only $format and $select are applicable to a single instance of a resource. Please remove that lead-in sentence and just list the 2 query parameters.

api-reference/v1.0/api/person_get.md
```diff
@@ -48,6 +44,42 @@ Do not supply a request body for this method.
...
+  "givenName": "givenName-value",
+  "id": "id-value",
+  "surname": "surname-value",
+  "birthday": "birthday-value",
+  "personNotes": "personNotes-value",
+  "isFavorite": true
+}
+```
+
 ### Perform a search
```
angelgolfer-ms: Please remove the examples "Perform a search" and "Select the fields to return in a filtered response". They query on the people nav property and don't index on it to return a single instance of person.
You can however, show a second example of GET /people/{id}?$select={property name}, by specifying real values for {id} and {property name}, and showing a response with real values (not pseudo values). Again, try the query on either Graph Explorer or Fiddler, and copy/paste the request and response to as a second example.

api-reference/beta/api/person_get.md
```diff
@@ -41,6 +41,42 @@ Do not supply a request body for this method.
 
 If successful, this method returns a `200 OK` response code and [person](../resources/person.md) object in the response body.
```
angelgolfer-ms: Please apply the same comments for the HTTP request and Optional query parameters sections I added below for the v1 version of person_get.md, to this beta version as well.

api-reference/beta/api/person_get.md
```diff
@@ -41,6 +41,42 @@ Do not supply a request body for this method.
...
 ## Examples
+### Get a person given the ID
```
angelgolfer-ms: Please remove the H3 heading under the Examples section, as detailed in my comments for the v1 version of person_get.md below. Please also provide a request and response with real values.

api-reference/beta/api/person_get.md
```diff
@@ -41,6 +41,42 @@ Do not supply a request body for this method.
...
+  "givenName": "givenName-value",
+  "id": "id-value",
+  "surname": "surname-value",
+  "birthday": "birthday-value",
+  "personNotes": "personNotes-value",
+  "isFavorite": true
+}
+```
+
 ### Browse
```
angelgolfer-ms: Because this topic is about GET /people/{id}, the examples of using optional query parameters without indexing into a people collection don't really apply. 
I suggest copying these examples over to the user_list_people.md topic, following the example that has no query parameters (that you added below).

api-reference/v1.0/api/person_get.md
```diff
@@ -18,25 +18,12 @@ One of the following permissions is required to call this API. To learn more, in
...
 
 ```http
-GET /me/people/?$search='{property_value}'
-GET /me/people/?$filter={person_property} eq '{property_value}'
+GET /me/people/{id}
+GET /users/{id | userPrincipalName}/people/{id}
 ```
 
 ## Optional query parameters
-|Name|Value|Description|
```
angelgolfer-ms: Hi Elizabeth, you might have misunderstood the notes in the tracking working item. Even though the original syntax section listed only using $search and $filter, GET person does support these 2 plus the other query parameters in the existing table. So while we remove the query parameters in the syntax section, the existing table in the Optional query parameters section should remain. 
On the other hand, because the syntax section no longer shows the property_value and person_property parameters, we should remove the Parameters section, as you have done here.

api-reference/v1.0/api/person_get.md
```diff
@@ -18,25 +18,12 @@ One of the following permissions is required to call this API. To learn more, in
...
-|$select|string|Comma-separated list of properties to include in the response. For optimal performance, only select the subset of properties needed.|
-|$skip|int|Skip the first n results, useful for paging. This is not supported when using *$search*.|
-|$top|int|Number of results to be returned.|
-
-## Parameters
-| Parameter |Type       |Description|
-|:----------|:----------|:----------|
-|property_value|String     |The value of the extended property to match. Required where listed in the **HTTP request** section.|
-|person_property|String    |The person property to match. Required where listed in the **HTTP request** section.|
+This method supports the `$format` and `$select` [OData Query Parameters](../../../concepts/query_parameters.md) to help customize the response.
```
angelgolfer-ms: Please rephrase this lead-in sentence to go in the Optional query parameters, preceding the table, and point customers to the concept topic in https://developer.microsoft.com/en-us/graph/docs/concepts/people_example.
Not to the general query parameters topic.

api-reference/v1.0/api/person_get.md
```diff
@@ -45,126 +32,99 @@ GET /me/people/?$filter={person_property} eq '{property_value}'
...
-           "websites": [],
-           "personType": [
-               {
-                    "class": "Person",
-                    "subclass": "OrganizationUser"
-                }
-            ]
-       }
-   ]
+    "id": "33b43a5b-87d6-41ec-91f8-a2610048105f",
```
angelgolfer-ms: I'd suggest showing a more typical case that represents a person and not an O365 group? The primary scenario we describe in our docs are "people who are relevant to the user".

api-reference/v1.0/api/person_get.md
```diff
@@ -45,126 +32,99 @@ GET /me/people/?$filter={person_property} eq '{property_value}'
...
 Do not supply a request body for this method.
+
 ## Response
-If successful, this method returns a `200 OK` response code and a [person](../resources/person.md) object in the response body. The response can contain one person instance or a collection of person instances. 
+If successful, this method returns a `200 OK` response code and a [person](../resources/person.md) object in the response body.
+
 ## Examples
-### Perform a search 
-The following request does a search for a person named Irene McGowan. 
+#### Request
```
angelgolfer-ms: Please label this "Request 1" and "Response 1", "Request 2" and "Response 2", for consistency with other topics that have multiple examples.

api-reference/v1.0/api/person_get.md
```diff
@@ -45,126 +32,99 @@ GET /me/people/?$filter={person_property} eq '{property_value}'
...
-
 ```http
-GET https://graph.microsoft.com/v1.0/me/people/?$search="Irene McGowan"
+GET https://graph.microsoft.com/v1.0/me/people/33b43a5b-87d6-41ec-91f8-a2610048105f
 ```
+#### Response
 
-The following example shows the response. 
+>**Note:** The response object shown here might be shortened for readability. All the properties will be returned from an actual call.
```
angelgolfer-ms: I am aware that Laura might have instructed you to put this as a note. I'll let you both work this out. FYI my concern is this note is currently auto-inserted and we know the second sentence is not necessarily true, or necessarily true over time. So calling this boilerplate sentence as a note takes this one step further in terms of alerting customers, and I am not sure this is a good move at this point of time before we have a plan in place to make sure the sentence always applies.

api-reference/v1.0/api/person_get.md
```diff
@@ -45,126 +32,99 @@ GET /me/people/?$filter={person_property} eq '{property_value}'
...
-            "scoredEmailAddresses": [
-                {
-                    "address": "Lorrief@contoso.onmicrosoft.com",
-                    "relevanceScore": 8.0
-                }
-            ]
-        }
-    ]
+    "id": "33b43a5b-87d6-41ec-91f8-a2610048105f",
+    "displayName": "Marketing"
```
angelgolfer-ms: Still suggest showing a person rather than a group in the response. I think if we want to show the capability of returning a group (as well as persons) in a response, we can do that in the examples for List people, which would show multiple person objects in a response.

api-reference/beta/api/person_get.md
```diff
@@ -17,17 +17,10 @@ One of the following permissions is required to call this API. To learn more, in
 <!-- { "blockType": "ignored" } -->
 ```http
 GET /me/people/{id}
-GET /users/{id}/people/{id}
+GET /users/{id | userPrincipalName}/people/{id}
 ```
 ## Optional query parameters
```
angelgolfer-ms: Same comments I entered for the v1 person_get.md topic apply here as well. Thanks.

api-reference/beta/api/person_get.md
```diff
@@ -38,127 +31,100 @@ GET /users/{id}/people/{id}
...
-```
-
-#### Selecting the fields to return in a filtered response
-
-You can combine the *$select* and *$filter* parameters to create a custom list of people relevant to the user and get only the fields that your application needs. 
-
-The following example gets the *DisplayName* and *EmailAddress* of people whose display name equals the specified name. In this example, only people whose display name equals "Nestor Kellum" are returned. 
-
+#### Request
+The following is an example of the request that gets the person who has this ID in the user's organization and restricts the response to the selected properties.
```
angelgolfer-ms: Specially calling this out - just like in the v1.0 topic, it might be better to show an example that returns a person rather than an O365 group, even though returning an O365 group is a valid response.

api-reference/beta/api/user_list_people.md
```diff
@@ -16,11 +16,11 @@ One of the following permissions is required to call this API. To learn more, in
 ## HTTP request
 <!-- { "blockType": "ignored" } -->
 ```http
+GET /me/people
 GET /users/{id | userPrincipalName}/people
 ```
 ## Optional query parameters
```
angelgolfer-ms: As commented for the person_get.md topics, please keep the original table in the "Optional query parameters" section that lists all the supported query parameters. The link should go to the people API concept topic (https://developer.microsoft.com/en-us/graph/docs/concepts/people_example), not the general query parameters topic, since querying is what people API is about. Please do this for both the v1 and beta List people topics.

api-reference/v1.0/api/user_list_people.md
```diff
@@ -21,15 +19,9 @@ One of the following permissions is required to call this API. To learn more, in
 GET /me/people
 GET /users/{id | userPrincipalName}/people
 ```
+
 ## Optional query parameters
-|Name|Value|Description|
```
angelgolfer-ms: Please keep this table and have the lead-in sentence point to the people API concept topic instead.

PR #1939, opened 2017-10-30
Fixing links

mlafleur:
Corrected some incorrect anchor links and cleaned up formatting a bit


Lauragra:
Hi Marc, 
Is this primarily a cosmetic fix, or was something broken in this topic? Can you please check the Appveyor report - looks like we have a broken link. Thanks for helping out!

mlafleur:
I was able to fix that link. 

There is currently an issue with the [search](https://developer.microsoft.com/en-us/graph/docs/concepts/query_parameters#search) anchor link. After digging a little be deeper, it looks like it only occurs in production. The markdown generation results in `<h2 id="search">search</h2>` which conflicts with the Search button in navigation (`<button id="search" aria-label="Search" class="c-glyph" ms.title='Search' data-bi-dnt="true"></button>`).

Lauragra:
Thanks, Marc. I can repro the issue; I'll file a bug. We had a similar issue with Examples H2s; the fix was to update the ID for the navigation element.

PR #2027, opened 2017-12-05
fixed wrong link

khalyomede:
OData Query Parameters link was pointing to the wrong page (to /concepts/people_example.md).


Lauragra:
Good catch, thanks!

PR #2047, opened 2017-12-08
Intune Graph API updates for 1712

BrucePerlerMS:
Changelog and TOC (SEOKeys-en-us) are updated with changes from Josh. Includes all generated API ref pages in the beta.


BrucePerlerMS:
After chatting again with Lance yesterday, I am using all absolute paths in the changelog updates. I still think that relative paths should work but, would rather have it work than be right. :)

PR #2053, opened 2017-12-11
Update terms-of-use.md

whjensen:
Added "service" in clause 3.c after Intune for grammatical correctness.

### DIFF
--- a/misc/terms-of-use.md
+++ b/misc/terms-of-use.md
@@ -72,7 +72,7 @@ You may NOT:
 
 	5. Violating applicable privacy laws and regulations.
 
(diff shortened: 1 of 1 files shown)


misc/terms-of-use.md
```diff
@@ -72,7 +72,7 @@ You may NOT:
 
 	5. Violating applicable privacy laws and regulations.
 
-c) **Accessing the Microsoft Intune through Graph API**
+c) **Accessing the Microsoft Intune Service through Graph API**
```
Lauragra: Please change "Graph API" to "the Microsoft Graph API".

whjensen: Done.

PR #2082, opened 2017-12-20
Test rebuild API ref by ...

BrucePerlerMS:
deleting all intune_* in resources and api, then copy in just the 1712 intune_* for resources and api from 1712 drop. This should create a build that only contains current api elements and will turn up flaws in the TOC: stale links, bad links due to name changes etc. 

This is a test to find flaws in the TOC through accumulation of errors over time. Will also unclutter API ref / resources published, removing unused / outdated files.


BrucePerlerMS:
#hold-off
This is a test and will not be merged.

PR #2083, opened 2017-12-20
test for clean build with name change

BrucePerlerMS:
replace all instances of _devicefe_ with _devices_


BrucePerlerMS:
#hold-off
This is only a test and will not be merged.

BrucePerlerMS:
Starting again after reviewing this process with Laura G. - targeting stage rather than master

PR #2090, opened 2017-12-22
Johnau msgdelta

JohnAustin-MSFT:
Added a deleted message to the incremental sync example response


concepts/delta_query_messages.md
```diff
@@ -92,7 +92,7 @@ The following example shows a series of 3 requests to synchronize a specific fol
 - [Sample second request](#sample-second-request) and [response](#sample-second-response)
 - [Sample third request](#sample-third-request) and [final response](#sample-third-and-final-response)
 
-See also what you'll do in the [next round](#the-next-round).
+See also what you'll do in the [Sample incremental mail folder synchronization request](#synchronize-the-same-folder-in-the-next-round).
```
angelgolfer-ms: Thanks for the addition of a second round of sync, John.
To better incorporate the additional round, we should clean up the lead-in sentence of this section a bit. Would you like me to do it after you commit and merge the other feedback (further below), or would you like to take my feedback and incorporate it at the same time?
Anyway, I think the following would describe the example in a more holistic way:
## Example to synchronize messages in a folder
The following example shows 2 rounds of synchronization of a specific folder which initially contains 5 messages. 
The first round involves a series of 3 requests to synchronize all 5 messages in the folder:
- (the 3 existing bullet points)...
After the first round, one of the messages is deleted, and another message is updated (John, please describe how, I couldn't tell from your example what actually changed in that message).
The [second round](#the-next-round) of synchronization returns exactly the delta (the deletion and update), without returning the other messages that have remained the same.

concepts/delta_query_messages.md
```diff
@@ -189,6 +189,28 @@ more messages to get from the folder.
     "@odata.context":"https://graph.microsoft.com/v1.0/$metadata#Collection(message)",
     "@odata.nextLink":"https://graph.microsoft.com/v1.0/me/mailfolders('AQMkADNkNAAAgEMAAAA')/messages/delta?$skiptoken=GwcBoTmPKILK4jLH7mAd1lLU",
     "value":[
+        {
+            "@odata.type": "#microsoft.graph.message",
+            "subject": "Microsoft Virtual Academy at Contoso",
```
angelgolfer-ms: Please include the @odata.etag for this message. Also the id property.

concepts/delta_query_messages.md
```diff
@@ -249,23 +272,38 @@ synchronization is complete for the time being for this folder. Save and use the
...
         {
-            "@odata.type":"#microsoft.graph.message",
-            "@odata.etag":"W/\"CQAAABYAAAARn2vdzPFjSbaPPxzjlzOTAAAEfYB8\"",
-            "subject":"You've joined the Customer Manager group",
-            "sender":{
-                "emailAddress":{
-                    "name":"Customer Managers team",
-                    "address":"customer_managers@contoso.onmicrosoft.com"
+            "@odata.type": "#microsoft.graph.message",
+            "subject": "Fabric CDN now available",
```
angelgolfer-ms: Please include the @odata.etag for this message.

concepts/delta_query_messages.md
```diff
@@ -249,23 +272,38 @@ synchronization is complete for the time being for this folder. Save and use the
...
+            "bccRecipients": [],
+            "replyTo": [],
+            "id": "AAMkADk0MGFkODE3LWEAAA="
+        }]
 }
 ```
 
 
-### The next round
+### Synchronize the same folder in the next round
```
angelgolfer-ms: Please rephrase the section title as "Synchronize messages in the same folder in the next round", just to be unambiguous. In general, "synchronize a folder" can mean sync-ing any property changes on the folder instance. In this case, we are specifically synchronizing the messages contained in a folder.

concepts/delta_query_messages.md
```diff
@@ -280,7 +318,62 @@ GET https://graph.microsoft.com/v1.0/me/mailfolders('AQMkADNkNAAAgEMAAAA')/messa
 Prefer: odata.maxpagesize=2
 ```
 
+### Sample incremental mail folder synchronization response
```
angelgolfer-ms: I think we can remove this section heading and present the additional example simply with a request and response. (I also updated the link above to this section to reflect the new section title.)

concepts/delta_query_messages.md
```diff
@@ -280,7 +318,62 @@ GET https://graph.microsoft.com/v1.0/me/mailfolders('AQMkADNkNAAAgEMAAAA')/messa
...
+
+The response contains a `deltaLink`. This indicates that all changes in the remote mail folder are now synchronized. One message was deleted and the other message was changed.
 
+>**Note:** The Microsoft Graph API v1.0 does not support all of the message properties that a user can change using Microsoft Outlook. In this example, a message's follow up status is not supported in the API. However, the changed message still appears in the response to a synchronization request.
```
angelgolfer-ms: Can we show a simpler example where the change to the message can be reflected in the properties that are returned in the GET /messages/delta call?
As for this case where a message is changed in the UI but the message in REST API doesn't reflect the change, I recall you talked about it. Did you have Sri confirming the expected behavior and what they plan to do if anything about this kind of cases? We should do that, before we surface that as a note just so the docs are aligned with the product team's plan.

PR #2120, opened 2018-01-09
Jan Rostering update and gitignore.

garethj-msft:
ChangeLog update for Jan Education rostering. Don't merge until main doc changes for Jan Rostering have been merged.

Also added a generic gitignore to the project so it can more easily be edited in VS.


api-reference/beta/api/educationschool_get_administrativeUnit.md
```diff
@@ -0,0 +1,74 @@
+# Get administrativeUnit
```
Lauragra: Please update the filename to use all lowercase (lowercase the U). Mixed case filenames result in broken URLs on our site. Thanks!

PR #2165, opened 2018-01-24
Updated correct permission policy for the call

wobba:
Renewing a groups expiration life cycle policy requires either `Directory.ReadWrite.All` or `Group.ReadWrite.All`.

Renewing also works fine using app-only, not only using delegated.


yinaa:
Thank you for the contribution. Document was already updated by author.

wobba:
Good stuff :)

PR #2236, opened 2018-02-12
Remove administrativeUnit and related topic and references from v1.0

JohnAustin-MSFT:
administrativeUnit has not been promoted to v1.0 and attempts to access it in the beta endpoint from the v1.0 endpoint will fail. References to the resource, the get_administrativeunit topic, and references to both of these entities are removed from the v.10 docs with this PR


garethj-msft:
LGTM

PR #2254, opened 2018-02-20
Minor edits to invitation topic.

Lauragra:


api-reference/beta/resources/groups-overview.md
```diff
@@ -107,10 +107,6 @@ To learn more about formulating membershipRules, see [Create attribute-based rul
 
 > **Note**: Dynamic membership rules requires the tenant to have a license at tier [Azure Active Directory Premium P1](https://azure.microsoft.com/en-us/pricing/details/active-directory/) or greater.
 
-## Other types of groups
```
angelgolfer-ms: I'm actually communicating with Pat and looping in Dan on this one, so please hold off removing this paragraph. It's basically the GET groups operation would include groups in Yammer, but that's all there is to the support for Yammer. There isn't much more customers can do with such groups, as the integration with Yammer is very partial. This section is currently pointing out one of the holes with Yammer groups, and that is,  even after you get such a group, you can't use the group conversations API to access conversations in a Yammer group.

api-reference/beta/resources/identityriskevent.md
```diff
@@ -3,13 +3,15 @@
...
-Complete information about risk events can be found in the [Azure AD Identity Protection documentation](https://azure.microsoft.com/en-us/documentation/articles/active-directory-identityprotection-risk-events-types/).
+
+* [Sign-ins from anonymous IP addresses](anonymousipriskevent.md)
+* [Sign-ins from malware-infected devices](malwareriskevent.md)
+* [Impossible travel to atypical locations](impossibletravelriskevent.md)
+* [Users with leaked credentials](leakedcredentialsriskevent.md)
+* [Sign-ins from suspicious IP addresses](suspiciousipriskevent.md)
+* [Sign-ins from unfamiliar locations](unfamiliarlocationriskevent.md)
+
+For more information about risk events, see the [Azure AD Identity Protection documentation](https://azure.microsoft.com/en-us/documentation/articles/active-directory-identityprotection-risk-events-types/).
```
angelgolfer-ms: The standard lead-in sentence to link out is exactly what Dan, Jackson, and I found not working in contexts like this. Maybe you can take a look at the guidance I wrote up in the CDK draft topic for concept topics and see if I got the message across successfully. (I'm not including a link on a public site but I assume you know how to look for the "Guidance for Microsoft Graph concept topics" article.)

api-reference/beta/resources/invitation.md
```diff
@@ -1,15 +1,15 @@
...
 The invitation process uses the following flow:
 
-* An invitation is created
-* An invitation is sent to the invited user (containing an invitation link)
-* The invited user clicks on the invitation link, signs in and redeems the invitation and creation of the user entity representing the invited user completes
-* The user is redirected to a specific page after redemption completes
+* An invitation is created.
+* An invitation is sent to the invited user (including an invitation link).
+* The invited user clicks the invitation link, signs in, and redeems the invitation. The user entity that represents the invited user is created.
+* The user is redirected to a specific page when redemption is finished.
```
angelgolfer-ms: I think there is technically one point that is different from the original and your rewrite - in the last 2 items (btw, should that list be a numbered list): the creation of the user entity marks the completion of the redemption. From the developer's point of view, that's a different point of time in the process than the user being redirected to another page in the user interface.
I don't know what significance there is with keeping this, but if I were to change something of this scope, I would consult with the owner. I never know, there might be some additional context or use cases about redemption.

api-reference/v1.0/resources/groups-overview.md
```diff
@@ -107,10 +107,6 @@ To learn more about formulating membershipRules, see [Create attribute-based rul
 
 > **Note**: Dynamic membership rules requires the tenant to have a license at tier [Azure Active Directory Premium P1](https://azure.microsoft.com/en-us/pricing/details/active-directory/) or greater.
 
-## Other types of groups
```
angelgolfer-ms: Same comment as in the beta version of this topic. Please hold off. I'm seeking confirmation for a rewrite of this section.

api-reference/v1.0/resources/invitation.md
```diff
@@ -1,13 +1,13 @@
...
 The invitation process uses the following flow:
 
-* An invitation is created
-* An invitation is sent to the invited user (containing an invitation link)
-* The invited user clicks on the invitation link, signs in and redeems the invitation and creation of the user entity representing the invited user completes
-* The user is redirected to a specific page after redemption completes
+* An invitation is created.
+* An invitation is sent to the invited user (including an invitation link).
+* The invited user clicks the invitation link, signs in, and redeems the invitation. The user entity that represents the invited user is created.
+* The user is redirected to a specific page when redemption finishes.
```
angelgolfer-ms: Same comments as in the beta version of this topic above.

api-reference/beta/resources/identityriskevent.md
Lauragra: Can you propose what context you want to set regarding this link? Seems like it should come from a SME because they best understand the context of what they're linking to an why. Who is the writer owner for this topic now - is it Lance?

api-reference/beta/resources/invitation.md
Lauragra: So would like the writer owner to make the call and update as needed. Originally I put Pat here as a reviewer but now I'm not sure....

api-reference/beta/resources/identityriskevent.md
dkershaw10: @mmcla Can you provide some context lead-in sentence here please?

api-reference/beta/resources/identityriskevent.md
```diff
@@ -39,7 +41,7 @@ Complete information about risk events can be found in the [Azure AD Identity Pr
 
 ## JSON representation
 
-Here is a JSON representation of the resource.
+The following is a JSON representation of the resource.
```
dkershaw10: So personally I think either sentence is redundant and should be removed from all topics.  What value does it add, to the section title?  Honestly?

api-reference/beta/resources/invitation.md
dkershaw10: I can provide guidance here. So...

1. Numbered list seems appropriate
2. For the second, generically, an invitation is sent to a user - either it's an email generated by our system, or  the client app sends a communication itself to the user (email, IM, WhatsApp, carrier pigeon, etc).  In all cases it **must** contain the invitation link.  Anyways, not really fussed on containing vs including, but not really sure why the change.
3. The third item is tricky.  POST /invitations actually has a side effect of creating the user entity.  So strictly speaking it exists after item one.  After item one, the app could actually start manipulating the created user (adding it to groups, assigning it to apps, assigning licenses etc).  BUT, it's only after the invited user redeems their invitation and signs in that the user entity is a "fully fledged" user entity that allows that user to sign in to this tenant.  I'm not sure how you want to capture this.

api-reference/beta/resources/invitation.md
```diff
@@ -27,17 +27,17 @@ Creating an invitation will return a redemption URL in the response (*inviteRede
 |invitedUserMessageInfo|[invitedUserMessageInfo](invitedusermessageinfo.md)|Additional configuration for the message being sent to the invited user, including customizing message text, language and cc recipient list.|
 |sendInvitationMessage|Boolean|Indicates whether an email should be sent to the user being invited or not. The default is false.|
 |inviteRedirectUrl|String|The URL user should be redirected to once the invitation is redeemed. Required.|
-|inviteRedeemUrl|String|The URL user can use to redeem his invitation. Read-Only|
-|invitedUserType|String|The userType of the user being invited. By default, this is Guest. You can invite as Member if you're are company administrator. |
-|status|String|The status of the invitation. Possible values: PendingAcceptance, Completed, InProgress, and Error|
+|inviteRedeemUrl|String|The URL user can use to redeem his invitation. Read-only.|
```
dkershaw10: Shouldn't we scan this for gender-neutral tone?  Shouldn't it say "redeem **their** license"?

api-reference/v1.0/resources/invitation.md
```diff
@@ -1,13 +1,13 @@
-# invitation manager
+# invitation resource type
 
-Use the invitation manager to create an invite, in order to add an external user to the organization. 
+Use the invitation resource to create an invitation, in order to add an external user to the organization.
```
dkershaw10: Keep it simple - can we remove the "in order"?

dkershaw10:
@Lauragra Are you going to make the updates here based on my feedback?  This has been in this state for too long.  Can we just get this pushed?

PR #2260, opened 2018-02-21
Clarified known bug with certain scopes

mlafleur:
There have been several cases of developers failing to catch the note regarding a known bug above the permissions table. In most cases the only hint that there was something unique about certain scopes was that one or more of them were _italicized_ (this was also inconsistent with some pages using **bold**). Since the result of this bug is that these scopes cannot be used, emphasizing them could result in implying they are preferred rather than that there is an issue with them. In order to clarify, I have replaced the _italic_/**bold** formatting with ~~strike-through~~ and moved the note bellow the table.


jasonjoh:
@Jackson-Woods I believe you added the note about the current issue with these scopes. It looks like the change here is to mark the ones that currently don't work with strikethrough rather than italics. If you're good with it we can merge.

PR #2282, opened 2018-03-02
Additional properties to fileservice beta

kevklam:
@daspek  to update with Activities docs


api-reference/beta/resources/sharepointids.md
```diff
@@ -29,6 +29,7 @@ Here is a JSON representation of the resource
     "listItemUniqueId": "string",
     "siteId": "string",
     "siteUrl": "url",
+    "tenantId": string",
```
daspek: probably my typo, but `string` needs to be quoted

PR #2300, opened 2018-03-08
passbook to wallet

barlanmsft:
passbook was renamed to wallet in 2015; unsure how that would impact "lockScreenBlockPassbook" but at least for the description we can update


lleonard-msft:
#sign-off

Note: these docs are automatically generated from the Graph API metadata schema.  Content changes may be overritten unless they're also made upstream.  I'm handling this case offline and also taking a backlog item to figure out how we can ensure such changes persist when the pages are re-generated for a new drop.

PR #2339, opened 2018-03-21
Added TOU APIs.

Lauragra:


ntonyho:
In general, the indentation for the entire xml is off
Consider fixing them all in the future. Tools like visual studio can help

PR #2412, opened 2018-04-07
Update group_update.md

guillaumemeyer:
The "visibility" property type cannot be "boolean" if possible values are: **Private**, **Public**...


jasonjoh:
Hmm. Apparently my edit tipped the threshold for requiring a CLA, so reverting. I'll make a separate PR for my changes.

PR #2443, opened 2018-04-12
New conceptual

KundanaP:
Pull request for adding EDU conceptual doc.


concepts/education-concept-overview.md
```diff
@@ -1,11 +1,79 @@
...
+
+The following are some of the scenarios that the roster APIs enable:
+
+- [Get all classes in a school](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationschool_list_classes)
+- [Get all users in a class](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationclass_list_members)
+- [Get all the classes I teach](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationuser_list_classes)
+
+
+### Enable your teachers to create assignments at the click of a buttion using assignments APIs (preview)
+Microsoft Teams in Office 365 is a digital hub that brings conversations, content, and apps together in one place for classrooms. Microsoft Teams provides a [rich set of extensbility points](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/apps/apps-overview), including creating Tabs, Connectors, Bots, and APIs in Microsoft Graph to work with assignments and submissions.
```
angelgolfer-ms: I think we should start by pointing out both Teams and EDU API let developers automate creating assignments and making submissions. It's like a topic sentence for this section. Otherwise it's kind of strange that in the same section, we start with Teams (which is not even EDU API), and without any transition, we just jump over to EDU API.

concepts/education-concept-overview.md
```diff
@@ -1,11 +1,79 @@
-# Microsoft Graph education API overview
+# Education API overview
```
angelgolfer-ms: @Lauragra , please don't modify these service-level overviews on a per topic basis. Yina commented there should be some consistency among these topics. I also incorporated our earlier discussions about how to best differentiate these titles from their API-level overview counterparts, and also from possibly other topics (e.g., where applicable, end-user or admin-based topics of the similar subjects), so that these service-level overviews have a chance being surfaced higher up in search.

Lauragra: This edit does reflect what my understanding of our agreed-upon topic titling convention was. I'm following up with you separately in mail.

concepts/education-concept-overview.md
```diff
@@ -1,11 +1,79 @@
...
+
+The following are some of the scenarios that the roster APIs enable:
+
+- [Get all classes in a school](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationschool_list_classes)
+- [Get all users in a class](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationclass_list_members)
+- [Get all the classes I teach](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationuser_list_classes)
+
+
+### Enable your teachers to create assignments at the click of a buttion using assignments APIs (preview)
```
garethj-msft: typo button

garethj-msft: Typo button

concepts/education-concept-overview.md
garethj-msft: It's unfortunate that we haven't made a business decision whether assignments API is coupled to Teams or not. But we haven't so this needs to stay ambiguous.  Maybe being explicit that Teams has a generic API for all scenarios and then the assignments-specific pieces which are dedicated to EDU?

concepts/education-concept-overview.md
```diff
@@ -1,11 +1,79 @@
...
+# Education API overview
 
-{1-2 sentences about the value of this feature/service to its target customers.} 
+The education API in Microsoft Graph enhances Office 365 resources with information that is relevant for education scenarios, including information about schools, classes, users (students and teachers), assignments, and submissions. This makes it easy for you to build solutions that integrate with educational resources for various school and classroom scenarios.
 
-## Why build education solutions on Microsoft Graph?
+The education API in Microsoft Graph provides access to classes, schools, users, assignments, submissions and more.
 
-{<=1.5 pages of value proposition for developers.}
+![EDU Graph Overview](images/EDUGraph.PNG)
```
garethj-msft: Did Kundana update this to be accurate?

concepts/education-concept-overview.md
```diff
@@ -1,11 +1,79 @@
...
+
+With assigment APIs your app can interact with the assignment service outside of Teams. Teams will handle distribution, due dates and grading while your system can provide a rich learning experience to students.
+Here are examples of a few scenarios enabled by assignments API:
+
+- [Add an Assignment that links to your application](https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/api/educationclass_post_assignments) . 
+- [Assign grades to individual students for assignments linked to your application](https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/api/educationsubmission_update).
+- [Create a student dashboard to show which assigments are due by when](https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/api/educationclass_list_assignments).
+
+
+## Enable school ITs to manage Identity and Roster sync using School Data Sync Management (preview)
```
garethj-msft: Can we say IT administrators rather that "ITs"?

concepts/education-concept-overview.md
garethj-msft: Yes, she did.

concepts/education-concept-overview.md
angelgolfer-ms: @garethj-msft , I perceive your last comment says in Teams API, there's "assignment-specific pieces which are dedicated to EDU". Can you confirm if bringing up Teams Tabs, Connectors, Bots, is relevant to EDU then? They could, but just not dedicated, is that right? Or, if they are not relevant, then we shouldn't talk about it in this topic. :-)

concepts/education-concept-overview.md
angelgolfer-ms: This is to provide the necessary context to unbranded services and features. We can carry on offline.

concepts/education-concept-overview.md
angelgolfer-ms: Right.

concepts/education-concept-overview.md
```diff
@@ -1,11 +1,79 @@
...
+
+- [Get all classes in a school](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationschool_list_classes)
+- [Get all users in a class](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationclass_list_members)
+- [Get all the classes I teach](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/educationuser_list_classes)
+
+
+### Enable your teachers to create assignments at the click of a buttion using assignments APIs (preview)
+Microsoft Teams in Office 365 is a digital hub that brings conversations, content, and apps together in one place for classrooms. Microsoft Teams provides a [rich set of extensbility points](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/apps/apps-overview), including creating Tabs, Connectors, Bots, and APIs in Microsoft Graph to work with assignments and submissions. 
+
+![Microsoft Teams assignments](images/AssignmentsInTeams.PNG)
```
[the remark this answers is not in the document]
angelgolfer-ms: @JohnAustin-MSFT , btw, can you please also scale down the screen shot a bit, or show only a portion (with the relevant UI) of that screen, so that the UI is still discernible.

PR #2444, opened 2018-04-12
Add legalAgeGroup, ageGroup and consentProvidedForMinor attributes

BenV-MSFT:


api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
+# legalAgeGroupClassification, ageGroup and consentProvidedForMinor
```
angelgolfer-ms: I suggest the title "Categorize users by age and consent". @dkershaw10 or others, you probably have more insight and can tweak this title further. Just so you get the idea - as a concept topic, the title needs to use spoken English and not programmability terms. You should also phrase the title based on the purpose of these properties - why devs would use them. The title should let developers key into the feature, doesn't have to represent the complete feature.

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
+# legalAgeGroupClassification, ageGroup and consentProvidedForMinor 
+
+These three properties are used by Azure AD Tenant Administrators & Application Developers to help ensure the user of an account is handled correctly based on the regulatory rules governing the user's country.
+
+The ageGroup and consentProvidedForMinor are set by the administrative owner of the account. The property can not be read outside of the tenant because each tenant is generally consider a seperate regulatory boundary.
+
+The legalAgeGroupClassification is used by application developers to restrict features and capabilities as required. This property is only available to applications in the Home Tenant of the user and Microsoft. 
+
+## legalAgeGroupClassification
```
angelgolfer-ms: You already listed the long property names in the first section, so that meets the SEO needs. For these subsections, I suggest using friendly spoken English terms to express each property. We did SEO experiments before in the O365 REST API doc set. Once we changed from exact API names to friendly English terms in the right-side TOC, the page clicks increased right away.

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+
+This property is used by applications to ensure the correct handling of a user based on their Legal Age Group. It is Read-Only and calculated based on the ageGroup and consentProvidedForMinor proprties.
+
+| Value	   | #	|Description|
+|:---------------|:--------|:----------|
+|undefined|0|Default value, no ageGroup has been set for the user.|
```
angelgolfer-ms: Is it necessary to document the numeric value that each enum constant is mapped to? We don't do that for all other enums in the MS Graph docs. If there's no special reason, then I suggest we remove the middle column of each of the 3 tables.

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+## consentProvidedForMinor
+| Value	   | #	|Description|
+|:---------------|:--------|:----------|
+|none|0|Default value, no ageGroup has been set for the user.|
+|granted|1|Consent has been obtained for the user to have an account.|
+|denied|2|Consent has not been obtained for the user to have an account.|
+|notRequired|3|The user is from a location that does not require consent.|
+
+
+## JSON representation
```
angelgolfer-ms: You can delete this JSON representation section. This is a concept topic.
I will comment in the user.md topic that you should add these 3 properties in the JSON representation section, as well as the HTML comment in that section, just like you would when you add any other properties to an entity in MS Graph.

api-reference/beta/resources/legalagegroupclassification.md
angelgolfer-ms: Also, please rename this file according to our file-naming conventions for concept topics. The guideline is to keep the  keywords in your topic title and hyphenate each word (note, use hyphen, not underscore). So if you use a topic title like the one I suggested above, I would suggest the filename to be classify-users-age-group-legal-consent.md. I don't suggest to keep it word-for-word. Over time, we may tweak the topic title to respond to metrics (page views, etc.) We're not going to change the topic filename every time we tweak the topic title. The two will possibly go out of sync anyway, so there's no point to keep every word in the filename, if we consider SEO as the main factor in deciding the filename.

dkershaw10: Per separate communication, I think this should all be folded into the user.md file, in a sub-section under Properties, with internal links.  I don't think this is conceptual material, and really if there is conceptual material, it should encompass more than just parental controls.

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
+# legalAgeGroupClassification, ageGroup and consentProvidedForMinor 
+
+These three properties are used by Azure AD Tenant Administrators & Application Developers to help ensure the user of an account is handled correctly based on the regulatory rules governing the user's country.
+
+The ageGroup and consentProvidedForMinor are set by the administrative owner of the account. The property can not be read outside of the tenant because each tenant is generally consider a seperate regulatory boundary.
```
dkershaw10: I don't think the second sentence is strictly necessary - or could be merged into the first sentence.  Not sure if you should italicize the property names.
How about: "The ageGroup and consentProvidedForMinor may be set by the company or user account administrators."

api-reference/beta/resources/legalagegroupclassification.md
dkershaw10: Agreed.  Also can you double check this?  When I query users for these properties (when they've not been set), the property is **actually** `null`. This goes for both writeable properties.  The 2 writeable properties are not required properties, so if they aren't provided (on creation, or an update), they will be `null`.  So I'm not sure what the first enum value actually means here?

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+
+The legalAgeGroupClassification is used by application developers to restrict features and capabilities as required. This property is only available to applications in the Home Tenant of the user and Microsoft. 
+
+## legalAgeGroupClassification
+
+This property is used by applications to ensure the correct handling of a user based on their Legal Age Group. It is Read-Only and calculated based on the ageGroup and consentProvidedForMinor proprties.
```
dkershaw10: Shouldn't this say "This read-only property should be used by applications...".  I'd then remove "Read-Only" from the next sentence.

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+
+| Value	   | #	|Description|
+|:---------------|:--------|:----------|
+|undefined|0|Default value, no ageGroup has been set for the user.|
+|MinorWithoutParentalConsent |1|(Reserved for future use) The user is consider a minor based on the regulations of their location but consent has not been obtained. User's in this state will be unable to Sign In.|
```
dkershaw10: consider -> considered? (and for the next row).
Also lower camel-case please (for this and the next row).

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+|MinorWithParentalConsent|2|The user is consider on the regulations of their location and consent has been obtained for them to use the account.|
```
dkershaw10: Please fix.  This sentence makes no sense.

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+|notAdult|3|The user is from a country that has regulations (US, KR, UK, EU) and user’s age is more than the upper limit of kid age (as per country) and less than lower limit of Adult age (as per country). So basically, teenagers are considered as NotAdult in regulated countries.|
```
dkershaw10: All countries have regulations :). Also tried to simplify the text.  Maybe "The user is from a country or region that has additional age-related regulations (United States, United Kingdom, European Union or South Korea), and the user's age is in between a *minor* and an *adult* age (as stipulated based on country or region).  Generally, this means that teenagers are considered as `NotAdult` in regulated countries."

api-reference/beta/resources/legalagegroupclassification.md
```diff
@@ -0,0 +1,71 @@
...
+|adult|3|The user should be a treated as an adult.|
+
+
+## ageGroup
```
dkershaw10: Needs an intro - optional property that can only be set by an administrator.

api-reference/beta/resources/legalagegroupclassification.md
dkershaw10: If you incorporate into the user topic, what I'd do is add an additional Request/Response example to the Get user topic, and maybe the Update user topic showing an example of the how these are set, and how it affects the calculated property.

api-reference/beta/resources/user.md
```diff
@@ -67,11 +67,12 @@ This resource supports:
 |:---------------|:--------|:------------|
 |aboutMe|String|A freeform text entry field for the user to describe themselves.|
 | accountEnabled|Boolean| **true** if the account is enabled; otherwise, **false**. This property is required when a user is created. Supports $filter.    |
-|assignedLicenses|[assignedLicense](assignedlicense.md) collection|The licenses that are assigned to the user. Not nullable.            |
+|ageGroup|[ageGroup](legalAgeGroupClassification.md)|This property is used to set the age group of the user.||assignedLicenses|[assignedLicense](assignedlicense.md) collection|The licenses that are assigned to the user. Not nullable.            |
```
dkershaw10: Please add the line break back in for assignedLicenses.
Based on other enums descriptions, the type should be String, and the description should list the allowed values, and how null is treated. The allowed values should be listed like: `undefined`, `minor`, `notAdult` or `adult`.  From the description you should link to more information.

api-reference/beta/resources/user.md
```diff
@@ -82,6 +83,7 @@ This resource supports:
 |id|String|The unique identifier for the user. Inherited from [directoryObject](directoryobject.md). Key. Not nullable. Read-only.|
 |interests|String collection|A list for the user to describe their interests.|
 |jobTitle|String|The user’s job title. Supports $filter.|
+|legalAgeGroupClassification|[legalAgeGroupClassification](legalAgeGroupClassification.md)|This property is used by applications to determine the Legal Age Group of the user. This property is Read-only and calculated based on [ageGroup](legalAgeGroupClassification.md) and [consentProvidedForMinor](legalAgeGroupClassification.md) properties.|
```
dkershaw10: Please get rid of spurious capitalizations.

dkershaw10:
@lleonard-msft Signing off - this LGTM. 
@angelgolfer-ms Can you also quickly review this?

PR #2484, opened 2018-04-19
Fix doc bugs in the schema extension lifecycle

dkershaw10:
Fixes incorrect statement for the InDevelopment state.


dkershaw10:
@lleonard-msft Fixing a doc bug. Signed-off.

PR #2499, opened 2018-04-23
Update mailfolder.md

Clartred:
There is no folder named ArchiveRoot, but there is Archive.
My bad that my comment wasn't really good explanation of the problem.
What I wanted to say that there isn't https://graph.microsoft.com/v1.0/me/mailFolders/ArchiveRoot endpoint. There is https://graph.microsoft.com/v1.0/me/mailFolders/Archive Endpoint.
So, ArchiveRoot folder name is wrong and it shouldn't be on the documention where it currently stands.


jasonjoh:
I'm going to close this PR as I added a list of well-known folders in #2895.

PR #2525, opened 2018-04-28
Webhooks create subscription cleanup beta

piotrci:
Cleaned up the article for webooks subscription creation. I have some open questions for @jesakowi and @dkershaw10 - see inline.


api-reference/beta/api/subscription_post_subscriptions.md
```diff
@@ -2,10 +2,10 @@
 
 > **Important:** APIs under the /beta version in Microsoft Graph are in preview and are subject to change. Use of these APIs in production applications is not supported.
 
-Subscribes a listener application to receive notifications when data on the Microsoft Graph changes.
+Subscribes a listener application to receive notifications when data on a Microsoft Graph resource changes.
 
 ## Permissions
```
piotrci: @jesakowi , this original section reads a bit cryptic to me. why do we say "suggested" permissions? Shouldn't we have a section that says what happens when you try to create a subscription without the required permission?

api-reference/beta/api/subscription_post_subscriptions.md
```diff
@@ -96,24 +96,26 @@ Content-length: 252
...
+## Receiving a notification
+When the subscribed resource changes, Microsoft Graph sends a notification to your notification endpoint. The endpoint must respond with status code of 200 or 204 (with no response body) within 30 seconds; otherwise the notification will be resent at exponentially increasing time intervals. Services that consistently take more than 30 seconds to respond may be throttled and receive a sparser notification set.
 
-Services may also return a 422 response from a notification, in which case the subscription will be automatically deleted and the stream of notifications will come to a halt.
+You may also respond with the status code of 422, in which case your subscription will be automatically deleted and the stream of notifications will come to a halt.
 
-Depending on the subscribed resource, an additional resourceData field may provide additional information.
+Below is an example of a notification payload. For some types of resources, an additional `resourceData` field may include additional information.
```
piotrci: @jesakowi , shouldn't we explain what the resourceData field is for?
Also, nothing here explains what the app should do to get their hands on the changed data - the notification payload does not contain much info. shouldn't we briefly describe the calling pattern in response to the notification - or is it described somewhere else already?

api-reference/beta/api/subscription_post_subscriptions.md
```diff
@@ -130,7 +132,8 @@ Depending on the subscribed resource, an additional resourceData field may provi
   }
 }
 ```
-When receiving notifications from Drive subscriptions the resourceData will be null and the [delta](driveitem_delta.md) API should be called to determine the changes that have occured. Here is an example of a Drive notification:
+
+Notifications from Drive subscriptions have a null value for the `resourceData` field; the [delta](driveitem_delta.md) API should be called to determine the changes that have occurred for this resource. Here is an example of a Drive notification:
```
piotrci: @jesakowi , personally, i don't know what to do to handle the Drive notification? call delta - with what? should we improve this section or remove it altogether?

api-reference/beta/api/subscription_post_subscriptions.md
dkershaw10: This is a bigger piece of work.  All of our API topics should have an errors section, and at some point we'll auto-generate this.  For now this is not required.

api-reference/beta/api/subscription_post_subscriptions.md
dkershaw10: Webhooks  (and resourceData) are described here: https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/resources/webhooks. It does have a bit about the calling pattern, but I think it could be more detailed.  Maybe that's the subject of a new PR.

api-reference/beta/api/subscription_post_subscriptions.md
```diff
@@ -96,24 +96,26 @@ Content-length: 252
   "creatorUserId": "8ee44408-0679-472c-bc2a-692812af3437"
 }
 ```
-## Subscription validation
-In order to to avoid mistaken subscriptions directing notifications to arbitrary URLs, the subscription notification endpoint must be capable of responding to a validation request. During processing of the `POST` to the `/subscriptions` endpoint, the Microsoft Graph will send a `POST` request back to your `notificationUrl` in the following form:
+## Notification endpoint validation
+The subscription notification endpoint (specified in the `notificationUrl` property) must be capable of responding to a validation request; validation is required to make sure that subscriptions are not created pointing to invalid endpoints (e.g. due to misspelling of the URL in the request).
```
dkershaw10: due to misspelling of the `notificationUrl` in the subscription request.

api-reference/beta/api/subscription_post_subscriptions.md
```diff
@@ -96,24 +96,26 @@ Content-length: 252
...
 ```http
 HTTP/1.1 200 OK
 Content-type: text/plain
 Content-length: 7
 <token>
 ```
-## Notification payload
-When the subscribed resource changes, the webhooks facility sends a notification to your notification URL with the following payload.  The notification endpoint must send a response of 200 or 204 with no response body within 30 seconds otherwise the notification attempt will be retried at exponentially increasing intervals.  Services that consistently take 30 seconds or more may be throttled and receive a sparser notification set.
+## Receiving a notification
+When the subscribed resource changes, Microsoft Graph sends a notification to your notification endpoint. The endpoint must respond with status code of 200 or 204 (with no response body) within 30 seconds; otherwise the notification will be resent at exponentially increasing time intervals. Services that consistently take more than 30 seconds to respond may be throttled and receive a sparser notification set.
```
dkershaw10: 30 seconds, otherwise - I don't think this is a place to use a semi-colon.  However, just because I'm British doesn't mean my grammar knowledge is good ;)

dkershaw10:
@piotrci Couple of minor comments.  Changes look good to me.

jasonjoh:
Per conversation with Peter, merging this PR into the webhooks branch

per conversation with Peter, merging his webhooks PR's in to the webhooks branch with my existing webhook PR's so we can reconcile any conflicts and have one central PR (#2686) to review.

PR #2526, opened 2018-04-28
Best practices draft

dkershaw10:


concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+This topic describes a series of best practices that can help your apps get the best out of Microsoft Graph - whether that's through best ways to learn about using Microsoft Graph, performance improvements, or making your app more reliable for end-users.
+
+TDOD: OTHER TOPICS - REMOVE BEFORE MERGE:
+
+- App registration best practices and key management best practices.  Not specific to Graph, but...
+- Building workflow automation apps
+
+## Using Graph Explorer to get to know the API
+
+The easiest and best way to start exploring and navigating the data available through Microsoft Graph is by using [Microsoft Graph Explorer](http://aka.ms/ge). Microsoft Graph Explorer lets you craft REST requests (with full CRUD support), adapt the HTTP request headers, and see the data responses. To help you get started, Graph Explorer also provides a set of sample queries.
```
aaudu: **nit**: https://aka.ms/ge

It's always recommended to have the https versions of links.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+GET https://graph.microsoft.com/v1.0/messages
+```
+
+Would return a response containing an `@odata.nextLink` property, if the result set exceeds the server-side page size limit.
+
+```json
+"@odata.nextLink": "https://graph.microsoft.com/v1.0/me/messages?$skip=23"
+```
+
+> Your application should **always** handle the possibility that the responses are paged in nature, and use the `@odata.nextLink` propertty to obtain the next paged set of results, until all pages of the result set have been read. The final page will not contain an `@odata.nextLink` property. You should include the entire URL in the `@odata:nextLink` property in your request for the next page of results, treating the entire URL as an opaque string.
```
aaudu: typo '@data.nextLink' property -- single 't' as opposed to double 'propertty'

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+|Not found| 404 | In certain cases, a requested resource may not be found. For example a resource may not exist, because it has not yet been provisioned (like a user's photo) or because it has been deleted. Some deleted resources *may* be fully restored within 30 days of deletion - such as user, group and application resources, so your application should also take this into account.||
+|Throttling|429|APIs may throttle at any time for a variety of reasons, so your application must **always** be prepared to handle 429 responses. This error response includes the *Retry-After* field in the HTTP response header. Backing off requests using the *Retry-After* delay is the fastest way to recover from throttling.| [Throttling](throttling.md) |
+|Service unavailable| 503 | This is likely because the services is busy. You should employ a back-off strategy similar to 429. Additionally, you should always make new retry requests over a new HTTP connection.|| 
+
+FOLLOW UP ON BEHAVIORS AROUND LACK OF SUBSCRIPTION OR LICENSE.
+
+### Evolvable enums
+
+TBD
```
aaudu: I know we can these Evolvable enums, but maybe we opt for Enum Evolution or similar.

[This doc](https://docs.corda.net/serialization-enum-evolution.html) from corda presents some good framing and structure.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Storing data locally
+
+Your application should ideally make calls to Microsoft Graph to retrieve data in real-time as necessary.  You should only cache or store data locally if required for a specific scenario, and if that use case is covered by your terms of use and privacy policy, and does not violate and [Microsoft Graph terms of use](../misc/terms-of-use.md) or any other terms of use.
```
aaudu: "necessary.  You should" ==> double space as opposed to single.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Optimizations: get only the data you need
+
+The general theme here is that for performance and even security or privacy reasons, you should only get the data your application really needs.
```
aaudu: "data your application really needs, **nothing more.**" ==> Adding nothing more to be even more explicit.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Use projections
+
+Choose only the properties your app really needs and no more, since this saves unnecessary network traffic and data processing in your application (and in the service).
```
aaudu: "app" versus "application"? 

Let's be consistent and use "application", so that seems to be your predominant reference.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+For example, when retrieving the messages of the signed-in user, you can specify that only the from and subject properties be returned:
+
+```http
+GET https://graph.microsoft.com/v1.0/me/messages?$select=from,subject
+```
+
+### Getting minimal responses
+
+For some operations, such as PUT and PATCH (and in some cases POST), if your app doesn't need to make use of a response payload, you can ask the API to return minimal data. NOTE: that some services already return a 204 No Content response for PUT and PATCH operations.
```
aaudu: "application" as opposed to "app"

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+> Use [delta query](delta_query_overview.md) to efficiently keep data up to date.
+
+### Using webhooks and delta query together
+
+Webhooks and delta query are often used better together, because if you use delta query alone, you need to figure out the right polling interval - too short and this may lead to empty responses which wastes resources, too long and you might end up with stale data. If you use webhook notifications as the trigger to make delta query calls, you get the best of both worlds.
+
+> Use [webhook notifications](../api-reference/v1.0/resources/webhooks.md) as the trigger to make delta query calls.  You should also ensure that your application has a backstop polling threshold, in case no notifications are triggered.
+
+## Other optimizations
```
aaudu: Why "other" and not just part of the overall performance and reliability theme?

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+> Use [batching](json_batching.md) where significant network latency can have a big impact on the performance.
+
+### Data compression
+
+To minimize network traffic, it's recommended that applications request that response data is compressed.
+
+> Request compression of responses using an HTTP request header:  *Accept-Encoding: gzip*.
+
+## Reliability and support
```
aaudu: Sriram: Do we think #1 & #2 need more meat. I'm reading this and wondering where I could set these, or some  of the practical ways I can implement this.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. Honor DNS TTL and set connection TTL to match it. This ensures availability in case of failovers.
+
+2. Open connections to all advertised DNS answers.
+
+3. Always log the *request-id* and *timestamp* from the HTTP response header. This is required when escalating issues or reporting issues in StackOverflow or to Microsoft Customer Support.
```
dan-silver: Maybe add a section on graph versions and link to https://developer.microsoft.com/en-us/graph/docs/concepts/versioning_and_support for more details?

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Authentication
+
+To access the data in Microsoft Graph your application will need to acquire an OAuth 2.0 access token, and present it to Microsoft Graph:
+
+- in the HTTP *Authorization* request header, as a *Bearer* token, or
+- in the graph client constructor when using a Microsoft Graph client library
+
+> Use the Microsoft Authentication Library API, [MSAL](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-libraries) to acquire the access token to Microsoft Graph.
+
```
dan-silver: Other ideas if we want to expand Graph auth:
- Apps should always index on user ids since the same email can be used for personal and work/schools accounts.
- Apps should treat *access* tokens as any string.  If you need other metadata like obtained scopes or expiration timestamp, these can be returned along with the token in interactive and non-interactive flows. Apps shouldn't take a dependency on the format or content of access tokens.

dkershaw10: Yeah - there's a whole auth best practices being developed now that includes the latter item.  I'm not sure how much to say about auth, because I wanted to keep this focused on Graph.  However, there's other generic app guidance that's not specific to Graph, so I'll think about it.
For the first one, yeah - in general it's good practice to index on the id, rather than email or UPN because id is immutable, whereas the other identifiers are not.  I'll see how to weave this in.

concepts/best-practices-concept.md
dkershaw10: Yeah - was getting lazy.  Will fix.

concepts/best-practices-concept.md
dkershaw10: Because these aren't part of "get only the data you need".  Alternatively I could just have a single mondo "Optimizations" section, and put all of these under that.  @aaudu Thoughts?

concepts/best-practices-concept.md
dkershaw10: When we asked Adam about this he knew how to set the connection TTL, but not how to find the DNS TTL.

concepts/best-practices-concept.md
dkershaw10: Good call.  Let me think about this.

concepts/best-practices-concept.md
aaudu: Makes sense. Perhaps, retitle to "Optimizations: send requests effectively", since the theme above is "Optimizations: get only the data you need"

concepts/best-practices-concept.md
dkershaw10: Not sure where to weave this in.  We can add something on this after we publish the first draft.

aaudu: SGTM as a later update. I think we don't rev enough yet that it's a serious concern, but definitely something we can help people think through.

concepts/best-practices-concept.md
```diff
@@ -0,0 +1,145 @@
...
+GET https://graph.microsoft.com/v1.0/me/messages?$select=from,subject
+```
+
+### Getting minimal responses
+
+For some operations, such as PUT and PATCH (and in some cases POST), if your application doesn't need to make use of a response payload, you can ask the API to return minimal data. NOTE: that some services already return a 204 No Content response for PUT and PATCH operations.
+
+> Request minimal representation responses using an HTTP  request header where appropriate:  *Prefer: return=minimal*. NOTE: for creation operations, this may not be appropriate because your application may expect to get the service generated `id` for the newly created object in the response.
+
+### Track changes: delta query and webhook notifications
```
angelgolfer-ms: @dkershaw10 , @jesakowi , for Build, the docs are referring what used to be called webhooks in the docs as push notifications. Jackson previously brought this up with @jesakowi and that was the agreement. 
AFAIK, other than the TOC and topic content, featured scenarios are using "push notifications" too. 
Shall we stay consistent? Or, if there a different opinion, we need to talk about this soon as content is being staged for publishing later this week.

PR #2534, opened 2018-04-30
Incorporate feedback for users article 

rloutlaw:
Adding in feedback received from reviewers of the Users article.


rloutlaw:
@Lauragra I had to put these changes here because the base article isn't yet in the Staging branch. LMK if you need it brought over wholesale.

PR #2582, opened 2018-05-04
Updating Security Overview text

jdallman:
- Changing wording in the overview to align with other messaging. 
- Removing the "Federated" text since it didn't fit the page flow. 
- Adding details & blog hyperlink to Authorization section to address a common customer issue & the solution.

### DIFF
--- a/api-reference/beta/resources/security-api-overview.md
+++ b/api-reference/beta/resources/security-api-overview.md
@@ -8,15 +8,11 @@ The security API in Microsoft Graph makes it easy to connect with and extend Int
 
 ## Why connect to the Intelligent Security Graph?
 
-By using the security API to connect with Microsoft and partner security solutions that are part of the Intelligent Security Graph, you can more readily realize and enrich the value of these solutions. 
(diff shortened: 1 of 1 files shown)


api-reference/beta/resources/security-api-overview.md
```diff
@@ -8,16 +8,12 @@ The security API in Microsoft Graph makes it easy to connect with and extend Int
 
 ## Why connect to the Intelligent Security Graph?
 
-By using the security API to connect with Microsoft and partner security solutions that are part of the Intelligent Security Graph, you can more readily realize and enrich the value of these solutions. 
+The security API provides a unified gateway to connect to security solutions from Microsoft and partners. Queries are federated to all security providers in Microsoft Graph and responses are aggregated. Use the security API to:
```
Lauragra: Generally we don't introduce subheadings with an introductory phrase and colon. I think that the intent is for the subheadings to answer the question Why connect to the Intelligent Security Graph?. Okay to delete "Use the security API to:"?

jdallman: No problem. Since this was already merged, I'll submit another pull

PR #2633, opened 2018-05-11
Fixed Inacurrate Quote Limits

mlafleur:
The "Limitations" secion of the webhooks document inaccuratly implied these limits applied to webhooks in general. They actually only apply to AAD based resources and are not applicable to others (i.e. Mail, Events, etc.).  This was causing some devs to make incorrect design assumptions (see https://stackoverflow.com/questions/50275993/how-to-increase-microsoft-graph-api-subscription-limits).

I also cleaned up some formatting inconsitencies (ie H1s following H2s, inconsitant language spesification, etc).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+
+.vscode/settings.json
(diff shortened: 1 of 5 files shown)


mlafleur:
Great catch on the code blocks. I'll add a rule to my md linter to catch this case in the future. I've also fixed the links (silly copy & paste error on my part).

PR #2634, opened 2018-05-11
Open extensions

jasonjoh:
Adding Outlook-specific considerations to open extensions docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Visual Studio Code config directory
+.vscode/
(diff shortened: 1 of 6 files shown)


mlafleur:
I think the error message here is pretty misleading. It looks like it's actually failing because some reference types are not showing up in the https://graph.microsoft.com/beta/$metadata schema at the moment:
- `microsoft.graph.educationSynchronizationOAuth1ConnectionSettings`
- `microsoft.graph.educationSynchronizationOAuth2ClientCredentialsConnectionSettings`

Maybe the AppVeyor tests should treat a missing reference as a warning (as it does missing properties from a reference)?

PR #2660, opened 2018-05-17
Update v1 webhooks with content from beta

piotrci:
We are about to publish new beta functionality (List subscriptions) into v1.0. This change updates v1.0 reference docs with the new functionality. A few minor fixes are also included for /beta articles (mostly, consistency).


Lauragra:
@piotrci , is the API available in v1.0 now? If so, we'll need an entry for List subscriptions in the v1.0 TOC.

PR #2663, opened 2018-05-18
Update list_create.md

richdizz:
If you try to provision a list by specifying the name, you get this error: "Cannot define a 'name' for a list as it is assigned by the server. Instead, provide 'displayName'". Using displayName instead works just fine.


Lauragra:
Fixed per PR #4624.

PR #2725, opened 2018-06-02
API changes to support feedback flow on Assignments

murtuzas:


api-reference/beta/api/educationsubmission_unsubmit.md
```diff
@@ -0,0 +1,65 @@
+# educationSubmission: submit
```
Lauragra: The title should be educationSubmission: unsubmit here.

PR #2794, opened 2018-06-20
Update bookingbusiness_list and bookingbusiness_list_services pages

fpintos:
This PR updates some details of the bookingbusiness_list and bookingbusiness_list_services pages.

In particular:
- the odata.context metadata in the sample response must use the 'beta' endpoint;
- GET bookingBusinesses returns _only_ id and displayName properties;
- GET bookingBusinesses supports a 'query' parameter, which limits the results.


api-reference/beta/api/bookingbusiness_list.md
```diff
@@ -24,6 +24,15 @@ GET /bookingBusinesses
 ## Optional query parameters
 This method supports the [OData Query Parameters](http://graph.microsoft.io/docs/overview/query_parameters) to help customize the response.
 
+This method also supports the following parameters:
+- **query**: a string that limits the results to businesses that match the given value.
```
angelgolfer-ms: @fpintos , is the query string matched against any particular properties, like the displayName? In other words, exactly how does GET determine there's a match?

fpintos: It performs an [ambiguous-name-resolution](https://msdn.microsoft.com/en-us/library/cc223243.aspx).

In this context, it looks at the id and displayName, although (other properties in the AAD)[https://en.wikipedia.org/wiki/Ambiguous_name_resolution] will also be considered, even if they are not directly exposed by this API. In other words, it is conceivable that a query for 'foo' might return a object that doesn't have 'foo' in its id or displayName, if the object in the AAD was manipulated outside the Bookings API; they are, after all, mailboxes in the O365 tenant.

In principle, the fact that we do an ANR is an implementation detail.

api-reference/beta/api/bookingbusiness_list.md
```diff
@@ -4,7 +4,7 @@
  
 Get a collection of [bookingbusiness](../resources/bookingbusiness.md) objects that has been created for the tenant. 
 
-This operation returns only the **id** and **displayName** of each Bookings business in the collection. For performance considerations, other properties are _returned_ as null even if they have been set. You can get the other properties of a Bookings business by specifying its **id** in a [GET](bookingbusiness_get.md) operation.
+This operation returns only the **id** and **displayName** of each Bookings business in the collection. For performance considerations, other properties are _not returned_ even if they have been set. You can get the other properties of a Bookings business by specifying its **id** in a [GET](bookingbusiness_get.md) operation.
```
angelgolfer-ms: I'd suggest after this paragraph, we add another paragraph to describe the use case of the query parameter. For example (feel free to update):
"Alternatively, without providing the **id** of a Bookings business, you can also try to match a business in a tenant using the `query` parameter. (Here, provide a link to the second example that shows how to use the query parameter.)  Matches are based on an ambiguous name resolution algorithm (here, provide a link to https://msdn.microsoft.com/en-us/library/cc223243.aspx.)"

api-reference/beta/api/bookingbusiness_list.md
```diff
@@ -24,6 +24,15 @@ GET /bookingBusinesses
 ## Optional query parameters
 This method supports the [OData Query Parameters](http://graph.microsoft.io/docs/overview/query_parameters) to help customize the response.
 
+This method also supports the following parameters:
```
angelgolfer-ms: Can also add a link to the second example noted below.

PR #2814, opened 2018-06-26
Update driveitem_post_children.md

r-henriques:
Fixed links in Response and Error Response sections.


jthake:
@angelgolfer-ms how can we test these links work after the HTML changes? These three links right at bottom were not in correct format and aren't displaying currently here https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/driveitem_post_children

Lauragra:
Thanks; I've added this fix to PR #3910, so closing this one.

PR #2818, opened 2018-06-26
Onedrive delta query omitted props tables

Jackson-Woods:
Note that OneDrive for Business does return size on deleted items, but value is always 0. OneDrive consumer omits the property entirely.


Lauragra:
Closing this PR and reopening a new one.

PR #2828, opened 2018-06-28
Corrected the URL for patch operation

paragbhand:
Please note that the patch operation doesn't work on URL https://graph.microsoft.com/beta/organization
instead it works for https://graph.microsoft.com/beta/organization/<tenant-id>


api-reference/v1.0/api/organization_update.md
```diff
@@ -52,7 +52,7 @@ If successful, this method returns a `204 No Content` response code.
 }-->
 
 ```http
-PATCH https://graph.microsoft.com/v1.0/organization
+PATCH https://graph.microsoft.com/v1.0/organization/<tenant-id>
```
dkershaw10: Can you change this to {id} (instead of <tenant-id>) please?
Also earlier in the topic there's an "HTTP Request" section, where you need to make the same change.

dkershaw10: Also can you make this update in the /beta docs too please?

Lauragra:
Closing as this has been fixed in the file. Thanks.

PR #2843, opened 2018-07-05
(DO NOT MERGE UNTIL 7/13) Microsoft Teams Get Channel messages API's

hamenon-ms:
Pull request for updating Graph Documentation with the GET channel messages and GET channel message replies API's. At a high level the changes are as follows

- Updating the chatmessage resource
- Adding the chatattachment, chatmention and chatreaction resources
- Adding the new API's for List all channel messages, get a channel message, list all replies to a message and get a reply to a channel message


hamenon-ms:
Used the wrong branch. Will update with the right one.

PR #2845, opened 2018-07-05
Update group visibility property value

dkershaw10:
Added the HiddenMembership value


dkershaw10:
@lleonard-msft Please review and merge.

PR #2851, opened 2018-07-09
Updated docs with new OneDrive max subscription time

piotrci:


tfrancois:
Please please please....can someone at Microsoft correct this please.  Everywhere in the documentation it states that the max subscription time is 3 days for user, calendar, contacts subscriptions, etc.  This is 4320 minutes = 72 hours = 3 days.  BUT!!!!! The error message that gets returned when you try to set this max value is:

`"error":{  
         "code":"ExtensionError",
         "message":"Subscription expiration can only be 4230 minutes in the future.",
         "innerError":{  
            "request-id":"48f13973-e805-4cfe-a011-161a005783df",
            "date":"2018-08-20T02:29:38"
         }
      }`

For whatever reason, the max threshold value is INCORRECTLY set to 4230 rather than 4320.  Why am I posting this here?! Because the docs have been updated for OneDrive to the correct value but not for the others.

Resource | Maximum Expiration Time
-- | --
Mail | 4230 minutes (3 days)
Calendar | 4230 minutes (3 days)
Contacts | 4230 minutes (3 days)
Group conversations | 4230 minutes (3 days)
Drive root items | **4320** minutes (**3 days**) => CORRECT VALUE

Is this carelessness on Microsoft's part?  Why the difference in minutes even though the threshold is stated as 3 days for all object types.  This creates issues for developers when trying to set a correct max subscription time via mathematical computation (again: 3 days = 72 hours = 4320 minutes) HENCE when you try new Date() + 4320 minutes the request returns the error above.  Can we just fix this ONCE and for all.  It's either 3 days OR 4230 minutes.  The 2 are not functional or mathematical equivalents.

PLEASE FIX THIS.  Thank you.

PR #2855, opened 2018-07-12
Added content about group membership changes via delta query

piotrci:
This addresses the feedback tracked here: Task 206020: Improve group example in delta query

@dkershaw10  can you have a look?

### DIFF
--- a/api-reference/beta/api/group_delta.md
+++ b/api-reference/beta/api/group_delta.md
@@ -14,7 +14,7 @@ One of the following permissions is required to call this API. To learn more, in
 |Application | Group.Read.All, Group.ReadWrite.All |
 
 ## HTTP request
-To begin tracking changes, you make a request including the delta function on the groups resource. 
+To begin tracking changes, you make a request including the delta function on the groups resource.
 
 <!-- { "blockType": "ignored" } -->
 ```http
@@ -36,11 +36,12 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
 ## Optional query parameters
 This method supports OData Query Parameters to help customize the response.
 
-- You can use a `$select` query parameter as in any GET request to specify only the properties your need for best performance. The 
-_id_ property is always returned. 
-- Delta query support `$select`, `$top`, and `$expand` for groups. 
+- You can use a `$select` query parameter as in any GET request to specify only the properties your need for best performance. The
+_id_ property is always returned.
+- Delta query support `$select`, `$top`, and `$expand` for groups.
 - There is limited support for `$filter` and `$orderby`:
   * The only supported `$filter` expression is for tracking changes on a specific object: `$filter=id+eq+{value}`. You can filter multiple objects. For example, `https://graph.microsoft.com/v1.0/groups/delta/?$filter= id eq '477e9fc6-5de7-4406-bb2a-7e5c83c9ffff' or id eq '004d6a07-fe70-4b92-add5-e6e37b8affff`. There is a limit of 50 filtered objects.
+- There is no support for `$search`.
 
 ## Request headers
 | Name       | Description|
@@ -59,9 +60,9 @@ If successful, this method returns `200 OK` response code and [group](../resourc
 - If a deltaLink URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
 
 See:</br>
-- [Using Delta Query](../../../concepts/delta_query_overview.md) for more details</br>
-- [Get incremental changes for groups](../../../concepts/delta_query_groups.md) for an example requests.</br>
-    
+- [Using Delta Query](../../../concepts/delta_query_overview.md) for more details.</br>
+- [Get incremental changes for groups](../../../concepts/delta_query_groups.md) for a more detailed walkthrough.</br>
+
 ## Example
 #### Request
 The following is an example of the request.
@@ -76,6 +77,8 @@ GET https://graph.microsoft.com/beta/groups/delta
 #### Response
 The following is an example of the response.
 >**Note:** The response object shown here might be shortened for readability. All the properties will be returned from an actual call.
+>
+> Note the presence of the *members@delta* property which includes the ids of member objects in the group.
 
 <!-- {
   "blockType": "response",
@@ -98,7 +101,17 @@ Content-type: application/json
       "groupTypes": [
         "groupTypes-value"
       ],
-      "mail": "mail-value"
+      "mail": "mail-value",
+      "members@delta": [
+               {
+                   "@odata.type": "#microsoft.graph.user",
+                   "id": "693acd06-2877-4339-8ade-b704261fe7a0"
+               },
+               {
+                   "@odata.type": "#microsoft.graph.user",
+                   "id": "49320844-be99-4164-8167-87ff5d047ace"
+               }
+      ]
     }
   ]
 }
--- a/api-reference/beta/api/group_delta.md
+++ b/api-reference/beta/api/group_delta.md
@@ -59,10 +59,6 @@ If successful, this method returns `200 OK` response code and [group](../resourc
 
 - If a deltaLink URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
 
-See:</br>
-- [Using Delta Query](../../../concepts/delta_query_overview.md) for more details.</br>
-- [Get incremental changes for groups](../../../concepts/delta_query_groups.md) for a more detailed walkthrough.</br>
-
 ## Example
 #### Request
 The following is an example of the request.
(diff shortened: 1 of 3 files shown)


api-reference/beta/api/group_delta.md
```diff
@@ -36,11 +36,12 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
 ## Optional query parameters
```
angelgolfer-ms: This comment pertains to a break to consistency from an earlier point, and I wonder if you mind fixing it at the same time: 
(1) Change ### Query parameters to ## Query parameters. 
(2) Change ## Optional query parameters to a subsection ### OData query parameters.  (That's to contrast the latter as OData query paremeters.)
(3) Add the word "optional" to the first sentence under this section to say "This method supports optional OData query parameters to help customize the response."

api-reference/beta/api/group_delta.md
```diff
@@ -59,9 +60,9 @@ If successful, this method returns `200 OK` response code and [group](../resourc
 - If a deltaLink URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
 
 See:</br>
```
angelgolfer-ms: I am also aware the following comment applies to something done some time ago that broke consistency; if you can fix this, that would be great: please move these 2 links to a ## See also section at the end of the topic. 

In the context of the delta function reference topic, the Response section is meant to concisely define what a single call of delta returns. This is similar to your Example request and response - it shows what a single call does. 

To make the point that typically, the scenario would require a series of calls, one can do that in the Query parameters section (which this topic does), and in the example section.

api-reference/v1.0/api/group_delta.md
```diff
@@ -33,9 +33,9 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
 ## Optional query parameters
```
angelgolfer-ms: Please apply my comments above, for the beta version of the topic, here as well.

api-reference/beta/api/group_delta.md
piotrci: Done

api-reference/beta/api/group_delta.md
piotrci: Moved the "see also stuff". I'm not sure I understand the rest of the comment: this reference article contains the very basic response scenario. The conceptual article - linked under See also - has a detailed walkthrough for how to implement the real world delta scenario. Is that the correct way to break down the complexity?

piotrci:
@angelgolfer-ms thanks for reviewing. does it look good now?

concepts/delta_query_groups.md
```diff
@@ -24,22 +24,23 @@ The following example shows a series  requests to track changes to groups:
...
 ## Initial request
 
-To begin tracking changes in the group resource, you make a request including the delta function on the group resource. 
+To begin tracking changes in the group resource, you make a request including the delta function on the group resource.
 
 Note the following:
 
-- The optional $select query parameter is included in the request to demonstrate how query parameters are automatically included in future requests.
+- The optional `$select` query parameter is included in the request to demonstrate how query parameters are automatically included in future requests.
+- The optional `$expand` query parameter is included to show how group members can be retrieved together with group objects. This allows tracking of membership changes, such as when users are added or removed from groups.
```
angelgolfer-ms: Again, the next comment pertains to some oversight that was introduced early on (I noticed from the history): would you mind updating delta_query_overview.md that still says, for users and groups, "$expand is not supported"?

concepts/delta_query_groups.md
```diff
@@ -133,15 +162,19 @@ Content-type: application/json
...
+- Group objects for which member objects have been added or removed.
 
 ``` http
 GET https://graph.microsoft.com/v1.0/groups/delta?$deltatoken=sZwAFZibx-LQOdZIo1hHhmmDhHzCY0Hs6snoIHJCSIfCHdqKdWNZ2VX3kErpyna9GygROwBk-rqWWMFxJC3pw
 ```
 
 ## deltaLink response
 
-If no changes have occurred, the same `deltatoken` is returned with no results.
+If no changes have occurred, a `dataLink` is returned with no results - the `value` property is empty. Make sure to replace the previous link in the application with the new one for use in future calls.
```
angelgolfer-ms: Do you mean a deltaLink instead of a dataLink?

concepts/delta_query_groups.md
```diff
@@ -154,7 +187,7 @@ Content-type: application/json
 }
 ```
 
-If changes have occurred, the same `deltatoken` is returned including a collection of changed groups.
+If changes have occurred, a collection of changed groups is included. The response also contains either a `nextLink` - in case there are multiple pages of changes to retrieve - or a `deltaLink`. You should implement the same pattern of following *nextLinks* as before and persist the final `deltaLink` for future calls.
```
angelgolfer-ms: I know what you mean with "*nextLinks*" :-) To format "nextLink" appropriately though, because the term is always "nextLink" and not the plural, we would need to rephrase the sentence as  "You should implement the same pattern of following the corresponding *nextLink* as before..."

concepts/delta_query_groups.md
```diff
@@ -163,15 +196,123 @@ Content-type: application/json
...
+
+Let's assume you are executing the following delta query - either to capture the initial full state of groups, or later on to get delta changes:
+
+``` http
+GET https://graph.microsoft.com/v1.0/groups/delta?$select=displayName,description&$expand=members
+```
+
+1. Microsoft Graph may return a response that contains just one group object, with a large list of members in the `members@delta` property:
+
+**1st page**
```
angelgolfer-ms: Please use "First page". FYI, if you're interested, the Microsoft Writing Style Guide (http://aka.ms/style) for numbers is here:
https://worldready.cloudapp.net/StyleGuide/Read?id=2700&topicid=33688
https://worldready.cloudapp.net/StyleGuide/Read?id=2700&topicid=36961

concepts/delta_query_groups.md
```diff
@@ -163,15 +196,123 @@ Content-type: application/json
...
+      ]
+    }
+    <...no more groups included - this group filled out the entire response...>
+  ]
+}
+```
+
+2. When you follow the `nextLink` you may receive a response again containing the same group object. The same property values will be returned but the expanded `members@delta` property now contains a different list of users.
+
+**2nd page**
```
angelgolfer-ms: Similar to last comment, please use "Second page".

concepts/delta_query_groups.md
```diff
@@ -163,15 +196,123 @@ Content-type: application/json
...
     }
+    <...no more groups included - this group filled out the entire response...>
   ]
 }
 ```
 
+3. Eventually, the entire member list will be returned in this fashion, and other groups will start showing up in the response.
+
+We recommend the following best practices to correctly handle this pattern:
+- Always follow `nextLink` and locally merge each group's state: as you receive responses related to the same group, use them build the full membership list in your application.
```
angelgolfer-ms: Please add "to" here: "... use them to build the full membership..."

concepts/delta_query_groups.md
```diff
@@ -163,15 +196,123 @@ Content-type: application/json
...
+- It is best not to assume a specific sequence of the responses. Assume that the same group could show up anywhere in the nextLink sequence and handle that in your merge logic.
```
angelgolfer-ms: Please format "nextLink" with ticks, like `nextLink`.

concepts/delta_query_groups.md
```diff
@@ -133,15 +162,19 @@ Content-type: application/json
 
 ## deltaLink request
 
-Using the `deltaToken` from the [last response](#final-nextlink-response), you will be able to get changed (by being added, deleted, or updated) groups since the last request.
+Using the `deltaLink` from the [last response](#final-nextlink-response), you will be able to get net new changes to groups since the last request. Changes include:
+- Newly created group objects.
+- Deleted group objects.
+- Group objects for which a property has changed (e.g. displayName has been modified).
```
angelgolfer-ms: Please format "displayName" as bold.

concepts/delta_query_groups.md
piotrci: Done.

concepts/delta_query_groups.md
piotrci: Done.

concepts/delta_query_groups.md
piotrci: Done.

concepts/delta_query_groups.md
piotrci: Done.

concepts/delta_query_groups.md
piotrci: Done.

concepts/delta_query_groups.md
piotrci: Done.

concepts/delta_query_groups.md
piotrci: Done.

PR #2903, opened 2018-07-23
Delta query changed properties only

piotrci:
#hold-off
This PR should only be merged after we enable the new functionality for the prefer:return=minimal header in production.

Description: delta queries for AAD objects (currently: users and groups) will soon support a new header that changes the default behavior regarding which properties are returned when using deltaLinks. The new behavior returns only changed properties since last delta token, and allows the caller to disambiguate between properties that have not changed and properties that have been set to null.

Things TBD: 
- similar updates to group_delta.md.
- Possibly, add *_delta.md to TOC <- should this be a separate PR? @PatAltimore , what do you think?


api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,105 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
+- There is limited support for `$filter`:
   * The only supported `$filter` expression is for tracking changes on a specific object: `$filter=id+eq+{value}`. You can filter multiple objects. For example, `https://graph.microsoft.com/v1.0/groups/delta/?$filter= id eq '477e9fc6-5de7-4406-bb2a-7e5c83c9ffff' or id eq '004d6a07-fe70-4b92-add5-e6e37b8affff`. There is a limit of 50 filtered objects.
-- There is no support for `$search`.
 
 ## Request headers
 | Name       | Description|
 |:---------------|:----------|
 | Authorization  | Bearer &lt;token&gt;|
 | Content-Type  | application/json |
+| Prefer | return=minimal </br></br>When specified, only the object properties whose values have changed are included in the response. See [this section](#properties-included-in-the-response) for more details. |
```
lleonard-msft: Use `<br>`, not `</br>` or `<br/>` (XHTML only).

piotrci: Fixed.

api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,105 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
 
 - You can use a `$select` query parameter as in any GET request to specify only the properties your need for best performance. The
-_id_ property is always returned.
-- Delta query support `$select`, `$top`, and `$expand` for groups.
-- There is limited support for `$filter` and `$orderby`:
+*id* property is always returned.
+- You can use `$expand=members` to get membership changes.
+- There is limited support for `$filter`:
   * The only supported `$filter` expression is for tracking changes on a specific object: `$filter=id+eq+{value}`. You can filter multiple objects. For example, `https://graph.microsoft.com/v1.0/groups/delta/?$filter= id eq '477e9fc6-5de7-4406-bb2a-7e5c83c9ffff' or id eq '004d6a07-fe70-4b92-add5-e6e37b8affff`. There is a limit of 50 filtered objects.
-- There is no support for `$search`.
```
angelgolfer-ms: In the last line, are we missing a single quote at the end? That is, the GUID should be enclosed by single-quotes?
'004d6a07-fe70-4b92-add5-e6e37b8affff'

api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,92 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
 
-- If a nextLink URL is returned, there are additional pages of data to be retrieved in the session. The application continues making requests using the nextLink URL until a deltaLink URL is included in the response.
+- If a `nextLink` URL is returned, there are additional pages of data to be retrieved in the session. The application continues making requests using the nextLink URL until a deltaLink URL is included in the response.
 
-- If a deltaLink URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
+- If a `deltaLink` URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
+
+### Properties included in the response
+
+Responses for `nextLink` always include all the properties of the object - the default set, or the set specified in the optional `$select` query parameter. This allows you to capture the full current state of the objects when initiating the delta cycle.
```
angelgolfer-ms: Is it accurate to say:
"Responses for `nextLink` always include the same set of properties as in the initial delta query request."
Sometimes customers only scan and see "always include all the properties of the object" without finishing the sentence, so maybe we should rephrase to avoid confusion?
If you want, you can then elaborate with another sentence about the default set or the $select parameter,  but I think it's ok not to (since it's explained in full in the overview topic) and stay focused on contrasting nextLink and deltaLink here.

api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,92 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
+
+Responses for `deltaLink` include values for only the properties that have changed since the time the `deltaLink` was issued. There are two behaviors you can choose from:
+
+#### Default: all properties included
+All properties selected in the initial delta query are always returned in the Json response, even if their values have not changed.
```
angelgolfer-ms: Please replace "Json" by "JSON" (in all instances if others exist).

api-reference/beta/api/group_delta.md
piotrci: Fixed.

api-reference/beta/api/group_delta.md
piotrci: Fixed.

api-reference/beta/api/group_delta.md
piotrci: Fixed.

api-reference/beta/api/group_delta.md
angelgolfer-ms: In general, for better readability, please use active voice whenever possible, instead of passive voice. Here, especially considering the lead-in sentence in our context:
>>
There are two behaviors you can choose from:
#### Default: include all properties
Return in the JSON response those properties that are selected in the initial delta query. This applies even if their values have not changed.

api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,92 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
+- You can use `$expand=members` to get membership changes.
+- There is limited support for `$filter`:
+  * The only supported `$filter` expression is for tracking changes on a specific object: `$filter=id+eq+{value}`. You can filter multiple objects. For example, `https://graph.microsoft.com/beta/groups/delta/?$filter= id eq '477e9fc6-5de7-4406-bb2a-7e5c83c9ffff' or id eq '004d6a07-fe70-4b92-add5-e6e37b8affff'`. There is a limit of 50 filtered objects.
 
 ## Request headers
 | Name       | Description|
 |:---------------|:----------|
 | Authorization  | Bearer &lt;token&gt;|
 | Content-Type  | application/json |
+| Prefer | return=minimal <br><br>When specified, only the object properties whose values have changed are included in the response. See [this section](#properties-included-in-the-response) for more details. |
```
angelgolfer-ms: Because the delta function can be called as part of a nextLink or deltaLink, and, (if I understand correctly) you can apply the Prefer header only with deltaLink, I think the description for the Prefer header should clarify that:

"Specifying this header with a request that uses a deltaLink would return only the object properties that have changed since the last round. Optional."

api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,92 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
 Do not supply a request body for this method.
 
 ## Response
 If successful, this method returns `200 OK` response code and [group](../resources/group.md) collection object in the response body. The response also includes a state token which is either a nextLink URL or a deltaLink URL.
 
-- If a nextLink URL is returned, there are additional pages of data to be retrieved in the session. The application continues making requests using the nextLink URL until a deltaLink URL is included in the response.
+- If a `nextLink` URL is returned, there are additional pages of data to be retrieved in the session. The application continues making requests using the nextLink URL until a deltaLink URL is included in the response.
 
-- If a deltaLink URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
+- If a `deltaLink` URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
```
angelgolfer-ms: Please also format the second instance of deltaLink with a set of tick marks.

angelgolfer-ms: I suggest instead of "For future requests, the application uses the `deltaLink` URL to learn about changes to the resource", say:
"Save and use the `deltaLink` URL to learn about changes to the resource in the next round."

api-reference/beta/api/group_delta.md
```diff
@@ -37,27 +37,92 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
+
+#### Alternative: only changed properties included
+Adding an optional request header - `prefer:return=minimal` - ensures that only the properties whose values have changed since the last delta sync are included in the Json respnose.
+
+> **Note:** The header can be added to a `deltaLink` request at any point in time in the delta cycle. The header only affects the set of properties included in the response and it does not affect how the delta query is executed.
+
+
+For example, an initial request selected 3 properties for change tracking:
+```http
+GET https://graph.microsoft.com/beta/groups/delta?$select=displayName,description,mailNickname
```
angelgolfer-ms: I am confused now - should this GET request use a deltaLink, something like:

Prefer: return=minimal
GET https://graph.microsoft.com/v1.0/groups/delta?$deltatoken=sZwAFZibx-LQOdZIo1hHhmmDhHzCY0Hs6snoIHJCSIfCHdqKdWNZ2VX3kErpyna9GygROwBk-rqWWMFxJC3pw

The request you have here starts all over again, there isn't a "last state" from a previous round, how can one compare and determine whether a property has changed or not?

api-reference/beta/api/user_delta.md
```diff
@@ -17,64 +17,121 @@ One of the following permissions is required to call this API. To learn more, in
...
-- If a nextLink URL is returned, there are additional pages of data to be retrieved in the session. The application continues making requests using the nextLink URL until a deltaLink URL is included in the response.
+- If a `nextLink` URL is returned, there are additional pages of data to be retrieved in the session. The application continues making requests using the nextLink URL until a deltaLink URL is included in the response.
 
-- If a deltaLink URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
+- If a `deltaLink` URL is returned, there is no more data about the existing state of the resource to be returned. For future requests, the application uses the deltaLink URL to learn about changes to the resource.
 
-See:</br>
-- [Using Delta Query](../../../concepts/delta_query_overview.md) for more details</br>
-- [Get incremental changes for users](../../../concepts/delta_query_users.md) for an example requests.</br>
+### Properties included in the response
```
angelgolfer-ms: Please apply the similar reorganization in this Response section, as in the group_delta.md Response section.

api-reference/beta/api/group_delta.md
piotrci: Fixed.

api-reference/beta/api/group_delta.md
piotrci: Fixed.

piotrci: Fixed.

api-reference/beta/api/group_delta.md
piotrci: deltaLinks do not include any $select parameters in the URL, because everything that was initially specified is encoded in the token itself. My goal was to show the initial request to explicitly show what was selected, and then show the response to a future deltaLink. i have made some changes to simplify and clarify.

api-reference/beta/api/user_delta.md
piotrci: Fixed.

piotrci:
#hold-off because the feature has not been deployed to production yet.

api-reference/beta/api/user_delta.md
angelgolfer-ms: Hm, Peter, wondering you have taken a look at the separate email with the Word document that I sent a day ago (7/31), that suggested how to move the examples to the Example section, etc.? Do you have an opinion on that please?

piotrci: @angelgolfer-ms , I have made these changes now. Can you please have a look? I'd like to merge this PR soon, because the new functionality is now available in production.

api-reference/beta/api/group_delta.md
```diff
@@ -113,9 +150,76 @@ Content-type: application/json
...
+  "name": "group_delta"
+}-->
+
+```http
+GET https://graph.microsoft.com/beta/groups/delta?$select=displayName,description,mailNickname
+```
+
+#### Response 2
+
+The following is an example of the response when using `deltaLink` obtained from the query initialization. Note that `description` and `mailNickname` have the value of `null` which means that they may have not changed or have been set to an empty value.
```
angelgolfer-ms: @piotrci , I just noticed this - does the following sentence match example 2? "The following is an example of the response when using `deltaLink` obtained from the query initialization."
Request 2 doesn't show any deltaToken; I thought a URL returned by deltaLink includes a deltaToken so that the next round can pick up where the previous round ended?

api-reference/beta/api/user_delta.md
angelgolfer-ms: @piotrci , thanks, looks good! I have only one technical [comment](https://github.com/microsoftgraph/microsoft-graph-docs/pull/2903/files#r222065015), we probably overlooked the issue before.
And just one more instance of deltaLink I spotted in the Response section of group_delta.md, and the same in user_delta.md, that are missing the tick marks. 

In general: API names as defined in the Graph metadata - bold, unless we're in a table with most words being API names, then we'll just leave them normal text (not bolded) for better readability. Other programmability terms, like OData query parameter names, our own HTTP request header names, etc., enclose them in tick marks.

Issue #1975, opened 2017-11-15
Getting profile photo by size is available in v1.0 but only documented in beta


dan-silver:
It looks like the section for `HTTP request to get the metadata for a specific photo size` is documented at https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/api/profilephoto_get but it's missing in the v1.0 page at https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/profilephoto_get.  To test this, try GET `https://graph.microsoft.com/v1.0/me/photos/64X64/$value` and you should see an image returned.

lleonard-msft:
@dan-silver - Just wanted to let you know that this fix has been checked in and should be public in 24 hours.  Once you confirm that's it public, would you mind closing the issue?

dan-silver:
Thanks @lleonard-msft!

PR #2924, opened 2018-07-30
Per issue 1975, copying size request examples from beta.

lleonard-msft:
Per Issue #1975, this copies in the HTTP request examples for `/size`.

@dkershaw10 - PTAL


lleonard-msft:
#sign-off, per Dan via phone (6 Aug 2018).

PR #2960, opened 2018-08-10
Update driveitem_put_content.md

Reclad:
Change return code for an existing file. (200 OK not 201 CREATED).


Lauragra:
Thanks. Fixed in PR #5196

PR #3002, opened 2018-08-21
Billbl ignore testdocs files

billbliss:
As discussed over email


billbliss:
Closing so I can do the PR against staging instead

PR #3003, opened 2018-08-22
Adding more directory entities to dq

piotrci:
Added: application, directoryRole, servicePrincipal
Updated delta overview and changelog


piotrci:
#hold-off until code deployed to production

concepts/delta_query_overview.md
```diff
@@ -62,12 +62,15 @@ Delta query is currently supported for the following resources.
 
 | **Resource collection** | **API** |
 |:------ | :------ |
+| Applications | [delta](../api-reference/beta/api/application_delta.md) function of the [application](../api-reference/beta/resources/application.md) resource (preview) |
```
angelgolfer-ms: Hi @piotrci , please append "(preview)" to the resource phrase under the Resource collection column, if the delta query support is only in preview. That is, as an example, instead of "Applications", do "Applications (preview)".

angelgolfer-ms: Please also update each of the 3 resource type topics to include: 
(a) A sentence in the preamble to call out the support. You can take a look at the [user](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/resources/user) topic as an example.
>> Using delta query to track incremental additions, deletions, and updates, by providing a delta function.

(b) In the Methods table of each of the three resources, add a table row for delta.

piotrci: Done.

PR #3008, opened 2018-08-23
Billbl ignore testdocs files

billbliss:
Resubmitting PR to staging instead of master.


billbliss:
Was already reviewed and accepted by Daron Spektor.

PR #3024, opened 2018-08-28
New column for MSA support in permissions reference

madansr7:
This is needed to help generate permissions in OpenAPI description. Also updated the "remarks" section wherever applicable.


concepts/permissions_reference.md
```diff
@@ -121,9 +122,7 @@ None.
 
 <br/>
 
-### Remarks
-
-_Calendars.Read.Shared_ and _Calendars.ReadWrite.Shared_ are only valid for work or school accounts. All other permissions are valid for both Microsoft accounts and work or school accounts.
+> **Note:** _Calendars.Read.Shared_ and _Calendars.ReadWrite.Shared_ are only valid for work or school accounts. All other permissions are valid for both Microsoft accounts and work or school accounts.
```
dkershaw10: Do we need to say this if it's in the table above?

concepts/permissions_reference.md
```diff
@@ -71,20 +71,21 @@ None.
...
-| _Bookings.ReadWrite.All_ | Allows an app to read and write Bookings appointments, businesses, customers, services, and staff on behalf of the signed-in user. Does not allow create, delete, or publish of Bookings businesses. | Intended for management applications that manipulate existing businesses, their services and staff members. Cannot create, delete, or change the publishing status of a booking business. Typical target user is the support staff of an organization.| No |
-| _Bookings.Manage_ | Allows an app to read, write, and manage Bookings appointments, businesses, customers, services, and staff on behalf of the signed-in user.  | Allows the app to have full access. <br>Intended for a full management experience. Typical target user is the administrator of an organization.| No |
+|   Permission    |  Display String   |  Description | Admin Consent Required | Microsoft Account supported |
```
dkershaw10: Have you had any of this staged yet?  I wonder how this looks with 5 columns on our doc site.  Preview on GitHub does a better job of sizing the columns.  But on our doc site, the column sizes seem to be determined by the title row headings. For example the final column today is as wide as "Admin Consent Required", even when the content in the column is minimal. The doc site does not seem to require the title heading to be on one line.

concepts/permissions_reference.md
```diff
@@ -103,12 +104,12 @@ None.
 
 #### Delegated permissions
 
-|Permission    |Display String   |Description |Admin Consent Required |
-|:-----------------------------|:-----------------------------------------|:-----------------|:-----------------|
-| _Calendars.Read_ |Read user calendars |Allows the app to read events in user calendars. |No |
-| _Calendars.Read.Shared_ |Read user and shared calendars |Allows the app to read events in all calendars that the user can access, including delegate and shared calendars. |No |
-| _Calendars.ReadWrite_ |Have full access to user calendars |Allows the app to create, read, update, and delete events in user calendars. |No |
-| _Calendars.ReadWrite.Shared_ |Read and write user and shared calendars |Allows the app to create, read, update and delete events in all calendars the user has permissions to access. This includes delegate and shared calendars.|No |
+|   Permission    |  Display String   |  Description | Admin Consent Required | Microsoft Account supported |
```
dkershaw10: Is there something shorter we can use than "Microsoft Account supported"?  Can the intro talk about: You can check the **MSA Supported** column for each permission group to determine whether a specific permission is valid for Microsoft accounts (MSA), work or school accounts, or both.

I know abbreviations are generally frowned upon, but...

Lauragra: We can't use MSA as an abbreviation, FYI.

concepts/permissions_reference.md
madansr7: Haven't tried it in staging yet but i can give it a try.

concepts/permissions_reference.md
madansr7: I had it as MSA previously but had to expand due to the same concerns of using abbreviations. I think your idea of including an intro paragraph will help set the context.

concepts/permissions_reference.md
madansr7: I left the note because it was there orginally. Wasn't sure if it was put in to address a specific problem but i agree taking it out .

madansr7: @dkershaw10, I removed the inconsistent note.

concepts/permissions_reference.md
madansr7: working to get this in staging.

concepts/permissions_reference.md
madansr7: It is advised to not abbreviate the Microsoft brand to MS so we will use "Microsoft account supported" column name.

PR #3035, opened 2018-08-30
Security API v1.0

edwardkoval:


api-reference/v1.0/resources/alertfeedbackenumtype.md
```diff
@@ -0,0 +1,12 @@
+# alertFeedback enum type
```
angelgolfer-ms: @edwardkoval , can you please take a look at the feedback I sent you last night about this bunch of enum topics? Basically, we are not creating new enum topics unless they contain more value than what we can already incorporate in the resource type topics where the enums are referenced. We can talk more about that offline  :-)

PR #3102, opened 2018-09-17
Immutable ID feature

jasonjoh:


jasonjoh:
@Lauragra I'm waiting on confirmation on when the schema updates will go live.

Lauragra:
Okay, that's fine; either way, we should plan to go to staging before we merge to master. Thanks!

PR #3129, opened 2018-09-20
update permissions and registration info for Calls and Online Meetings

satyakonmsft:


api-reference/beta/resources/calls-api-overview.md
```diff
@@ -1,46 +1,92 @@
-﻿# Working with calls in Microsoft Graph
+# Working with Calls and Online Meetings in Microsoft Graph
```
Lauragra: This is not a proper noun so should not be capitalized.

concepts/calling-testing.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+> **IMPORTANT**: The Certificate Thumbprint provided above should match the Service FQDN. That is why the DNS entries are required.
+
+## Next Steps
+
+Your bot can now run locally and all the flows work from your localhost.
+
+## Caveats
+
```
kanchm: Add that this is working only for cloud services and not for local service fabric instance.

PR #3132, opened 2018-09-21
Clean up linefeeds and add references to new calls and meetings APIs

billbliss:


billbliss:
Added links to calling and online meeting APIs from Teams content, did some cleanup while I was in there.

PR #3153, opened 2018-09-24
Add information about automatically generated display names

MadsSibeko:


api-reference/beta/api/contact_update.md
```diff
@@ -47,7 +47,7 @@ In the request body, supply the values for relevant fields that should be update
 |children|String||
 |companyName|String|The name of the contact's company.|
 |department|String|The contact's department.|
-|displayName|String|The contact's display name.|
+|displayName|String|The contact's display name. Note that later updates to other properties may cause an automatically generated value to overwrite the DisplayName value you have specified.|
```
bullmastiffo: Maybe we should add information, that API user can preserve the existing value by supplying it in update call body.

angelgolfer-ms: Please use "displayName" (lower camel case) instead of "DisplayName". The case needs to reflect what's actually used in the API name.

api-reference/beta/resources/contact.md
```diff
@@ -79,7 +79,7 @@ Here is a JSON representation of the resource
 |companyName|String|The name of the contact's company.|
 |createdDateTime|DateTimeOffset|The time the contact was created. The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`|
 |department|String|The contact's department.|
-|displayName|String|The contact's display name.|
+|displayName|String|The contact's display name. You can specify the display name with Create or Update. Note that later updates to other properties may cause an automatically generated value to overwrite the display name value you have specified.|
```
angelgolfer-ms: Please rephrase the first sentence as "You can specify the display name in a create or update operation." Please add a relative link to "create" pointing at the POST operation topic (../api/user_post_contacts.md), and a relative link for "update" to point to the PATCH operation topic (../api/contact_update.md).
Generally, in documentation, please use lower case instead of upper case. See the Microsoft Cloud Style Guide on [Capitalization](https://worldready.cloudapp.net/Styleguide/Read?id=2696&topicid=25357). API names must match the actual case (and in the case of Graph, most API names are in lower camel case).

angelgolfer-ms:
HI @MadsSibeko , thanks for the update. I reviewed your changes. Sent you one last question by email; please take a look.

PR #3211, opened 2018-10-04
Add readOnly property to directoryDefinition

MickiMau5:
Add the readOnly property to the definition


dkershaw10:
@lleonard-msft Sign-off

PR #3213, opened 2018-10-04
Update query_parameters.md - expand example update

MIchaelMainer:
The link to Graph Explorer was missing the ?


Lauragra:
Good catch!

PR #3247, opened 2018-10-11
Add directory permissions to list groups endpoint

fiznool:
You can also use the permissions `Directory.Read.All` and `Directory.ReadWrite.All` to access the `GET /groups` endpoint.


dkershaw10:
@lleonard-msft Signed-off.

PR #3248, opened 2018-10-11
Make the POST example valid

jsweiler:
If you post the current example you get an error. I've found that the change suggested here https://stackoverflow.com/a/48609918/2171587 works fine.

### DIFF
--- a/api-reference/beta/api/application_post_applications.md
+++ b/api-reference/beta/api/application_post_applications.md
@@ -45,8 +45,11 @@ Content-type: application/json
 Content-length: 67
 
 {
-  "allowPublicClient": true,
-  "displayName": "Display name"
+  "displayName": "MyCoolApp",
+  "passwordCredentials": [{
+    "endDateTime": "2020-10-19T17:59:59.53Z",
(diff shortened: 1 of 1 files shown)


dkershaw10:
@sureshja Please take a look at this request. Typically the POST topics have a "required properties" section - this topic is missing this section.  While Jeff's suggestion is good, `passwordCredential` is not  a required property.  So it might be better, if we want to show this, to have an intro line something like - "Here is an example of creating a web app registration, with a secret." and put in all the fields for that type of app.
@jsweiler NOTE: the property descriptions in the docs are **now** correct (see https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/resources/passwordcredential)

dkershaw10:
@sureshja Please take a look at this PR.

Lauragra:
@sureshja , did you have a chance to take a look at this PR? Do we want to merge it or do we need changes?

Lauragra:
@davidmu1 , can you please help drive resolution on this PR? Thanks.

davidmu1:
Suresh is updating application articles. Closing this PR.

PR #3253, opened 2018-10-12
Update/aad group visibility details

lleonard-msft:
Adding block describing the effects of specific **.visibility** values.

@dkershaw10, @Lauragra, and @angelgolfer-ms: Please take a look (PTAL)


api-reference/beta/resources/group.md
```diff
@@ -114,6 +114,33 @@ This resource supports:
 |unseenCount|Int32|Count of posts that the current  user has not seen since his last visit.|
 |visibility|String| Specifies the visibility of an Office 365 group. Possible values are: **Private**, **Public**, **HiddenMembership**, or empty (which is interpreted as **Public**).|
```
angelgolfer-ms: @lleonard-msft , can you please double-check if the enum values are capitalized? All programmability names in Graph should be lower camel case. If the product team made them upper camel case, per Yina, we need the product team to change the case in the metadata.

api-reference/beta/resources/group.md
```diff
@@ -114,6 +114,33 @@ This resource supports:
...
 
+### Visibility property definitions
```
angelgolfer-ms: I'd suggest a title like "The visibility of a group". Basically we're elaborating the possible values of an enumeration. We usually don't call those values out as "definitions". :-)

api-reference/beta/resources/group.md
```diff
@@ -114,6 +114,33 @@ This resource supports:
...
+
+The visibility of Unified groups (`"grouptypes" = ["Unified"]`) can be controlled through the **visibility** property. Here's the impact of each supported value:
```
angelgolfer-ms: If there is an enum defined for these possible values, then to be consistent, you don't need the lead-in sentence. Please refer to the example in the v1.0 mailtips resource topic in my last comment.

If there is no enum involved here, then I suggest a simpler lead-in sentence that uses more commonly used terms, to align with Microsoft Voice:
"Here's what visibility means for Office 365 groups."
BTW, if you were to keep "unified groups", the 'u' should be lower case please.
I'd also suggest using a table, just to tighten the vertical space. I'd suggest the column headers "Value" and "Description". BTW if this indeed involves an enum, then yes, we'll have to use a table instead of paragraphs and lists, and we'll be using those column headers for consistency.

api-reference/beta/resources/group.md
```diff
@@ -114,6 +114,33 @@ This resource supports:
...
+
+- `HiddenMembership`
+
+    - Owner permission is needed to join the group.  
+    - Non-members cannot view the contents of the group.  
+    - Non-members cannot see the members of the group.   
+    - Global/company administrators, user administrators, and helpdesk administrators can view the membership of the group.
+    - Group appears in the global address book (GAL).
+
+Visibility can be set only when a group is created; it is not editable.
```
angelgolfer-ms: I suggest you roll the info of these last two sentences into the property description. That would be consistent with other group properties that have this level of detail in their property description table cell, e.g., displayName, groupTypes. Then we can dedicate this section just for the possible values of visibility.

api-reference/beta/resources/group.md
lleonard-msft: Fixed.

api-reference/beta/resources/group.md
lleonard-msft: @angelgolfer-ms - Reviewing the metadata, I see group.visibility defined as an "Edm.String."  Further, there is no relevant enum listed in enums.md.

Feedback incorporated.

api-reference/beta/resources/group.md
lleonard-msft: Feedback incorporated.

api-reference/beta/resources/group.md
lleonard-msft: Changes as requested, though I believe some clarity has been lost.

Feedback incorporated.

api-reference/v1.0/resources/group.md
```diff
@@ -99,11 +99,22 @@ This resource supports:
...
 |securityEnabled|Boolean|Specifies whether the group is a security group. If the **mailEnabled** property is also true, the group is a mail-enabled security group; otherwise it is a security group. Must be **false** for Office 365 groups. Supports $filter.|
 |unseenCount|Int32|Count of posts that the current  user has not seen since his last visit.|
-|visibility|String| Specifies the visibility of an Office 365 group. The possible values are: **Private**, **Public**, **HiddenMembership**, or empty (which is interpreted as **Public**).|
+|visibility|String| Specifies the visibility of an Office 365 group. Possible values are: `private`, `public`, or `hiddenmembership`; blank values are treated as public.  See [Group visibility options](#group-visibility-options) to learn more.<br>Visibility can be set only when a group is created; it is not editable.<br>Visibility is supported only for unified groups; it is not supported for security groups.|
```
dkershaw10: the values for visibility are case insensitive (on set).  Graph returns them initial caps camel-casing, for this property anyways

dkershaw10:
@lleonard-msft Signed-off

PR #3255, opened 2018-10-12
Delta for directory objects

piotrci:
We are porting AAD Graph functionality into MS Graph. it allows a single delta query on 3 entities: user, group, orgContact

https://blogs.msdn.microsoft.com/aadgraphteam/2013/04/08/differential-query-in-windows-azure-active-directory-graph/


api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
+# directoryObject: delta
```
dkershaw10: @lleonard-msft told me not to put colons in our titles in my latest PR.  We have this everywhere, but not sure if we are trying to shift this.  Might want to update the title.

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+Get newly created, updated, or deleted directory objects of the following types: [user](../../../api-reference/beta/resources/user), [group](../../../api-reference/beta/resources/group) and [organizational contact](../../../api-reference/beta/resources/orgcontact), in a single delta query. See [Track changes](../../../concepts/delta_query_overview.md) for details.
+
+## Permissions
+
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](../../../concepts/permissions_reference.md).
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | Directory.Read.All, Directory.AccessAsUser.All  |
+|Delegated (personal Microsoft account) | User.Read, User.ReadWrite    |
```
dkershaw10: This is kind of weird.  Firstly does delta really work for personal accounts?  Second, even if it did, directoryObjects make little sense in the consumer world.

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+
+Adding an optional request header - `prefer:return=minimal` - results in the following behavior:
+
+- If the property has changed, return the property in the JSON response.
+- If the property has been set to an empty value, return the property value as null.
+- If the property has not changed, do not include the property in the JSON response. (Different from the default behavior.)
+
+> **Note:** The header can be added to a `deltaLink` request at any point in time in the delta cycle. The header only affects the set of properties included in the response and it does not affect how the delta query is executed. See the [third example](#request-3) below.
+
+### Example
```
dkershaw10: This should be an H2.
Request/Responses should be H3s

concepts/delta_query_overview.md
```diff
@@ -1,22 +1,22 @@
...
+3. When the application needs to learn about changes to the resource, it makes a new request using the `deltaLink` URL received in step 2. This request *may* be made immediately after completing step 2 or when the application checks for changes.
+4. Microsoft Graph returns a response describing changes to the resource since the previous request, and either a `nextLink` URL or a `deltaLink` URL.
 
->**Note:** Entities derrived from `directoryObject` (such as users and groups) support "sync from now" scenarios. This allows you to skip steps 1 and 2 above (if you are not interested in retrieving the full state of the resource) and ask for the latest `deltaLink` instead. Append `$deltaToken=latest` to the `delta` function and the response will contain a `deltaLink` and no resource data.
+>**Note:** Entities stored in Azure Active Directory (such as users and groups) support "sync from now" scenarios. This allows you to skip steps 1 and 2 above (if you are not interested in retrieving the full state of the resource) and ask for the latest `deltaLink` instead. Append `$deltaToken=latest` to the `delta` function and the response will contain a `deltaLink` and no resource data.
```
dkershaw10: I think we normally refer to these things as Resources rather than Entities.

dkershaw10:
@lleonard-msft Added a few minor comments, but otherwise if those are addressed, this is good, so signed-off.

api-reference/beta/api/directoryobject_delta.md
piotrci: We have this in all delta articles so I'll leave it as is. We can fix them all if needed.

api-reference/beta/api/directoryobject_delta.md
piotrci: Good catch. Copy and paste error on my side.

api-reference/beta/api/directoryobject_delta.md
piotrci: Done.

concepts/delta_query_overview.md
piotrci: Done.

api-reference/beta/api/directoryobject_delta.md
angelgolfer-ms: @dkershaw10 , if you're referring to the topic titles like "application: delta", FYI this is by design. This is how Sudhi's tool generates topics for actions and functions. It's also in the onboarding docs (can provide you URL if you can't find it :-)).

angelgolfer-ms: However, @piotrci , my question related to this is, is there going to be a delta function added for directoryObject? If not, we should expose this content as a concept topic instead.

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+|:---------------|:----------|
+| Authorization  | Bearer &lt;token&gt;|
+| Content-Type  | application/json |
+| Prefer | return=minimal <br><br>Specifying this header with a request that uses a `deltaLink` would return only the object properties that have changed since the last round. Optional. |
+
+## Request body
+
+Do not supply a request body for this method.
+
+### Response
```
angelgolfer-ms: @piotrci , this "Response" section should be an H2. Please also adjust its sub-sections and make them H3's, not H4's.

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### Alternative: return only the changed properties
+
+Adding an optional request header - `prefer:return=minimal` - results in the following behavior:
+
+- If the property has changed, return the property in the JSON response.
+- If the property has been set to an empty value, return the property value as null.
+- If the property has not changed, do not include the property in the JSON response. (Different from the default behavior.)
+
+> **Note:** The header can be added to a `deltaLink` request at any point in time in the delta cycle. The header only affects the set of properties included in the response and it does not affect how the delta query is executed. See the [third example](#request-3) below.
```
angelgolfer-ms: @piotrci , please change the link to point to the second example, not the third. Your second example is the one that uses the Prefer header now.

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+      <...response trimmed for brevity...>
+    },
+    <...response trimmed for brevity...>
+  ]
+}
+```
+
+### Request 2
+
+The next example shows the use of the alternative minimal response behavior:
```
angelgolfer-ms: @piotrci , please explicitly call attention to the Prefer header in this example. Some customers just scan the topic and look for the sample code. They may not notice there's a Prefer header in the sample request.
I would suggest wording it like this:
"The next example uses the `Prefer: return=minimal` header to get only the changed properties."

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+      "businessPhones": "12345"
+    },
+    <...response trimmed for brevity...>
+  ]
+}
+```
+
+### Request 3
+
+The next example shows the initial request using the `isOf` operator to filter out only user and group entities:
```
angelgolfer-ms: Please bold "user" and "group" because they are entities in the Graph API. 
Please also rephrase the request description as:
"The next example shows the initial request using the `$filter` query option with the `isOf` operator to get instances of the **user** or **group** entity that have been updated."

api-reference/beta/api/directoryobject_delta.md
```diff
@@ -0,0 +1,299 @@
...
+  "name": "directoryobject_delta"
+}-->
+
+```http
+GET https://graph.microsoft.com/beta/directoryObjects/delta?$filter=isOf('Microsoft.Graph.User')+or+isOf('Microsoft.Graph.Group')
+```
+
+### Response 3
+
+The following is an example of the response when using `deltaLink` obtained from the query initialization. Note that only user and group objects are returned:
```
angelgolfer-ms: @piotrci , in contrast with the 2 examples above, I'd suggest to replace the last sentence:
"Note that only **user** and **group** objects that have been updated are returned, and note that all the properties of these objects are returned."

api-reference/beta/api/directoryobject_delta.md
angelgolfer-ms: Also, for the request and response sections of all 3 examples, please do not use the ':' at the end of the lead-in paragraph. Please use the period, like the first request and response.

concepts/delta_query_overview.md
```diff
@@ -1,22 +1,22 @@
...
+3. When the application needs to learn about changes to the resource, it makes a new request using the `deltaLink` URL received in step 2. This request *may* be made immediately after completing step 2 or when the application checks for changes.
+4. Microsoft Graph returns a response describing changes to the resource since the previous request, and either a `nextLink` URL or a `deltaLink` URL.
 
->**Note:** Entities derrived from `directoryObject` (such as users and groups) support "sync from now" scenarios. This allows you to skip steps 1 and 2 above (if you are not interested in retrieving the full state of the resource) and ask for the latest `deltaLink` instead. Append `$deltaToken=latest` to the `delta` function and the response will contain a `deltaLink` and no resource data.
+>**Note:** Resources stored in Azure Active Directory (such as users and groups) support "sync from now" scenarios. This allows you to skip steps 1 and 2 above (if you are not interested in retrieving the full state of the resource) and ask for the latest `deltaLink` instead. Append `$deltaToken=latest` to the `delta` function and the response will contain a `deltaLink` and no resource data.
```
angelgolfer-ms: @piotrci - the next suggestion from me is to try to make the DQ overview topic more scannable. I know this is not something you added in this PR, but nonetheless, I find overall our DQ support story has quite a few exceptions here and there, I hope the topic flow would make the story a bit easier to pick up. 
Please make the existing note about "sync from now" a sub-section (H3) instead. Assuming the "sync from now" term is a reasonable term to use to describe this capability, please integrate the existing note like the following:
"### Sync from now
As an alternative to the typical call pattern, you can start tracking changes without doing a full initial GET of the resource.  This capability applies to resources stored in Azure Active Directory  (such as **user** and **group** instances). Instead of  steps 1 and 2 above, ask for the latest deltaLink instead. Append `$deltaToken=latest` to the **delta** function and the response will contain a `deltaLink` and no resource data. "

I think we should include an example. Assuming it's a short request and response, it's weight is ok for the overview topic. At this point of the overview topic, customers may not have learned what a typical round is like, the current level of description may be not too helpful, so a short example would explain better. Alternatively you can add an example in one of the delta function topics, and link to that example, please.

PR #3302, opened 2018-10-26
Updates in Planner overview and planner task

dzhavat:
The reason I have removed the anchor tags around the titles in Planner overview is because it looks like these extra tags break the shortcuts displayed in the sidebar for that particular document (that's my theory at least). Like in this screenshot:

image

As you can see the `Track changes using delta query` and `Populate the object cache for delta queries` titles are missing from the sidebar and they are the only ones that are wrapped in an `a` tag.


Lauragra:
Closing this PR. I don't repro the first issue. The second issue I will open a PR to fix in both the beta and v1.0 topics. Thanks.

PR #4309, opened 2019-04-03
Added descriptions and syntax for $value.

angelgolfer-ms:


api-reference/beta/api/attachment-get.md
```diff
@@ -19,6 +21,11 @@ An attachment can be one of the following types:
 All these types of attachment resources are derived from the [attachment](../resources/attachment.md)
 resource.
 
+### Get the raw contents of a file or item attachment
+You can append the path segment `/$value` to get the raw contents of a file or item attachment. For a file attachment, the content type is based on its original content type. For a message-type item attachment, the raw contents include the MIME representation of that message. 
+
+Attempting to get the `$value` of a reference attachment returns HTTP 405.
```
angelgolfer-ms: @svpsiva , can you please confirm if HTTP 405 is the intended response code, as opposed to HTTP 501 as you mentioned before? I got HTTP 405 in Graph Explorer.

svpsiva: Hi,

405 is the intended response code.

svpsiva: 405 is the correct response code. Pls correct.

api-reference/beta/api/attachment-get.md
```diff
@@ -19,6 +21,11 @@ An attachment can be one of the following types:
 All these types of attachment resources are derived from the [attachment](../resources/attachment.md)
 resource.
 
+### Get the raw contents of a file or item attachment
+You can append the path segment `/$value` to get the raw contents of a file or item attachment. For a file attachment, the content type is based on its original content type. For a message-type item attachment, the raw contents include the MIME representation of that message.
```
svpsiva: Item attachments can be of multiple types. They could be a Message, Contact, Event or Task. We can just change this to "For an item attachment, the raw contents include the MIME representation of that item. "

angelgolfer-ms: @svpsiva , I have further questions about the format of the raw content for item types other than message. For example, do contact attachments get returned in MIME format? I showed an example in an email to you on 4/8. Please take a look and re-confirm.

PR #4325, opened 2019-04-04
Update chatMessage documentation to match code

nkramer:


nkramer:
Oddly enough we had the change log entry for removing the deleted property, but never actually removed it from the reference page.

Lauragra:
@nkramer , this is just a correction, so no update to the changelog needed?

PR #4332, opened 2019-04-08
Fix group-post-events.md

rykoma:
Fixed incorrect request and response body.


Lauragra:
@dkershaw10 , please review.

dkershaw10:
@Lauragra Same again - this is about Outlook calendar events.

Lauragra:
@angelgolfer-ms , can you please review this PR? Also, if accepted. I assume we need to make the same update to the beta topic.

rykoma:

angelgolfer-ms:
@rykoma , thank you for creating this pull request.
There are a few issues in this pull request:
(1) The example response has an etag value that needs escaping the double quote. This causes an API Doctor error.
(2) Even though you updated the example request with more real property values, the example response doesn't contain most of the properties that would be typically included in the response. Normally it would be more helpful to have the response reflect what's being set in the request body.
(3) The PR includes only v1 but not beta.

For the above issues, I decided to do a follow-up on the examples for v1 and beta myself: PR #4634 .
Sorry for the delay in responding and looking forward to your continued support.

rykoma:
@angelgolfer-ms 
Thank you for responding to multiple pull requests that were stuck!
I really appreciate your cooperation.

PR #4333, opened 2019-04-08
Fix group-update-event.md

rykoma:
Fixed incorrect request and response body.


Lauragra:
@dkershaw10 , please review.

dkershaw10:
@Lauragra You need someone from the outlook team to review this.  This isn't my area.

Lauragra:
@angelgolfer-ms , can you please review this PR? Also, if accepted. I assume we need to make the same update to the beta topic.

rykoma:

api-reference/v1.0/api/group-update-event.md
```diff
@@ -46,31 +46,37 @@ The following is an example of the request.
 }-->
 ```http
 PATCH https://graph.microsoft.com/v1.0/groups/{id}/events/{id}
+Prefer: outlook.timezone="Pacific Standard Time"
```
angelgolfer-ms: Hi @rykoma , thank you for your PR.
Thanks for pointing out that the return code on success is HTTP 200 and not HTTP 204.
There are 2 areas in your PR that are incorrect:
(a) The Prefer: outlook.timezone request header should be used in only [GET ](https://docs.microsoft.com/en-us/graph/api/event-get?view=graph-rest-beta&tabs=cs) event and [list ](https://docs.microsoft.com/en-us/graph/api/user-list-events?view=graph-rest-beta&tabs=cs) events operations, which would display the timezone property of the start and end properties as the time zone specified in the Prefer header. 
That Prefer header doesn't do anything for the PATCH operation.

(b) The way that you specified the etag value in line 74, without escaping the backstroke character, is causing an API Doctor error. In general, PRs must pass API Doctor in order to be merged.

api-reference/v1.0/api/group-update-event.md
```diff
@@ -46,31 +46,37 @@ The following is an example of the request.
...
+  "@odata.type": "microsoft.graph.event"
 } -->
 ```http
-HTTP/1.1 204 No Content
+HTTP/1.1 200 OK
+
+{
+	"@odata.context":"https://graph.microsoft.com/v1.0/$metadata#groups('7fe8323e-1b24-406c-924c-9d9ccf2bc13d')/events/$entity",
+	"@odata.etag":"W/"P8tzXbof6Eyv1DvnV4lZKgAAAAAOoQ=="",
```
angelgolfer-ms: @rykoma , if you look at the [API Doctor log](https://dev.azure.com/msrobots/apidoctor/_build/results?buildId=3668), you'd find that in line 74, the value of etag, the double quotes inside the string should be escaped with a backstroke.

rykoma:
@angelgolfer-ms 
Thank you very much for the review, and I'm sorry for the trouble. Thank you for your cooperation.

angelgolfer-ms:
> @angelgolfer-ms
> Thank you very much for the review, and I'm sorry for the trouble. Thank you for your cooperation.

@rykoma , no problem, thank you for your patience and we look forward to your continued support!
BTW, I just merged PR #4633 to fix the issue you uncovered in the response code, and gave a real REST example. The updates should be live on the next publishing day (weekday around 3pm PST).

PR #4343, opened 2019-04-09
Fix calendargroup-post-calendars.md

rykoma:
Fix incorrect request body.


rykoma:

api-reference/v1.0/api/calendargroup-post-calendars.md
```diff
@@ -70,9 +70,7 @@ Content-type: application/json
 Content-length: 78
 
 {
-  "name": "name-value",
-  "color": {
-  }
+  "name": "name-value"
```
angelgolfer-ms: @rykoma , thanks for the PR. The default REST code examples are generated and are known to be not necessarily correct with the request bodies, so thanks for bringing this issue up with this example.
I merged your pull request and it will go live on the next publishing day (weekday around 3pm PST).
FYI I will proceed to finish this example:
(1) Your fix is only in v1.0. This fix applies to both v1 and beta. I will add the fix to beta as well.
(2) It's a good call for you to remove setting color this way as the property should always have a value that's part of the enum calendarColor. However, it would be more helpful if the example response is also updated to show a valid color property value. For this reason, I will follow up with a PR to show a more realistic example response.

rykoma: @angelgolfer-ms, thank you for merging my PR.
I understood your comment. Thank you very much for following up this PR.

angelgolfer-ms: Follow up PR is #4632 .

PR #4377, opened 2019-04-11
Adds risk history details + new reference topic

cloudhandler:


api-reference/beta/resources/riskyuserhistoryitem.md
```diff
@@ -14,9 +14,9 @@ ms.prod: "microsoft-identity-platform"
 
 | Property       | Type    | Description |
 |:---------------|:--------|:------------|
-| userId         | string  |             |
```
angelgolfer-ms: @cloudhandler , 
Please add a description to the beginning of the topic (around line 12) for this riskyUserHistoryItem entity type as well. 
Since you're adding a GET method in this PR (the metadata doesn't reveal this level of info consistently), please add a "Methods" section to this topic and list the get method. You can please look at other published topics and copy their layout for the "Methods" section, like message.md.
Please also fill in the YAML header of the file (lines 3 and 4).

api-reference/beta/api/riskyusershistory-get.md
```diff
@@ -0,0 +1,96 @@
+---
```
angelgolfer-ms: According to the beta metadata, there's only the riskyUserHistoryItem entity, there's no riskyUserHistory. 

Since the related navigation property, "history", that's been added to riskyUser returns a **collection** of history items and not a single item, I think you'll have to add both a "Get" topic (which you have here), and a "List" topic.

Other than "get" and "list", are there any other methods supported by riskyUserHistoryItem? Are there "update" and "delete"? 

For now, let's switch back to what we're sure of - "get" and "list". Please provide 2 files with the following filenames and topic titles:
For "get": riskyuserhistoryitem-get.md, "Get history item of riskyUser"
For "list": riskyuser-list-history.md, "List history of riskyUser".

To proceed, you might want to rename the file riskyuserhistory-get.md to riskyuser-list-history.md. My comments below will assume that you're making this the file for "list". You'll please separately create another file for "get" and we'll review that one afterwards.

api-reference/beta/api/riskyusershistory-get.md
```diff
@@ -0,0 +1,96 @@
...
+title: "Get history of riskyUser"
```
angelgolfer-ms: Please change to "List history of riskyUser", both in lines 2 and 8.

api-reference/beta/api/riskyusershistory-get.md
```diff
@@ -0,0 +1,96 @@
...
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | IdentityRiskyUser.Read.All    |
+|Delegated (personal Microsoft account) | Not supported.    |
+|Application | IdentityRiskyUser.Read.All |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+GET /riskyUsers('{userid}')/history
```
angelgolfer-ms: I was informed some time ago (by @daspek ) that we decided to use the "ID by segment" notation, that is, instead of identifying an id into a collection by what you have here in brackets, riskyUsers('{userId}'), we're supposed to identify this by a segment: riskyUsers/{userId}.
AFAIK both should work, the point here is to standardize on one across the docs.

api-reference/beta/api/riskyusershistory-get.md
```diff
@@ -0,0 +1,96 @@
...
+|:----------|:----------|
+| Authorization  | Bearer {token}. Required. |
+| Workbook-Session-Id  | Workbook session ID that determines whether changes are persisted. Optional.|
+
+## Request body
+Do not supply a request body for this method.
+
+## Response
+
+If successful, this method returns a `200 OK` response code and a [riskyUsersHistoryItem](../resources/riskyuserhistoryitem.md)  in the response body.
```
angelgolfer-ms: Line 43 the response section should read: 
"... and a collection of [riskyUsersHistoryItem](../resources/riskyuserhistoryitem.md) objects in the response body."

api-reference/beta/api/riskyusershistory-get.md
```diff
@@ -0,0 +1,96 @@
...
+
+{
+  "id": "f61fad1b-cf1b-4ce5-ba25-957ac919ddeb636699887174968737",
+  "isDeleted": false,
+  "isGuest": false,
+  "riskLevel": "high",
+  "riskState": "atRisk"
+  "riskDetail": "none",
+  "riskLastUpdatedDateTime": "2018-08-16T03:58:37.4968737Z",
+  "userDisplayName": "John Nash",
```
angelgolfer-ms: Is "John Nash" a name that's returned by the CELA fictictious names tool please? 
I'm not familiar with that name and that's why I asked.
In general, we should use only fictitious names generated by CELA's took: https://microsoft.sharepoint.com/sites/LCAWeb/Home/Copyrights-Trademarks-and-Patents/Trademarks/Fictitious-Names-Finder

api-reference/beta/api/riskyusershistory-get.md
```diff
@@ -0,0 +1,96 @@
...
+}-->
+```http
+GET https://graph.microsoft.com/beta/riskyUsers('f61fad1b-cf1b-4ce5-ba25-957ac919ddeb')/history
+```
+##### Response
+Here is an example of the response.
+<!-- {
+  "blockType": "response",
+  "truncated": true,
+  "@odata.type": "microsoft.graph.riskyUser"
```
angelgolfer-ms: Please add the line, say after line 59, 
"isCollection": true,

This tells API Doctor to expect a collection of objects, not a single object, in the example response payload.

api-reference/beta/api/riskyusershistory-get.md
angelgolfer-ms: Please also format the body of the example response as an array of objects, using the value keyword and the square brackets, and not as a single item, like you have it here.
If necessary, you can find a published topic like user-list-messages.md to see how to format the response payload.
That being said, in general, it would be best if you can try the API in say Graph Explorer with some test tenant that you guys might have, or a demo tenant that already has sample data, and copy and paste the actual request and response to the docs. Then there's less chance of typos, etc.

api-reference/beta/api/riskyusershistory-get.md
angelgolfer-ms: And before I forget, after you confirm the exact methods that riskyUserHistoryItem supports and adds those method topics, we should also update the TOC.
If it turns out that only the "list" and "get" topics are in scope and your team is not planning to add more methods any time soon, you can choose to add them to the existing TOC node for "Risky users" and not add a separate node for history. If we just use the existing "Risky users" node, then please add them this way:
Risky users
  List
  Get
  List history
  Get history item
  Confirm compromised
  Dismiss 



 (we should also modify the TOC title of the existing "Dismiss" TOC title, since we're dimissing the

angelgolfer-ms: I also noticed an existing topic title issue for "Confirm compromised" and "Dismiss".
Please update the topic titles for the corresponding topics to use singular riskUser, not plural:
Change from "Confirm riskyUsers compromised" to "Confirm riskyUser compromised"
Change from "Dismiss riskyUsers" to "Dismiss riskyUser".

cloudhandler: Thanks Angela. The only two methods we support are get and lsit for history. How do I update the TOC?

api-reference/beta/api/riskyusershistory-get.md
cloudhandler: updated to an approved fictitious name, Helen Poulsen

cloudhandler:
@angelgolfer-ms - Are there any additional changes I need to make? Thanks!

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,98 @@
...
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | IdentityRiskyUser.Read.All    |
+|Delegated (personal Microsoft account) | Not supported.    |
+|Application | IdentityRiskyUser.Read.All |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+LIST /riskyUsers/{userid}/history
```
dkershaw10: LIST -> GET
{userid} -> {id}  
NOTE if your API supports UPN as id, then it should be {id | userPrincipalName}

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+
+## Request headers
+| Name      |Description|
+|:----------|:----------|
+| Authorization  | Bearer {token}. Required. |
+| Workbook-Session-Id  | Workbook session ID that determines whether changes are persisted. Optional.|
```
dkershaw10: Remove this line.

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,98 @@
...
+  "blockType": "response",
+  "truncated": true,
+  "isCollection": true,
+  "@odata.type": "microsoft.graph.riskyUser"
+} -->
+```http
+HTTP/1.1 200 OK
+Content-type: application/json
+
+{
```
dkershaw10: This is not a list response.  Please look at other list examples.
There should be 
```json
{
  "values": [
     {
         <record data>
     }
  ]
}
```

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | IdentityRiskyUser.Read.All    |
+|Delegated (personal Microsoft account) | Not supported.    |
+|Application | IdentityRiskyUser.Read.All |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+GET /riskyUsers/{userid}/history
```
dkershaw10: syntax is not correct. 
Should be:  `GET /riskyUsers/{userid}/history/{id}`

Also please check with Jeffrey on whether this is supported - in API review it says GET is not supported, but list is.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Get history item of riskyUser"
+description: "Retrieve the history of a **riskyUser** object."
+localization_priority: Normal
+author: "cloudhandler"
+ms.prod: "microsoft-identity-platform"
+---
+# Get riskyUser History
```
dkershaw10: lower case the History

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,98 @@
+---
+title: "List history of riskyUser"
+description: "Retrieve the risk history"
+localization_priority: Normal
+author: "cloudhandler"
+ms.prod: "microsoft-identity-platform"
+---
+# List riskyUser History
```
dkershaw10: lower case the History

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,98 @@
...
+```http
+GET https://graph.microsoft.com/beta/riskyUsers/f61fad1b-cf1b-4ce5-ba25-957ac919ddeb/history
+```
+##### Response
+Here is an example of the response.
+<!-- {
+  "blockType": "response",
+  "truncated": true,
+  "isCollection": true,
+  "@odata.type": "microsoft.graph.riskyUser"
```
angelgolfer-ms: Line 62 - please change it to:
"@odata.type": "microsoft.graph.riskyUserHistoryItem"

Lines 61 and 62 tell API Doctor the type (which is a collection of riskyUserHistoryItem) that is returned in the example response payload.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: @cloudhandler , aside from dkershaw10's comment (that the response payload for a list operation is a collection of objects rather than a single object), please double-check if this operation returns indeed a collection of riskyUserHistoryItem objects. I would expect to see a collection of objects, and each object would just have the 3 properties userId, initiatedBy, and acitivity. There shouldn't be properties of a riskyUser object in the response.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Retrieve the history of a **riskyUser** object.
```
angelgolfer-ms: @cloudhandler , if you verify that there is indeed a GET operation for a riskyUserHistoryItem (in other words, this topic should exist), then, line 12 - please change that to:
"Get a riskyUserHistoryItem object of a riskyUser."

And, please provide links to riskyUserHistoryItem  and riskyUser in that description.
In general, in a topic, we need to provide a link to the first mention of any programmability element (resource, method) in the Graph API.

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: @cloudhandler , further refinement for line 8:
Please title this topic as "Get riskyUserHistoryItem".

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+```http
+GET /riskyUsers/{userid}/history
+```
+
+
+## Request headers
+| Name      |Description|
+|:----------|:----------|
+| Authorization  | Bearer {token}. Required. |
+| Workbook-Session-Id  | Workbook session ID that determines whether changes are persisted. Optional.|
```
angelgolfer-ms: Remove line 36 please.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Get history item of riskyUser"
+description: "Retrieve the history of a **riskyUser** object."
```
angelgolfer-ms: Line 3: please change to "Get a history item of a riskyUser object".
Alternatively, you can just use the same description as in line 12. For SEO, you can vary between the "description" value and actual topic title a bit.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Response
+
+If successful, this method returns a `200 OK` response code and a collection of  [riskyUserHistoryItem](../resources/riskyUserHistoryItem.md) objects in the response body.
+## Example
+##### Request
+Here is an example of the request.
+<!-- {
+  "blockType": "request",
+  "name": "get_riskyuser",
```
angelgolfer-ms: Even though the "name" is totally controlled by the author, it's generally good practice to customize the "name" value in line 49 to match what your example tries to do. For example:
"name": "get_riskyuser_historyitem",
One benefit is in case API Doctor identifies an error in this example, its error message would include this "name" value, and you can quickly identify which example has the error.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+  "sampleKeys": ["c2b6c2b9-dddc-acd0-2b39-d519d803dbc3"]
+}-->
+```http
+GET https://graph.microsoft.com/beta/riskyUsers/c2b6c2b9-dddc-acd0-2b39-d519d803dbc3/history
```
angelgolfer-ms: Line 53 - the example request needs to be consistent with the "HTTP request" section. In this case, if this operation is indeed supported, then you need to specify a history item ID at the end of the query URL.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+```
+##### Response
+Here is an example of the response.
+<!-- {
+  "blockType": "response",
+  "truncated": true,
+  "isCollection": true,
+  "@odata.type": "microsoft.graph.riskyUser"
```
angelgolfer-ms: Please remove line 60 - this operation, if supported, returns a single object, not an array of objects.

Line 61 - please change it to:
"@odata.type": "microsoft.graph.riskyUserHistoryItem"

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+} -->
+```http
+HTTP/1.1 200 OK
+Content-type: application/json
+
+{
+  "id": "c2b6c2b9-dddc-acd0-2b39-d519d803dbc3",
```
angelgolfer-ms: Lines 68-77 - the body of the response should represent a riskyUserHistoryItem object, which has the 3 properties userId, initiatedBy, and activity.
What you have there right now is a riskyUser object.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,90 @@
...
+  "userDisplayName": "Helen Poulsen",
+  "userPrincipalName": "helen@contoso.com"
+}
+```
+
+<!-- uuid: 8fcb5dbc-d5aa-4681-8e31-b001d5168d79
+2015-10-25 14:57:30 UTC -->
+<!-- {
+  "type": "#page.annotation",
+  "description": "Get riskyUsers",
```
angelgolfer-ms: Please update line 85 as same as the topic title.

api-reference/beta/api/riskyusers-confirmcompromised.md
```diff
@@ -6,7 +6,7 @@ localization_priority: Normal
 ms.prod: microsoft-identity-platform
 ms.date: 03/20/2019
 ---
-# Confirm riskyUsers compromised
+# Confirm riskyUser compromised
```
angelgolfer-ms: Line 9: Please retitle the topic as:
riskyUser: confirmCompromised

For reference topics, we generally try to display the related object, action, or function names. This is to help set the expectations that the topic is a reference topic rather than a conceptual topic. 
The TOC title can use a friendly name, like under the riskyUser node, "Confirm compromised", like we have it live now.
This is also how our stub topic generation tool would generate the topic title.

In case you weren't aware, the onboarding docs has guidance for topic titling: https://msgo.azurewebsites.net/add/document/guidelines/file-naming.html

api-reference/beta/api/riskyusers-dismiss.md
```diff
@@ -7,7 +7,7 @@ ms.prod: microsoft-identity-platform
 ms.date: 03/20/2019
 ---
 
-# Dismiss riskyUsers
+# Dismiss riskyUser
```
angelgolfer-ms: Similar to the confirmCompromised action topic, please change the topic title in line 10 as:
riskyUser: dismiss

api-reference/beta/resources/riskyuserhistoryitem.md
```diff
@@ -1,22 +1,31 @@
...
 # riskyUserHistoryItem resource type
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+Represents the risk history of Azure AD users as determined by Azure AD Identity Protection. 
+
+## Methods
+
+| Method   | Return Type|Description|
+|:---------------|:--------|:----------|
+|[Get riskyUser History](../api/riskyuserhistoryitem-get.md) | [riskyUser](riskyUser.md) collection|List the risk history of an Azure AD user.|
```
angelgolfer-ms: Line 18: if you verify that there is a GET riskyUserHistoryItem opertaion supported, then, please change the "Method" column to say:
Get riskyUserHistoryItem

The "Return type" column should be:
riskyUserHistoryItem 
and please provide a link for that.

And the Description column should use the same description for that operation, which would be something like:
"Get a riskyUserHistoryItem object of a riskyUser."
Please include links to riskyUserHistoryItem and riskyUser.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: Line 8: For better contrast with the topic title of the "get riskyUserHistoryItem" topic, I'd suggest changing the topic title of this topic to:
"List history of riskyUser"

api-reference/beta/resources/riskyuserhistoryitem.md
```diff
@@ -1,22 +1,31 @@
...
+|[List riskyUser History](../api/riskyuser-list-history.md) | [riskyUser](riskyUser.md)|Get the risk history of an Azure AD user.|
```
angelgolfer-ms: Line 19: Please change the "Method" column to say:
List history
The "Return type" column should be:
riskyUserHistoryItem collection
and please provide a link for that.
And the Description column should use the same description for that operation, which would be something like:
"Get the risk history of a riskyUser object."
Please include links to riskyUserHistoryItem and riskyUser.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: Also, keeping "List history" as the first part of the topic title matches better with the TOC title which is "List history". So changed the topic title to "List history of riskyUser".

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,98 @@
...
+
+If successful, this method returns a `200 OK` response code and a collection of [riskyUsersHistoryItem](../resources/riskyuserhistoryitem.md) objects in the response body.
+
+## Example
+##### Request
+Here is an example of the request.
+<!-- {
+  "blockType": "request",
+  "name": "get_userriskhitsory",
+  "sampleKeys": ["c2b6c2b9-dddc-acd0-2b39-d519d803dbc3"]
```
angelgolfer-ms: The value for "sampleKeys" is a collection of IDs that you specify in the request URL.
In your example, @cloudhandler , because you specify a query URL with f61fad1b-cf1b-4ce5-ba25-957ac919ddeb as the GUID of a riskyUser, you should replace "c2b6c2b9-dddc-acd0-2b39-d519d803dbc3" by " f61fad1b-cf1b-4ce5-ba25-957ac919ddeb".

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: I got HTTP 404 when I did that.

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: The example response payload should be a riskyUserHistoryItem object. Lines 67-78 looks like a riskyUser object to me.

api-reference/beta/resources/riskyuserhistoryitem.md
angelgolfer-ms: Line 18: After verifying that at least for now, the get history item operation returns 404, I am removing the table row in line 18.
If @cloudhandler later confirms that operation is supported, please add it back and incorporate the above comments.

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,141 @@
...
+description: "Retrieve the risk history"
+localization_priority: Normal
+author: "cloudhandler"
+ms.prod: "microsoft-identity-platform"
+---
+# List history of riskyUser
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Get the risk history of a [riskyUser](../resources/riskyuser.md) object.
```
dkershaw10: Nit: I thought we should be saying resource rather than object? Not that fussed to be honest.

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,141 @@
...
+
+>**Note:** Using the riskyUsers API requires an Azure AD Premium P2 license.
+
+## Permissions
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | IdentityRiskyUser.Read.All    |
```
dkershaw10: I believe IdentityRiskUser.ReadWrite.All would also work. It's available for app-only too.

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,141 @@
...
+|Delegated (personal Microsoft account) | Not supported.    |
+|Application | IdentityRiskyUser.Read.All |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+LIST /riskyUsers/{id}/history
```
dkershaw10: LIST -> GET
There's no such HTTP verb as LIST.

api-reference/beta/api/riskyuser-list-history.md
```diff
@@ -0,0 +1,141 @@
...
+  "@odata.type": "microsoft.graph.riskyUserHistoryItem"
+} -->
+```http
+HTTP/1.1 200 OK
+Content-type: application/json
+
+{
+    "@odata.context": "https://graph.microsoft.com/beta/$metadata#riskyUsers('41a31b00-3b3b-42d9-8f1c-6d4f14e74c69')/history",
+    "value": [
+        {
```
dkershaw10: Nit: Not sure that you need to show more than 2 items in the collection here.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,94 @@
+---
+title: "Get history item of riskyUser"
+description: "Get a history item of a riskyUser object."
+localization_priority: Normal
+author: "cloudhandler"
+ms.prod: "microsoft-identity-platform"
+---
+# Get riskyUserHistoryItem
```
dkershaw10: The titles aren't consistent.
Previous one for LIST says "List history of riskyUser"

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,94 @@
...
+Get a [riskyUserHistoryItem](../resources/riskyuserhistoryitem.md) object of a [riskyUser](../resources/riskyuser.md).
+
+>**Note:** Using the riskyUsers API requires an Azure AD Premium P2 license.
+
+## Permissions
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | IdentityRiskyUser.Read.All    |
```
dkershaw10: Same again with the permissions as above.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,94 @@
...
+| Name      |Description|
+|:----------|:----------|
+| Authorization  | Bearer {token}. Required. |
+
+## Request body
+Do not supply a request body for this method.
+
+## Response
+
+If successful, this method returns a `200 OK` response code and a collection of  [riskyUserHistoryItem](../resources/riskyuserhistoryitem.md) objects in the response body.
```
dkershaw10: This doesn't return a collection, but a single item. Please fix.

api-reference/beta/api/riskyuserhistoryitem-get.md
```diff
@@ -0,0 +1,94 @@
...
+## Example
+##### Request
+Here is an example of the request.
+<!-- {
+  "blockType": "request",
+  "name": "get_riskyuser_historyitem",
+  "sampleKeys": ["41a31b00-3b3b-42d9-8f1c-6d4f14e74c69"]
+}-->
+```http
+GET https://graph.microsoft.com/beta/riskyUsers/41a31b00-3b3b-42d9-8f1c-6d4f14e74c69/history/41a31b00-3b3b-42d9-8f1c-6d4f14e74c69
```
dkershaw10: Example looks suspicious because the 2 ids in the URL are the same.  Please make them different GUIDs

api-reference/beta/api/riskyusers-confirmcompromised.md
```diff
@@ -1,18 +1,18 @@
...
 ---
-# Confirm riskyUsers compromised
+# riskyUser: confirmCompromised
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
 >**Note:** The riskyUsers API requires an Azure AD Premium P2 license.
 
-Confirm a [riskyUser](../resources/riskyuser.md) object as compromised. This will set the targeted user's risk level to high.
+Confirm one or more [riskyUser](../resources/riskyuser.md) objects as compromised. This action sets the targeted user's risk level to high.
```
dkershaw10: In the ODATA world, "action" is correct, but in Graph docs, don't we call this a "method"?

api-reference/beta/resources/riskyuserhistoryitem.md
```diff
@@ -1,22 +1,30 @@
...
 # riskyUserHistoryItem resource type
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+Represents the risk history of Azure AD users as determined by Azure AD Identity Protection. 
+
+## Methods
+
+| Method   | Return Type|Description|
+|:---------------|:--------|:----------|
+|[List history](../api/riskyuser-list-history.md) | [riskyUserHistoryItem](riskyuserhistoryitem.md) collection|Get the risk history of an Azure AD user.|
```
dkershaw10: This is missing the other methods:
confirmCompromised and dismiss.  They should be added to the TOC too IMHO.

dkershaw10: @angelgolfer-ms Should this be here on the history item, or should it be on riskyUsers, as a method and navigation property?

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: Yes, certainly. I'll fix that.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: I'm adding IdentityRiskUser.ReadWrite.All to the Delegated (work or school) row and the Application row of the permissions table then.
I'll do the same to the riskyuserhistoryitem-get.md topic too, logically, just in case later on the product team verifies there's support and we don't overlook that permission before publishing that topic.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: I'd generally agree with you on that, @dkershaw10 .
In this case, however, I found it intriguing/ interesting that all history items have their id property based on the risky user's ID, one of the history items having its id the same as the risky user ID, and the rest of  the history items having ids that are risky user ID appended with more characters. So including a multiple of the latter shows the "pattern".
I don't know if there's any significance (as I only know the part of the feature that we're documenting), but it's worth @cloudhandler evaluating whether it's beneficial to elaborate in the docs about the history item ID value.

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: Yes, added that.

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: Yup, overlooked that.

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: Just to be clear, @dkershaw10 , the request here is what I tried in Graph Explorer, but the response payload is made-up. Per my email and email attachment, I actually got an HTTP 404 for this example request in Graph Explorer. I am only making up the example just in case this is a solely service-side bug and not a doc bug, then this topic is in a more ready state to go (to save @cloudhandler 's time and any chance we overlook anything missing in this topic before we publish).
Back to the example request - as you may have noticed, I got these ID values from the response payload of the list history example. Based on what we have documented, and assuming we support a get operation for history item, I would expect to be able to "get" any of the 3 history items returned from the list operation. It's a fact that the first history item from that list operation has the history item id property the same as the risky user's ID (userID property of the history item).
If indeed there's a special reason that a GET on that history item would fail, then we should document it.
Nonetheless, at this point so close to the publishing cutoff, unless someone can verify there's no problem with the GET operation, I am still planning NOT to include the riskyuserhistoryitem-get.md topic in the TOC nor riskyuserhistoryitem,md.

api-reference/beta/api/riskyusers-confirmcompromised.md
angelgolfer-ms: Yes, I agree with you too. This is my understanding:

- Technically "methods" refers to the CRUD operations. Methods do not include actions and functions.
- In the API reference topic template, however, the use of "Methods" as a section title is a loose application of the term for conciseness.

There are other places in the docs where authors have made the level of distinction between actions/functions and methods. _In general_, I think making a distinction has a benefit in setting expectations whether the API is using a nav property...

api-reference/beta/resources/riskyuserhistoryitem.md
angelgolfer-ms: Yes, I'll add this to the riskyUser resource topic Methods table too.
The history property is already in the "Relationships" section in that topic.

api-reference/beta/resources/riskyuserhistoryitem.md
angelgolfer-ms: I think Sarah took care of this.

api-reference/beta/api/riskyusershistory-get.md
angelgolfer-ms: I added List history to the TOC.
For now, we leave Get history item out of the TOC because your engineer confirmed it's not supported.

api-reference/beta/api/riskyusershistory-get.md
angelgolfer-ms: Just a note - I changed the syntax line to use GET /riskyUsers/{id}/history.
That works in Graph Explorer. It's also in the format that we're supposed to specify in the docs, generally.

api-reference/beta/api/riskyusershistory-get.md
angelgolfer-ms: Done.

api-reference/beta/api/riskyusershistory-get.md
angelgolfer-ms: Done.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: UPN didn't work for me. So I didn't add UPN.
The other two - done.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: done.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: Done.

api-reference/beta/api/riskyusers-confirmcompromised.md
angelgolfer-ms: Done.

api-reference/beta/api/riskyusers-dismiss.md
angelgolfer-ms: Done.

api-reference/beta/resources/riskyuserhistoryitem.md
angelgolfer-ms: Done with updating line 19. Removed line 18.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: Done.

api-reference/beta/api/riskyuser-list-history.md
angelgolfer-ms: Fixed finally!

api-reference/beta/api/riskyuserhistoryitem-get.md
angelgolfer-ms: For the topic titles, we try to use programmability names rather than friendly names.
So based on this principle, "Get riskyUserHistoryItem" is obvious.
"List history of riskyUser" follows that pattern too, as history is a programmability name representing a collection of riskyUserHistoryItem objects. 
Just like when we use the topic title "List messages". The topic generation tool knows to use "messages" and generates that topic title because that's the nav property name (and it just so happens to be friendly and grammatically correct).
Then there's the "of riskyUser" part of the topic title. That's more a human-added thing for clarity and SEO.
Related to this subject is TOC titles - we are more liberal with using "friendly" names rather than programmability names in the TOC :-)

PR #3919, opened 2019-02-04
Check all fixes and fixes to errors raised by apidoc

finsharp:

PR #4390, opened 2019-04-12
Verify beta documentation against latest apidoctor

finsharp:
Changes Test-Docs.ps1 to do the following

- Merged [PR3919](https://github.com/microsoftgraph/microsoft-graph-docs/pull/3919) into staging branch.
- Changed Test-Docs.ps1 to use ApiDoctor from github develop version. 
- Changed Test-Docs.ps1 to point to beta folder of documentation because pointing to the root causes errors to occur due to duplicate resources. (Exist in beta and v1.0).


api-reference/beta/api/accessreview-addreviewer.md
```diff
@@ -8,7 +8,7 @@ ms.prod: "microsoft-identity-platform"
 
 # Add accessReview reviewer
 
-[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md) ]
```
angelgolfer-ms: I'm removing this trailing space. The staging site confirms the space breaks the display of the beta disclaimer.

api-reference/beta/api/application-update.md
```diff
@@ -52,7 +52,7 @@ In the request body, supply the values for relevant fields that should be update
 |preAuthorizedApplications|[preAuthorizedApplication](../resources/preauthorizedapplication.md) collection| Lists applications and requested permissions for implicit consent. Requires an admin to have provided consent to the application. preAuthorizedApplications do not require the user to consent to the requested permissions. Permissions listed in preAuthorizedApplications do not require user consent. However, any additional requested permissions not listed in preAuthorizedApplications require user consent. |
 |requiredResourceAccess|[requiredResourceAccess](../resources/requiredresourceaccess.md) collection|Specifies resources that this application requires access to and the set of OAuth permission scopes and application roles that it needs under each of those resources. This pre-configuration of required resource access drives the consent experience. Not nullable.|
 |tags|String collection| Custom strings that can be used to categorize and identify the application. |
-|web|[web](../resources/web.md)| Specifies settings for a web application. |
+|web|[webApplication](../resources/webApplication.md)| Specifies settings for a web application. |
```
angelgolfer-ms: Just noting that webApplication.md is the name (with the incorrect case) of the file in the master branch, this PR focuses on getting the "GitHub version" of API Doctor to work and does not have a goal to fix all observed problems in the existing docs.

api-reference/beta/api/channel-get-message.md
```diff
@@ -95,8 +95,7 @@ Content-length: 201
   ],
   "mentions": [
       {
-          "type": "user",
-          "id": "id-value ",
+          "id": "Int32",
```
angelgolfer-ms: Just so we're aware, line 98 is in an example, not in a JSON definition. In an example payload, the returned id should be an integer value, without quotes in the JSON payload. It's confusing to customers if we present IDs that are integers this way in pseudo code.
Assuming API Doctor called this out as an issue (and that's why we're trying to change this) - I think part of the issue is actually in chatmessagemention.md. I'm changing the JSON definition there, from "id": "Int32" to "id": 1024, which is consistent with how the stub generator tool generates JSON definitions that contain integer-type properties.
Then in API topics like channel-get-message.md which specify response payloads that include a chatMessageMention type, I am changing the "id" property from "id": "Int 32" to a real example, like "id": 100.

This change will apply to a few topics below in this PR that also include the chatMessageMention in the response payload.

Let's see if API Doctor has a problem with this. I believe content-wise, this is the appropriate approach.

api-reference/beta/api/channel-get-messagereply.md
```diff
@@ -93,8 +93,7 @@ Content-length: 201
   ],
   "mentions": [
       {
-          "type": "user",
-          "id": "id-value ",
+          "id": "Int32",
```
angelgolfer-ms: As explained above for chat-get-message.md, line 96 here is in an example. I am changing the "id" property from "id": "Int 32" to a real example, like "id": 100.

api-reference/beta/api/channel-list-messagereplies.md
```diff
@@ -96,8 +96,7 @@ Content-length: 201
         ],
         "mentions": [
             {
-                "type": "user",
-                "id": "id-value ",
+                "id": "Int32",
```
angelgolfer-ms: As explained above for chat-get-message.md, line 99 here is in an example. I am changing the "id" property from "id": "Int 32" to a real example, like "id": 100.

api-reference/beta/api/channel-list-messages.md
```diff
@@ -97,8 +97,7 @@ Content-length: 201
         ],
         "mentions": [
             {
-                "type": "user",
-                "id": "id-value ",
+                "id": "Int32",
```
angelgolfer-ms: As explained above for chat-get-message.md, line 100 here is in an example. I am changing the "id" property from "id": "Int 32" to a real example, like "id": 100.

api-reference/beta/api/chartseries-list-points.md
```diff
@@ -39,7 +39,7 @@ Do not supply a request body for this method.
 
 ## Response
 
-If successful, this method returns a `200 OK` response code and collection of [ChartPoints](../resources/chartpoint.md) objects in the response body.
+If successful, this method returns a `200 OK` response code and collection of [workbookChartPoints](../resources/workbookchartpoint.md) objects in the response body.
```
angelgolfer-ms: The original link text "ChartPoints" should have been singular. I am changing our correction to the link text workbookChartPoint, singular.

api-reference/beta/api/chartseries-post-points.md
```diff
@@ -33,11 +33,11 @@ POST /workbook/worksheets/{id|name}/charts(<name>)/series(<undefined>)/points
 | Workbook-Session-Id  | Workbook session Id that determines if changes are persisted or not. Optional.|
 
 ## Request body
-In the request body, supply a JSON representation of [ChartPoints](../resources/chartpoint.md) object.
+In the request body, supply a JSON representation of [workbookChartPoints](../resources/workbookchartpoint.md) object.
```
angelgolfer-ms: This API creates a single workbookChartPoint object. The original link text are all plural workbookChartPoints and are incorrect. I am correcting the link text to singular.

api-reference/beta/api/directory-deleteditems-user-owned.md
```diff
@@ -91,7 +83,7 @@ Content-length: 1249
           {
               "@odata.type": "#microsoft.graph.group",
               "id": "bfa7033a-7367-4644-85f5-95aaf385cbd7",
-              "deletedDateTime": "2018-04-01T12:34:56Z",
+              "deletedDateTime": "2018-04-01T12:39:16Z",
```
angelgolfer-ms: Since we're correcting the deletedDateTime value to be a time later than the createdDateTime, I am going to change it to be a later time, rather than the same time, as the createDateTime value. That's a more plausible scenario.

I'm also changing line 48 "Request body" section, to use lower case for "group", in the phrase "... `group` is currently the only supported value."

api-reference/beta/api/message-post-attachments.md
```diff
@@ -82,7 +82,7 @@ Content-length: 142
 {
   "@odata.type": "#microsoft.graph.fileAttachment",
   "name": "smile",
-  "contentBytes": "R0lGODdhEAYEAA7"
+  "contentBytes": "base64:R0lGODdhEAYEAA7"
```
angelgolfer-ms: @daspek , @finsharp , can either of you please explain why API Doctor expects a change like the above (line 85)?

For now, so that we can merge PR #4390 , I am going to make this change the same as what's in message-post-attachments.md in v1. I would expect the requirements would be the same between v1 and beta. There, in the same example request payload, we have:
"contentBytes": "base64R0lGODdhEAYEAA7"
instead of (with the colon):
"contentBytes": "base64:R0lGODdhEAYEAA7"

angelgolfer-ms: I'm also replacing the value of "contentBytes" in the response payload. to be consistent as the request:
"contentBytes": "base64R0lGODdhEAYEAA7"
Right now, in the response payload, "contentBytes": "R0lGODdhEAYEAA7".

After making that change in the response payload, I am going to remove the second suppression, as the error in that suppression quotes the actual value of contentBytes being R0lGODdhEAYEAA7.

api-reference/beta/api/rangeformat-get.md
```diff
@@ -57,7 +57,7 @@ Here is an example of the response. Note: The response object shown here may be
 <!-- {
   "blockType": "response",
   "truncated": true,
-  "@odata.type": "microsoft.graph.rangeFormat"
+  "@odata.type": "microsoft.graph.workbookRangeFont"
```
angelgolfer-ms: I am fixing this as workbookRangeFormat.

angelgolfer-ms: I am curious how specifying the response payload as of type microsoft.graph.workbookRangeFont would not have triggered API Doctor error.
The properties in the example response payload are not properties of microsoft.graph.workbookRangeFont, and should have triggered API Doctor.

api-reference/beta/api/schedule-post-schedulinggroups.md
```diff
@@ -50,6 +50,9 @@ If successful, this method returns a `201 Created` response code and a [scheduli
 The following is an example of the request.
 <!-- {
   "blockType": "request",
+   "optionalProperties": [
```
angelgolfer-ms: Lines 53-55 are not applicable to the HTML comment block preceding an example request. We specify optionalProperties only when defining the JSON of resources. I am removing these 3 lines.

api-reference/beta/api/schedule-post-schedulinggroups.md
```diff
@@ -74,6 +77,7 @@ The following is an example of the response.
 <!-- {
   "blockType": "response",
   "truncated": true,
+  "baseType":"microsoft.graph.changeTrackedEntity",
```
angelgolfer-ms: I think we can specify the baseType when defining the JSON of a resource, but we don't define baseType in the HTML comment block preceding an example response. I am removing line 80.

angelgolfer-ms: Note: by verifying the changes in this topic, I noticed that the "JSON representation" section for schedulinggroup.md is actually incorrect. 
That section inappropriately specifies example values for the properties of schedulingGroup. I am going to change that section to defining the JSON of schedulingGroup.

api-reference/beta/api/securescorecontrolprofiles-list.md
```diff
@@ -77,11 +78,11 @@ Content-type: application/json
             "actionUrl": "actionUrl.value",
             "controlCategory": "controlCategory.value",
```
angelgolfer-ms: In verifying this section (the example response payload for list secureScoreControlProfile), I noticed that the JSON definition for secureScoreControlProfile in securescorecontrolprofiles.md is actually incorrect. It's got a few duplicates, and the properties don't match what's in the beta metadata. So I updated the JSON definition for secureScoreControlProfile in securescorecontrolprofiles.md, by matching it with the current beta metadata.

api-reference/beta/api/securescorecontrolprofiles-list.md
```diff
@@ -77,11 +78,11 @@ Content-type: application/json
...
             "title": "title.value",
-            "deprecated": "deprecated.value",
+            "deprecated": true,
             "implementationCost": "implementationCost.value",
             "lastModifiedDateTime": "lastModifiedDateTime.value",
-            "maxScore": "maxScore.value",
-            "rank": "rank.value",
+            "maxScore": 12,
```
angelgolfer-ms: I'm changing the example values of "maxScore" and "rank", which are defined to be double and int32 respectively.

api-reference/beta/api/sitepage-list.md
```diff
@@ -55,7 +69,7 @@ Content-type: application/json
     "value": [
         {
             "id": "5",
-            "eTag": "{8BEE5ABE-49B9-431B-AEBA-C96D6DEF44E3},4",
+            "eTag": "{8BEE5ABE-49B9-431B-AEBA-C96D6DEF44E3}",
```
angelgolfer-ms: @finsharp , @daspek , can you please confirm if the change for eTag here (line 72) is appropriate?

I did some superficial browsing of the SharePoint and OneDrive API topics, it looks to me that the eTag property value is a string that itself is delimited by double quotes, and that takes the format of a GUID enclosed by curly brackets, followed by a comma, followed by an integer. That is, something like the following:
"eTag": "\"{B93F682B-DBBC-42DD-802D-5D3DD0CBF371},2\"",

However, I see in the existing sitepage-create.md and sitepage-get.md, the eTag property values are slightly different, like the following, without its own double quote delimiters and curly brackets around the GUID:
"eTag": "75bc70e2-6587-45be-8493-c99a956b2e05,7"

Based on the above, I think what is shown above in line 72 in this topic:
"eTag": "{8BEE5ABE-49B9-431B-AEBA-C96D6DEF44E3}"
 is least consistent with any other existing example. 
So for the time being, I am going to change the instances of eTag in sitepage-list.md to include their own sets of double-quote delimiters:
"eTag": "\"{8BEE5ABE-49B9-431B-AEBA-C96D6DEF44E3},4\""

Please let me know if you know what the format should be.

api-reference/beta/api/sitepage-list.md
```diff
@@ -127,8 +141,8 @@ Content-type: application/json
             ]
         },
         {
-            "id": 2,
-            "eTag": "75bc70e2-6587-45be-8493-c99a956b2e05,7",
+            "id": "2",
+            "eTag": "{75bc70e2-6587-45be-8493-c99a956b2e05}",
```
angelgolfer-ms: I am changing the eTag value:
 "eTag": "\"{75bc70e2-6587-45be-8493-c99a956b2e05},7\"",

api-reference/beta/api/timeoffreason-put.md
```diff
@@ -93,15 +93,7 @@ Content-length: 401
...
-  "lastModifiedBy": {
-    "application": null,
-    "device": null,
-    "conversation": null,
-    "user": {
-      "id": "366c0b19-49b1-41b5-a03f-9f3887bd0ed8",
-      "displayName": "John Doe"
-    }
-  }
+  "lastModifiedBy": { "@odata.type":"microsoft.graph.identitySet"}
```
angelgolfer-ms: This replacement of the value of lastModifiedBy is not appropriate. This is an example, not the JSON definition. I am goign to reverse the change back to what it was before.

api-reference/beta/api/worksheet-get.md
```diff
@@ -39,7 +39,7 @@ Do not supply a request body for this method.
 
 ## Response
 
-If successful, this method returns a `200 OK` response code and [Worksheet](../resources/worksheet.md) object in the response body.
+If successful, this method returns a `200 OK` response code and [workBookWorksheet](../resources/workbookworksheet.md) object in the response body.
```
angelgolfer-ms: I am changing the link text to [workbookWorksheet] (lower case 'b').

api-reference/beta/api/worksheet-list.md
```diff
@@ -39,7 +39,7 @@ Do not supply a request body for this method.
 
 ## Response
 
-If successful, this method returns a `200 OK` response code and collection of [Worksheet](../resources/worksheet.md) objects in the response body.
+If successful, this method returns a `200 OK` response code and collection of [workBookWorksheet](../resources/workbookworksheet.md) objects in the response body.
```
angelgolfer-ms: To correct all instances of the inappropriate case for workBookWorksheet, I searched and there are still 4 instances of link text that are misspelt:
workbook-list-worksheets.md
worksheet-list.md
worksheet-update.md
worksheetcollection-add.md

I am replacing each instance in the above listed 4 files by workbookWorksheet.

api-reference/beta/resources/changetrackedentity.md
```diff
@@ -0,0 +1,37 @@
...
+Here is a JSON representation of the resource.
+<!--{
+  "blockType": "resource",
+  "@odata.type": "microsoft.graph.changeTrackedEntity",
+  "baseType":"microsoft.graph.entity",
+  "abstract":true
+}-->
+``` json
+{
+    "createdDateTime":"DateTimeOffset",
```
angelgolfer-ms: I am changing createdDateTime and lastModifiedDateTime in the JSON representation section to be "String (timestamp)". That's consistent with other properties that are of the same Edm.DateTimeOffset type.

api-reference/beta/resources/chatmessagemention.md
```diff
@@ -29,9 +29,9 @@ The following is a JSON representation of the resource.
 
 ```json
 {
-  "id": "string (identifier)",
+  "id": "Int32",
```
angelgolfer-ms: I am changing "id" in the JSON definition to "id": 1024. This is consistent with how the stub topic generator generates properties that are of the Edm.Int32 type.

PR #4396, opened 2019-04-15
eketo - adding docs for revoke sign-in sessions rename

eketo-msft:


api-reference/beta/api/user-revokesigninsessions.md
```diff
@@ -0,0 +1,77 @@
...
+
+# user: revokeSignInSessions
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Invalidates all of the user's refresh tokens issued to applications (as well as session cookies in a user's browser), by resetting the **signInSessionsValidFromDateTime** user property to the current date-time. Typically, this operation is performed (by the user or an administrator) if the user has a lost or stolen device.  This operation would prevent access to any of the organization's data accessed through applications on the device without the user first being required to sign in again. In fact, this operation would force the user to sign in again for all applications that they have previously consented to, independent of device.
+
+For developers, if the application attempts to redeem a delegated access token for this user by using an invalidated refresh token, the application will get an error. If this happens, the application will need to acquire a new refresh token by making a request to the authorize endpoint, which will force the user to sign in.
+
+## Permissions
```
dkershaw10: @eketo-msft The format doesn't look correct.  This should be a table.  Please look at some of the other topics for examples.

api-reference/beta/api/user-revokesigninsessions.md
```diff
@@ -0,0 +1,77 @@
...
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
++ For an application to allow the signed in user to invalidate applications they've consented to: User.ReadWrite, Directory.ReadWrite.All, Directory.AccessAsUser.All
```
dkershaw10: You can add some additional qualifying information like this under the table.

api-reference/beta/api/user-revokesigninsessions.md
eketo-msft: I'm just renaming existing stuff to unblock parity.  This shouldn't block...

dkershaw10: OK - fair enough. I'll approve and we can fix later with the PR to v1.0.

Lauragra: @dkershaw10 , @eketo-msft , are you saying that you want to publish the content with an inconsistent Permissions table? Not sure I understand eketo's comment. Is there some urgency to publish this content such that we can't have a consistent user experience?

eketo-msft: I'm saying that the permission table is already inconsistent, and the rename of the property/action should not be blocked by that.  We aren't introducing the inconsistency with this change.

Lauragra: Since we're now aware of the issue from the original file, is there a reason why we can't make the fix now, rather than have the inconsistency continue in the documentation? In the future, non-standard content might trigger validation failures, so this would become a blocking issue.

eketo-msft: We don't know what the right fix is yet.  An open issue should be created to track this, and then it can be fixed orthogonally.

api-reference/beta/api/user-revokesigninsessions.md
```diff
@@ -0,0 +1,78 @@
...
+
+Invalidates all the refresh tokens issued to applications for a user (as well as session cookies in a user's browser), by resetting the **signInSessionsValidFromDateTime** user property to the current date-time. Typically, this operation is performed (by the user or an administrator) if the user has a lost or stolen device. This operation prevents access to the organization's data through applications on the device by requiring the user to sign in again to all applications that they have previously consented to, independent of device.
+
+If the application attempts to redeem a delegated access token for this user by using an invalidated refresh token, the application will get an error. If this happens, the application will need to acquire a new refresh token by making a request to the authorize endpoint, which will force the user to sign in.
+
+## Permissions
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type                        | Permissions (from least to most privileged)              |
```
Lauragra: @eketo-msft , it's better if we make the content consistent before we publish.  Otherwise, this can cause issues for us down the road. I formatted the bullets into the standard table structure. Can you please review? Thanks.

eketo-msft: Looks good, thank you.

PR #4393, opened 2019-04-13
@mentions fixes

nkramer:

PR #4405, opened 2019-04-16
Fixes

nkramer:


api-reference/beta/resources/team.md
```diff
@@ -83,10 +85,27 @@ The following is a JSON representation of the resource.
   "funSettings": {"@odata.type": "microsoft.graph.teamFunSettings"},
   "internalId": "19:...big.number...@thread.skype",
```
angelgolfer-ms: @nkramer , should we remove line 86 "internalId"? It's not in the beta metadata.

api-reference/beta/resources/team.md
```diff
@@ -83,10 +85,27 @@ The following is a JSON representation of the resource.
...
   "isArchived": false,
-  "webUrl": "https://...longUrl..."
+  "webUrl": "https://...longUrl...",
```
angelgolfer-ms: Please set "webUrl": "string (URL)".
API Doctor needs to find the type of a property in the "JSON representation" in order to validate values assigned to that property in an example payload.
I suggest this string as this seems to be how the stub generator tool would generate the JSON definition for a property that's defined as a string and is a URL.

api-reference/beta/resources/team.md
```diff
@@ -83,10 +85,27 @@ The following is a JSON representation of the resource.
...
+  "displayName": "string",
+  "description": "string",
+  "classification": "string",
+  "specialization": "string",
+  "visibility": "string"
 }
 
 ```
+<Property Name="displayName" Type="Edm.String"/>
```
angelgolfer-ms: I think we forgot to remove lines 97-107? :-)

api-reference/beta/resources/team.md
angelgolfer-ms: If "internalId" is not supposed to be there anymore, then we should also remove it from line 51 in the Properties table as well.

nkramer: No, we really are supposed to have a property called internalId.

api-reference/beta/resources/team.md
nkramer: fixed

api-reference/beta/resources/team.md
nkramer: fixed

nkramer:
feedback addressed

api-reference/beta/resources/team.md
angelgolfer-ms: @nkramer , @Lauragra , since @nkramer says internalId is supposed to be documented, then please follow up to add the property to the metadata.
Also, I am going to correct in the "JSON representation" section that internalId is of type string . I am going to add this fix to PR #4393, just so in the future, having internalId in a response payload will not trigger API Doctor.

PR #4501, opened 2019-04-30
Event Series Behavior in Calendar View Delta Function

RobertEng:
Add documentation on recurring event series for the delta function of calendar views. Namely:
- No indication is given for a deleted occurrence
- Series won't be split on multiple delta pages

These conclusions were found via experimentation.


Lauragra:
@angelgolfer-ms , please review and approve this PR.

PR #4526, opened 2019-05-04
Expanding the role of Graph in M365.

angelgolfer-ms:


concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
 
 # Overview of Microsoft Graph
 
-Microsoft Graph is the gateway to data and intelligence in Microsoft 365. Microsoft Graph provides a unified programmability model that you can use to take advantage of the tremendous amount of data in Office 365, Enterprise Mobility + Security, and Windows 10. 
+Microsoft Graph is the gateway to data and intelligence in Microsoft 365. It provides a unified programmability model to let you access the tremendous amount of data in Office 365, Windows 10, and Enterprise Mobility + Security. You can use Microsoft Graph to leverage the wealth of data and build apps for organizations and consumers that interact with millions of users.
```
Lauragra: In the second sentence, why did we revise "you can use" to "let you access"? Since programmability model is singular, the verb should be lets; however, we try to avoid that construction, per the following guideline: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/l/let-lets Also, in the last sentence, can we please change "leverage", per the following guidance: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/l/leverage

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
 
+In the Microsoft 365 platform, three main components facilitate the access and flow of data:
 
-## What's in the graph?
-Microsoft Graph exposes REST APIs and client libraries to access data on the following:
+- The Microsoft Graph API offers a single endpoint, `https://graph.microsoft.com`, to provide access to rich, people-centric data and insights exposed as resources of Microsoft 365 services. You can use REST APIs or SDKs to access the endpoint and build apps that support scenarios spanning across productivity, collaboration, education, security, identity, access, device management, and much more.
+- [Microsoft Graph data connect](#access-microsoft-graph-data-at-scale-using-microsoft-graph-data-connect) provides a set of tools to streamline secure and scalable delivery of Microsoft Graph data to popular Azure data stores. This cached data serves as data sources for Azure development tools that you can use to build intelligent applications.
+- Microsoft Graph connectors (private preview) works in the incoming direction, delivering external data into Microsoft Graph services and applications and enhancing custom experiences.
```
Lauragra: Connectors is plural, so we should change "works" to "work".

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
 
-- Azure Active Directory
-- Office 365 services: SharePoint, OneDrive, Outlook/Exchange, Microsoft Teams, OneNote, Planner, and Excel
-- Enterprise Mobility and Security services: Identity Manager, Intune, Advanced Threat Analytics, and Advanced Threat Protection.
-- Windows 10 services: activities and devices
-- Education
+Together, the Microsoft Graph API, data connect, and connectors (private preview) power the Microsoft 365 platform. The ability to access Microsoft Graph data and other datasets to derive insights and analytics lets you extend Microsoft 365 experiences, or build unique intelligent applications.
```
Lauragra: We should have a comma after unique.

Lauragra: We can also revise the second sentence to focus the action on the user - "With the ability to access Microsoft Graph data and other datasets to derive insights and analytics, you can extend Microsoft 365 experiences, or build unique, intelligent applications."

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
 
-You can also get valuable insights and intelligence about the data from Microsoft Graph. For example, you can get the popular files [trending around](/graph/api/resources/insights-trending?view=graph-rest-beta) a particular user, or [get the most relevant people](/graph/api/user-list-people?view=graph-rest-beta) around a user.
+## What's in Microsoft Graph?
+Microsoft Graph exposes REST APIs and client libraries to access data on the following Microsoft 365 services:
+
+- Azure Active Directory
+- Education
+- Enterprise Mobility and Security services: Advanced Threat Analytics, Advanced Threat Protection, Identity Manager, and Intune 
+- Office 365 services: Delve, Dynamics 365 Business Central, Excel, Microsoft Bookings, Microsoft Teams, OneDrive, OneNote, Outlook/Exchange, Planner, and SharePoint
```
Lauragra: We're grouping Dynamics 365 Business Central as an Office 365 service? We might want to make that a separate bullet point.

concepts/overview.md
angelgolfer-ms: First comment: "that you can use to take advantage of" is way too wordy for what we're trying to say. It's also against the Microsoft Voice principle of "crisp and clear".
Second comment - no opinion as long as it's another word replacement that doesn't add more words :-) I don't think "take advantage of" here is appropriate though.

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
 ![An image showing the primary resources and relationships that are part of the graph](images/microsoft-graph.png)
 
 ## What can you do with Microsoft Graph? 
 
-You can use Microsoft Graph to build experiences around the user's unique context to help them be more productive. Imagine an app that...
+> [!VIDEO https://www.youtube-nocookie.com/embed/PI9NO5rayiY]
+
+Use Microsoft Graph to build experiences around the user's unique context to help them be more productive. Imagine an app that...
+
+- Looks at your next meeting and helps you prepare for it by providing profile information for attendees, including their job titles and managers, as well as information about the latest documents they're working on, and people they're collaborating with
```
Lauragra: These bullet points complete the introductory sentence and therefore should end in periods.

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
+- Scans your calendar, and suggests the best times for the next team meeting
+- Fetches the latest sales projection chart from an Excel file in your OneDrive and lets you update the forecast in real time, all from your phone
+- Subscribes to changes in your calendar, sends you an alert when you’re spending too much time in meetings, and provides recommendations for the ones you can miss or delegate based on how relevant the attendees are to you
+- Helps you sort out personal and work information on your phone; for example, by categorizing pictures that should go to your personal OneDrive and business receipts that should go to your OneDrive for Business
+- Leverages at-scale Office 365 data and analytics so that decision makers can unlock valuable insights into time allocation and collaboration patterns that impact business productivity
```
Lauragra: Replaces "Leverages" with a clear word or phrase - i.e. "Makes use of".

concepts/overview.md
angelgolfer-ms: I was using the term as a feature/service name. 
If we look at that as a service, then it should be treated as one unit and hence third-person singular.

concepts/overview.md
angelgolfer-ms: Right. Please add it.

concepts/overview.md
angelgolfer-ms: It is. It's part of Office 365, a certain edition. I have it in my demo tenant for that edition of O365 (whatever edition it is).

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
+- Brings custom business data into Microsoft Graph, indexing it to make it searchable along with data from Microsoft 365 services
+
+Pick the first scenario about researching meeting attendees as an example. With Microsoft Graph API, you can:
```
Lauragra: We should either say "With the Microsoft Graph API", or "With Microsoft Graph".

concepts/overview.md
```diff
@@ -7,48 +7,70 @@ localization_priority: Priority
...
+
+1. Get the email addresses of the [meeting event](/graph/api/resources/event?view=graph-rest-1.0) attendees
```
Lauragra: These steps should also end in periods.

concepts/overview.md
angelgolfer-ms: Right. I only remembered the second piece of advice in the style guide. Please add them back.

concepts/overview.md
angelgolfer-ms: Can you find a better word replacement for "leverage"?
It's more impactful than "take advantage of" or "make use of".
Same sentiment for "affect" replacing "impact on" (missing the "on").

PR #4537, opened 2019-05-07
separated the PIM node into Azure AD and Azure Resources

CelesteDG:


api-reference/beta/toc.yml
```diff
@@ -7865,9 +7865,10 @@
           href: api/policy-list-assigned.md
         - name: List applied to applications
           href: api/policy-list-appliesto.md
-      - name: Privileged Identity Management
-        href: resources/privilegedidentitymanagement-root.md
+      - name: Privileged Identity Management - Azure AD
```
dkershaw10: I looked at the review site, and this doesn't look so good.  The top level node look too long :(
Also the topics names underneath don't seem very consistent.

Lauragra: @dkershaw10 , what do you recommend? Possibly we can shorten to use PIM.

Lauragra:
@CelesteDG , closing this one since we implemented the TOC update in a separate PR.

CelesteDG:
Thank you, @Lauragra  :)

PR #4590, opened 2019-05-14
Fixing some issues with the data connect docs

tlenig:
Received some feedback that our links are broken in the get-started section and fixing the name of the ToC


Lauragra:
@tlenig , FYI, I already fixed the links. You can verify here: https://review.docs.microsoft.com/en-us/graph/data-connect-get-started?branch=master. This fix will go live today.

concepts/toc.yml
```diff
@@ -90,7 +90,7 @@
     - name: Customer booking (preview)
       href: booking-concept-overview.md
       displayName: Bookings, Microsoft Bookings
-    - name: Data access
+    - name: Data connect
```
Lauragra: FYI, we're leaving the title of the node as "Data access". Data connect is the only child node right now, but we'll be adding docs for connectors here too, so we'll have two peer nodes.

tlenig: Ah interesting, are we concerned about putting both data connect and connectors in the same section of our docs? The names of the products are already so close, associating them in the same section might make that even more confusing, at least from my perspective

Lauragra: Yes, the naming similarity is a bridge that we will need to cross when we get there. ;) Note that in the Learn section of the TOC, we don't use product or feature names, we lead with the capability, so Data access is more consistent with that approach.

tlenig: Understood makes sense thanks!

PR #4592, opened 2019-05-14
Add version info to outboundConnectionsRequired and outboundConnectio…

kagillig:
…nsBlocked

Intune was updated to only apply these settings to Windows versions 1809 and up so this updates the docs to reflect these changes. I also plan on making the changes in the service code so they get pulled in automatically, but wanted to make these changes now so they are in line with the service.


Lauragra:
@rolyon as an FYI. I'll merge this one in as it sounds like it will be pushed upstream as well.

PR #4643, opened 2019-05-20
chat membership APIs

nkramer:


api-reference/beta/api/chatmembership-get.md
```diff
@@ -0,0 +1,82 @@
+---
+title: "Get Chat Conversation Membership"
+description: "Retrieve a member of a chat."
+author: "nkramer"
+localization_priority: Priority
+ms.prod: "microsoft-teams"
+---
+
+# Get Chat Member
```
Lauragra: The title here should be Get chatMembership, The title pattern is to use the verb + resource name and follow lower case camel casing for the resource name. Also, please make sure that the title metadata matches the H1 title.

nkramer: fixes

PR #4664, opened 2019-05-22
Update data-connect-overview.md

tlenig:
Adding information about the Terms of Use and Privacy statements needed to be reviewed and agreed to before using Graph features

### DIFF
--- a/concepts/data-connect-overview.md
+++ b/concepts/data-connect-overview.md
@@ -10,7 +10,7 @@ ms.prod: "data-connect"
 
 Microsoft Graph data connect brings Office 365 data and Microsoft Azure resources to you through a toolset. You can use Microsoft Graph data connect to build intelligent applications that access valuable data using the best development tools offered by Microsoft, all within the Microsoft secure cloud. All businesses are interested in improving productivity, which means that building products that make knowledge workers more productive presents tremendous opportunities. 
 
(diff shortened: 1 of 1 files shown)


concepts/data-connect-overview.md
```diff
@@ -10,7 +10,7 @@ ms.prod: "data-connect"
 
 Microsoft Graph data connect brings Office 365 data and Microsoft Azure resources to you through a toolset. You can use Microsoft Graph data connect to build intelligent applications that access valuable data using the best development tools offered by Microsoft, all within the Microsoft secure cloud. All businesses are interested in improving productivity, which means that building products that make knowledge workers more productive presents tremendous opportunities. 
 
-You can take advantage of Microsoft Graph data connect if you are:
+You can take advantage of Microsoft Graph data connect if you agree to the [Microsoft Graph Terms of Use](https://developer.microsoft.com/en-us/graph/terms-of-use) and the [Microsoft Privacy Statement](https://go.microsoft.com/fwlink/p/?LinkId=123161) and are:
```
Lauragra: This looks fine to me. FYI, we'll have a new URL for TOU shortly; we'll put a redirect in place so this will be okay but we should get it updated eventually.

tlenig: Understood thanks Laura! This is to satisfy a CELA requirement so good to know it will redirect to the new link

PR #4697, opened 2019-05-27
Code Snippets Generation - Support for Custom Queries

andrueastman:


api-reference/beta/includes/get-list-item-Cs-snippets.md
```diff
@@ -6,8 +6,13 @@ description: "Automatically generated file. DO NOT MODIFY"
 
 GraphServiceClient graphClient = new GraphServiceClient( authProvider );
 
+var queryOptions = new List<QueryOption>()
+{
+	new QueryOption("expand", "fields")
```
darrelmiller: The `expand` query parameter is the same as $expand.  In OData 4.01 the $ was considered optional.  Ideally this should map to the standard expand function.

PR #4699, opened 2019-05-28
Added scenario table to choose auth provider page

darrelmiller:


concepts/sdks/choose-authentication-providers.md
```diff
@@ -5,11 +5,25 @@ localization_priority: Normal
...
-
-## Authorization code OAuth flow
+# Choose a Microsoft Graph authentication provider based on scenario
+
+Authentication providers implement the code required to acquire a token using the Microsoft Authentication Library (MSAL); handle a number of potential errors for cases like incremental consent, expired passwords, and conditional access; and then set the HTTP request authorization header. The following table lists the set of providers that match the scenarios for different [application types](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-app-types).
+
+|Scenario | Flow/Grant | Audience | Provider|
+|--|--|--|--|--|
```
dkershaw10: Table is busted and doesn't render properly.  I think you have 4 columns, but this row indicates 5?

concepts/sdks/choose-authentication-providers.md
```diff
@@ -186,7 +200,7 @@ Not applicable.
 
 ---
 
-## Device code OAuth flow
+##  <a name="DeviceCodeProvider"/>Device code Provider
```
Lauragra: Provider -> provider

concepts/sdks/choose-authentication-providers.md
```diff
@@ -223,7 +237,7 @@ Not yet available. Please vote for or open a [Microsoft Graph feature request](h
 
 ---
 
-## Integrated Windows flow
+##  <a name="IntegratedWindowsProvider"/>Integrated Windows Provider
```
Lauragra: Provider -> provider

concepts/sdks/choose-authentication-providers.md
```diff
@@ -260,7 +274,7 @@ Not applicable.
 
 ---
 
-## Interactive flow
+##  <a name="InteractiveProvider"/>Interactive Provider
```
Lauragra: Provider -> provider

concepts/sdks/choose-authentication-providers.md
darrelmiller: Ahh, it did have 5 and we merged two.  I'll fix this.

concepts/sdks/choose-authentication-providers.md
```diff
@@ -5,11 +5,25 @@ localization_priority: Normal
...
-
-## Authorization code OAuth flow
+# Choose a Microsoft Graph authentication provider based on scenario
+
+Authentication providers implement the code required to acquire a token using the Microsoft Authentication Library (MSAL); handle a number of potential errors for cases like incremental consent, expired passwords, and conditional access; and then set the HTTP request authorization header. The following table lists the set of providers that match the scenarios for different [application types](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-app-types).
+
+|Scenario | Flow/Grant | Audience | Provider|
+|--|--|--|--|
+|[Single Page App](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-spa-acquire-token)| Implicit | Delegated Consumer/Org |[Implicit Provider](#ImplicitProvider) |
+| [Web App that calls web APIs](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-web-app-call-api-acquire-token) | Authorization Code | Delegated Consumer/Org | [Authorization Code Provider](#AuthCodeProvider) |
```
Lauragra: The empty cells in the Scenario column could be confusing when glancing at the table. You can solve this in a few ways: Repeat the scenario in each empty cell; merge the cells into a single row per scenario with bullets (might be less clear that way), or use empty rows as "labels" for the scenario rows, as you can see in the Methods table in this topic: https://docs.microsoft.com/en-us/graph/api/resources/message?view=graph-rest-1.0. First option is probably the most straightforward.

Issue #4674, opened 2019-05-23
What's the maximum length of subscription for user and group


dkershaw10:
The table does not document subscription length maximum for users and groups. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cbe0590e-070b-cf56-cbc7-1dcb8a266380
* Version Independent ID: 2d912536-afc5-60f1-e381-2eef892cbded
* Content: [subscription resource type - Microsoft Graph v1.0](https://docs.microsoft.com/en-us/graph/api/resources/subscription?view=graph-rest-1.0#maximum-length-of-subscription-per-resource-type)
* Content Source: [api-reference/v1.0/resources/subscription.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/v1.0/resources/subscription.md)
* Product: **non-product-specific**
* Technology: **microsoft-graph**
* GitHub Login: @piotrci
* Microsoft Alias: **MSGraphDocsVteam**

piotrci:
PR created: https://github.com/microsoftgraph/microsoft-graph-docs/pull/4707

PR #4707, opened 2019-05-28
Updated expiration times for directory entities

piotrci:
Fixing issue: https://github.com/microsoftgraph/microsoft-graph-docs/issues/4674


api-reference/beta/resources/subscription.md
```diff
@@ -60,6 +60,7 @@ Here is a JSON representation of the resource.
 
 | Resource            | Maximum Expiration Time  |
 |:--------------------|:-------------------------|
+| User, group, other directory entities   | 4230 minutes (under 3 days)    |
```
dkershaw10: I think we call them "resources" rather than entities in the docs...

dkershaw10:
This is related to #4674

PR #4748, opened 2019-06-04
fix for git issue 4705

davidmu1:


davidmu1:
#sign-off

api-reference/beta/api/user-post-users.md
```diff
@@ -13,6 +13,9 @@ ms.prod: "microsoft-identity-platform"
 Create a new user.
 The request body contains the user to create. At a minimum, you must specify the required properties for the user. You can optionally specify any other writable properties.
 
+>[!NOTE]
+>If you need to enable the creation of external users, see the [invitation manager](../resources/invitation.md).
```
Lauragra: @davidmu1 , I think the title of the link should be "invitation", and the article topic should be "invitation resource type". Can you pls. verify and update?

davidmu1: @Lauragra , the title of the article that is being linked to is "invitation manager". I don't understand what you mean by "invitation resource type", can you clarify?

Lauragra: @davidmu1 , yes, I think that the topic you're linking to is mis-titled. It's a resource topic, and the name of the resource is "invitation", so for consistency the topic title should be "invitation resource type". I think we should take a look at the topic, as perhaps it's trying to serve the purpose of both a reference a conceptual topic: https://docs.microsoft.com/en-us/graph/api/resources/invitation?view=graph-rest-1.0. Also, I'm not sure it's clear how this is related to creating external users.

davidmu1:
Renamed the invitation article, changed the link text, and added a sentence saying the external users must be created through an invitation.

#sign-off

davidmu1:
oops, signed off before pushing changes.

PR #4755, opened 2019-06-05
Add revoke sign-in sessions docs to v1.0

eketo-msft:


api-reference/v1.0/api/user-revokesigninsessions.md
```diff
@@ -0,0 +1,91 @@
...
+
+##### Response
+<!-- {
+  "blockType": "response",
+  "truncated": true
+} -->
+```http
+HTTP/1.1 204 No Content
+```
+#### SDK sample code
```
dkershaw10: I think this stuff gets automatically inserted once it's built.  Otherwise this reference will probably fail.

concepts/changelog.md
```diff
@@ -13,6 +13,13 @@ For details about known issues with Microsoft Graph APIs, see [Known issues](kno
 
 ## June 2019
 
+### Directory APIs
+
+| **Change type** | **Version** | **Description**                          |
+| :-------------- | :---------- | :--------------------------------------- |
+| Addition | v1.0 | Add the **signInSessionsValidFromDateTime** property on the [user](/graph/api/resources/user?view=graph-rest-1.0) resource. |
```
dkershaw10: Add -> Added

api-reference/v1.0/api/user-revokesigninsessions.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "user: revokeSignInSessions"
```
Lauragra: @eketo-msft , Do you want to expose this new API in the v1.0 TOC? Note that the beta API topic is not exposed in the TOC, currently.

eketo-msft: Hmm, @Lauragra, this is a good question.  We don't tend to put our service actions in the TOC at all.  I'll talk to @dkershaw10 about this, maybe we'll do it in a bulk change for many of our service action but I don't think we'll do it in the short term.

Lauragra: Okay, I'll merge this for now and we can update later if we decide to.

eketo-msft: Thanks @Lauragra !

PR #4774, opened 2019-06-07
fix for git issue 4705

davidmu1:


davidmu1:
The GitHub issue asked to add a link to the invitation article to make developers aware of the way to create an external user account. Changed the title of the invitation article. Added a sentence saying that to create an external user, an invitation must be used.

#sign-off

api-reference/beta/resources/invitation.md
```diff
@@ -1,16 +1,16 @@
 ---
-title: "invitation manager"
-description: "Use the invitation manager to create an invite, in order to add an external user to the organization. "
+title: "invitation resource type"
```
Lauragra: @davidmu1 , I just noticed that this file is also in v1.0. Can you please make the same update to the v1.0 version of the file?

api-reference/beta/resources/invitation.md
```diff
@@ -1,16 +1,16 @@
...
 ms.prod: "microsoft-identity-platform"
 ---
 
-# invitation manager
+# invitation resource type
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Use the invitation manager to create an invite, in order to add an external user to the organization. 
+The invitation is used to create an invite, in order to add an external user to the organization.
```
Lauragra: We're using a definite article here which might be unclear. Resource topics often begin with a verb like "Represents". Suggested revision: "Represents an invitation that is used to add external users to an organization."

PR #4806, opened 2019-06-11
Adding branding changelog

ankitachandak:
Adding Organizational branding element in changelog.


dkershaw10:
Removing and closing for now.  There's a different PR for this, and anyways, we might be holding this for a bit as we might change some of the API shape.

PR #4836, opened 2019-06-14
Add note that Teams SendMessage APIs are not for migration

nkramer:


api-reference/beta/api/channel-post-messagereply.md
```diff
@@ -12,6 +12,8 @@ ms.prod: "microsoft-teams"
 
 Create a new reply to a [message](../resources/chatmessage.md) in a specified [channel](../resources/channel.md).
 
+> **Note**: This API is not intended for data migration, as it does not have the throughput necessary for typical migrations.
```
Lauragra: @nkramer , can we revise this note to the following? "We don't recommend that you use this API for data migration. It does not have the throughput necessary for a typical migration."

nkramer: Done

Issue #4771, opened 2019-06-07
Not applicable for Mail enabled security groups


bartvermeersch:
The first line says: "Use this API to remove a member from an Office 365 group, a security group, or a mail-enabled security group through the members navigation property. You can remove users or other groups." 

But you can't add or remove members in mail-enabled security groups using the MS graph at this moment.
error="Unable to update the specified properties for objects that have originated within an external service."

See https://github.com/microsoftgraph/microsoft-graph-docs/issues/546 and https://github.com/microsoftgraph/microsoft-graph-docs/issues/4210 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4a1141e6-3381-6433-0fdc-187b65eb299d
* Version Independent ID: 5f05d149-8a88-5dea-35a3-02ade58fa9aa
* Content: [Remove member - Microsoft Graph v1.0](https://docs.microsoft.com/en-us/graph/api/group-delete-members?view=graph-rest-1.0&tabs=cs)
* Content Source: [api-reference/v1.0/api/group-delete-members.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/v1.0/api/group-delete-members.md)
* Product: **groups**
* Technology: **microsoft-graph**
* GitHub Login: @dkershaw10
* Microsoft Alias: **MSGraphDocsVteam**

davidmu1:
Sentence was revised.

PR #4838, opened 2019-06-14
fix for git issue 4771

davidmu1:


davidmu1:
@dkershaw10 - Dan, can you verify this change? I removed mention of the ability to remove other groups. Based on GitHub issue https://github.com/microsoftgraph/microsoft-graph-docs/issues/4771. Thanks.

api-reference/v1.0/api/group-delete-members.md
```diff
@@ -1,13 +1,13 @@
...
-description: "Use this API to remove a member from an Office 365 group, a security group, or a mail-enabled security group through the **members** navigation property. You can remove users or other groups."
+description: "Use this API to remove a member from an Office 365 group through the **members** navigation property."
 localization_priority: Priority
 author: "dkershaw10"
 ms.prod: "groups"
 ---
 
 # Remove member
-Use this API to remove a member from an Office 365 group, a security group, or a mail-enabled security group through the **members** navigation property. You can remove users or other groups.
+Use this API to remove a member from an Office 365 group through the **members** navigation property.
```
Lauragra: Revise as "Remove a member from an Office 365 group via the **members** navigation property.

Lauragra:
@davidmu1 , does this change apply to beta as well as v1.0?

api-reference/beta/api/group-delete-members.md
```diff
@@ -10,7 +10,7 @@ ms.prod: "groups"
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Use this API to remove a member from an Office 365 group, a security group, or a mail-enabled security group through the **members** navigation property. You can remove users or other groups.
+Use this API to remove a member from an Office 365 group via the **members** navigation property. You can remove users or other groups.
```
dkershaw10: Actually O365 groups don't support nesting, so groups is not an option for removal.  You could remove a user, device or servicePrincipal.

davidmu1: Made introductions in beta and v1 consistent with: "Use this API to remove a member from a group via the **members** navigation property."

Lauragra:
@davidmu1 , @dkershaw10 , are we good to merge this one?

api-reference/beta/api/group-delete-members.md
angelgolfer-ms: @dkershaw10 , @krbain , @davidmu1 , I would like clarification too, please, as one of the original writers for groups and looking at this from the customer's perspective. 2 comments/questions:

(A) We have documented the group entity in Microsoft Graph as something that can be an Office 365 group, security group, or mail-enabled security group. And unless stated otherwise, the CRUD operations for group should apply to these types of groups as well, and that's why most of the other method topics under "Group" list these group types just to be clear. Is there a reason we want to remove the group types in this topic? 

In case there is, then would that apply to other group topics as well? Otherwise it would be really confusing. Or I should say, we will get back to the confusing state - at one point in the beginning, we did make a special effort to identify the types of groups that are supported by the group entity in Microsoft Graph. There was internal and external confusion whether the group entity represents only Office 365 groups, as some of the group capabilities indeed apply to only Office 365 groups.

(B) In particular, per Dan's comment, security groups allow having a user or a group as a member, whereas Office 365 groups don't. If we want to clarify the delete behavior, it might be better to keep the first sentence in the description which identifies the applicable types of groups, and then make clarification on which kind of members can be removed for which types? In other words, make the existing second sentence clearer?

bartvermeersch: The issue is that CRUD operations on Microsoft Graph are only implemented for Office 365 groups. Deleting a member from a mail enabled security group through the Microsoft Graph throws an error.

PR #4867, opened 2019-06-19
att targetResource must be targetResources to work

jefersonsv:
The attribuite targetResource don't work with $filter parameter. To use target Resource attribute you must to use targetResources (with 's' at end).

And the JSON of SDK sample code has the attribute targetResource but the correct is targetResources also.


jefersonsv:
What is the branch that I must to do the pull request, it's master?

Lauragra:
Master is the correct target branch. Thanks for the PR. @dkershaw10 or @dhanyahk , can you please review and verify? @darrelmiller for the SDK update.

darrelmiller:
@Lauragra Once the example response is updated, next time we generate the snippets, the correction will be propagated.  I don't believe there is anything we need to do on the SDK snippet generation side.

Lauragra:
@dkershaw10 , @dhanyahk , do you want to sign off on this before I merge? Thanks.

Lauragra:
@dhanyahk , please sign off on this PR.

PR #4871, opened 2019-06-19
added a note to specify OData is only available on GET requests

baywet:
https://stackoverflow.com/questions/56657077/how-to-filter-response-with-query-parameters-on-post-methods-on-microsoft-graph/56673331#56673331


Lauragra:
Thanks for your submission. @piotrci , please review this PR.

baywet:
Hey Saurabh!
Agreed, and today the service implements OData 4.0. Tomorrow when the service implements OData 4.01 we can update this sentence to include PUT and PATCH. I'll be happy to submit another PR then :)

concepts/query-parameters.md
```diff
@@ -9,6 +9,8 @@ localization_priority: Priority
 
 Microsoft Graph supports optional query parameters that you can use to specify and control the amount of data returned in a response. The support for the exact query parameters varies from one API operation to another, and depending on the API, can differ between the v1.0 and beta endpoints. 
 
+>**Note:** OData queries are only supported on GET HTTP requests.
```
darrelmiller:
```suggestion
>**Note:** OData query parameters are only supported on GET HTTP requests.
```

baywet:
Thanks for the suggestion @darrelmiller !

angelgolfer-ms:
@baywet , thanks for your contribution. Can you please do the following modifications, and then we can close on this pull request:

1. Can you please add the text to the section "OData system query options", instead of to the top of the topic? This is because this topic covers other query parameters that are not part of the OData 4.0 list of query options, in the section "Other query parameters". We'd like to reserve some flexibility to the topic organization.
2. Can you please further clarify the support in the context of OData 4.0? I suggest the following, using active voice:
">**Note:** OData 4.0 supports system query options in only GET operations."

The section will look like this:
=======
**OData system query options**
A Microsoft Graph API operation might support one or more of the following OData system query options. These query options are compatible with the OData V4 query language.

**Note**: OData 4.0 supports system query options in only GET operations. 

Click the examples to try them in [Graph Explorer](https://developer.microsoft.com/graph/graph-explorer).

=======

Note that you should please remove the existing "note" formatting for the last sentence "Click the examples to try them in Graph Explorer", for better focus.

I suggest making the OData 4.0 context clearer so to accommodate Microsoft Graph partner services gradually supporting OData 4.01, which allows system query options in PUT and PATCH operations. When such support happens, I anticipate the specific operation topics would document it. That way, this generalized [query parameters](https://docs.microsoft.com/en-us/graph/query-parameters) topic would not become confusing right away.

baywet:
@angelgolfer-ms thanks for the suggestions, done

PR #4899, opened 2019-06-22
Reorganized what's new.

angelgolfer-ms:


concepts/whats-new-overview.md
```diff
@@ -0,0 +1,72 @@
...
+description: "What's currently new in Microsoft Graph"
+author: "angelgolfer-ms"
+localization_priority: Priority
+---
+
+# What's new in Microsoft Graph
+
+Did you know that some new features in Microsoft Graph originate as popular requests from the developer community? 
+
+Let's walk through how new features get released, what's currently new in Microsoft Graph, and how you can give us your ideas.
```
angelgolfer-ms: @Lauragra , you had a comment "Some text is written in first person plural,..." If you're referring to the "Let's", I see the following in our [style guide](https://styleguides.azurewebsites.net/StyleGuide/Read?id=2696&topicid=29324):
>>
Project friendliness
Use contractions: it’s, you’ll, you’re, we’re, let’s. 

To provide a fuller context to the guidance about the first person plural, I agree in general, with the exception of trying to projecting friendliness, we should avoid using the first person plural, for the following reason stated [here](https://styleguides.azurewebsites.net/StyleGuide/Read?id=2700&topicid=25524):
>>
First-person plural, which often uses the pronoun we, can feel like a daunting corporate presence—the opposite of Microsoft's modern voice. 

I think in this particular topic, the use of "let's" is a friendlier choice than the robot sounding "The following sections walk through how features are released,..."

I can change the last phrase to "... how you can share your ideas" to avoid the "us", It's after all a community we'd like to build,

PR #4914, opened 2019-06-25
Fix header area of aadUserConversationMember

iamdavidfrancis:
Header data is badly formatted for aaduserconversationmember resource page.


Lauragra:
Good catch - thanks.

PR #5009, opened 2019-07-04
Fix Wrong Parameter on One Note Tab Configuration

AriGunawan:
Should pass groupId instead of sectionsUrl


Lauragra:
@clearab , please review this PR.

PR #5047, opened 2019-07-09
Tg/msgraph custom request

titusgicheru:
- How to Send Custom HTTP Request using GraphClientFactory


concepts/sdks/custom-requests.md
```diff
@@ -0,0 +1,80 @@
...
+
+It provides native HTTP Client Support and therefore supports sending HTTP requests and receiving HTTP responses from a resource identified by a URI. 
+
+## When should you use this?
+
+Sometimes, the functionality that you want to use isn't a part of the client library. In this case, you can still use the client library to make your life easier.
+
+
+### Benefits of GraphClientFactory
+- The **GraphClientFactory** provides ability to send arbitrary requests
```
darrelmiller: GraphClientFactory doesn't allow you to send arbitrary requests.  It allows you to create a HttpClient.

concepts/sdks/custom-requests.md
```diff
@@ -0,0 +1,80 @@
...
+It provides native HTTP Client Support and therefore supports sending HTTP requests and receiving HTTP responses from a resource identified by a URI. 
+
+## When should you use this?
+
+Sometimes, the functionality that you want to use isn't a part of the client library. In this case, you can still use the client library to make your life easier.
+
+
+### Benefits of GraphClientFactory
+- The **GraphClientFactory** provides ability to send arbitrary requests 
+- The client library can authenticate your requests and provide you the serializer.
```
darrelmiller: It doesn't have any knowledge of the serializer

concepts/sdks/custom-requests.md
```diff
@@ -0,0 +1,80 @@
...
+## Sample Code
+### Get MailboxSettings
+
+    MailboxSettings mailboxSettings = GetCustomRequestAsync<MailboxSettings>("me/MailboxSettings").GetAwaiter().GetResult();
+
+### Generic Custom Request
+
+    private static async Task<T> GetCustomRequestAsync<T>(string resourcePath)
+    {
+        _graphClient = GraphServiceClientProvider.GetAuthenticatedClient();
```
darrelmiller: How can this depend on a static field?  Where is that static field defined?

concepts/sdks/custom-requests.md
```diff
@@ -0,0 +1,80 @@
+# Send Custom HTTP Request using GraphClientFactory
```
MIchaelMainer: Let's use sentence casing for headings. Also, let's make this topic more generic so we can add information for all platforms into this topic. I expect that we may want a section a for each platform in this topic since there is some explanation needed.

The title should be something like:
# Send custom HTTP requests

concepts/sdks/custom-requests.md
```diff
@@ -0,0 +1,80 @@
...
+
+The **GraphClientFactory** provides ability to send arbitrary requests and get back a HttpResponseMessage. The response can then be handled by a ResponseHandler to deserialize the response object.
```
MIchaelMainer: In the .Net section, which should be an H2. Don't add anymore subsections, delete all of the other subsection titles.. All of the information about the GraphClientFactory can be contained in a single section.

> The **GraphClientFactory** provides ability to send arbitrary requests and get back a HttpResponseMessage. 

That is incorrect. The **GraphClientFactory** provides an HttpClient object preconfigured with default middleware.

concepts/sdks/custom-requests.md
MIchaelMainer: We don't need the GraphServiceClient for this example. Just use the GraphClientFactory with an AuthenticationProvider to get an HttpClient. I'd add a code comment that specifies what features are on the HttpClient: RedirectHandler, RetryHandler, CompressionHandler, AuthenticationHandler.

* We should have just one example. The MailboxSettings is a good example (get rid of the GraphServiceClient part of it).
* No need to make the example method static.
* Explicitly type variables.

Lauragra:
@titusgicheru , @MIchaelMainer , @darrelmiller , FYI, I'm closing this PR. Please open a new one if it's still needed.

PR #5052, opened 2019-07-10
chore: Remove Appveyor config

nschonni:
Looks like build moved to Azure Pipelines


Lauragra:
@nschonni , thanks for this PR. We're verifying internally that we can delete this file and we'll take care of this on our end. Thanks.

PR #5062, opened 2019-07-10
fix: Relative redirect URL in redirect

nschonni:


.openpublishing.redirection.json
```diff
@@ -26774,6 +26774,16 @@
             "source_path": "concepts/training.md",
             "redirect_url": "/graph/tutorials",
             "redirect_document_id": true
+        },
```
Lauragra: We need to keep the two index files in the folders. Can you please remove the file deletion from this PR?

nschonni: Curious, is it because the URL might change between versions? That's the main reason I've seen before why some repos need the manual redirect files to stay in place. The other was because content was about to be added to the file, but that doesn't appear to be the case here.

nschonni: Sorry, just as an example this is what I mean for the changing every release part https://github.com/dotnet/docs/blob/master/docs/csharp/whats-new/index.md
They use another piece of metadata `ms.custom: "updateeachrelease"`, but it isn't something the Docs script understands 😦

Lauragra: I was told that our build system looks for an index file at the root of the folder, and these index files are in our contextual TOCs, so we want to leave the files in place so that we don't risk breaking our builds. Thanks.

nschonni: Is it worth trying on the review.docs.microsoft.com to see if that is an issue, or does that not really work for these changes? (sorry, I can't see those URLs since they're internal to MS)

PR #5068, opened 2019-07-10
Updates to appRole field descriptions

psignoret:
Note on character limitations in `value` and added a note and description for `origin`.


api-reference/beta/resources/approle.md
```diff
@@ -39,12 +39,14 @@ Here is a JSON representation of the resource
...
 |displayName|String|Display name for the permission that appears in the admin consent and app assignment experiences.|
-|id|Guid|Unique role identifier inside the **appRoles** collection.|
-|isEnabled|Boolean|When creating or updating a role definition, this must be set to **true** (which is the default). To delete a role, this must first be set to **false**.  At that point, in a subsequent call, this role may be removed.|
-|value|String|Specifies the value of the roles claim that the application should expect in the authentication and access tokens.|
+|id|Guid|Unique role identifier inside the **appRoles** collection. When creating a new app role, a new Guid identifier must be provided. |
+|isEnabled|Boolean|When creating or updating an app role, this must be set to **true** (which is the default). To delete a role, this must first be set to **false**.  At that point, in a subsequent call, this role may be removed.|
+|origin|String| Read-only. Specifies if the app role is defined on the Application object or on the ServicePrincipal object. Must _not_ be included in any POST or PATCH requests. |
```
dkershaw10: Remove last sentence.  You've already said this is read-only.

api-reference/beta/resources/approle.md
```diff
@@ -39,12 +39,14 @@ Here is a JSON representation of the resource
...
+|value|String|Specifies the value which will be included in the `roles` claim in authentication and access tokens. Must not exceed 120 characters in length. Allowed characters are `:` `!` `#` `$` `%` `&` `'` `(` `)` `*` `+` `,` `-` `.` `/` `:` `;` `<` `=` `>` `?` `@` `[` `]` `^` `+` `_` <code>&#96;</code> `{` `|` `}` `~`, as well as characters in the ranges `0-9`, `A-Z` and `a-z`. Any other character, including the space character, are not allowed.  |
```
dkershaw10: This isn't rendering properly.  I think you need to escape the |  (as md thinks that's the end of the table row, and you lose everything after it...)

psignoret: It renders OK in the preview, I think it's just the GitHub markdown that doesn't work.

api-reference/beta/resources/approle.md
psignoret: Normally you can include other read-only properties (unchanged) when you PATCH. You actually have to _remove_ this property (even if you don't change the value) for it to succeed.

api-reference/beta/resources/approle.md
psignoret: Escaped additional characters, just in case (now renders in both GitHub preview and Docs preview).

dkershaw10: Yeah - looks better.  Nice fix.
In fact it's SO good, can you make a similar update to https://docs.microsoft.com/en-us/graph/api/resources/oauth2permissiongrant?view=graph-rest-beta please?

:)

PR #5071, opened 2019-07-10
fix for git issue 4226

davidmu1:


api-reference/beta/api/governanceresource-register.md
```diff
@@ -2,60 +2,72 @@
 title: "Register governanceResource"
 description: "Register an unmanaged governanceResource object in PIM."
 localization_priority: Normal
+author: davidmu1
```
Lauragra: The title here and in the H1 should be "governanceResource: register".

davidmu1: @Lauragra - I did see this format described in the guidelines as applying to Functions. I'm assuming that you are suggesting the change because you recognized this topic as being related to a Function. Can you tell me how you recognized that?

Lauragra: I did a search in the beta metadata: 
<Action Name="register" IsBound="true">
<Parameter Name="bindingParameter" Type="Collection(microsoft.graph.governanceResource)"/>
<Parameter Name="externalId" Type="Edm.String" Unicode="false"/>
</Action>

PR #5082, opened 2019-07-11
Remove references to preauthentication in OneNote since these are not…

dconger-microsoft:
… supported in Graph.


davidconger:
@manjushm-microsoft & @jewan-microsoft

PR #5088, opened 2019-07-11
fix: Unordered list formatting

nschonni:
Unordered list indentation


concepts/security-splunk-siemintegration.md
```diff
@@ -61,10 +61,6 @@ Security alerts are highly privileged data typically viewable only by security r
 
 3. Prepare a diagnostic setting request JSON file like the following:
 
-<!-- {
```
Lauragra: Please preserve the HTML comments in files.

nschonni: I can try and add it back, but it would have to be indented at least, because it breaks the OL parsing calculation

nschonni: I added it back, but indented it so it doesn't confuse the markdown parser

PR #5111, opened 2019-07-12
Added new userScopeFilterUri example

linclim:
Single user selection is a common use case, adding an example in the documentation


Lauragra:
@tlenig , please review/ approve.

tlenig:
@linclim can we also specify that you can run on just your mailbox if you specify only one group with one member using the allowedGroupId parameter? Something in the first paragraph like "Specify your user selection in the SourceDataSet of the Azure Data Factory copy activity. To run on a list of groups, add a new field **allowedGroups** under **typeProperties** and set this to a list of up to 10 groups' **Object Ids** separated by commas. If no groups are specified by default, data will be extracted for the entire organization. You can run on one user by creating one group, adding one user to it and adding the **Object Id** of that group to the **allowedGroups** field.

Lauragra:
@tlenig , please review this PR and let us know if we can merge. Thanks!

PR #5114, opened 2019-07-12
Add documentation for expiratinDateTime and password

ificator:
`permission.md` documentation makes reference to `password` and `expirationDateTime` but the `createLink` documentation does not. The schema is published, so you can call the API as expected, the documentation just doesn't align. This PR fixes that.


Lauragra:
@ifcator, pls. include an entry for the changelog for the new properties. What is your timeline for merging this PR?

ificator:
It isn't a high priority, but we've had a couple of questions that would have been answered if this documentation was present. What's the usual timeline for such a change?

Lauragra:
I set it to ready to merge and we can publish on Monday.

PR #5119, opened 2019-07-14
chore: Remove untracked files

nschonni:
Files are gitignored by still exist in the git index.
Ran `git rm -rf --cached .` and then `git add .`.
The _theme folder appears to be an orphan gitsubmodule


Lauragra:
@jasonjoh , can you pls. verify that we can delete these files? Thanks.

PR #5156, opened 2019-07-17
fix for workitem 750279

davidmu1:


Lauragra:
@clearab, please review and sign off.

clearab:
@davidmu1 There is a pending change that I think models what you're going for with the change better here: https://github.com/microsoftgraph/microsoft-graph-docs/blob/d20bd4b43db4c22dd952ce14c273adf1beb5b9dc/api-reference/beta/toc.yml

Can you take a look at that and let me know if you think that will work?

davidmu1:
@clearab : Fine with me. I will close this PR and the workitem. Are you publishing the change soon?

Issue #5167, opened 2019-07-18
Mute is not in the table of contents


nkramer:
add it to toc.yaml

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9b6d5c51-529b-8d06-273b-c04859878b67
* Version Independent ID: d7040fd7-ebb2-d127-1da6-9706a5c941fc
* Content: [call: mute - Microsoft Graph beta](https://docs.microsoft.com/en-us/graph/api/call-mute?view=graph-rest-beta&tabs=http#feedback)
* Content Source: [api-reference/beta/api/call-mute.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/call-mute.md)
* Product: **microsoft-teams**
* Technology: **microsoft-graph**
* GitHub Login: @VinodRavichandran
* Microsoft Alias: **MSGraphDocsVteam**

dplotnikov:
@nkramer and @VinodRavichandran I've made a PR, feel free to merge it

ananmishr:
@dplotnikov I'm closing this issue since the PR has been merged in

PR #5195, opened 2019-07-22
Mute is not in the table of contents

dplotnikov:
https://github.com/microsoftgraph/microsoft-graph-docs/issues/5167

### DIFF
--- a/api-reference/beta/toc.yml
+++ b/api-reference/beta/toc.yml
@@ -811,8 +811,10 @@
             href: api/participant-invite.md
         - name: Change screen sharing role
           href: api/call-changescreensharingrole.md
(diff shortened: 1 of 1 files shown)


api-reference/beta/toc.yml
```diff
@@ -811,8 +811,10 @@
             href: api/participant-invite.md
         - name: Change screen sharing role
           href: api/call-changescreensharingrole.md
-        - name: Unmute self
+        - name: Unmute itself
```
Lauragra: Unmute application and Mute application might be better TOC titles for these APIs.

dplotnikov: @Lauragra I've just made this change

PR #5198, opened 2019-07-22
toc testing version B

davidmu1:


api-reference/beta/toc.yml
```diff
@@ -585,6 +585,134 @@
...
+        - name: List sign-ins
+          href: api/signin-list.md
+        - name: Get sign-in
+          href: api/signin-get.md
+      - name: Provisioning logs
+        href: resources/provisioningobjectsummary.md
+        items:
+        - name: List directory provisioning
+          href: api/provisioningobjectsummary-list.md
+    - name: Authentication and security
```
dkershaw10: Wonder whether this should be called, Authentication, security and compliance, or Authentication and identity compliance.

I worry that putting security in the title will confuse folks with the Security Graph.

davidmu1: It seems that most of the API under the node is related to authentication, maybe we should just use Authentication?

PR #5270, opened 2019-07-24
update to clarify that this only works for enterprise users.

BarbaraSelden:
per Dan Kershaw answer to queston on SO@MS


api-reference/v1.0/api/user-list-licensedetails.md
```diff
@@ -8,7 +8,7 @@ ms.prod: "microsoft-identity-platform"
 
 # List licenseDetails
 
-Retrieve a list of licenseDetails objects.
+Retrieve a list of licenseDetails objects for enterprise users.
```
Lauragra: Please make the same update to the beta version of the topic. Thanks.

PR #5278, opened 2019-07-25
Adding doc page to post chat messages

nkramer:
The page for the API already exists, and at least one customer is found it, but it's not linked to from anywhere. I added one link but we should also link to it from the table of contents and from the chat resource page.


api-reference/beta/api/chat-post-messages.md
```diff
@@ -12,6 +12,7 @@ doc_type: "apiPageType"
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
 Create a new [message](../resources/chatmessage.md) in the specified [chat](../resources/chat.md).
+Note there is no API to create a new chat, nor an efficient way to find an existing chat short of enumerating all of the user's chats.
```
Lauragra: Should you say this in the context of this API, i.e.: "This API does not create a new chat or allow to to find an existing chat. To find an existing chat, you must enumerate through all the user's chats." I'm missing the context for why we're talking about finding an existing chat in a create topic, however.

clearab: updated the verbiage

PR #5292, opened 2019-07-26
Update site-get.md

tposzytek:
Small issue: was "A **site** is addressed be a unique identifier" and I changed it to "A **site** is addressed by a unique identifier". be to by.


Lauragra:
Good catch! Thanks.

PR #5294, opened 2019-07-26
Change notifications for Teams messages, including message properties (preview)

piotrci:
This PR is still work in progress. We are finalizing the implementation with Teams, and several sections of the doc are subject to change. We will keep updating as the details are finalized.


concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,455 @@
...
+
+- Teams messages (preview)
+  - New or changed messages in Team channels in the entire organization (tenant): `/teams/allMessages`
+  - New or changed messages in all chats in the entire organization (tenant): `/chats/allMessages`
+
+## Creating a subscription
+
+To have resource data included in change notifications, you need to provide additional properties when creating a subscription:
+
+- @@@Name may change@@@**includeProperties** set to `true` to explicitly request resource data.
```
Lauragra: This section includes placeholders - please finalize.

piotrci:
@Lauragra , this is a pretty large doc here. Can you help us with the review so we can sign off on it so it is ready to go? for now we want to keep the doc un-published, until the business team is ready, but it would be good to have it fully signed off and ready to go on short notice.

api-reference/beta/resources/subscription.md
```diff
@@ -52,13 +56,16 @@ Here is a JSON representation of the resource.
...
 | id | string | Unique identifier for the subscription. Read-only. |
 | applicationId | string | Identifier of the application used to create the subscription. Read-only. |
 | creatorId | string | Identifier of the user or service principal that created the subscription. If the app used delegated permissions to create the subscription, this field contains the id of the signed-in user the app called on behalf of. If the app used application permissions, this field contains the id of the service principal corresponding to the app. Read-only. |
+| includeResourceData | boolean | Optional. When set to `true`, change notifications will [include resource data](/graph/webhooks-with-resource-data) (such as content of a Teams message). | 
+| encryptionCertificate | string | Optional. Required when `includeResourceData=true`. A base64-encoded representation of a certificate with a public key used to encrypt resource data in notifications. |
```
angelgolfer-ms: Please rephrase and reformat the sentence in rows 67 and 68, from:
Required when `includeResourceData=true`.
to:
Required when **includeResourceData** is true.

angelgolfer-ms: That is, remove the enclosing tick marks, and bold the property name.

api-reference/beta/resources/subscription.md
```diff
@@ -71,6 +78,7 @@ Here is a JSON representation of the resource.
 | Group conversations | 4230 minutes (under 3 days)    |
 | Drive root items    | 4230 minutes (under 3 days)    |
 | Security alerts     | 43200 minutes (under 30 days)  |
+| Teams messages (all teams, all channels)     | 60 minutes (1 hour)  |
```
angelgolfer-ms: Global: Please use the term "chat message" to represent a chatMessage instance in friendly, non-programmatic wording. Please do not use "Teams message" or "message".

Please make this change in other instances; I won't be commenting on every one of them. Thanks.

concepts/changelog.md
```diff
@@ -13,6 +13,12 @@ For details about known issues with Microsoft Graph APIs, see [Known issues](kno
 
 ## October 2019
 
+### Change notifications (webhooks)
+
+| **Change type** | **Version**   | **Description**                          |
+| :-------------- | :------------ | :--------------------------------------- |
+| Addition | beta | Added support for subscribing to notifications that include resource data. Currenttly supported resources are Teams messages in channels and chats. Subscribing apps will have to implement additional authorization and decryption code to benefit from this feature. For more details, see [Set up change notifications for Teams messages, including message properties (preview)](webhooks-with-resource-data.md).|
```
angelgolfer-ms: Maybe you can rephrase the first two sentences in the Description column as:
"Added support for subscribing to notifications that include resource data, on chatMessage {include link to the chatMessage topic} instances that happen in a channel {include link to the channel topic} or chat {include link to the chat topic}".

concepts/webhooks.md
```diff
@@ -18,6 +18,10 @@ After Microsoft Graph accepts the subscription request, it pushes notifications
 > [!div class="nextstepaction"]
 > [Build a webhook app with .NET Core](/graph/tutorials/change-notifications)
 
+By default, change notifications do not contain resource data, other than the `id`. If the app requires resource data, it makes calls to Microsoft Graph APIs to get the full resource. In this article, we use the **user** resource as an example for working with notifications.
+
+An app can also subscribe to change notifications that include resource data, to avoid having to make additonal API calls to access the data. Such apps will need to implement extra code to handle the requirements of such notifications, specifically: responding to subscription lifecycle notifications, validating the authenticity of notifications, and decrypting the resource data. We are gradually extending support for these notifications to additional resource types. For details about how to work with these notificatios, see [Set up notifications for Teams messages, including message properties](webhooks-with-resource-data.md).
```
angelgolfer-ms: Line 23: Please rephrase "We are gradually extending support for these notifications to additional resource types" as
"We are gradually extending support for this type of notifications to additional resource types".

BTW, is this approach of doing notifications called "rich notifications", returning more data in the notification to save the developer an extra call? Apple and Google show examples of their rich notifications in iOS and Chrome, with the additional data benefiting the end user.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+- Teams messages (preview)
+  - New or changed messages in Team channels in the entire organization (tenant): `/teams/allMessages`
+  - New or changed messages in all chats in the entire organization (tenant): `/chats/allMessages`
+
+## Creating a subscription
+
+To have resource data included in change notifications, you need to provide additional properties when creating a subscription:
+
+- **includeResourceData** set to `true` to explicitly request resource data.
+- An optional `$select` operator in the resource path to select the properties to be included.
```
angelgolfer-ms: Line 33: since this topic is about getting property values along with a notification, are we showing an example of appending $select and some properties, to the (partial) URL assigned to the resource property in a create subscription call?

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+- [Validate](#validating-the-authenticity-of-notifications) the authenticity of notifications as having originated from Microsoft Graph.
+- [Provide](#decrypting-resource-data-from-change-notifications) a public encryption key and use a private key to decrypt resource data received through notifications.
+
+This article walks through the details, using the Team **messages** resource as an example.
+
+## Supported resources
+
+The following resources support change notifications that include resource data:
+
+- Teams messages (preview)
```
angelgolfer-ms: Line 24: Please name the specific resource here. 
I'd suggest the following rewording for lines 22-24:
The Microsoft Teams chatMessage {link to the chatMessage entity topic on first reference like here} resource supports change notifications that include resource data. That includes:
- New or changed chatMessage {bold chatMessage on subsequent references like this one} instances in a specific Teams channel {link to the channel resource topic}. To subscribe to message changes in a channel, specify `/teams/{id}/channels/{id}/messages` in the resource property.
-  New or changed chatMessage {bold chatMessage on subsequent references like this one} instances in a specific Teams chat {link to the chat resource topic}. To subscribe to message changes in a chat, specify `/chats/{id}/messages` in the resource property.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+  - New or changed messages in a specific Teams channel: `/teams/{id}/channels/{id}/messages`
+  - New or changed messages in a specific Teams chat: `/chats/{id}/messages`
+  
+## Creating a subscription
+
+To have resource data included in change notifications, you need to provide additional properties when creating a subscription:
```
angelgolfer-ms: Line 30: Please rephrase:
To have change notifications include resource data, you must specify the following properties, in addition to those that are always specified when [creating a basic change subscription](https://docs.microsoft.com/en-us/graph/webhooks?view=graph-rest-beta#creating-a-subscription):

concepts/webhooks-with-resource-data.md
angelgolfer-ms: BTW, maybe there's something I don't get from this topic yet - so if one doesn't specify the $select and any specific properties of chatMessage, are we still getting any "resource data" from a notification? 
I have thought that the "resources data" we're talking about is indeed specific property values of the chatMessage entity. Can you please clarify, @piotrci ?

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+  "resource": "/teams/{id}/channels/{id}/messages",
+  "includeResourceData": true,
+  "encryptionCertificateId": "<customId>",   
+  "encryptionCertificateThumbprint": "<thumbprintFromTheCertificate>",
+  "expirationDateTime": "2019-09-19T11:00:00.0000000Z",
+  "clientState": "<secretClientState>"
+}
+```
+ 
+> **Important:** Use the same hostname for both notifications URLs.
```
angelgolfer-ms: Since lines 75-80 are adjacent to one another, I suggest we style it with just one "Note" and use bullet points.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+> **Note:** You need to validate both notification endpoints as described in [the generic notification article](webhooks.md#managing-subscriptions).
```
angelgolfer-ms: Line 77: Please rephrase:
"You need to validate both notification endpoints as described for [creating a basic subscription](webhooks.md#managing-subscriptions)."

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+If you choose to use the same URL for both endpoints you will receive and respond to two validation requests.
+
+> **Note:** You cannot update (`PATCH`) the existing subscriptions to add the **lifecycleNotificationUrl** property. You should remove such existing subscriptions, and create new subscriptions and specify the **lifecycleNotificationUrl** property.
```
angelgolfer-ms: Line 80: I suggest to say:
"You cannot update an existing subscription to add the... You should remove the existing subscription, create a new subscription, and specify the... property".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Subscription lifecycle notifications
+
+Subscription lifecycle notifications inform you about actions you need to take in order to maintain an uninterrupted flow of notifications. Unlike resource change notifications, which inform about an instance of a resource changing, these notifications tell you about the subscription itself, and its current state in the lifecycle.
+
+These notifications will be delivered to the **lifecycleNotificationUrl**. You should identify the type of notification, and take the corresponding action to ensure that the change notifications continue to flow.
+
+### Re-authorization challenges
+
+The `reauthorizationRequired` notification informs you that a subscription must be re-authorized to maintain the flow of data.
```
angelgolfer-ms: Line 90: Please introduce the concept of "reauthorization required" notification this way (please note the styling as well):
"When the **lifecycleEvent** property in a notification indicates `reauthorizationRequired`, you must re-authorize the subscription to maintain the data flow."

angelgolfer-ms: About styling:
In general, properties and other API elements like entity, complex type, function, action, should be styled as bold.
Enum values should be styled by enclosing them in tick marks.
BTW, I am aware that some of the earlier identity content follow a slightly different styling convention (which is btw suggested by the docs.microsoft.com folks), that styling convention is not the same as the Microsoft Writing Style Guide or the developer division style guide that Microsoft developer docs have used for decades. Just so we know :-) 
We got consensus with the identity team that we should converge and follow the latter.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+Subscription lifecycle notifications inform you about actions you need to take in order to maintain an uninterrupted flow of notifications. Unlike resource change notifications, which inform about an instance of a resource changing, these notifications tell you about the subscription itself, and its current state in the lifecycle.
+
+These notifications will be delivered to the **lifecycleNotificationUrl**. You should identify the type of notification, and take the corresponding action to ensure that the change notifications continue to flow.
+
+### Re-authorization challenges
+
+The `reauthorizationRequired` notification informs you that a subscription must be re-authorized to maintain the flow of data. 
+
+You can create a long-lived subscription (e.g. 3 days), and resource data notifications will start flowing to the **notificationUrl**. However, at any point in time, Microsoft Graph may require that you re-authorize the subscription to prove that you still have access to resource data in case the conditions of access have changed since the subscription was created. Some examples of changes that may impact your apps access to data are:
```
angelgolfer-ms: BTW, if we agree to refer to this feature as "rich notifications", then line 92 would read:
"You can create a long-lived subscription (e.g. 3 days), and rich notifications will start flowing to the **notificationUrl**. "

angelgolfer-ms: Line 92: Please rephrase:
"When you create a long-lived subscription (for example, one that lasts for 3 days), resource data flows to the location indicated by **notificationUrl**. However, … access to resource data, in case conditions of access have changed since the subscription was created. The following are examples of changes that affect your access to data:"

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+- a tenant admin may revoke your app's permissions to read the resource
+- the user providing the authentication token to your app (interactive scenarios) may be subject to dynamic policies based on their location, device state, risk assesment, etc. - see this article for more information on [Azure AD conditional access policies](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/overview). For example, if the user changes their physical location, they may no longer be allowed to access the data, and your app will not be able to re-authorize the subscription.
```
angelgolfer-ms: Line 95: Please rephrase:
"The user who provides the authentication token... based on factors including their location, device state, or risk assessment. As an example, if the user changes their physical location, ...re-authorize the subscription. For more information on dynamic policies that control access, see [Azure AD conditional access policies](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/overview).

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+Given an active, non-expired subscription, the flow looks as follows:
+
+1. Microsoft Graph decides that a subscription requires re-authorization to maintain the resource notification flow.
+    1. The reasons for this may vary from resource to resource, and may change over time. You do not need to understand the specific cases in which re-authorization events occur, you only need to respond to them.
+
+2. Microsoft Graph sends an authorization challenge notification to the **lifecycleNotificationUrl**
+    1. Note that the flow of resource notifications may continue for a while, giving you extra time to respond. However, eventually resource notification delivery will be paused, until you take the required action.
```
angelgolfer-ms: Line 103: "... will pause, until you take the required action".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+3. You can respond to this notification in two ways:
+    1. Re-authorize the subscription; this does not extend the subscription's expiry date.
```
angelgolfer-ms: Line 106: "... extend the expiry date of the subscription".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+    2. Renew the subscription; this both re-authorizes and extends the expiry date.
+
+  Note: both actions require you to present a valid authentication token; same as when creating or renewing a subscription.
+
+4. If you successfully re-authorize or renew the subscription, resource notifications will continue. Otherwise, resource notifications will remain paused.
```
angelgolfer-ms: Line 111: "... will continue to pause."

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+    
+#### reauthorizationRequired notification example
```
angelgolfer-ms: Line 113: Please title the section as "Example: notification to re-authorize"

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+- a tenant admin may revoke your app's permissions to read the resource
+- the user providing the authentication token to your app (interactive scenarios) may be subject to dynamic policies based on their location, device state, risk assesment, etc. - see this article for more information on [Azure AD conditional access policies](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/overview). For example, if the user changes their physical location, they may no longer be allowed to access the data, and your app will not be able to re-authorize the subscription.
+
+Given an active, non-expired subscription, the flow looks as follows:
+
+1. Microsoft Graph decides that a subscription requires re-authorization to maintain the resource notification flow.
```
angelgolfer-ms: @Lauragra , do you mind reformatting the list from line 99 to 111? We shouldn't have an embedded "list" with a single list item, which happens under #1 and #2.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+```
+
+A few things to note about this type of notification:
+- The `"lifecycleEvent": "reauthorizationRequired"` field designates this notification as an authorization challenge.
+- The notification does not contain any information about a specific resource, because it is not related to a resource change, but to the subscription state change.
+- Similar to resource notifications, lifecycle notifications may be batched together (in the **value** array), each with a possibly different **lifecycleEvent** value. Process each notification in the batch accordingly.
+
+#### Actions to take
+
+1. [Acknowledge](webhooks.md#notifications) the receipt of the notification, by responding to the POST call with `202 - Accepted`.
```
angelgolfer-ms: Line 136: Please make the link for "Acknowledge" to point to the particular section titled "Processing the notification", accessible by this URL https://docs.microsoft.com/en-us/graph/webhooks?view=graph-rest-1.0#processing-the-notification.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+2. [Validate](webhooks.md#notifications) the authenticity of the notification.
```
angelgolfer-ms: Line 137: Same change for the link for "Validate" - create a relative link the effectively points to https://docs.microsoft.com/en-us/graph/webhooks?view=graph-rest-1.0#processing-the-notification.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+3. Ensure the app has a valid access token to take the next step. 
+> **Note:** If you are using one of the [authentication libraries](https://docs.microsoft.com/azure/active-directory/develop/reference-v2-libraries) they will handle this for you by either reusing a valid cached token, or obtaining a new token, including asking the user to login again (e.g. with a new password). Note that obtaining a new token may fail, since the conditions of access may have changed, and the caller may no longer be allowed access to the resource data.
```
angelgolfer-ms: Line 139: Please don't format this as a note. @Lauragra , can we format that paragraph in 139 as a paragraph under item #3? Right now, the indentation is not correct either.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+4. Call either of these two APIs - if successful, both will have the affect of resuming the resource notification flow.
+    1. To re-authorize the subscription, without extending its expiration date, call the `/reauthorize` action:
+```http
+POST  https://graph.microsoft.com/beta/subscriptions/{id}/reauthorize
```
angelgolfer-ms: @piotrci , is "reauthorize" indeed an action in the CSDL? If so, please create an API topic for it.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+Content-type: application/json
+```
+  2. To re-authorize and renew the subscription at the same time, perform a regular renew action:
```
angelgolfer-ms: Line 147: please rephrase:
"To authorize and renew the subscription at the same time, extend the expiration date-time for the subscription:

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+```http
+PATCH https://graph.microsoft.com/beta/subscriptions/{id}
+Content-Type: application/json
+
+{
+  "expirationDateTime": "2019-09-21T11:00:00.0000000Z",
+}
+```
+
+> **Note:** This action may fail, because the authorization checks performed by the system may deny the app or the user access to the resource. It may be necessary for the app to obtain a new access token from the user to successfully reauthorize a subscription. You may retry these actions later, at any time, for example when the conditions of access have changed. Any resource changes in the time period from when the lifecycle notification was sent, to when the app re-creates the subscription successfully, will be lost. The app will need to fetch those changes on its own.
```
angelgolfer-ms: Line 157: please do not format that as a note. Up to this point, there are many notes in this article, making the call-out less and less effective.
Just so it's clear this paragraph in line 157 applies to the second choice, we can start by saying:
"Authorizing and renewing the subscription in the same step may fail, because..."

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+  ]
+}
+```
+
+A few things to note about this type of notification:
+- The `"lifecycleEvent": "reauthorizationRequired"` field designates this notification as an authorization challenge.
+- The notification does not contain any information about a specific resource, because it is not related to a resource change, but to the subscription state change.
+- Similar to resource notifications, lifecycle notifications may be batched together (in the **value** array), each with a possibly different **lifecycleEvent** value. Process each notification in the batch accordingly.
+
+#### Actions to take
```
angelgolfer-ms: Line 134: I'd suggest keep each heading to remind customer where they are, because this is a long article with many sections and sub-sections. So changing the section header from "Actions to take" to "Actions to re-authorize".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+Content-Type: application/json
+
+{
+  "expirationDateTime": "2019-09-21T11:00:00.0000000Z",
+}
+```
+
+> **Note:** This action may fail, because the authorization checks performed by the system may deny the app or the user access to the resource. It may be necessary for the app to obtain a new access token from the user to successfully reauthorize a subscription. You may retry these actions later, at any time, for example when the conditions of access have changed. Any resource changes in the time period from when the lifecycle notification was sent, to when the app re-creates the subscription successfully, will be lost. The app will need to fetch those changes on its own.
+
+#### Implementation guidelines
```
angelgolfer-ms: Line 159: Suggest to use the header "Guidelines for implementing re-authorization".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+2. Do not make assumptions about the frequency at which you may see challenges for your subscriptions, it is subject to change. The main reason for these signals is for you, the app developer, not to have to know all the cases in which re-authorization of your subscription is required; instead, Microsoft Graph tells you when you need to take action. Your app should be ready to handle these challenges anywhere from once every few minutes per each subscription to very rarely, only for some subscriptions.
+
+### Future-proof your code
+
+In the future Microsoft Graph will add more types of subscription lifecycle notifications. They will be posted to the same endpoint: **lifecycleNotificationUrl**, but they will have a different value under **lifecycleEvent** and may contain a slightly different schema and properties, specific to the scenario for which they will be issued.
+
+You should implement your code in a future-proof way so it does not break when Microsoft Graph introduces new types of lifecycle notifications. We recommend the following approach:
```
angelgolfer-ms: Line 171: Second sentence which starts with "we". We generally avoid first person plural in docs. Please say: "The following are some guidelines:"

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+1. Explicitly identify each notification as an event that you support, using the **lifecycleEvent** property. For example, look for the `"lifecycleEvent": "reauthorizationRequired"` property to identify a specific event, and handle it.
+
+1. In your app, ignore any lifecycle events that the app does not recognize, and log them to gain awareness.
+
+1. Watch for announcements of notifications for new scenarios, as there may be more types of lifecycle notifications in the future.
```
angelgolfer-ms: It would be helpful to point customers to the Graph blog site here (https://developer.microsoft.com/en-us/graph/blogs/).

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+For basic notifications, which do not contain resource data, this validation can be simplified and based on the **clientState** value, as described [here](webhooks.md#notifications). That is acceptable, because you rely on additional trusted Microsoft Graph calls to get access to resource data, and therefore the impact of any spoofing attempts is limited. A more thorough validation is recommended before you process resource data delivered via notifications.
+
+### Validation tokens in the notification
+
+A notification contains an additional property - **validationTokens** - containing an array of JWT tokens generated by Microsoft Graph; there will be a single token for each distinct app and tenant pair for whom there is an item in the **value** array. Keep in mind that notifications may contain a mix of items for various apps and tenants that subscribed using the same **notificationUrl**.
```
angelgolfer-ms: @piotrci , does the validationTokens property exist _only_ in notifications that include resource data?
Or is this property applicable to basic notifications and notifications with resource data alike?
Either way, I think we should clarify that in line 189 when we introduce the property.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+	],
+	"validationTokens": [
+		"eyJ0eXAiOiJKV1QiLCJhb...",
+    "cGlkYWNyIjoiMiIsImlkc..."
+	]
+}
+```
+### How to validate
+
+If you are new to token validation, this [blog article](http://www.cloudidentity.com/blog/2014/03/03/principles-of-token-validation/) contains a useful overview. We recommend that you use an SDK, such as Microsoft's [System.IdentityModel.Tokens.Jwt](https://www.nuget.org/packages/System.IdentityModel.Tokens.Jwt/) library for .NET, or a 3rd party library for a different platform.
```
angelgolfer-ms: Line 217: please do not use "we" in documentation. You can rephrase this paragraph:
"If you are new to token validation, refer to this [blog article](http://www.cloudidentity.com/blog/2014/03/03/principles-of-token-validation/) for a useful overview. Use an SDK, such as Microsoft's [System.IdentityModel.Tokens.Jwt](https://www.nuget.org/packages/System.IdentityModel.Tokens.Jwt/) library for .NET, or a 3rd party library for a different platform.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+> **Note:** Make sure to always send a `202 - Accepted` status code in in the response to the notification; do that before validating the notification (e.g. if you store notifications in queues for later processing) or after (if you process them on the fly) - even if validation failed. Accepting a notification prevents unnecessary delivery retries and it also prevents any potential rogue actors from finding out if they passed or failed validation. You can always choose to ignore an invalid notification after you have accepted it.
+
```
angelgolfer-ms: Please add a lead-in sentence before the numbered list in line 221-234. For example, a sentence like the following:
"Use the following steps to validate tokens and apps that generate tokens:"

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+3. Validate that the token was issued for your app that is subscribing to change notifications.
+  - Validate the "audience" in the token matches your app id.
+  - If you have more than one app receiving notifications, make sure to check for multiple ids
+
+> **Note:** The above steps are part of standard validation logic in JWT token libraries and can typically be executed as a singlee function call. 
+
+4. **Critical**: Validate that the app that generated the token represents the Microsoft Graph change notification publisher. This ensures that notifications are not sent by a different app that is not Microsoft Graph
+  - Check that the **appid** property in the token matches the expected value of `0bf30f3b-4a52-48df-9a82-234910c4a086`.
+
+This validation should be performed on every JWT token in the **validationTokens** array. If any tokens fail, you should consider the notification suspicious and investigate further.
```
angelgolfer-ms: Line 236: Per Microsoft Voice, please use active voice and trim redundant words. I suggest:
"Perform validation on every JWT token in the **validationTokens** array. If any tokens fail, consider the notification suspicious and investigate further. "

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+### Managing encryption keys
+
+1. You need to obtain a certificate with a pair of assymetric keys. The certificate can be self-signed - Microsoft Graph does not verify the certificate issuer, but only uses the public key for encryption. We recommend [Azure Key Vault](https://docs.microsoft.com/azure/key-vault/key-vault-whatis) as the solution to create, rotate and securily manage certificates.
+  - the key must be of type `RSA`
+  - the key size must be between 2048 and 4096 bits
+
+2. Export the certificate **ensuring that only the public key** is included. The certificate must be exported as a Base64-encoded X.509 certificate.
```
angelgolfer-ms: Line 269 - is the bolding for emphasis?
I suggest a slight reword, to reduce chance of confusion for Loc:
"Export the certificate, and include only the public key. Make sure to export the certificate as a base64-encoded, X.509 certificate".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+3. When creating the subscription:
+  - Provide the certificate in the **encryptionCertificate** property; use the Base64-encoded content the certificate was exported in.
```
angelgolfer-ms: Line 272: please keep each sentence short and reduce the use of semi-colons. We typically use semi-colons between two self sufficient sentences if there is a close relationship between the sentences. 
Also, please note "Base64" should be lower caps "base64".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+  - Provide your own identifier in the **encryptionCertificateId**. It will allow you to match your certificates to the notifications you receive and can also be used to retrieve certificates from your certificate store. The length of this id can be up to 128 characters.
+
+4. The private key must be securily held and managed by you. Your notification processing code will need to have access to the private key to decrypt resource data.
+
+#### Rotating keys
+
+It is a recommmended practice to periodically change your assymetric keys to minimize the risk of a private key becoming compromised. If you want to introduce a new pair of keys, follow these steps:
```
angelgolfer-ms: Line 279: "To minimize the risk of a private key becoming compromised, periodically change your assymetric keys. If necessary, follow these steps to introduce a new pair of keys:"

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+### How to decrypt
+
+To optimize performance, Microsoft Graph uses a two-step encryption process:
+  - It generates a single use symmetric key, and uses it to encrypt resource data.
+  - It uses the public assymetric key - provided by you - to encrypt the symmetric key and includes it in the notification.
+
+To decrypt resource data, your app should perform the reverse steps, using the properties under **encryptedContent** in each notification item:
+
+1. Use the **encryptionCertificateId** property to identify the certificate that should be used.
```
angelgolfer-ms: Line 298: "... to identify the certificate to use".

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+2. Initialize an RSA cryptographic component (such as the .NET [RSACryptoServiceProvider](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsacryptoserviceprovider.decrypt?view=netframework-4.8)) with the private key.
+
+3. Use it to decrypt the symmetric key delivered in the **dataKey** property of each item in the notification.
+  - Note: the decryption algorithm should use the OAEP padding.
```
angelgolfer-ms: Line 303, I believe you should please expand OAEP on first reference, like the following:
"...should use  Optimal Asymmetric Encryption Padding (OAEP)."

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+
+4. Use the symmetric key to calculate the HMAC-SHA256 signature of the value in **data**.
+  - Compare it to the value in **dataSignature**. If they do not match, assume the payload has been tampered with and do not decrypt it.
+
+5. Use the symmetric key with an AES cryptographic component (such as the .NET [AesCryptoServiceProvider](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.aescryptoserviceprovider?view=netframework-4.8)) to decrypt the content in **data**.
```
angelgolfer-ms: Line 308: please expand AES as "Advanced Encryption Standards (AES)" on first mention.

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
+---
+title: "Set up change notifications for Teams messages, including message properties (preview)"
+description: "Microsoft Graph uses a webhook mechanism to deliver change notifications to clients. Notifications can include resource properties."
+author: "piotrci"
+localization_priority: Priority
+---
+
+# Set up change notifications for Teams messages, including message properties (preview)
```
angelgolfer-ms: Line 8 the topic title: @piotrci , if you agree that we should use "rich notifications" to refer to this approach, then I'd suggest the topic title:
"Set up rich notifications for channel and chat messages in Microsoft Teams (preview)"
And the TOC title:
"Rich notifications for Microsoft Teams (preview)"

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+  - Padding: PKCS7
+  - Cipher mode: CBC
+  - The "initialization vector" must be set by copying the first 16 bytes of the symmetric key used for decryption.
+
+6. The decrypted value is a string json representation of the resource included in the notification.
+
+> **Note:** always assume that the symmetric key is different for each item in the notification.
+
+#### Sample notification with encrypted resource data
+
```
angelgolfer-ms: @piotrci , I think it would be really helpful to describe for the example below, what the actual returned resource data is (in the encryptedContent property). Something like, I have to guess:
>> This example returns encrypted property values of the chatMessage specified by the id (teams('88cbc8fc-164b-44f0-b6a6-b59b4a1559d3')/channels('19:8d9da062ec7647d4bb1976126e788b47@thread.tacv2')/messages('1565293727947')/replies('1565293727947'))

Is that correct, btw?

concepts/webhooks-with-resource-data.md
```diff
@@ -0,0 +1,426 @@
...
+This article walks through the details, using the Team **messages** resource as an example.
+
+## Supported resources
+
+The following resources support change notifications that include resource data:
+
+- Teams messages (preview)
+  - New or changed messages in a specific Teams channel: `/teams/{id}/channels/{id}/messages`
+  - New or changed messages in a specific Teams chat: `/chats/{id}/messages`
+
```
angelgolfer-ms: Per my comment around line 33 below, I'd suggest we clarify, after specifying either of these specific resource types, what kind of resource data and from which properties would an app get the actual data. Even though this is at the beginning of the topic when customers do not know the details how this is done, we can describe it in a high level way and link to specific sections so that customers can refer to the details when they re-read the article (which is almost assumed to happen for a long article like this).

So for example, we can say (assuming you are ok with using the term "rich notification"):
>>
In a rich notification, you can find the following actual resource data, for a chatMessage instance in scope for the subscription:
(1) The id and type information of the chat message instance.
(2) All the other property values of the chatMessage instance, encrypted as specified in the notification. See an example {link to the section #sample-notification-with-encrypted-resource-data} that shows the encrypted data returned in the encryptedContent property.
(3) Specific properties of the chatMessage instance. This is optional and can be specified as part of the **resource** property URL when creating the subscription. Such property values are also returned encrypted in the encryptedContent property.

@piotrci , can you confirm if this is how things work please, so that we can add the clarification up top? Thanks.

concepts/webhooks-with-resource-data.md
piotrci: We specifically decided not to use rich notifications, a long time ago. I think the feedback around style was to give descriptive names to features like these, as opposed to create feature specific terms.

concepts/webhooks-with-resource-data.md
piotrci: I was planning to make this a generic conceptual article for all resources, not just teams. chat messages will be used as an example, similarly to how we use another resource for the basic notification conceptual doc.
For that reason I think we should list all resources entity by entity. @angelgolfer-ms , do you think we should change the title to make it generic?

concepts/webhooks-with-resource-data.md
piotrci: I'd rather leave it the way it is. assume this section will eventually list all resources from various workloads that support this feature.

concepts/webhooks-with-resource-data.md
piotrci: I'm not sure i understand the question. Here is how the feature works:
- resource properties are included in the notification. they are always encrypted under the encryptedContent property.
- what properties are included, depends: some workloads will require a $select statement and only those properties will be included. other workloads (Teams) don't support $select (or had to deprioritize it) so all properties are included.
I'm not sure I want to explain this nuance to the developer right now.

concepts/webhooks-with-resource-data.md
angelgolfer-ms: Ok, I get that you'd like to make the addition less Teams-specific. 
I'm adding the following, so that customers don't have to wade through 60% of the long article to find exactly what resource data they can get from a notification.

"Change notifications include the following resource data in the payload:

- ID and type of the resource instance, returned in the resourceData property.
- All the property values of that resource instance, encrypted as specified in the subscription, returned in the encryptedContent property.
- As an alternative to returning full resource data, specific properties of that resource, encrypted in the encryptedContent property. To get only specific properties, you would have specified them as part of the resource URL in the subscription, using a $select parameter."

concepts/webhooks-with-resource-data.md
angelgolfer-ms: I see.

PR #5327, opened 2019-07-31
fix for git issue 5282

davidmu1:


andrueastman:
@davidmu1 
Apologies for not mentioning this as well. 
Ideally, we should not add the two lines [here](https://github.com/microsoftgraph/microsoft-graph-docs/pull/5327/files#diff-465f8093fb5eb2ccd4a5ea1a132ca084R134) and [here](https://github.com/microsoftgraph/microsoft-graph-docs/pull/5327/files#diff-465f8093fb5eb2ccd4a5ea1a132ca084R134) that have the tabs syntax unless they are accompanied by other tabs. 

The lines look like as below.
`# [HTTP](#tab/http)` 

As this will only create one tab group (i.e for http for now) there is no need to set them now and therefore removing the two lines should fix the warnings above. 
The tabs will be added automatically as when example snippets are generated.

api-reference/beta/api/user-update.md
```diff
@@ -84,9 +85,12 @@ add, update, or delete your own app-specific data in custom properties of an ext
 ## Response
 
 If successful, this method returns a `204 No Content` response code.
-## Example
-##### Request
-Here is an example of the request.
+
+## Examples
+
+### Request 1
```
Lauragra: @davidmu1 , when there's multiple examples, we create H3s for each example. So you would have this heading: ### Example 1: {description}. Then you make the Request and Response headings H4s. Please apply the same to the numbered examples in this PR. Thanks.

davidmu1: Changed.

davidmu1:
#sign-off

api-reference/beta/api/user-update.md
```diff
@@ -84,9 +85,16 @@ add, update, or delete your own app-specific data in custom properties of an ext
 ## Response
 
 If successful, this method returns a `204 No Content` response code.
-## Example
-##### Request
-Here is an example of the request.
+
+## Examples
+
+### Example 1
```
Lauragra: @davidmu1 , can you please fix the examples? The H3 headings need to have a descriptor, so for this one: "Example 1: Update properties of the signed in user". Then for the Request and Response headings, remove the number. This is the standard structure to apply throughout.

davidmu1: Fixed.

davidmu1:
#sign-off

PR #5316, opened 2019-07-30
add AccessReview.ReadWrite.Membership as an app permission

markwahl-msft:
update Access Reviews API in beta with new app permission AccessReview.ReadWrite.Membership

PR #5328, opened 2019-07-31
add accessreview.readwrite.membership as a delegated permission

markwahl-msft:
Following on to #5316 , this PR updates the Azure AD Access Reviews API in beta with new delegated permission AccessReview.ReadWrite.Membership


markwahl-msft:
#sign-off

PR #5331, opened 2019-07-31
Update the Calendar docs to mention that they work with Group.Read.All.

nkramer:
There are probably additional Calendar pages that should be updated as well, I'm not the person to do that.


nkramer:
I expect so but I have not confirmed.

angelgolfer-ms:
> I expect so but I have not confirmed.

No problem, @nkramer . I researched into the other few topics and can update them in a separate PR.

api-reference/beta/api/calendar-get.md
```diff
@@ -25,9 +25,9 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type      | Permissions (from least to most privileged)              |
 |:--------------------|:---------------------------------------------------------|
-|Delegated (work or school account) | Calendars.Read    |
+|Delegated (work or school account) | Calendars.Read, Group.Read.All (for calendars in groups), Group.ReadWrite.All (for calendars in groups)    |
 |Delegated (personal Microsoft account) | Calendars.Read    |
-|Application | Calendars.Read |
+|Application | Calendars.Read, Group.Read.All (for calendars in groups), Group.ReadWrite.All (for calendars in groups) |
```
angelgolfer-ms: BTW, @nkramer , per the documented [known issue](https://docs.microsoft.com/en-us/graph/known-issues#groups) for groups, and questions online, group calendars and events do not support delegated personal accounts and application permissions. Can you re-confirm line 30 please?
For now, my PR #5368  assumes that group calendars and events do not support application permission.

PR #5371, opened 2019-08-06
Update teamstab-add.md

AlexanderHolmeset:
Added example on how to add webpage as tab, since example 1 can be confusing since it looks like a website, but is an application.


Lauragra:
@clearab , please review and sign off. Thanks.

clearab:
Thank you, but this is already documented here: https://docs.microsoft.com/en-us/graph/teams-configuring-builtin-tabs

PR #5510, opened 2019-08-22
Calendar url fix

madansr7:
Calendar is a single navigation property so having an ID on it would be incorrect. The correct path should be "calendars" which is a collection navigation property.


Lauragra:
@angelgolfer-ms , please review. This HTTP syntax doesn't exist in the beta version of this topic.

PR #5515, opened 2019-08-23
Update create-file-attachment-from-message-v1-javascript-snippets.md

Abrissirba:
The example produces a 422 response. After removing the wrapped body it works


Lauragra:
@andrueastman , @darrelmiller , please review and approve.

darrelmiller:
@andrueastman It looks like our code generator is making this same mistake for all create and update operations.  We should fix this with a change to the generator instead of hand correcting.  We can close this issue once the fix is in place.

andrueastman:
The issue is already tracked [here](https://github.com/microsoftgraph/microsoft-graph-explorer-api/issues/88). We should see the docs corrected in the course of the week,

andrueastman:
Closing this as the docs have been updated [here](https://github.com/microsoftgraph/microsoft-graph-docs/pull/5533) after a bug fix in the snippet generator.

PR #5577, opened 2019-08-30
Update tablecolumn.md

grangeryy:
Excel graph supports to set the table column name. So it shouldn't be read-only.


api-reference/v1.0/resources/tablecolumn.md
```diff
@@ -32,7 +32,7 @@ Represents a column in a table.
 |:---------------|:--------|:----------|
 |id|string|Returns a unique key that identifies the column within the table. This property should be interpreted as an opaque string value and should not be parsed to any other type. Read-only.|
 |index|int|Returns the index number of the column within the columns collection of the table. Zero-indexed. Read-only.|
-|name|string|Returns the name of the table column. Read-only.|
+|name|string|Returns the name of the table column.|
```
Lauragra: Generally when we update a v1.0 topic, we should make sure that the same change is applied in the beta version. I'm not finding a tablecolumn resource topic in beta. There is a workbookTableColumn resource that only exists in beta and v1.0. Can you please double-check to make sure the documentation is consistent?

Lauragra: I checked the metadata - in both beta and v1.0 the resource name is "workbookTableColumn". Please rename this topic, and then make the update to the name property in the workbookTableColumn topics in beta as well.

grangeryy:
Notice I forgot to correct the links. Just finished and please have a review:)

PR #5629, opened 2019-09-06
Calendar examples backslash to forward slash

mmascolino:
In the calendar examples, converted the backslashes to forward slashes.


Lauragra:
Thanks!

PR #5654, opened 2019-09-10
Update private channel docs

clearab:


api-reference/beta/api/channel-post.md
```diff
@@ -105,6 +106,62 @@ Content-length: 201
 }
 ```
 
+### Example 2: Create private channel on behalf of user
```
phecdalouie: Can we make "Create private channel on behalf of user" Example 3 and include a new Example 2 just covering how to "Create a private channel"?

This basic question comes up quite often so I want to make sure I can point devs to this doc when it comes up. Even thought they should be able to infer this from the two examples you have so far.

clearab: I mean, that example wouldn't really add any additional information beyond what is already available. Do you really think it is necessary?

api-reference/beta/toc.yml
```diff
@@ -9512,6 +9512,12 @@
           href: api/channel-post-messages.md
         - name: Reply to message
           href: api/channel-post-messagereply.md
+        - name: Update the role of a member in channel
```
phecdalouie: Nit: Can we tweak the order of the TOC? Lead with "Add" first, then "Remove", lastly followed by "Update", since this is likely the order developers will implement and use these APIs.

api-reference/beta/toc.yml
```diff
@@ -9512,6 +9512,12 @@
...
+          href: api/conversationmember-update.md
+        - name: Remove a member from a chat
+          href: api/conversationmember-delete.md
+        - name: Add a member to a chat
```
phecdalouie: Can we change the titles of these to say "Add a member to **channel**" and "Remove a member from a **channel**". 

This avoids the confusion with a similar feature in the client experience where you could add/remove users from a group chat, which we don't yet support in our APIs.

phecdalouie:
For the "add conversationMember" doc 
https://docs.microsoft.com/en-us/graph/api/conversationmember-add?view=graph-rest-beta&tabs=http

I still think we should add at least an example or mention that you can pass in either nothing for roles (which would default add the user a member), and support for roles = [owners] which would add the user as an owner.

PR #5706, opened 2019-09-17
Update workbook.md

grangeryy:
Update the Excel workbook


api-reference/v1.0/includes/snippets/csharp/get-workbookapplication-csharp-snippets.md
```diff
@@ -0,0 +1,13 @@
+---
```
Lauragra: @grangeryy , did you intend to add these snippets as part of the PR? I thought these files were autogenerated via the weekly snippet automation.

api-reference/v1.0/api/workbookapplication-calculate.md
```diff
@@ -0,0 +1,103 @@
+---
```
Lauragra: @grangeryy , just to double check, did you commit and sync these files via git (VS Code, GitHub application, git shell) or did you upload them directly? We need to make sure that we don't upload files without using git to ensure that we don't have issues with line endings.

Lauragra:
@grangeryy , I set this to Ready to merge because it looks like the APIs are already in v1.0. It's set to Content review so that I know to review it. We need to resolve my questions/feedback before we can publish. Thanks.

api-reference/v1.0/api/workbookapplication-calculate.md
grangeryy: Hi @Lauragra, I update load these files directly, not through git. So what should be the proper steps I should take next?

api-reference/v1.0/includes/snippets/csharp/get-workbookapplication-csharp-snippets.md
grangeryy: @Lauragra, actually I'm not quite sure about what these snippets are. The thing is, we have documentation for beta version but not for v1.0 version. My intention is to have the same docs for v1.0 so I copied files from beta and modified them for v1.0. At first, I didn't include these snippets, but the system check didn't pass so I added them. What's your suggestion here?

api-reference/v1.0/api/workbookapplication-calculate.md
Lauragra: @grangeryy , in the future, you can use git to upload files - either git shell, the github desktop app, or VSCode. We discovered an issue with the line endings when MD files are uploaded directly rather than committed via Git.

api-reference/v1.0/includes/snippets/csharp/get-workbookapplication-csharp-snippets.md
Lauragra: @darrelmiller , @andrueastman , our snippets generation process will pick up the new files in v1.0 right? Just to confirm, we can remove the snippet files that were added in this PR.

darrelmiller: Yes, the snippet generation process will add them automatically.

PR #5765, opened 2019-09-20
Resolving customer issues

### LINKED ISSUE #5572: Delta Queries only work for some tenants
I have wasted a month already being chased from pillar to post by MPN Support.
We are busy developing a service to sync user profile images from Microsoft Graph to our system for clients using our SaaS platform.
The list of clients includes 200 clients and 4’000’000 Microsoft users about are affected by this.
 
Doing full sync daily from Microsoft Graph is suicide on our end and it will be murder on the Microsoft Graph infrastructure as we run these syncs globally across the globe for all our clients.
To get around us having to hammer the Graph service, we opted to run full sync once per client, store the delta token and then from that point on wards we only sync updates from the delta every 30 minutes.
 
During development, we tested this against one of our test tenants and it all went well. The full sync completes and from that point on wards we would see delta results if a user updated their profile image.
We then started testing against our primary tenant and some client tenants and very quickly found that the delta query was not returning results on updates on some tenants. This brought the entire project plan to a halt and as a result, we cannot sync profile images, forcing to stay on the current plan of over querying Graph with live queries.

piotrci:
Fixes #5572


api-reference/beta/api/user-delta.md
```diff
@@ -283,6 +283,46 @@ Content-type: application/json
   ]
 }
 ```
+
+## Limitations
```
Lauragra: @piotrci , can we add this information about the limitations to the concept overview topic for Delta query(https://docs.microsoft.com/en-us/graph/delta-query-overview)? This is a non-standard H2 section for an API topic, and is more conceptual information than we generally cover in the reference.

piotrci: @Lauragra , I thought about that, too, but we get so many questions about the user entity and how it is affected by this problem, that I thought it would be better to have it here. but yes, i'll move it to the overview, and we can always reference from here, if needed. thanks!

concepts/delta-query-overview.md
```diff
@@ -113,6 +113,45 @@ Delta query is currently supported for the following resources.
 
 > \*\* The usage pattern for Planner resources is similar to other supported resources with a few differences.  For details, see [Track changes for Planner](/graph/api/planneruser-list-delta?view=graph-rest-beta).
 
+### Limitations
+
+Some resources contain properties that are stored outside of the main data store where the resource is mastered. Currently, those properties are not supported as part of change tracking; a change to one of those properties will not result in an object showing up in the delta query response. Currently, only the properties stored in the main data store trigger changes in the delta query.
```
angelgolfer-ms: I was @-mentioned in issue 5572, so I am looking at this :-). Thanks for creating this PR, @piotrci . Just 2 questions:
(a) In line 118, can you use another phrase for "... where the resource is mastered"? "mastered" is not clear, and Loc will have problem with that phrase.
(b) Your example of the skills property of the user entity - how does that apply to the customer question in issue 5572, about user photo? The photo property is a nav property of user. If one applies the same syntax (doing a $select to get the skills property on the users collection) to the photo navigation property, that is:

`GET https://graph.microsoft.com/v1.0/users/?$select=photo`

one gets the same response as in getting a collection of users.

piotrci: I have addressed (a). 

Re (b) yes, that example is slightly different. I chose a more generic/common example. I will add a clarification that navigation properties are not supported either.

piotrci: @angelgolfer-ms , please have a look at the changes I made. i now created two sub sections for each limitation

piotrci:
@Lauragra , can we please merge this one to resolve the Git hub issues/questions? Let me know if any further changes are needed. thanks!

PR #5772, opened 2019-09-23
Toolkit v1

elisenyang:
- Move get started content to a separate page
- Add overview content
- Add Person-Card documentation


concepts/toolkit/get-started.md
```diff
@@ -0,0 +1,190 @@
+---
+title: "Get Started with the Microsoft Graph Toolkit"
+description: "Use the Microsoft Graph Toolkit by referencing the loader directly, or by installing the npm package."
+localization_priority: Normal
+author: nmetulev
+---
+
+# Get started
```
Lauragra: @elisenyang , it looks like this is essentially the content from the overview topic, moved into a separate topic. Can we keep the information in the overview rather than here? Otherwise, the the overview topic is a bit sparse, and it's unclear where we'll surface this in the TOC.

elisenyang: @Lauragra sure, I moved it back and combined it with the overview page

PR #5773, opened 2019-09-23
Intune app only auth perms doc update

davidmu1:


davidmu1:
#sign-off

Lauragra:
@davidmu1 , does this change merit a changelog update?

davidmu1:
@Lauragra I do not think it does, just a clarification in permissions.

Lauragra:
@davidmu1 , how is it a clarification? It looks like you're adding application permissions for Intune. That's generally something that we would call out to customers in the changelog.

davidmu1:
@Lauragra I don't agree with what you are saying, but if you provide an example of the type of entry you think should be added, I will go ahead and add it.

Lauragra:
We have entries in the changelog like this: 
Addition | Beta | Added delegated and application permissions for reading organizational contacts: OrgContact.Read.All

PR #5780, opened 2019-09-24
Update Request header specification

piricca:
1. Request header is not optional for group setting update. This operation requires authentication.
2. The Authorization header require the token only. Adding token type (Bearer) doesn't actually works (reference: https://icm.ad.msft.net/imp/v3/incidents/details/132155485/home).


Lauragra:
@dkershaw10 , please review and sign off.

PR #5788, opened 2019-09-24
Add new reviewed properties to EduAssignments docs

dipakboyed:
Add new, already reviewed and approved properties to the Beta educationAssignment resource docs.


Lauragra:
@dipakboyed , this PR looks like a dupe of #5789 except for the new educationmakecoderesource topic. Unless you need to publish that topic separately from the new property, please combine into one PR; otherwise, update the PR to include only changes that aren't in #5789.

PR #5804, opened 2019-09-26
hosted content

clearab:


Lauragra:
@clearab , does this change warrant a changelog entry?

clearab:
> @clearab , does this change warrant a changelog entry?

It is actually already in the changelog, but the ToC and outdated articles were never cleaned up.

https://docs.microsoft.com/en-us/graph/changelog#teamwork-microsoft-teams-4

PR #5810, opened 2019-09-26
fixed typo

zeroeagle1:


Lauragra:
Good catch, thanks. Made a fix and will merge.

PR #5813, opened 2019-09-26
Adding two new MS Learn modules

jthake-msft:


jthake-msft:
FYI @jasonjoh  just adding the new MS Learn modules here.

PR #5836, opened 2019-09-30
Update conversation-post-threads.md

paddymccallan:
https://github.com/microsoftgraph/microsoft-graph-docs/pull/1497#issue-131304053


clearab:
These are two different endpoints, with different purposes.

paddymccallan:
@clearab In the graph-explorer its uses the end-point I requested for change. You can see from the change [here](https://github.com/microsoftgraph/microsoft-graph-docs/pull/1497/commits/dd3ed603187ef1c3da351c4d60424c548cfe9658). The current documentation for this does not result in posting a thread to chat - Although it does return a 201. Thanks.
image - Graph-explorers end-point does create the thread.

PR #5857, opened 2019-10-01
author change for intune v1.0 articles

davidmu1:


davidmu1:
This is only an author metadata change.

#sign-off

PR #5863, opened 2019-10-02
2 new props for group exposure in Outlook.

angelgolfer-ms:


api-reference/beta/resources/group.md
```diff
@@ -111,6 +111,8 @@ This resource supports:
...
 |groupTypes|String collection| Specifies the group type and its membership.  <br><br>If the collection contains `Unified` then the group is an Office 365 group; otherwise it's a security group.  <br><br>If the collection includes `DynamicMembership`, the group has dynamic membership; otherwise, membership is static.  <br><br>Returned by default. Supports $filter.|
 |hasMembersWithLicenseErrors|Boolean| Indicates whether there are members in this group that have license errors from its group-based license assignment. <br><br>This property is never returned on a GET operation. You can use it as a $filter argument to get groups that have members with license errors (that is, filter for this property being **true**).|
+|hideFromAddressLists |Boolean |True if the group is not displayed in certain parts of the Outlook user interface: in the **Address Book**, in address lists for selecting message recipients, and in the **Browse Groups** dialog for searching groups; false otherwise. Default value is **false**. <br><br>Returned only on $select.|
```
angelgolfer-ms: Please note the descriptions for lines 114 and 115 avoid a phrase like "Indicates whether the group is {verb} in {certain parts of the UI}, for 2 reasons:
- I was suggested the verb "hidden" in this construct. However, "hidden" would repeat an significant term in the property name, so we should avoid using similar terms in the description to be effective.
- The managing editor suggested to use the verb "displayed", in which case we would be adding a level of negation if we simply say "indicates if the group is not displayed in {certain UI}". In natural language, it's clearer to say "true if the group is not displayed in {certain UI}…" BTW, technically speaking, "true if... otherwise false" is actually the appropriate construct to document Boolean properties in the Microsoft Developer Division style guide over the years.

api-reference/beta/resources/group.md
```diff
@@ -111,6 +111,8 @@ This resource supports:
...
+|hideFromOutlookClients |Boolean |True if the group is not displayed in Outlook clients, such as Outlook for Windows and Outlook on the web, false otherwise. Default value is **false**. <br><br>Returned only on $select.|
```
angelgolfer-ms: Attempting to keep an exhaustive list has the downside if your team increases support on a new Outlook client, this property description would more than likely get out-of-date with nobody noticing it for a while.
I'll let your team decide whether or not to enumerate.

PR #5868, opened 2019-10-03
Update Targeted Users

a-baabde:
PG informed the endpoint can only be used with accounts with mailbox. 
So it doesn't work with contact users.

### DIFF
--- a/api-reference/v1.0/resources/user-settings.md
+++ b/api-reference/v1.0/resources/user-settings.md
@@ -12,7 +12,8 @@ doc_type: resourcePageType
 The current user settings. 
 To learn how to get or update user settings, see [Get settings](../api/user-get-settings.md) and [Update settings](../api/user-update-settings.md).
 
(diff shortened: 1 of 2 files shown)


api-reference/v1.0/resources/user-settings.md
```diff
@@ -12,6 +12,9 @@ doc_type: resourcePageType
 The current user settings. 
 To learn how to get or update user settings, see [Get settings](../api/user-get-settings.md) and [Update settings](../api/user-update-settings.md).
 
+> [!NOTE]
+> This endpoint works only with mail users. You can't use this endpoint with contacts.
```
dkershaw10: I wouldn't even say mail users - just users.

dkershaw10:
@jthake-msft Do you know who owns this?  This is not an AAD feature.

dkershaw10:
@jthake-msft Do you know who owns this?  This is not an AAD feature.

PR #5879, opened 2019-10-04
Added missing relatedContacts property

mlafleur:
The `educationUser` documentation previously included the `relatedContacts` property that was inadvertently removed. This PR adds it back in (as well as some minor formatting/cleanup)


api-reference/beta/api/educationuser-delta.md
```diff
@@ -68,7 +68,8 @@ POST https://graph.microsoft.com/v1.0/education/me/delta
 
 The following is an example of the response.
 
->**Note:** The response object shown here might be shortened for readability. All the properties will be returned from an actual call.
```
Lauragra: FYI, for this boilerplate statement, we keep the old note formatting rather than using the Docs syntax. That is to keep the emphasis on other alerts on the page.

PR #5913, opened 2019-10-09
Update the protected API list

nkramer:


api-reference/beta/api/conversationmember-get.md
```diff
@@ -23,6 +23,9 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account)|Not supported|
 |Application| For **user** or **chat** resource:<br/>Chat.Read.All, Chat.ReadWrite.All<br/><br/>For **channel** resource:<br/>Group.Read.All, Group.ReadWrite.All |
```
RamjotSingh: While you are here can you also update we need User.Read.All in app and User.Read in user context along with Chat.X. That way we can make code side change later.

api-reference/beta/api/conversationmember-list.md
```diff
@@ -23,6 +23,9 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account)|Not supported|
```
RamjotSingh: Same here.

PR #5963, opened 2019-10-15
Headers are both required

smbolster:
Optional request headers is conflicting with the fact that both headers are actually required.


Lauragra:
Thanks. I updated the header for consistency (we can list required & optional headers).

PR #5972, opened 2019-10-15
Change aadConversationMember to aadUserConversationMember

DHerls:
When adding a user to a private channel or chat the @odata.type parameter should be microsoft.graph.aadUserConversationMember, not microsoft.graph.aadConversationMember


clearab:
Fixed in a different PR.

PR #6017, opened 2019-10-18
Add description and note for Group.Selected permission

psignoret:


concepts/permissions-reference.md
```diff
@@ -565,8 +565,9 @@ For more complex scenarios involving multiple permissions, see [Permission scena
...
-| _Group.Read.All_ | Read all groups | Allows the app to read memberships for all groups without a signed-in user. Also allows the app to read calendar, conversations, files, and other group content for all groups. > **NOTE:** that not all group API supports access using app-only permissions. See [known issues](known-issues.md) for examples. | Yes |
-| _Group.ReadWrite.All_ | Read and write all groups | Allows the app to create groups, read and update group memberships, and delete groups. Also allows the app to read and write calendar, conversations, files, and other group content for all groups. All of these operations can be performed by the app without a signed-in user. > **NOTE:** that not all group API supports access using app-only permissions. See [known issues](known-issues.md) for examples.| Yes |
+| _Group.Read.All_ | Read all groups | Allows the app to read memberships for all groups without a signed-in user. Also allows the app to read calendar, conversations, files, and other group content for all groups. <br>> **NOTE:** that not all group API supports access using app-only permissions. See [known issues](known-issues.md) for examples. | Yes |
```
dkershaw10: The \> NOTE thing doesn't work inside a table.  Instead do a line break (or double line break) using \<br\> or \<br\>\<br\>

Same is true in the next entry.  Your new Entry does it correctly :)

concepts/permissions-reference.md
```diff
@@ -565,8 +565,9 @@ For more complex scenarios involving multiple permissions, see [Permission scena
...
+| _Group.ReadWrite.All_ | Read and write all groups | Allows the app to create groups, read and update group memberships, and delete groups. Also allows the app to read and write calendar, conversations, files, and other group content for all groups. All of these operations can be performed by the app without a signed-in user. <br><br> **NOTE:** that not all group API supports access using app-only permissions. See [known issues](known-issues.md) for examples.| Yes |
+| _Group.Selected_ |    Access selected groups | Allows the app to access only the selected groups it has been granted access to, without a signed-in user. The app will have access to basic properties of the group (name, email address and description). The level of access to the selected groups can range from read access to basic group member properties, to read and write access to all content within the group, such as files, events and notes. <br><br> **NOTE: This permission is part of an upcoming feature. It should not be used. If granted, it will not grant the application any permissions.** | Yes | No |
```
Lauragra: Why are we adding the permission before the feature is available? Is this associated with a PR for the new feature?

psignoret:
The feature which relies on this permission requires the permission to be defined for testing and validation. Defining the permission necessarily makes it public-facing, and customers who noticed this have been asking for more details. We're documenting here for those folks, to save them the trouble of a support call. Full details and documentation for the new capability will come later.

concepts/permissions-reference.md
psignoret: The feature which relies on this permission requires the permission to be defined for testing and validation. Defining the permission necessarily makes it public-facing, and customers who noticed this have been asking for more details. We're documenting here for those folks, to save them the trouble of a support call. Full details and documentation for the new capability will come later.

Lauragra: Adding @Jackson-Woods as we discussed this offline. If this permission is part of a feature that is shipping for Ignite, it would be better if we can document it as part of the documentation for that feature - i.e., include this update with docs PR for resource-specific consent. When will we have that PR and what is the timeline for shipping the feature?

psignoret: We need this documentation updated now, since the permission is already released and visible and customers have already started opening support cases asking about it. When the feature this refers to is released, this doc will be updated and the note will be updated.

PR #6041, opened 2019-10-21
Replace @odata:nextLink with correct @odata.nextLink

mcmevans:
The property Key is @odata.nextLink, but some of the references were incorrect and were @odata:nextLink


Lauragra:
Thank you!

PR #6049, opened 2019-10-22
Checkmemberobjectsv1 1

davidmu1:


api-reference/v1.0/api/device-checkmemberobjects.md
```diff
@@ -0,0 +1,112 @@
+---
+title: "device: checkMemberObjects"
+description: "Check for membership in a list of groups, directory roles, or administrative units for the specified device object."
```
dkershaw10: Any references to AdministrativeUnits (including the permissions) **must** be removed.  This object is not yet in v1.0

PR #6056, opened 2019-10-23
removed Directory.Read.All from the list of perms for List domainNameReferences

adimitui:


adimitui:
#sign-off

adimitui:
[\@davidmu1](https://github.com/davidmu1) **Signed-off.**

eketo-msft:
@adimitui, https://docs.microsoft.com/en-us/graph/permissions-reference doesn't have Domain listed at all?  Can we fix that at the same time?

concepts/permissions-reference.md
```diff
@@ -459,6 +459,15 @@ For more complex scenarios involving multiple permissions, see [Permission scena
 
 ---
 
+## Domain permissions
```
Lauragra: If this is a new permission, we should have an entry in the changelog calling this out as well.

eketo-msft: It is not new, the resource doc has mentioned it for a long time but this specific permissions reference doc was missed.

Lauragra: Okay, great. Approving.

PR #6063, opened 2019-10-23
group isAssignableRole property

davidmu1:


api-reference/beta/api/group-post-groups.md
```diff
@@ -262,11 +279,97 @@ Content-type: application/json
...
     "resourceProvisioningOptions": [],
     "securityEnabled": false,
+    "securityIdentifier": "S-1-12-1-1905728287-1207447622-870010782-555555555",
+    "theme": null,
     "visibility": "Public",
     "onPremisesProvisioningErrors": []
 }
 ```
 
+### Example 3: Create a group that can be assigned to an Azure AD role
```
abhijeetsinha: It might be helpful to add a bit of detail about these role-enabled groups. Something like this - 

A group with isAssignableToRole property set to 'true' cannot be of dynamic membership type and cannot have an owner. For more information, see [Using a group to manage Azure AD role assignments](https://go.microsoft.com/fwlink/?linkid=2103037)

api-reference/beta/api/group-post-groups.md
```diff
@@ -138,21 +140,27 @@ HTTP/1.1 201 Created
...
 {
+   "@odata.context": "https://graph.microsoft.com/beta/$metadata#groups/$entity",
 	 "id": "45b7d2e7-b882-4a80-ba97-10b7a63b8fa4",
 	 "deletedDateTime": null,
 	 "classification": null,
 	 "createdDateTime": "2018-12-22T02:21:05Z",
 	 "description": "Self help community for golf",
 	 "displayName": "Golf Assist",
 	 "expirationDateTime": null,
+   "externalGroupIds": [],
```
dkershaw10: Not sure about documenting the external* properties or the membershipTypes property further down.  Can you remove them please? Not sure about the status of those properties?  We do say that the example may not contain all values, so we should just remove these for now.  Same for the other examples please.

dkershaw10:
We need to get this wrapped up.  This property is now available in the service.

abhijeetsinha:
@davidmu1 - We need to update this doc as well - https://docs.microsoft.com/en-us/graph/api/group-getmemberobjects?view=graph-rest-1.0&tabs=http

Currently, it says that "Note: Groups cannot be members of directory roles, so no directory roles will be returned."

apraamsm:
i was playing around with the isAssignableToRole property for new groups using the beta API and constantly get this error when creating a group which has set isAssignableToRole to true

{
  "error": {
    "code": "Request_BadRequest",
    "message": "Tenant is not enabled for creating groups assignable to role.",
    "innerError": {
      "request-id": "f0daf797-9b6c-4cf0-aa45-cacdfc1950b1",
      "date": "2020-01-17T20:10:54"
    }
  }
}

Any thoughts?

Lauragra:
@davidmu1 , the new property is now available on the group resource. Can you please incorporate feedback and get this PR ready to publish? Thanks.

davidmu1:
Property addition is still in private preview

PR #6066, opened 2019-10-23
Updating Invitation resource type

aybankaMSFT:
We are adding a new resetRedemption Property. The PR is to reflect this change.


Lauragra:
@aybankaMSFT , please include a changelog entry for the new property.

PR #6085, opened 2019-10-25
Add AdditionalData dictionary

Pyrobolser:
Hi,

I understand this file is auto-generated and shouldn't be modified but I cannot find in the contributor guide how to properly update those snippets.
We need the **AdditionalData** dictionary for the code to compile and work properly.

### DIFF
--- a/api-reference/v1.0/includes/snippets/csharp/create-folder-csharp-snippets.md
+++ b/api-reference/v1.0/includes/snippets/csharp/create-folder-csharp-snippets.md
@@ -12,7 +12,10 @@ var driveItem = new DriveItem
 	Folder = new Folder
 	{
 	},
-	@microsoft.graph.conflictBehavior = "rename"
+	AdditionalData = new Dictionary<string, object>
+	{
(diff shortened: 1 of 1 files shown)


Lauragra:
@darrelmiller , please take a look. Thanks.

darrelmiller:
@Pyrobolser Thank you identifying the issue.  The snippet generators are here https://github.com/microsoftgraph/microsoft-graph-explorer-api/tree/dev/CodeSnippetsReflection/LanguageGenerators  @andrueastman   Could you take a look at this issue?  I think when we find a property that starts with an @ symbol, we can't update the object directly and we need to add it to the additional data dictionary.

andrueastman:
@Pyrobolser Thanks again for identifying this issue.
I have created this [issue](https://github.com/microsoftgraph/microsoft-graph-explorer-api/issues/145) on the generator and will work to have it resolved asap.

I will close this PR and have the issue resolved there so that the docs may be updated.
cc @darrelmiller @Lauragra

PR #6088, opened 2019-10-25
draft text and example

rwike77:


concepts/permissions-reference.md
```diff
@@ -40,6 +40,72 @@ If the signed-in user is a guest user, depending on the permissions an app has b
 
 With the appropriate permissions, the app can read the profiles of users or groups that it obtains by following links in navigation properties; for example, `/users/{id}/directReports` or `/groups/{id}/members`.
 
+## Return shell objects for inaccessible objects
+
+[Groups](/graph/api/resources/group) are members of [Roles](/graph/api/resources/directoryrole). Queries for Roles.members return Groups and queries for groups.memberOf return Role objects.  If an application has access to read some but not all all source/target object types to a given link, the application gets a 200 response and a collection of objects.  Complete objects are returned for the object types that the application has permissions to read.  A shell object is returned for any object that the application does not have access to read.  The shell object contains the object type and ID in the case of access denial (similar to what is returned if you don't $expand the object).
```
abhijeetsinha: @rwike77  - Today Azure AD groups cannot be assigned to roles. I am working on this functionality but it is coming in CY2020. I don't think it would be right to mention it.

abhijeetsinha: cc: @VinceSmith

rwike77: @abhijeetsinha , thanks.  I removed the text about groups being assigned to roles.

concepts/permissions-reference.md
```diff
@@ -40,6 +40,71 @@ If the signed-in user is a guest user, depending on the permissions an app has b
 
 With the appropriate permissions, the app can read the profiles of users or groups that it obtains by following links in navigation properties; for example, `/users/{id}/directReports` or `/groups/{id}/members`.
 
+## Limited information returned for inaccessible member objects
+Container objects such as groups and roles support members of various types, for example users and service principals. When an application queries the membership of a container object and does not have permission to read a certain type, members of that type are returned but with limited information.  The application receives a 200 response and a collection of objects.  Complete information is returned for the object types that the application has permissions to read.  For the object types which the application does not have permission to read, only the the object type and ID are returned (similar to what is returned if you don't $expand the object).
```
psignoret: I suggest removing "(similar to what is returned if you don't $expand the object)". I don't think that adds any clarity here.

concepts/permissions-reference.md
```diff
@@ -40,6 +40,71 @@ If the signed-in user is a guest user, depending on the permissions an app has b
...
+
+This is applied to all relationships that are of [directoryObject](/graph/api/resources/directoryobject) type (not just member links). Examples include `/groups/{id}/members`, `roles/{id}/members` or `me/ownedObjects`.
+
+For example, let's say an application has [User.Read.All](#user-permissions) and [Group.Read.All](#group-permissions) permissions for Microsoft Graph.  An Active Directory group has been created and that group contains a user and a service principal.  The application calls list group members.  The application has access to the user object in the group, but not the service principal object.  In the response, the user object and all of it's properties are returned. For the service principal object, however, only limited information is returned.  The data type and object ID are returned for the service principal, but all other properties have a value of *null*. Apps without permission will not be able to use the ID to get the actual object.
```
psignoret: Just "A group has been created...", instead of "An Active Directory group has been created...".

psignoret: "list group members" should link to relevant doc: https://docs.microsoft.com/graph/api/group-list-members

concepts/permissions-reference.md
```diff
@@ -40,6 +40,71 @@ If the signed-in user is a guest user, depending on the permissions an app has b
...
+
+```http
+GET https://graph.microsoft.com/1.0/groups/{id}/members HTTP/1.1
```
psignoret: Should be ".../v1.0/...", not ".../1.0/...".

concepts/permissions-reference.md
```diff
@@ -40,6 +40,71 @@ If the signed-in user is a guest user, depending on the permissions an app has b
...
+```
+
+The following is the JSON response:
+
+```json
+{
+    "@odata.context":"https://graph.microsoft.com/1.0/$metadata#directoryObjects",
+    "value":[
```
psignoret: In v1.0, not all properties are returned for users (and maybe for groups?) unless they are explicitly `$select`ed. The example should either explicitly request some attributes (and only those should be part of the example response), or the user entity in the example response should only include the attributes which are returned by default.

concepts/permissions-reference.md
dkershaw10: For the opening sentence, we should at least have something that is technically correct:
Container objects such as groups and roles support members of various types, for example users and service principals.
->
Container objects such as groups support members of various types, for example users and service principals.

dkershaw10: @Lauragra Should we use the term **resource** or **object**, or are we happy with either?

dkershaw10: Not only that, I don't believe that is correct either.  GET groups/{id}/members/$ref  will kind of give that response, but that's actually busted in Microsoft Graph, so I wouldn't go there either.
+1 to remove whats in the parentheses.

concepts/permissions-reference.md
```diff
@@ -40,6 +40,71 @@ If the signed-in user is a guest user, depending on the permissions an app has b
 
 With the appropriate permissions, the app can read the profiles of users or groups that it obtains by following links in navigation properties; for example, `/users/{id}/directReports` or `/groups/{id}/members`.
 
+## Limited information returned for inaccessible member objects
+Container objects such as groups and roles support members of various types, for example users and service principals. When an application queries the membership of a container object and does not have permission to read a certain type, members of that type are returned but with limited information.  The application receives a 200 response and a collection of objects.  Complete information is returned for the object types that the application has permissions to read.  For the object types which the application does not have permission to read, only the the object type and ID are returned (similar to what is returned if you don't $expand the object).
+
+This is applied to all relationships that are of [directoryObject](/graph/api/resources/directoryobject) type (not just member links). Examples include `/groups/{id}/members`, `roles/{id}/members` or `me/ownedObjects`.
```
dkershaw10: roles -> directoryRoles  (there's no entity set called `roles`)

Not sure directoryRoles is the best example - I think we've only JUST added the ability for directoryRoles to have groups as members, but it's not yet documented.
How about `/users/{id}/memberOf`?

concepts/permissions-reference.md
```diff
@@ -40,6 +40,71 @@ If the signed-in user is a guest user, depending on the permissions an app has b
 
 With the appropriate permissions, the app can read the profiles of users or groups that it obtains by following links in navigation properties; for example, `/users/{id}/directReports` or `/groups/{id}/members`.
 
+## Limited information returned for inaccessible member objects
```
dkershaw10: Are we not planning to update each topic that can return a heterogeneous collection of object types (inheriting from directoryObject) to hint that partial responses may be given, together to a reference to this section to get further information?

concepts/permissions-reference.md
rwike77: Done, thanks.

concepts/permissions-reference.md
rwike77: Made both updates, thanks.

concepts/permissions-reference.md
rwike77: My bad, good catch.

concepts/permissions-reference.md
rwike77: Sounds reasonable, I made the change.

concepts/permissions-reference.md
rwike77: Yes, we're planning to update those topics as well.  I need to work with David to get those articles updated (I'm not familiar with the process for updating api ref).  I'll get some text out for review this afternoon.

concepts/permissions-reference.md
rwike77: OK, made that change.

concepts/permissions-reference.md
rwike77: OK, I'll get the sample updated.  Thanks.

dkershaw10:
@rwike77 I wonder whether we also need to update the changelog.  While strictly the API schema has not changed at all, behavior and functionality IS changing here...

concepts/permissions-reference.md
rwike77: Updated the sample.

Lauragra:
@rwike77 , what is the status of this PR? Can you take care of the merge conflicts, or close this PR and create a new one?

PR #6154, opened 2019-10-31
Update place.md

vrod9429:


Lauragra:
@vrod9429 , what is the status of this PR? Should it be published? Looks like we have a validation error and we also need to move the changelog entry into the April section. Can you please take care of that?

angelgolfer-ms:
@Lauragra , the updates this PR attempted to make are already in the live docs.
@vrod9429 probably forgot to close this PR.

PR #6244, opened 2019-11-07
allowCreatePrivateChannels added to the request

markgort86:
allowCreatePrivateChannels added to the request, this because the response doesn't return this property. But when using this property in the actual call this setting will definitely changed when creating a new team.


laujan:
I reviewed and checked with Nick. This is okay to merge.

PR #6246, opened 2019-11-07
Added property allowCreatePrivateChannels

markgort86:
Missing property allowCreatePrivateChannels added to the documentation


phecdalouie:
@Lauragra This looks like a good doc update, looks like we missed adding this property to this specific page when we launched private channels.

PR #6286, opened 2019-11-12
Update group-post-groups.md

AlexanderHolmeset:
Possible to set visibility when creating a group


dkershaw10:
See comment in #6287

PR #6287, opened 2019-11-12
Update group-post-groups.md

AlexanderHolmeset:
possible to set visibility when creating group


dkershaw10:
Not sure that this change is something that should be approved.  For the properties section of a create topic, I believe we only show the **required** properties. @Lauragra can you confirm?  **visibility** is not a required property on group AFAIK.

PR #6309, opened 2019-11-13
Adding app-only permissions to read CA and named location policies

elcarlosadrian:
Adding app-only permissions to read CA and named location policies


Lauragra:
@elcarlosadrian , does this permission update warrant an entry in the changelog?

elcarlosadrian:
> @elcarlosadrian , does this permission update warrant an entry in the changelog?

Added the changelog entry

PR #6319, opened 2019-11-14
Updated businessFlowTemplate id character casing

seantleonard:
Letters changed to lower case to reflect formatting accepted by Graph API. Upper case letters lead to result values not showing. 

Review: @markwahl-msft

### DIFF
--- a/api-reference/beta/api/accessreview-list.md
+++ b/api-reference/beta/api/accessreview-list.md
@@ -45,7 +45,7 @@ If successful, this method returns a `200, OK` response code and an array of [ac
 
 ## Examples
 ##### Request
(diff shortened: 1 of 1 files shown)


markwahl-msft:
Unfortunately, this is not the problem.  The value shown in the documentation came from a call to retrieve the businessFlowTemplates - the customer who uses the recommended pattern to retrieve business flow templates and enumerate reviews of that template with that template's ID will encounter that issue..  The problem is that one AR call provides the values in upper case but another AR call only allows the values in lower case, whereas AR graph is case insensitive.

seantleonard:
Understood, thanks for the input.

PR #6322, opened 2019-11-15
Add CallOptions and KeepAlive docs.

zihzhan:
Update doc for lobbyByPass in outgoingCallOptions.
Update Keep Alive doc.


api-reference/beta/resources/call.md
```diff
@@ -45,6 +45,8 @@ Although the same identity cannot be invited multiple times, it is possible for
 | [Mute](../api/call-mute.md)                                       | [commsOperation](commsoperation.md)               | Mute self in the call.                       |
 | [Unmute](../api/call-unmute.md)                                   | [commsOperation](commsoperation.md)               | Unmute self in the call.                     |
 | [ChangeScreenSharingRole](../api/call-changescreensharingrole.md) |                                                   | Start and stop sharing screen in the call.   |
+| **Others**
```
ananmishr: Instead of "Others" I feel like keepAlive should be moved up to where Get call and Delete call are. What are your thoughts on this @paywuAtMicrosoft ?

PR #6336, opened 2019-11-15
update access reviews businessflowtemplates

markwahl-msft:
In graph beta, the business flow template ID and descriptions returned by /businessFlowTemplates have been updated. (These are used in access reviews). Changing the examples to match the current values.


markwahl-msft:
#sign-off

PR #6341, opened 2019-11-18
Added error codes/description to document

mariahhart:


api-reference/beta/api/notifications-post.md
```diff
@@ -5,6 +5,7 @@ localization_priority: Normal
 ms.prod: "notifications"
 doc_type: apiPageType
 author: "merzink"
+updates made by: "mariahhart"
```
Lauragra: Please don't add new attributes to the YAML headers; we include only the attributes defined in this topic: https://msgo.azurewebsites.net/add/document/guidelines/metadata.html.

api-reference/beta/api/notifications-post.md
```diff
@@ -70,7 +71,7 @@ Content-type: application/json
 	"ios",
 	"android"
         ]
-    },
+ Not supported.    },
```
Lauragra: This doesn't look right - what is your intention here?

api-reference/beta/api/notifications-post.md
```diff
@@ -46,6 +47,23 @@ In the request body, supply a JSON representation of a [notification](../resourc
 ## Response
 If successful, this method returns a `201 Created` response code that indicates that the notification was successfully created and stored. The notification will be subsequently fanned-out to all specified endpoints with a valid subscription. 
 
+The following table lists the possible error and response codes that can be returned.
```
Lauragra: FYI, we don't add custom H2 sections to API topics. I moved the error codes table into the Response section. I made a few minor edits to the descriptions here.

mariahhart: Thanks for the note, sorry about that.

api-reference/beta/api/notifications-post.md
mariahhart: Got it.

api-reference/beta/api/notifications-post.md
mariahhart: That must've been added by accident. I'll update it.

PR #6346, opened 2019-11-18
fix for git issue 5883

davidmu1:


api-reference/beta/resources/invitation.md
```diff
@@ -34,7 +34,7 @@ Creating an invitation will return a redemption URL in the response (*inviteRede
...
-|invitedUserEmailAddress|String|The email address of the user being invited. Required.|
+|invitedUserEmailAddress|String|The email address of the user being invited. Required. The following special characters are not permitted in the email address:<br><ul><li>Tilde (~)</li><li>Exclamation point (`!`)</li><li>At sign (`@`)</li><li>Number sign (`#`)</li><li>Dollar sign (`$`)</li><li>Percent (`%`)</li><li>Circumflex (`^`)</li><li>Ampersand (`&`)</li><li>Asterisk (`*`)</li><li>Parentheses (`( )`)</li><li>Hyphen (`-`)</li><li>Plus sign (`+`)</li><li>Equal sign (`=`)</li><li>Brackets (`[ ]`)</li><li>Braces (`{ }`)</li><li>Backslash (`\`)</li><li>Slash mark (`/`)</li><li>Pipe (`|`)</li><li>Semicolon (`;`)</li><li>Colon (`:`)</li><li>Quotation marks (`"`)</li><li>Angle brackets (`< >`)</li><li>Question mark (`?`)</li><li>Comma (`,`)</li></ul><br>However, the following exceptions apply:<br><ul><li>A period (`.`) or a hyphen (`-`) is permitted anywhere in the user name, except at the beginning or end of the name.</li><li>An underscore (`_`) is permitted anywhere in the user name. This includes at the beginning or end of the name.</li></ul>|
```
dkershaw10: This is busted, because you need to escape the pipe \| character in your list.  Same below.
that said, why are we putting this rubbish into this topic?  Can we reference the email address RFC and be done?

davidmu1: @dkershaw10 - What email address RFC are you referring to? I believe for this type 'rubbish', it is usually better to just list it in the topic rather than forcing a user to jump somewhere else and try to relate what they are seeing to what they need to do.

davidmu1: @dkershaw10 - I'm not sure it is busted. If you view the file with the viewer on the files page, it does look broken, but if you click the View link to the staging site on the conversation page, it looks correct.

davidmu1:
#sign-off

PR #6357, opened 2019-11-19
fix for git issue 5922

davidmu1:


davidmu1:
Syncing up the beta and v1.0 articles.
 
#sign-off

PR #6358, opened 2019-11-19
More specific on permission scope

jthake-msft:


concepts/permissions-reference.md
```diff
@@ -806,7 +806,7 @@ For more complex scenarios involving multiple permissions, see [Permission scena
 |   Permission    |  Display String   |  Description | Admin Consent Required | Microsoft Account supported |
 |:----------------|:------------------|:-------------|:-----------------------|:--------------|
 | _Mail.Read_ |    Read user mail | Allows the app to read email in user mailboxes. | No | Yes
-| _Mail.ReadBasic_ |    Read user basic mail | Allows the app to read the signed-in user's mailbox except body, previewBody, attachments and any extended properties. Does not include permissions to search messages. | No | Yes
+| _Mail.ReadBasic_ |    Read user basic mail | Allows the app to read the signed-in user's mailbox except Body, BodyPreview, UniqueBody, Attachments, ExtendedProperties, and Extensions. Does not include permissions to search messages. | No | Yes
```
angelgolfer-ms: @Lauragra , @jthake-msft , hm, did we review some description like this before... the description should be more specific regarding the types of items (i.e., messages, not every item type in the mailbox) that this permission applies to. Also, to be clear, the description should include the exact property names, with the spelling and lower camel case that match the schema:
"Allows the app to read message items in the signed-in user's mailbox except the body, bodyPreview, and uniqueBody properties, and the attachments, extensions, multiValueExtendedProperties, and singleValueExtendedProperties relationships. Does not include permissions to search messages."

If we find this too long and has to match somewhat the shorter description in the Azure portal, maybe we can make it less precise and say:
"Allows the app to read email in the signed-in user's mailbox except body, bodyPreview, uniqueBody, attachments, extensions, and any extended properties. Does not include permissions to search messages."

I think that would still be clear enough for existing Outlook developers.

Lauragra: FYI, I made this update in PR #6367 .

PR #6359, opened 2019-11-19
Places API updates.md

vrod9429:


concepts/permissions-reference.md
```diff
@@ -1122,7 +1122,8 @@ For more complex scenarios involving multiple permissions, see [Permission scena
 
 |   Permission    |  Display String   |  Description | Admin Consent Required | Microsoft Account supported |
 |:----------------|:------------------|:-------------|:-----------------------|:--------------|
-| _Place.Read.All_ |Read all company places |Allows the app to read company places (conference rooms and room lists) for calendar events and other applications. |No | No |
+| _Place.Read.All_ |Read all company places |Allows the app to read company places (conference rooms and room lists) for calendar events and other applications. |Yes | No |
```
angelgolfer-ms: @vrod9429 , I am aware Place.Read.All is not new, but now I think of it, why do we say "for calendar events"? The place entity is not integrated with the event entity or location complex type, is it?

vrod9429: you are right that it really isn't associated with the event because it can exist independently of calendar. What about like this?
"Allows the app to read company places (conference rooms and room lists) for Exchange and other applications" 
The key is that conf rooms are mailboxes in exchange.

angelgolfer-ms: The new description looks fine to me, thanks!

vrod9429: yes i made changes

api-reference/beta/api/place-update.md
```diff
@@ -40,7 +40,7 @@ PATCH /places/{id | emailAddress}
 
 ## Request body
 
-In the request body, supply the values for relevant fields that should be updated. Existing properties that are not included in the request body will maintain their previous values or be recalculated based on changes to other property values. For best performance, don't include existing values that haven't changed.
+In the request body, supply the values for relevant fields that should be updated. Only one instance of a place resource (room or roomList) can be updated at a time. In the request body, use @odata.type to specify the type of place, and include the properties of that type to update. Existing properties that are not included in the request body will maintain their previous values or be recalculated based on changes to other property values. For best performance, don't include existing values that haven't changed.
```
angelgolfer-ms: Just a formatting note: please enclose @odata.type with the tick marks, like this: `@odata.type`. Hopefully the last comment for this PR :-)

FYI, in case you're wondering what the guideline is in general - for a programmability element in the docs, we need to format them differently from normal text (i.e., the rest of the docs). If it's part of the Graph API, then bold it. If it's not part of the Graph API, like @odata.type, enclose it with tick marks. The exception (only an exception because the stub topic generation tool has made that an exception but it's still within reason) is in the generated tables like "Methods" or "Properties" table, because so many programmability elements are mentioned in all the columns, if we start bolding them, that might make the content harder to scan.

This Microsoft Style Guide describes the bolding part: https://docs.microsoft.com/en-us/style-guide/developer-content/formatting-developer-text-elements

PR #6371, opened 2019-11-20
checkobjects v1

davidmu1:


davidmu1:
@Lauragra: The warnings on the PR are very odd. It seems like checks are being made by APIDoctor on filesthat were previously checked in to the repository that are not part of this PR. None of the warnings apply to the changed files.

#sign-off

PR #6380, opened 2019-11-21
Limited info for inaccessible objects

rwike77:


api-reference/beta/api/administrativeunit-list-members.md
```diff
@@ -25,6 +24,8 @@ One of the following permissions is required to call this API. To learn more, in
 
 > Note: To list the members of a hidden membership in an administrative unit, the Member.Read.Hidden permission is required.
 
+When an application queries the membership of a container object and does not have permission to read a certain type, [limited information is returned for inaccessible member objects](/graph/permissions-reference#limited-information-returned-for-inaccessible-member-objects).
```
Lauragra: Global: I think it would be more clear to revise the note as follows:
When an application queries the membership of a container object and does not have permission to read a certain type, members of that type are returned but with limited information. For details, see [Limited information returned for inaccessible member objects].

I recommend that you create an include file for this note, and add the include to each topic, so that it's easy to make updates to the text if needed in the future.

rwike77: Good idea, thanks.

concepts/permissions-reference.md
```diff
@@ -39,6 +39,50 @@ If the signed-in user is a guest user, depending on the permissions an app has b
 
 With the appropriate permissions, the app can read the profiles of users or groups that it obtains by following links in navigation properties; for example, `/users/{id}/directReports` or `/groups/{id}/members`.
 
+## Limited information returned for inaccessible member objects
+Container objects such as groups support members of various types, for example users and service principals. When an application queries the membership of a container object and does not have permission to read a certain type, members of that type are returned but with limited information.  The application receives a 200 response and a collection of objects.  Complete information is returned for the object types that the application has permissions to read.  For the object types which the application does not have permission to read, only the the object type and ID are returned.
```
dkershaw10: Similar to the blog post, I would stick with objects that are in v1.  So remove service principals and add in groups and devices, and then align it to your example below.

dkershaw10: What if the app has permissions, but the user does not?  Is the experience the same?

dkershaw10: I like your paragraph here.  However, it's not just conceptually container objects.  Any heterogeneous collection/array *could* have this behavior.  In most cases it's visible through relationships (navigation property collections) on a resource. We might consider changing here, but I can live with what you've written.

api-reference/includes/limited_info.md
```diff
@@ -0,0 +1,5 @@
+
+<!-- markdownlint-disable MD041-->
+
+When an application queries the membership of a container object and does not have permission to read a certain type, members of that type are returned but with limited information. For details, see [Limited information returned for inaccessible member objects](/graph/permissions-reference#limited-information-returned-for-inaccessible-member-objects).
```
dkershaw10: This is only applicable to Azure AD, and the sentence makes it sound like it's MIcrosoft Graph wide behavior.
Not sure if we should be more explicit and say:  When an application queries relationships that are of [directoryObject](/graph/api/resources/directoryobject) type collections...
That way it's clear that this behavior is more specific to directory/Azure AD...

api-reference/beta/api/directory-deleteditems-list.md
```diff
@@ -16,6 +16,9 @@ Retrieve a list of recently deleted items from [deleted items](../resources/dire
 Currently, deleted items functionality is only supported for the [application](../resources/application.md), [group](../resources/group.md) and [user](../resources/user.md) resources.
 
 ## Permissions
```
dkershaw10: Per email, not sure this needs updating.  A cast is required for this API call, so you get a homogeneous response, so that means you can pick the matching permission, and unlikely to face the same challenge

api-reference/beta/api/administrativeunit-list-members.md
```diff
@@ -25,6 +24,8 @@ One of the following permissions is required to call this API. To learn more, in
 
 > Note: To list the members of a hidden membership in an administrative unit, the Member.Read.Hidden permission is required.
 
+[!INCLUDE [limited info](../../includes/limited_info.md)]
```
dkershaw10: I think API doctor is complaining because your link is wrong.  I think it has one too many ..
Should be ../includes/limited_info.md I think

api-reference/beta/api/application-list-owners.md
```diff
@@ -22,6 +22,8 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account) | Not supported.    |
```
dkershaw10: OK - you've missed one **really** important point here in **every** topic.
Now that we will return a response with data, rather than the 403, we can tell developers to request least privileged permissions.  For this example, if I'm only interested in users that are owners, then all I need is User.ReadBasic.All or User.Read.All.  If I want service principals, then Application.Read.All.  If both, then I would specify both permissions.  These are all less privileged than Directory.Read.All.

api-reference/beta/api/administrativeunit-list-members.md
Lauragra: If we're fixing the includes, link, I recommend that we also update the file name for the includes to use a hyphen - "limited-info" - and change the file name in the includes link too. That's consistent with our file naming convention.

Lauragra:
@rwike77 , @dkershaw10 , I agree that this PR warrants an entry in the changelog, so please add that. Thanks.

api-reference/beta/api/administrativeunit-list-members.md
rwike77: I changed the include file name to "limited-info.md" and updated all the includes links.  The links should be correct.  The includes are rendering OK in the preview builds, but API Doctor has these errors:
Error: /api/administrativeunit-list-members.md:
    Exception processing links.
System.ArgumentException: Link Definition was null. Link text: !INCLUDE [limited info](../../includes/limited-info.md)

Not sure what's going wrong...

api-reference/beta/api/directory-deleteditems-list.md
rwike77: Removed the include from this article.

api-reference/includes/limited_info.md
rwike77: Sounds good, I made the change.

api-reference/beta/api/administrativeunit-list-members.md
Lauragra: I think for includes the brackets need to include the file name, so you can try adding the hyphen there: [limited-info].

rwike77: I don't know why API doctor is complaining about the include link.  The link is correct and formatted exactly the same as other include links.  The content is even rendering properly in the doc builds...

concepts/permissions-reference.md
rwike77: I updated to use v1 objects (similar to the blog)

api-reference/includes/limited-info.md
```diff
@@ -0,0 +1,5 @@
+
+<!-- markdownlint-disable MD041-->
+
+When an application queries relationships that are of directoryObject type collections and does not have permission to read a certain type, members of that type are returned but with limited information. For details, see [Limited information returned for inaccessible member objects](/graph/permissions-reference#limited-information-returned-for-inaccessible-member-objects).
```
dkershaw10: Suggest a change to the 1st sentence.  Won't be upset if you reject it.

When an application queries relationships that are of directoryObject type collections and does not have permission to read a certain type,
->
When an application queries a relationship that returns a directoryObject type collection, if it does not have permission to read a certain derived type (like device),

dkershaw10: Not sure (until we update all the topic's permissions sections here) if we want to add something like:
With this behavior applications can start to request the least privileged permissions they need, rather than rely on the set of Directory.* permissions.

concepts/permissions-reference.md
```diff
@@ -39,6 +39,51 @@ If the signed-in user is a guest user, depending on the permissions an app has b
 
 With the appropriate permissions, the app can read the profiles of users or groups that it obtains by following links in navigation properties; for example, `/users/{id}/directReports` or `/groups/{id}/members`.
 
+## Limited information returned for inaccessible member objects
+Container objects such as groups support members of various types, for example users and device. When an application queries the membership of a container object and does not have permission to read a certain type, members of that type are returned but with limited information.  The application receives a 200 response and a collection of objects.  Complete information is returned for the object types that the application has permissions to read.  For the object types which the application does not have permission to read, only the the object type and ID are returned.
```
dkershaw10: users and device
->
user and device
OR
users and devices

rwike77: Good catch, thanks.  Fixed.

api-reference/includes/limited-info.md
rwike77: That's fine, thanks.  Updated.

rwike77: I think it would be helpful to add a hint, though we do need to update all the reference articles with the specific permissions.  I updated the include.  @davidmu1  FYI

rwike77:
#sign-off

PR #6392, opened 2019-11-22
Add KeepAlive docs.

zihzhan:


api-reference/beta/api/call-keepalive.md
```diff
@@ -0,0 +1,76 @@
...
+description: "Make a request to this API every 15 minutes to ensure that an ongoing call stays alive."
+author: "VinodRavichandran"
+localization_priority: Normal
+ms.prod: "cloud-communications"
+doc_type: apiPageType
+---
+
+# call: keepAlive
+
+Make a request to this API every 15 minutes to ensure that an ongoing call stays alive. A call that has not received this request within this time interval will be considered to be inactive and then subsequently ended.
```
Lauragra: I suggest that we revise the first sentence so that we don't anthropomorphize the call - i.e. "... ensure that the ongoing call stays active". The name of the API is a little unfortunate; a better name would have been "keepActive", for example.

Revise the second sentence to be more concise: "A call that does not receive this request within the time interval is considered inactive and will subsequently end."

(Same comment as on v1.0 PR.)

api-reference/beta/api/call-keepalive.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Permissions
+One of the following permissions may be required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+| Permission type | Permissions (from least to most privileged) |
+| :-------------- | :------------------------------------------ |
+| Delegated (work or school account)     | Not Supported        |
+| Delegated (personal Microsoft account) | Not Supported        |
+| Application     | None                                        |
```
Lauragra: Permissions?

api-reference/beta/api/call-keepalive.md
```diff
@@ -0,0 +1,79 @@
...
+The following is an example of a request.
+
+<!-- {
+  "blockType": "request",
+  "name": "keep-alive"
+}-->
+
+```http
+POST https://graph.microsoft.com/beta/communications/calls/2e1a0b00-2db4-4022-9570-243709c565ab/keepAlive
+Authorization: Bearer <Token>
```
Lauragra: Remove Authorization header.

api-reference/beta/api/call-keepalive.md
VinodRavichandran: Regarding the API name, 'keepAlive', that's a good feedback. We will note that for future API names. Note - @ananmishr, @zhengni-msft , @paywuAtMicrosoft

'KeepAlive' is standard term for 'Connections' (https://en.wikipedia.org/wiki/Keepalive), which was also used for Calls (https://tools.ietf.org/html/rfc6223). This term is well understood by developers, which is why we decided to go with term.

I will update the description with your suggestion.

api-reference/beta/api/call-keepalive.md
VinodRavichandran: This API, similar to bunch of other APIs that we have, depends on the Create or Answer call APIs; which has permissions. Once an App creates a call, it receives a CallID and is able to do this action on that CallID.

api-reference/beta/api/call-keepalive.md
```diff
@@ -0,0 +1,78 @@
...
+description: "Make a request to this API every 15 minutes to ensure that an ongoing call stays alive."
+author: "VinodRavichandran"
+localization_priority: Normal
+ms.prod: "cloud-communications"
+doc_type: apiPageType
+---
+
+# call: keepAlive
+
+A call that does not receive this request within the time interval (15 minutes) is considered inactive and will be subsequently ended.
```
ananmishr: Please change the description to
A call that does not receive this request within 45 minutes will be considered inactive and subsequently ended. 

It's important to make sure that at least one successful request is made within 45 minutes of the previous request, or the start of the call.

We recommend that a request is sent in shorter time intervals (e.g. every 15 minutes). Please make sure that these requests are successful to prevent the call from timing out and ending.

Attempting to send a request to a call that has already ended will result in a `404 Not-Found` error. The resources related to the call should be cleaned up on the application side.

PR #6393, opened 2019-11-22
Add KeepAlive docs. V1

zihzhan:


api-reference/v1.0/api/call-keepalive.md
```diff
@@ -0,0 +1,76 @@
...
+description: "Make a request to this API every 15 minutes to ensure that an ongoing call stays alive."
+author: "VinodRavichandran"
+localization_priority: Normal
+ms.prod: "cloud-communications"
+doc_type: apiPageType
+---
+
+# call: keepAlive
+
+Make a request to this API every 15 minutes to ensure that an ongoing call stays alive. A call that has not received this request within this time interval will be considered to be inactive and then subsequently ended.
```
Lauragra: I suggest that we revise the first sentence so that we don't anthropomorphize the call - i.e. "... ensure that the ongoing call stays active". The name of the API is a little unfortunate; a better name would have been "keepActive", for example.

Revise the second sentence to be more concise: "A call that does not receive this request within the time interval is considered inactive and will subsequently end."

api-reference/v1.0/api/call-keepalive.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Permissions
+One of the following permissions may be required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+| Permission type | Permissions (from least to most privileged) |
+| :-------------- | :------------------------------------------ |
+| Delegated (work or school account)     | Not Supported        |
```
Lauragra: Are any permissions required for this API?

api-reference/v1.0/api/call-keepalive.md
```diff
@@ -0,0 +1,80 @@
...
+The following is an example of a request.
+
+<!-- {
+  "blockType": "request",
+  "name": "keep-alive"
+}-->
+
+```http
+POST https://graph.microsoft.com/v1.0/communications/calls/2e1a0b00-2db4-4022-9570-243709c565ab/keepAlive
+Authorization: Bearer <Token>
```
Lauragra: FYI, we don't generally include the Authorization header in the request examples.

api-reference/v1.0/api/call-keepalive.md
VinodRavichandran: Regarding the API name, 'keepAlive', that's a good feedback. We  will note that for future API names. Note - @ananmishr, @zhengni-msft , @paywuAtMicrosoft 

'KeepAlive' is standard term for 'Connections' (https://en.wikipedia.org/wiki/Keepalive), which was also used for Calls (https://tools.ietf.org/html/rfc6223). This term is well understood by developers, which is why we decided to go with term.

I will update the description with your suggestion.

api-reference/v1.0/api/call-keepalive.md
VinodRavichandran: This API, similar to bunch of other APIs that we have, depends on the Create or Answer call APIs; which has permissions. Once an App creates a call, it receives a CallID and is able to do this action on that CallID.

api-reference/v1.0/api/call-keepalive.md
VinodRavichandran: Will remove it.

api-reference/v1.0/api/call-keepalive.md
Lauragra: Okay, and to verify, "Not supported" conveys a different meaning than "None", right?

VinodRavichandran: @zhengni-msft  @zihzhan, can you answer this question? I think we should keep it consistent with 'None', if there is no strong reason for 'Not supported'

zihzhan: For Delegated (work or school account) and Delegated (personal Microsoft account).
These two scenarios are applied with user token, which is rejected when calling the API. Hence, they are **Not Supported**. **Not Supported** is different than **None**.

api-reference/v1.0/api/call-keepalive.md
zihzhan: Thanks for the feedback! Address those comments in new iteration.

zihzhan: Thanks for the feedback! Address those comments in new iteration.

api-reference/v1.0/api/call-keepalive.md
zihzhan: Thanks for the feedback! Address those comments in new iteration.

zihzhan:
> Update the call resource topic to list the new method and add an entry to the TOC.
The PR adding V1 call resource haven't merged yet.
@ananmishr Can you please help to add an entry to api-reference/v1.0/api/call-keepalive.md in the call resource topic list?

api-reference/v1.0/api/call-keepalive.md
```diff
@@ -0,0 +1,75 @@
...
+description: "Make a request to this API every 15 minutes to ensure that an ongoing call stays alive."
+author: "VinodRavichandran"
+localization_priority: Normal
+ms.prod: "cloud-communications"
+doc_type: apiPageType
+---
+
+# call: keepAlive
+
+Make a request to this API every 15 minutes to ensure that an ongoing call stays alive. A call that does not receive this request within the time interval is considered inactive and will subsequently end.
```
ananmishr: @zhengni-msft what do you think about this?

Let's change the description to 

"Make a request to this API every 15 to 45 minutes to ensure that an ongoing call stays alive. A call that does not receive this request within this time interval is considered inactive and will subsequently end. 

A notification indicating that the call has ended will be sent. Due to network issues, the notification may not always be successfully sent out. After the call has ended, requests to this API will result in a `404 Error`. "

zhengni-msft: My recommendation: "A call that does not receive this request within 45 minutes (KeepAlive Timeout) is considered inactive and will subsequently end. Recommend to call this API with shorter interval (for example, 15 minutes) and handle failure properly to ensure at least one keep alive request reaches Graph successfully before hitting timeout. '404 Not-Found‘ will return in response if a call has ended because of timeout, hangup, or failures. In that case application should stop sending keep alive request and clean-up the resources of the call in application sdie. "

Lauragra:
Closing as this is a duplicate.

PR #6469, opened 2019-12-03
Microsoft Graph Toolkit 1.1 docs

nmetulev:
Documentation update for the 1.1 release of the Microsoft Graph Toolkit.

Adds the mgt-get component and the Proxy provider. Updates all existing documentation with new updates.


nmetulev:
Good question @Lauragra on the change log? Is there a way to leverage the release notes we author on GitHub - or should it be a separate change log in docs?

Lauragra:
@nmetulev , it would be good to have an entry in the changelog, here: https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/concepts/changelog.md. We can also link to the release notes in the changelog entry for details. That way it will also be listed in the monthly What's new topic.

PR #6172, opened 2019-11-01
add accessPackageResourceRequest

markwahl-msft:
Introduce to entitlement management in Microsoft Graph beta more operations to manage resource and their roles relationships to catalogs and access packages.

PR #6566, opened 2019-12-11
updated PR for entitlement management add accessPackageResourceRequest

markwahl-msft:
This PR adds to the documentation for beta/identityGovernance/entitlementManagement a new  object accessPackageResourceRequest, and APIs to be able to manage through Graph resources in access package catalogs and resource roles in access packages.  

This PR replaces #6172 and addresses the comments from that PR (one of the comments was that a POST message in that PR was missing a response body, which needed a deployment to fix and moved the completion of the docs from November to December.  This PR now includes the response body.)


markwahl-msft:
#sign-off

PR #6575, opened 2019-12-12
Update Postman environment file location

jeremie0:


Lauragra:
It looks like we fixed this in another PR - thanks!

PR #6584, opened 2019-12-12
Remove notes on access token format

psignoret:
The access token for Microsoft Graph should be an opaque value, the format is not important here.


psignoret:
#sign-off

PR #6401, opened 2019-11-23
Added Learn and Concept Pages for Cloud Communications

ananmishr:

PR #6614, opened 2019-12-16
Updated overview page

ananmishr:


Lauragra:
@ananmishr , this topic is part of PR #6401 ; do you want to merge this in separately and remove it from that PR?

concepts/cloud-communications-concept-overview.md
```diff
@@ -0,0 +1,119 @@
...
+## See Also
+Learn about [Delegated and Application Permissions](https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-permissions-and-consent)<br/>
+View and Manage [Permissions for Calls](/graph/permissions-reference#calls-permissions)<br/>
+View and Manage [Permissions for Online Meetings](/graph/permissions-reference#online-meetings-permissions)
+
+## API Reference
+[Communications API in Microsoft Graph (v1.0)](https://docs.microsoft.com/en-us/graph/api/resources/communications-api-overview?view=graph-rest-1.0)<br/>
+[Communications API in Microsoft Graph (beta)](https://docs.microsoft.com/en-us/graph/api/resources/calls-api-overview?view=graph-rest-beta)
+
+
```
ananmishr: @Lauragra I don't have a "Next Steps" header yet since the other pages are coming soon. I wanted to get this page out today and the rest out by Wednesday or Thursday

ananmishr:
> @ananmishr , this topic is part of PR #6401 ; do you want to merge this in separately and remove it from that PR?

Yes, I'm going to merge this separately to make sure that this is in and I'll update the file again in PR #6401

concepts/cloud-communications-concept-overview.md
```diff
@@ -0,0 +1,121 @@
...
+- Learn about [Delegated and application permissions](https://docs.microsoft.com/azure/active-directory/develop/v1-permissions-and-consent)<br/>
+- View and Manage [Permissions for calls](/graph/permissions-reference#calls-permissions)<br/>
+- View and Manage [Permissions for online meetings](/graph/permissions-reference#online-meetings-permissions)
+
+## API reference
+[Communications API in Microsoft Graph (v1.0)](https://docs.microsoft.com/en-us/graph/api/resources/communications-api-overview?view=graph-rest-1.0)<br/>
+[Communications API in Microsoft Graph (beta)](https://docs.microsoft.com/en-us/graph/api/resources/calls-api-overview?view=graph-rest-beta)
+
+
+## Feedback
```
Lauragra: FYI, I think we should remove this section, because all pages on Docs have a Feedback mechanism already, and having this duplication will be confusing to customers. 
image

ananmishr: Can I rename it to "API Feedback?" or "Cloud Communications Feedback?"

Lauragra: We can figure out a different way to present the information; there's also a larger question here - should we be consistent in how we ask for feedback across the various workloads? The **This product** link in the above image goes to [this page](https://developer.microsoft.com/en-us/graph/support), which provides general links to UV, GitHub, & SO. For CC, we're sending them to a samples repo for feedback, which is inconsistent with the rest of Graph. Unless we list these links on every CC page, you'll get inconsistent results, because some users will use the default Graph feedback mechanisms.

ananmishr: I see, so any forms or links that we provide should be in every cloud communications page including the API pages as well as the other learn/concept pages? Since the learn/concept pages are in a different path from the other API pages, I thought that it would be ok to add these in

Lauragra: What if we just change the heading? See the edit I checked in; I changed the heading to Support, to differentiate it from Feedback.

PR #6617, opened 2019-12-16

JarbasHorst:
Changed Graph permission from Application to Delegated since Presence.Read and Presence.Read.All only supports Delegated permissions.

fixed: #6613 

Consider updating the changelog since presence permissions were added in December, but are not listed in the Changelog!


Lauragra:
Thanks; I added the missing Permissions changelog entry in PR #6619.

PR #6399, opened 2019-11-23
Titus/recordresponserename

tituscha:
Documentation changes for renaming record to recordResponse and joinUrl to joinWebUrl.

PR #6620, opened 2019-12-16
changed from 2 min to 1 min - recordResponse

ananmishr:


api-reference/v1.0/api/call-record.md
```diff
@@ -16,7 +16,7 @@ For further information on how to handle operations, please review [commsOperati
 
 >**Note:** This is only supported for [calls](../resources/call.md) which are initiated with [serviceHostedMediaConfig](../resources/servicehostedmediaconfig.md).
 
-This action is not intended to record the entire call. The maximum length of recording is 1 minute. The recording is not saved permanently by the Cloud Communications Platform and is discarded shortly after the call ends. The bot must download the recording promptly after the recording operation finishes by using the recordingLocation value that's given in the completed notification.
+This action is not intended to record the entire call. The maximum length of recording is 2 minutes. The recording is not saved permanently by the Cloud Communications Platform and is discarded shortly after the call ends. The bot must download the recording promptly after the recording operation finishes by using the recordingLocation value that's given in the completed notification.
```
Lauragra: In the beta topic we say that the maximum length is 5 minutes- pls verify?

ananmishr: We are going to bring it down to 2 minutes soon for privacy reasons. We'll update the beta docs as well and have a note where we say that we've reduced it from 5 min to 2 min. (https://github.com/microsoftgraph/microsoft-graph-docs/pull/6399)

PR #6624, opened 2019-12-17
Add Christmas disclaimer to protected APIs

nkramer:


concepts/teams-protected-apis.md
```diff
@@ -28,5 +28,6 @@ The following APIs are currently protected:
 >[Send message](/graph/api/channel-post-messages?view=graph-rest-beta) is not a protected API.
 
 To request access to these protected APIs, complete the following [request form](http://aka.ms/teamsgraph/requestaccess). 
-We review access requests every Wednesday and deploy approvals every Friday.
+We review access requests every Wednesday and deploy approvals every Friday, 
+except the weeks of Thanksgiving, Christmas, and New Year's.
```
Lauragra: This isn't particularly globalized. I wonder if you want to be more generic, such as "except during holiday weeks in the U.S." Do you also want to add something like "If you submit your request during a holiday, we'll review your request the following week."

nkramer: Updates made. It's a little vague what constitutes a "major US holiday", but that's not necessarily a bad thing since we may or may not deploy on e.g. Fourth of July depending on what day of the week it falls, and I don't want to post a complete calendar that we have to maintain.

PR #6625, opened 2019-12-17
Update subscription-post-subscriptions.md

nkramer:


api-reference/beta/api/subscription-post-subscriptions.md
```diff
@@ -33,6 +33,8 @@ Creating a subscription requires read permission to the resource. For example, t
 |[security alert](../resources/alert.md) | SecurityEvents.ReadWrite.All | Not supported | SecurityEvents.ReadWrite.All |
 |[user](../resources/user.md) | User.Read.All | User.Read.All | User.Read.All |
 
+> **Note:** Soon chatMessage subscriptions will require [encryption](/graph/webhooks-with-resource-data), and subscription creation will fail if [encryptionCertificate](../resources/subscription.md) is not specified.
```
Lauragra: Is it possible to provide a more specific timeline? Soon sounds a little vague. Is there a specific action they need to take?

nkramer: I've added the best timeline we have (January) – action should be to do the encryption thing by specifying encryptionCertificate.

PR #6782, opened 2020-01-13
Update group-post-groups.md

JarbasHorst:
Added missing description property.


Lauragra:
@JarbasHorst , please make the same update to the beta version of the topic: https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/group-post-groups.md. Thanks!

JarbasHorst:
Hi @Lauragra, I've only updated the description property without max. supported length, since it is still under review.

PR #6810, opened 2020-01-15
Ruoyingl/graph doc

RuoyingLiang:
Range object has been adding new properties. The example in the document is get all properties. It may cause perf issue if user try to get large data set. So update the document to recommend user specific the properties they needed. Please feel free to update if you think there is better way to organize the content and express.

### DIFF
--- a/api-reference/v1.0/api/range-get.md
+++ b/api-reference/v1.0/api/range-get.md
@@ -50,7 +50,7 @@ Here is an example of the request.
   "name": "get_range"
 }-->
 ```msgraph-interactive
-GET https://graph.microsoft.com/v1.0/me/drive/items/{id}/workbook/names/{name}/range
+GET https://graph.microsoft.com/v1.0/me/drive/items/{id}/workbook/names/{name}/range?$select=address,addressLocal,cellCount,valueTypes
 ```
 # [C#](#tab/csharp)
 [!INCLUDE [sample-code](../includes/snippets/csharp/get-range-csharp-snippets.md)]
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/api/range-get.md
```diff
@@ -71,7 +71,7 @@ GET https://graph.microsoft.com/v1.0/me/drive/items/{id}/workbook/names/{name}/r
 ---
 
 ##### Response
-Here is an example of the response. Note: The response object shown here may be truncated for brevity. All of the properties will be returned from an actual call.
+Here is an example of the response. Note: If there is no optional query parameters, all of the properties will be returned from an actual call. The recommandation is to specific which properties are needed instead of get all properties for performance perspective.
```
Lauragra: Please revise the second sentence as follows:
>**Note:** If the request does not include any optional query parameters, all the properties will be returned. For best performance, we recommend that you specify only the properties that you need.

RuoyingLiang: Updated. Thank you! P.S. I also add another string in the request part in Example session. Please help to review. Thanks~~

RuoyingLiang:
Updated the string accordingly.

api-reference/v1.0/api/range-get.md
```diff
@@ -42,15 +42,15 @@ Do not supply a request body for this method.
 If successful, this method returns a `200 OK` response code and [Range](../resources/range.md) object in the response body.
 ## Example
 ##### Request
-Here is an example of the request.
+Here is an example of the request. Note: the example uses optional query parameters to customize the response.
```
Lauragra: Do you want to make the same change to the range-get topic in /beta? FYI, it will be easier to process PRs if you create a personal branch on the repo instead of submitting from your fork. I don't have access to edit files in your fork.

PR #6832, opened 2020-01-16
Add missing column information for places permissions

keesschollaart81:
I believe the documentenation of the places graph permissions is missing data in the 'Display Name' column. This causes the values of the other columns to shift left resulting in the wrong perception as an admin-consent is required!


Lauragra:
LGTM, thanks!

Issue #6852, opened 2020-01-20
Valid email


MasonTorres:
Is "At sign (@)" really not an allowed character in an email address?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 15c0170d-e214-ce6f-b7b4-589960e9962d
* Version Independent ID: bc805140-9ebe-a4ef-54e7-90f2054b8b83
* Content: [invitation resource type - Microsoft Graph v1.0](https://docs.microsoft.com/en-us/graph/api/resources/invitation?view=graph-rest-1.0#feedback)
* Content Source: [api-reference/v1.0/resources/invitation.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/v1.0/resources/invitation.md)
* Product: **microsoft-identity-platform**
* Technology: **microsoft-graph**
* GitHub Login: @davidmu1
* Microsoft Alias: **MSGraphDocsVteam**

PR #6858, opened 2020-01-21
Update invitation.md

JarbasHorst:
Removed "@" sign from list as it is a permitted character in an email address.

Fixes: #6852


RamonEbdonSS:
Just as FYI, if you're merging an update to this doc, the Pipe character should also be escaped.

@davidmu1 fixed, but then reverted this change in [f68c7d7](https://github.com/microsoftgraph/microsoft-graph-docs/commit/f68c7d79d5271e7d08892263f10c9b641d549b10#diff-aa85ac22bcdd91ddb7c1a85069d15c78)

JarbasHorst:
Hi @RamonEbdonSS ,
I've escaped the pipe and removed the hyphen since hyphens are allowed in an email address.

PR #6860, opened 2020-01-21
Update calendar-get.md

Harini84:
Graph beta Documentation for Calendar HexColor and event TransactionId.

TransactionId is supported in Create Event but not in update event.

Update to 'isAllDay' property as it expects start and end time to be mid-night and in same time-zone.


api-reference/beta/resources/event.md
```diff
@@ -134,6 +134,7 @@ Here is a JSON representation of the resource
 |showAs|String|The status to show. Possible values are: `free`, `tentative`, `busy`, `oof`, `workingElsewhere`, `unknown`.|
 |start|[DateTimeTimeZone](datetimetimezone.md)|The start date, time, and time zone of the event. By default, the start time is in UTC.|
 |subject|String|The text of the event's subject line.|
+|transactionId|String|A custom identifier specified by client app for the service to deduplicate calls in case of client retries to create an event. This is typically used by client app to handle low network connectivity and time out scenarios|
```
angelgolfer-ms: Line 137: I suggest rephrasing as follows:
"A custom identifier specified by a client app for the server to avoid redundant operations in case of client retries to create the same event. This is useful when low network connectivity causes the client to time out before receiving a response from the server for the client's preceding create request."

The term "deduplicate" is technically applicable to data storage scenarios. While I can guess what we're trying to say using that term, it might be confusing to Localization if we use the term here.

api-reference/beta/api/event-update.md
```diff
@@ -73,6 +73,8 @@ In the request body, supply the values for relevant fields that should be update
 | start|DateTimeTimeZone|The start date, time, and time zone of the event. |
 | subject|String|The text of the event's subject line.|
 
+TransactionId is supported only in create event and not in update event request.
```
angelgolfer-ms: I see that you'd like to point out that once the transactionId value is specified in POST (create event), then the value cannot be changed anymore.
Please do it this way:

- In the event.md topic, add the following to the description of the transactionId property:

"Once you set transactionId when creating an event, you cannot change transactionId in a subsequent update."

- Please style the two instances of transactionId as bold. 
- Please make "creating an event" a link; the link URL is ../api/user-post-events.md.
- Please make "update" a link, with the link URL ../api/event-update.md.

- Please remove the remark here in line 76, in the event-update.md topic, for consistency across other resources. 

In general, the "Request body" section in PATCH topics lists the properties that can be updated. Authors do not include those properties that cannot be updated. So if one starts specifying properties that cannot be updated, that may throw doubt whether other properties _not_ included in the Request body table are overlooked (with no mention in the table or separately).

api-reference/beta/resources/event.md
angelgolfer-ms: Also, is it optional to set this property? I guess so, such that one doesn't have to set it when creating an event?
If it's optional, please add the single-word sentence "Optional" to the very end of the description for the transactionId property.

api-reference/beta/api/calendar-post-events.md
```diff
@@ -372,6 +372,163 @@ Content-type: application/json
 }
 ```
```
angelgolfer-ms: @Harini84 , if the POST request body doesn't specify the transactionId, would that property be returned in the response payload?
If so, what value, is it empty or null?
Also, if that property is always returned, please update each existing example that creates an event, to have the example response include that property.

angelgolfer-ms: Reminder:
For creating events, the metadata had the stub generator create two separate topics - calendar-post-events.md, and user-post-events.md.
Please update the Examples section of user-post-events.md as well.

angelgolfer-ms: One suggestion:
Since from the developer customer's perspective, using the transactionId is just a matter of assigning a property when creating an event, in calendar-post-events.md, as an alternative to adding an entire example 3, you can use example 1, and call out that example "assigns the transactionId property for efficient processing", or something to that effect.

Your title for Example 1 can be:
"Example 1: Create an event in a specific calendar and assign the event a transactionId value"

The advantage of that is to make the overall topic shorter, better for customer scanning. When a topic is long - each example response is easily two pages on a monitor screen - the customer would likely not scroll to the end to go through all the examples.
Internally, we limit the number of examples we need to maintain over time - for instance, the next time we add another property to event, instead of updating the response payload of two examples in calendar-post-events.md, we'll have to update three examples.

Typically when we add a property to an existing entity, we update each existing example by adding the property to the response payload as appropriate, unless there is a more compelling customer scenario like organizing online meetings, in which case we also create a new example.

api-reference/beta/api/calendar-post-events.md
```diff
@@ -372,6 +372,163 @@ Content-type: application/json
...
+        "address":"adelev@contoso.onmicrosoft.com",
+        "name": "Adele Vance"
+      },
+      "type": "required"
+    }
+  ],
+   "transactionId":"7E163156-7762-4BEB-A1C6-729EA81755A7"
+}
+```
+# [C#](#tab/csharp)
```
angelgolfer-ms: Even though it's better not to add a separate example 3 but tag onto to example 1, just an FYI for future purposes, when we add a REST example, we shouldn't include the lines 421 to 431. Internally, there's a tool that runs weekly to generate / re-generate language-specific code snippets.
The downside of including these lines before the regeneration is, as soon as this topic is published, customers would see the language-specific tabs, but the code snippets shown would not have anything to do with example 3, which would be confusing. If we exclude lines 421-431, customers would just not see the language-specific tabs.

api-reference/beta/resources/event.md
Harini84: Changes done.

api-reference/beta/api/event-update.md
Harini84: Changes incorporated as per comment

api-reference/beta/api/calendar-post-events.md
Harini84: > @Harini84 , if the POST request body doesn't specify the transactionId, would that property be returned in the response payload?
> If so, what value, is it empty or null?
> Also, if that property is always returned, please update each existing example that creates an event, to have the example response include that property.

No, only if 'transactionId' is specified in POST request, it will be returned in response payload. Else it will not be returned.

Harini84: Updated to remove example 3 and specify in example 1 as per above comment

Harini84: > Reminder:
> For creating events, the metadata had the stub generator create two separate topics - calendar-post-events.md, and user-post-events.md.
> Please update the Examples section of user-post-events.md as well.

user-post-events specifically describes allowNewTimeProposals and location. Hence not updating it as the generic update is done in Example 1 in calendar-post-event.

Harini84:
@angelgolfer-ms 
Incorporated the above review comments. Pleaseupdate if any further changes required.

angelgolfer-ms:
> @angelgolfer-ms
> Incorporated the above review comments. Please update if any further changes required.

@Harini84 , did you do a commit to your remote branch and build? For example, in your branch on Github, I still see the same description for transactionId in event.md as before my review. Please confirm, update your remote branch, let me know, and I can take a look.

api-reference/beta/resources/event.md
angelgolfer-ms: I still see the same description for transactionId as before. 
Taking into account some other related info you provided in this review, I enhanced the description as follows:

"A custom identifier specified by a client app for the server to avoid redundant `POST` operations in case of client retries to create the same event. This is useful when low network connectivity causes the client to time out before receiving a response from the server for the client's prior create-event request. Once you set transactionId when creating an event, you cannot change transactionId in a subsequent update. This property is only returned in a response payload if an app has set it. Optional."

api-reference/beta/api/calendar-post-events.md
angelgolfer-ms: The [first example ](https://docs.microsoft.com/en-us/graph/api/user-post-events?view=graph-rest-beta&tabs=http#example-1-create-an-event) in user-post-events.md is just as plain-vanilla create event example. It's not dedicated to showing allowNewTimeProposals alone. It's ok and beneficial to show setting transactionId there as well, similar to calendar-post-events.md.

Just so you know, the fact that we have got 2 separate files, user-post-events.md and calendar-post-events.md is just to separate those scenarios/ syntax that creates an event in a specific user calendar into its own file. It's purely a human decision early on in 2015 (not certain of the reason as I wasn't around), and we could have combined all the scenarios/syntax that create an event in a user calendar in just one file, not two. user-post-events.md already includes some of the syntax for creating an event in a specific calendar. And more importantly, user-post-events.md is exposed in 2 places in the table of contents (TOC):
1. Users >> Calendar >> Create event 
2. Calendar >> Event >> Create event

Both are higher up in the table of contents than the single TOC entry point for calendar-post-events.md:
Calendar >> Calendar >> Create event

So it would be beneficial to customers to expose transactionId in an example in user-post-events.md as well. You can do that to example 1 in user-post-events.md, not to proliferate the number of examples just for one property. Thanks!

Harini84:
> > @angelgolfer-ms
> > Incorporated the above review comments. Please update if any further changes required.
> 
> @Harini84 , did you do a commit to your remote branch and build? For example, in your branch on Github, I still see the same description for transactionId in event.md as before my review. Please confirm, update your remote branch, let me know, and I can take a look.

Hi @angelgolfer-ms ,
The remote commit/ push request is completed. Could see the changes.

image.

Please let me know if anything else to be done from my side.

Thanks,
Harini

api-reference/beta/resources/event.md
Harini84: @angelgolfer-ms 
So, should I copy the above description and do the commit/ push changes again or is the update already done? Please clarify.

Harini84: @angelgolfer-ms 
I have copied the above new text. Please check the latest commit.

api-reference/beta/api/calendar-post-events.md
Harini84: @angelgolfer-ms 
As can be observed the first example in user-post-events.md is for explaining 'Prefer: outlook.timezone' and it is not a generic example.
Are you sure of updating this example to explain transactionID? Please confirm.

If this is the example to be used, then please provide the text for request 'Here is an example of the request. It uses the `Prefer: outlook.timezone` request header to specify the time zone for the **start** and **end** times in the response.'

api-reference/beta/api/calendar-post-events.md
```diff
@@ -61,7 +61,7 @@ If successful, this method returns `201 Created` response code and [event](../re
 ### Example 1: Create an event in a specific calendar
 
 #### Request
-The following example creates an event in the specified calendar of the signed-in user's.
+The following example creates an event in a specific calendar and assign the event an optional transactionId value
```
angelgolfer-ms: Line 64: Please add period at the end of the sentence.

api-reference/beta/api/event-update.md
```diff
@@ -59,7 +59,7 @@ In the request body, supply the values for relevant fields that should be update
 | categories|String|The categories associated with the event.|
 | end|DateTimeTimeZone|The date, time, and time zone that the event ends. |
 | importance|String|The importance of the event. Possible values are: `low`, `normal`, `high`.|
-| isAllDay|Boolean|Set to true if the event lasts all day.|
+| isAllDay|Boolean|Set to true if the event lasts all day. If true, irrespective of whether single day or multi-day event, start and end time must be set to mid-night and in same   time-zone.|
```
angelgolfer-ms: Line 62: 
Please hyphenate "single-day" as it's used as an adjective to "event".
Please rephrase the last part of the description, remove extra spaces between "same" and "time-zone", and spell "mid-night" as one word, "time-zone" as two words. The latter is consistent with the rest of the Graph docs, and the IANA web site. 
Rephrase: "... set to midnight and be in the same time zone."

api-reference/beta/resources/event.md
```diff
@@ -110,7 +110,7 @@ Here is a JSON representation of the resource
 |hasAttachments|Boolean|Set to true if the event has attachments.|
 |id|String| Unique identifier for the event. [!INCLUDE [outlook-beta-id](../../includes/outlook-beta-id.md)] Read-only. |
 |importance|String|The importance of the event. Possible values are: `low`, `normal`, `high`.|
-|isAllDay|Boolean|Set to true if the event lasts all day.|
+|isAllDay|Boolean|Set to true if the event lasts all day. If true, irrespective of whether single day or multi-day event, start and end time must be set to mid-night and in same time-zone.|
```
angelgolfer-ms: Line 113: same comments for that property description as in the v1 event.md topic.

api-reference/beta/resources/event.md
angelgolfer-ms: Looks fine now, thanks!

Harini84:
@angelgolfer-ms 
Please complete this review so that we can proceed with code check-in as well for Graph beta.

@sochowdh

api-reference/beta/api/calendar-post-events.md
angelgolfer-ms: In the beta version of user-post-events.md, please rephrase as follows:
"Example 1: Create an event in the specified time zone, and assign the event an optional transactionId value"

Then in both the v1 and beta versions, add a third sentence to the request description:
"Here is an example of the request. It uses the Prefer: outlook.timezone request header to specify the time zone for the start and end times in the response. It also sets the transactionId property to reduce unnecessary retries on the server."

api-reference/beta/api/calendar-post-events.md
Harini84: Modified as per feedback.

api-reference/beta/resources/event.md
Harini84: Modified as per feedback.

Harini84:
> Just a handful of remaining comments, some of them are straight-forward fixes. Thanks.

@angelgolfer-ms 
Modifications made as per feedback. Please let me know if any further details needed.

Harini84:
@angelgolfer-ms 
Awaiting on update.

Harini84:
@angelgolfer-ms 
This is set to 'Ready to merge' on Friday. Kindly update when will the request will be merged.

PR #6898, opened 2020-01-24
Update team-get.md

JarbasHorst:
Added discoverySettings since it is part of the v1.0/teams/{id} response


laujan:
I reviewed and checked with Nick.  This is okay to merge.

PR #6918, opened 2020-01-28
Changed table so it's more logical

WithHolm:
Both the request type and the permission type was in the same column. changed this so 'request' had its own column.
This should also be updated at all of the Intune-Shared-* documents.

### DIFF
--- a/api-reference/beta/api/intune-shared-windowsupdatestate-list.md
+++ b/api-reference/beta/api/intune-shared-windowsupdatestate-list.md
@@ -18,15 +18,14 @@ List properties and relationships of the [windowsUpdateState](../resources/intun
 ## Prerequisites
 One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
 
-|Permission type|Permissions (from most to least privileged)|
-|:---|:---|
-|Delegated (work or school account)||
-| &nbsp; &nbsp; **Device configuration** | DeviceManagementConfiguration.ReadWrite.All, DeviceManagementConfiguration.Read.All|
-| &nbsp; &nbsp; **Software Update** | DeviceManagementConfiguration.ReadWrite.All, DeviceManagementConfiguration.Read.All|
-|Delegated (personal Microsoft account)|Not supported.|
-|Application||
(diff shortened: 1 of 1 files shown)


WithHolm:
could i also add that this way of handling multiple scopes would do it aswell: https://docs.microsoft.com/en-us/graph/api/calendar-get?view=graph-rest-1.0&tabs=http

api-reference/beta/api/intune-shared-windowsupdatestate-list.md
```diff
@@ -18,15 +18,14 @@ List properties and relationships of the [windowsUpdateState](../resources/intun
...
-|:---|:---|
-|Delegated (work or school account)||
-| &nbsp; &nbsp; **Device configuration** | DeviceManagementConfiguration.ReadWrite.All, DeviceManagementConfiguration.Read.All|
-| &nbsp; &nbsp; **Software Update** | DeviceManagementConfiguration.ReadWrite.All, DeviceManagementConfiguration.Read.All|
-|Delegated (personal Microsoft account)|Not supported.|
-|Application||
-| &nbsp; &nbsp; **Device configuration** | DeviceManagementConfiguration.ReadWrite.All, DeviceManagementConfiguration.Read.All|
-| &nbsp; &nbsp; **Software Update** | DeviceManagementConfiguration.ReadWrite.All, DeviceManagementConfiguration.Read.All|
+Request|Permission type|Permissions (from most to least privileged)|
+|:--|:---|:---|
```
davidmu1: I'm not sure we want to do it this way. Might be better to have two tables. Will defer to what @Lauragra and @angelgolfer-ms think.

Lauragra: I think this is okay in one table. We do have the ability to customize the Permissions tables without causing a validation error.

Lauragra:
@davidmu1 , can we get this applied to the Intune file source generation? Thanks.

Lauragra:
@WithHolm , thank you for your contribution. Because we have multiple files with a similar table structure in our doc set, we're going to take a look at how we're organizing this information overall. We'll factor in your feedback and we plan to make a global update in the future, so for now we'll close this PR. Thanks again for contributing.

PR #6921, opened 2020-01-28
Update CSS supportability

a-baabde:
Stating the supportability limits for the tutorials from Microsoft CSS. 
Tutorials should be supported by the GitHub project community who created them. Microsoft only can help with Microsoft languages and not anything else.

### DIFF
--- a/concepts/tutorials.md
+++ b/concepts/tutorials.md
@@ -32,6 +32,8 @@ The following tutorials build an app that authenticates a user and accesses date
 
 Don't see your preferred language or platform? Request a tutorial for it on [UserVoice](https://microsoftgraph.uservoice.com/forums/920506-microsoft-graph-feature-requests).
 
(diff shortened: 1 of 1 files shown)


jasonjoh:
@a-baabde Was there some confusion with a customer that prompted this? All of our tutorials already ask customers to file issues on the corresponding GitHub repo if they find a problem with the content.

a-baabde:
@jasonjoh, Yes several customers downloaded the sample solutions from the tutorials then they come back to us asking about how to make them work in languages that's not supported by Microsoft like Python and Ruby. Customers argue that the tutorials are on Microsoft documentation so we should be supporting this.

jasonjoh:
Well, they are Microsoft documentation. Are you saying that CSS won't support a developer if they're using Python or Ruby? I would think that you would still provide support for the API itself, right?

jasonjoh:
Closing this as we're replacing this page.

PR #6929, opened 2020-01-28
reverted the TOC and changelog to remove policy additions

CelesteDG:
Removed the policy entries in the changelog and reverted the TOC to remove the policy API topics (sources were kept)

### DIFF
--- a/api-reference/beta/toc.yml
+++ b/api-reference/beta/toc.yml
@@ -8912,67 +8912,25 @@
           items:
           - name: Get member objects
             href: api/orgcontact-getmemberobjects.md
-      - name: Policy
+      - name: Policies
         href: resources/policy-overview.md
         items:
-        - name: Activity-based timeout policy
-          href: resources/activitybasedtimeoutpolicy.md
-          items:
-          - name: Create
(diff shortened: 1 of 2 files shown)


api-reference/beta/toc.yml
```diff
@@ -8912,67 +8912,25 @@
           items:
           - name: Get member objects
             href: api/orgcontact-getmemberobjects.md
-      - name: Policy
+      - name: Policies
```
dkershaw10: This should still say Policy, and the href should be to policy.md.

dkershaw10: This should still say Policy, and the href should be to policy.md.

api-reference/beta/toc.yml
```diff
@@ -8915,64 +8915,22 @@
       - name: Policy
         href: resources/policy-overview.md
```
dkershaw10: No - resources/policy.md

The policy-overview.md is a new topic for the new API.

CelesteDG:
Confirmed changes are rendering fine. Merging to master.

PR #6962, opened 2020-01-30
- fixes #6376: adds a note regarding possible delays before DQ reflects changes

### LINKED ISSUE #6376: Delta Query - Possible Delays
There can be a delay between when a change is made and it is available in a delta query. I've seen it take ~15 minutes in a test tenant with only 15 or so users for delta to show a group membership change for example.

---
#### Document Details

* ID: ac7ca25a-037d-5202-debc-92c06a2c8480
* Version Independent ID: 8ebb4dbc-76a1-00fd-e740-87eda29a40f0
* Content: [Known issues with Microsoft Graph - Microsoft Graph](https://docs.microsoft.com/en-us/graph/known-issues#delta-query)
* Content Source: [concepts/known-issues.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/concepts/known-issues.md)
* GitHub Login: @MSGraphDocsvTeam

baywet:


concepts/delta-query-overview.md
```diff
@@ -160,6 +160,10 @@ This tells you that the **skills** property is not supported for delta query on
 
 Navigation properties are not supported. For example, you cannot track changes to the users collection that would include changes to their **photo** property; **photo** is a navigation property stored outside of the user entity, and changes to it do not cause the user object to be included in the delta response.
 
+#### Processing Delays
+
+Developers should expect varying delays between the time a change is made in the interface or via APIs and the time this change is reflected by the delta query.
```
angelgolfer-ms: May I suggest a rephrase:
"Expect varying delays between the time a change is made to a resource instance, which can be through an app interface or API, and the time the tracked change is reflected in a delta query response."
Allow me to elaborate:
- Our docs should use second-person imperative where possible. Therefore, please drop the subject "Developers should", when we're giving directives to the target customer (developers).
- The other tweaks are just for clarity. Customers often only scan and so we can't assume they're always tuned in to the context prior to this new section.

angelgolfer-ms: There are some formatting issues for this section titled [Limitations](https://review.docs.microsoft.com/en-us/graph/delta-query-overview?branch=pr-en-us-6962#limitations), and its sub-sections:

- Please change the "Limitations" section from a level 3 section heading to a level 2 section heading. Based on the contents of this section, it shouldn't be a sub-section of the "Supported resources" section. Being a level 2 section would also bubble it up on the right-side TOC, helping customer to discover this section.

- Following the order of levelling, please elevate the 3 sub-sections of "Limitations" (e.g., "Properties stored outside of the main data store", "Navigation properties", and your new section "Processing delays") from level 4 sections to level 3 sections.

I hope you don't mind these levelling issues that you inherited from the past and fix them? Thanks!

concepts/delta-query-overview.md
```diff
@@ -160,6 +160,10 @@ This tells you that the **skills** property is not supported for delta query on
 
 Navigation properties are not supported. For example, you cannot track changes to the users collection that would include changes to their **photo** property; **photo** is a navigation property stored outside of the user entity, and changes to it do not cause the user object to be included in the delta response.
 
+#### Processing Delays
```
angelgolfer-ms: Please use sentence case capitalization "Processing delays".
This is our capitalization guidance: https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28939
In short, only marketing approved product / service names can use "title style case".

concepts/delta-query-overview.md
baywet: not at all. Thanks for spending the time reviewing the changes and the insights on how to improve the the next time, I'll try to keep that in mind.

baywet:
#sign-off

PR #7022, opened 2020-02-07
Updated list with list title

frankchen76:
Per conversation with Jeremy Thake from our FTE "Microsoft Graph Field Community", I updated the list get with list title.


Lauragra:
@frankchen76 , please update the beta version of this topic as well. Thanks.

PR #7043, opened 2020-02-10
Corrected some inconsistencies

xom-pschiess:
Endpoint was listed as POST in the description, but DELETE in the example.
The HTTP example incorrectly listed a body, which is not a part of the API
The HTTP example was referencing the v1.0 endpoint


Lauragra:
@davidmu1 , can you please fix the v1.0 version of this file as well? It seems to be using POST in the HTTP request section.

PR #7044, opened 2020-02-10
Fixing doc issues for large attachments

svpsiva:


api-reference/beta/api/attachment-createuploadsession.md
```diff
@@ -13,7 +13,7 @@ doc_type: "apiPageType"
 
 Create an upload session that allows an app to iteratively upload ranges of a file, so as to attach the file to the specified [message](../resources/message.md).
 
-Use this approach to attach files of sizes between 3 MB and 150 MB to a **message**. To attach files of sizes under 4 MB, simply [POST on the attachments navigation property](message-post-attachments.md). 
+Use this approach to attach a file to a **message** when the size is between 3 MB and 150 MB. To attach a file that's smaller than 3 MB,  [POST on the attachments navigation property](message-post-attachments.md).
```
angelgolfer-ms: Please add "file", as in "... attach a file to a message when the file size is between 3 MB and 150 MB", to disambiguate whether it's the file size or message size we're talking about.

api-reference/beta/api/attachment-createuploadsession.md
```diff
@@ -77,12 +77,10 @@ If successful, this method returns `201, Created` response code and a new [uploa
 
 ## Examples
 
-The following is an example of how to call this API.
+The following example shows how to call this API.
```
angelgolfer-ms: Since we bother to edit the boilerplate introductory text, let's change the introductory text to something that reinforces learning, which would be helpful if the customer is scanning and doesn't necessarily read everything in the topic:

"The following example shows how to create an upload session, that you can use in subsequent file upload operations."

concepts/outlook-large-attachments.md
```diff
@@ -10,9 +10,11 @@ ms.prod: "outlook"
 
 Depending on the size of the file, you can choose one of two ways to attach a file to a [message](/graph/api/resources/message?view=graph-rest-beta):
 
-- If the file size is under 4 MB, you can do a single [POST on the attachments navigation property of the message](/graph/api/message-post-attachments?view=graph-rest-beta). The successful `POST` response includes the ID of the file attached to the message.
+- If the file size is under 3 MB, you can do a single [POST on the attachments navigation property of the message](/graph/api/message-post-attachments?view=graph-rest-beta). The successful `POST` response includes the ID of the file attached to the message.
 - If the file size is between 3MB and 150MB, create an upload session, and iteratively use `PUT` to upload ranges of bytes of the file until you have uploaded the entire file. A header in the final successful `PUT` response includes a URL with the attachment ID. 
 
+You can repeat this process to upload multiple file attachments to the same **message**. 
+
```
angelgolfer-ms: Line 16: Because the PUT operations (that use the same uploadUrl) are indeed repetitive, maybe we should avoid the term "repeat" in line 16, if we mean customers should create a new upload session for each large file attachment.

"To attach multiple large files to a message, create a separate upload session and use the similar process for each large file."

concepts/outlook-large-attachments.md
```diff
@@ -252,5 +254,10 @@ DELETE https://outlook.office.com/api/beta/Users('a8e8e219-4931-95c1-b73d-62626f
 ```http
 HTTP/1.1 204 No content
 ```
+## Errors
+
+### ErrorAttachmentSizeShouldNotBeLessThanMinimumSize
+
+This error indicates that **uploadSession** was used for a file smaller than 3 MB in size. If the file size is under 3 MB, you should do a single [POST on the attachments navigation property of the message](/graph/api/message-post-attachments?view=graph-rest-beta). The successful `POST` response includes the ID of the file attached to the message.
```
angelgolfer-ms: This error can be returned in step 1, when trying to create the upload session, is that correct?

If so, please clarify in the first sentence. I can suggest:
"This error results when attempting to [create an upload session](https://docs.microsoft.com/en-us/graph/api/attachment-createuploadsession?view=graph-rest-beta&tabs=http) to attach a file smaller than 3 MB. If the file size is under 3 MB, …"

Please link "create an upload session" to the reference topic for the create upload session method. In the actual MD file, please specify your link as /graph/api/attachment-createuploadsession?view=graph-rest-beta. Do not use an http link.

concepts/outlook-large-attachments.md
svpsiva: There are two different ways to upload, one each for large and small attachments. The repetitive process that we refer to includes both.

angelgolfer-ms: In this case, we can say this instead:
"To attach multiple files to a message, choose the approach for each file based on its file size and attach them individually."

api-reference/beta/api/attachment-createuploadsession.md
```diff
@@ -13,7 +13,7 @@ doc_type: "apiPageType"
 
 Create an upload session that allows an app to iteratively upload ranges of a file, so as to attach the file to the specified [message](../resources/message.md).
 
-Use this approach to attach files of sizes between 3 MB and 150 MB to a **message**. To attach files of sizes under 4 MB, simply [POST on the attachments navigation property](message-post-attachments.md). 
+Use this approach to attach a file to a **message** when the file size is between 3 MB and 150 MB. To attach a file that's smaller than 3 MB,  [POST on the attachments navigation property](message-post-attachments.md).
```
angelgolfer-ms: @svpsiva , can you clarify please - does POST on the attachments nav property still support files between 3MB and 4MB? 
If we technically changed the upper limit from 4MB to 3MB, that's a breaking change, isn't it? We would need to update the v1 and beta versions of event-post-attachments.md and message-post-attachments.md.

Or is the change in line 16 just a recommendation, and we didn't change the actual upper limit of 4 MB?

angelgolfer-ms: I got the explanation in the GitHub issue thread with customers on outlook-large-attachments.md.

PR #7049, opened 2020-02-10
Alezhan onenote postpages

alezhan:
Updating the OneNote API documentation to reflect new page limit we put in place for the POST pages endpoint


api-reference/v1.0/api/onenote-post-pages.md
```diff
@@ -48,6 +48,13 @@ The body can contain HTML placed directly in the request body, or it can contain
 
 If successful, this method returns a `201 Created` response code and the new [page](../resources/page.md) object in the response body.
 
+## Limitations
```
Lauragra: This is non-standard content for a reference topic. I recommend that you add the content to this concept topic: https://docs.microsoft.com/en-us/graph/onenote-create-page. Then, you can link to it from the operation topic. Be sure to add the reference to the /beta operation topic as well. This also warrants an entry in the changelog.

alezhan: Added the content to the concept topic and linked the concept topic to reference topics for both v1 and beta

PR #7162, opened 2020-02-21
Update activitystatistics.md

madehmer:
remove info about date range for ID property


madehmer:
@angelgolfer-ms -per Mathew Aniyan's 2/20 email, would you please review and approve this minor change please? Thanks!

api-reference/beta/resources/activitystatistics.md
```diff
@@ -35,18 +35,18 @@ For example, the ID "email_2019-08-10_2019-08-12" represents the emailActivitySt
...
+| [List activityStatistics](../api/activitystatistics-list.md) | [activityStatistics](activitystatistics.md) | Retrieve the properties for the collection of activity statistics for a user, for the last complete week. |
 
 ## Properties
 
 | Property     | Type        | Description |
 |:-------------|:------------|:------------|
-|activity|analyticsActivityType| The type of activity for which statistics are returned. The possible values are: `call`, `chat`, `email`, `focus`, and `meeting`.|
-|duration|Duration|Total hours spent on the activity. The value is represented in ISO 8601 format for durations.|
-|endDate|Date|Date when the activity ended, expressed in ISO 8601 format for calendar dates. For example, the property value could be "2019-07-03" that follows the YYYY-MM-DD format.|
-|id|String| Read-only ID for the activity, which represents `{activity}_{startdate}_{enddate}`.|
```
angelgolfer-ms: @madehmer , I understand the purpose of this PR. I think it's (just) ok to change line 47 to what it is.

In Microsoft Graph today, there are contexts where an API exposes a value that the product team doesn't want customers to customize or re-engineer. They would call that out. An example is [here](https://docs.microsoft.com/en-us/graph/outlook-large-attachments?tabs=http#step-1-create-an-upload-session), where we tell customers not to customize the upload URL and use it as is for the purpose it's meant for.

If that's where your product team/marketing team are coming from too, then you might want to consider explicitly saying something like "Do not parse or customize the value for your scenarios." 
When customers see the actual values for the id property in the example in list activityStatistics topic, it would be obvious to most of them what the different parts mean anyway.

madehmer: Thank you Angela for your time on this!

madehmer:
@angelgolfer-ms -thanks for your feedback! I've added your suggested note to the ID property.

madehmer:
@Lauragra -Hey, Angela has approved these few changes. Would you please push to master and publish to live for me? Thanks!

PR #7169, opened 2020-02-22
Updated case for  initiatedBy/app/displayName

ramseygr:
Incorrect case causing failure on api call.


api-reference/v1.0/api/directoryaudit-list.md
```diff
@@ -50,7 +50,7 @@ This method supports the following OData query parameters to help customize the
 | initiatedBy/user/displayName                                 | eq                  |
 | initiatedBy/user/userPrincipalName                           | eq, startswith      |
 | initiatedBy/app/appId                                        | eq                  |
-| initiatedBy/app/appDisplayName                               | eq                  |
+| initiatedBy/app/displayName                                  | eq                  |
```
Lauragra: If this change is accurate, we need to make the same change to the /beta version of the topic.

PR #7207, opened 2020-02-26
fix for git issue 6253

davidmu1:


api-reference/v1.0/api/groupsetting-update.md
```diff
@@ -115,29 +122,47 @@ Content-length: 173
...
+HTTP/1.1 204 No Content
+```
 
-# [Java](#tab/java)
-[!INCLUDE [sample-code](../includes/snippets/java/update-groupsetting-java-snippets.md)]
-[!INCLUDE [sdk-documentation](../includes/snippets/snippets-sdk-documentation-link.md)]
+### Example 2: Update a group-specific setting
 
----
+In this example, the first `{id}` in the request is the identifier of the group, and the second `{id}` is the identifier of the group settings object.
```
dkershaw10: @davidmu1 You might want to consider maybe having descriptive ids:  `groupId` and `settingId`?  @Lauragra @angelgolfer-ms Thoughts?

api-reference/v1.0/api/groupsetting-update.md
```diff
@@ -48,30 +48,37 @@ In the request body, supply the values for relevant fields that should be update
 
 If successful, this method returns a `204 No Content` response code.
 
-## Example
+## Examples
+
+### Example 1: Update a tenant-wide setting
```
dkershaw10: It's still a group setting, so I would say "Update a tenant-wide group setting"

api-reference/v1.0/api/groupsetting-update.md
angelgolfer-ms: If I were doing this, I would actually specify real-looking IDs in place of any placeholder IDs. The reason is in math theory, placeholders serve generalization. Examples show specific instances. IMHO, we're losing grasp of the purpose of an example  when we introduce generalization to it.
In practice, there's value in showing real ID values, for example, I have seen customers asking what IDs look like for the user entity. If our examples generally use real-looking ID values, customers wouldn't have to dig all over the place.

That being said, there are authors who use plain {id} as placeholder. And, 4+ years ago Yina and Sudhi decided that we're not bothering to annotate what {id} is as they think it's intuitive. So if you're going to use a placeholder, then please just adhere to the existing placeholder approach - just use {id}, do not annotate, and do not bother to vary.

davidmu1: I think I'll go ahead and leave it as is. Although I think it is pretty obvious which id is which in the request, I don't think it hurts anything to have the sentence. Putting in actual GUIDs really doesn't help know what the GUID relates to.

api-reference/v1.0/api/groupsetting-update.md
davidmu1: Changed.

PR #7237, opened 2020-02-29
Throttling limits for teams

nkramer:


concepts/teams-throttling-limits.md
```diff
@@ -0,0 +1,78 @@
+---
+title: "Microsoft Graph throttling limits and boundaries"
+description: "Throttling limits the number of concurrent calls to a service to prevent overuse of resources. Microsoft Graph is designed to handle a high volume of requests. If an overwhelming number of requests occurs, throttling helps maintain optimal performance and reliability of the Microsoft Graph service."
+localization_priority: Priority
+author: "nkramer"
+---
+
+# Limits and boundaries for Microsoft Teams
```
Lauragra: The H1 and the title attribute are not consistent. To make sure we get the important keywords into the title, and front-load the more important keywords, suggested title: "Throttling limits for Microsoft Teams APIs in Microsoft Graph".

concepts/teams-throttling-limits.md
```diff
@@ -0,0 +1,78 @@
...
+
+Throttling limits the number of concurrent calls to a service to prevent overuse of resources.
```
Lauragra: The introduction should make it clear that this article covers throttling for Teams APIs specifically, so that it isn't confused with the more general throttling article.

api-reference/beta/toc.yml
```diff
@@ -32,6 +32,8 @@
       href: /graph/json-batching?context=graph/api/beta
     - name: Throttling
       href: /graph/throttling?context=graph/api/beta
+    - name: Teams throttling limits
```
Lauragra: This is probably not the right place to surface the topic in the TOC. We'll want to include this in the TOC file in the concepts folder, as a child topic of the Teams concept topic: https://docs.microsoft.com/en-us/graph/teams-concept-overview. As needed, you can link to it from reference topics.

concepts/teams-throttling-limits.md
```diff
@@ -0,0 +1,83 @@
...
+| GET messaging for an app/tenant | 5 requests/second |
+| POST messaging for an app/tenant | 2 requests/second |
+| POST channel messages for a tenant | 5000 messages/day per channel |
+| POST channel messages for a user | 3000 messages/day per channel |
+
+The following scenarios are not supported:
+- Using Teams as a log file. When your app sends a message, you should expect a person to read it.
+- Migrating messages from other messaging products or installations. The [Send message](/api/channel-post-messages?view=graph-rest-beta) API does not have the throughput necessary for a typical migration, and we are unable to adjust limits to assist migrations.
+
+## Group limits
```
baywet: The section and the following one (AAD service limits) are not about throttling but rather about usage, maybe we should find another place to document those so this article is specifically about throttling?

concepts/teams-throttling-limits.md
```diff
@@ -0,0 +1,83 @@
+---
+title: "Throttling limits for Microsoft Teams APIs in Microsoft Graph"
+description: "Throttling limits the number of concurrent calls to a service to prevent overuse of resources. Microsoft Graph is designed to handle a high volume of requests. If an overwhelming number of requests occurs, throttling helps maintain optimal performance and reliability of the Microsoft Graph service."
+localization_priority: Priority
+author: "nkramer"
+---
+
+# Throttling limits for Microsoft Teams APIs in Microsoft Graph
```
baywet: I do feel like we shouldn't make a Teams specific throttling limits page but rather a Microsoft Graph throttling limits broken down by  service. The reasoning being: hopefully at some point all services should provide information and guidance on that subject.
In that optic, we should probably move the existing limits with have from Outlook here as well.
https://review.docs.microsoft.com/en-us/graph/throttling#service-specific-limits

concepts/teams-throttling-limits.md
```diff
@@ -0,0 +1,87 @@
...
+Apps that violate these terms of use may be subject to additional throttling or suspended entirely.
+
+## General limits
+
+| Teams request type                                   | Value      |
+|-------------------------------------------|------------|
+| Graph API calls for an app/tenant         | 15000 requests/10 seconds |
+| GET request for resources (except messages) |60 rps/app/tenant, 600 rps/app/all tenants |
+| GET /teams/```{team-id}```|  30 rps |
+| POST/PUT/PATCH/DELETE /teams/```{team-id}```|  15 rps |
```
AkJo: There seems to be repetition of the POST across 41 and 42.
Also, we don't really have a PUT /teams/team-id at this point. Perhaps, we should remove the PUT reference.

AkJo: Do we not want make the distinction between rps/app/tenant vs rps/app/all tenants for these as well?

concepts/teams-throttling-limits.md
```diff
@@ -0,0 +1,87 @@
...
+| POST /teams/```{team-id}``` | 6 rps |
+
```
AkJo: We should add an additional note referring to the CS limitations, something along the lines of:

The above numbers apply to limits at the API level. However, our downstream dependencies place additional limits on the number of concurrent requests that can be made on a specific team resource. At any given time, a maximum of 4 requests per sec per app can be issued on any specific team or any resource contained within that specific team. For example, an app can issue 30 GET requests on 30 different team ids in one second; however, it can issue only 4 requests per sec when all the requests are a GET on the same team id or different channels within the same team id and so on.

nkramer:
replaced by https://github.com/microsoftgraph/microsoft-graph-docs/pull/7492/files

PR #7263, opened 2020-03-03
ChannelMessage.Read.All

nkramer:


bhartono:
@nkramer i verified that all of the api you are updating will check for channelmessages.read.all

PR #7289, opened 2020-03-05
Updating all /conditionalAccess/ paths to prefix with /identity/

elcarlosadrian:
We are moving the /conditionalAccess/ segment under /identity/


andrueastman:
@dkershaw10 @elcarlosadrian 
Ideally we shouldn't be editing the code snippets directly but I have gone through the changes and this looks good to me and I am okay with having this merged in. 
If there are any issues, the tooling will correct it when it runs the next time in a couple of days.

Lauragra:
@elcarlosadrian , what is the merge status of this PR? Thanks.

PR #7302, opened 2020-03-06
External ID for Online Meetings (beta)

ananmishr:


api-reference/beta/api/onlinemeeting-createorget.md
```diff
@@ -0,0 +1,195 @@
+---
+title: "onlineMeeting: createOrGet"
```
Lauragra: Please add an entry to the TOC for the new method and add this method to the Methods table in the onlineMeeting resource topic.

PR #7356, opened 2020-03-12
I added information about sending attachment via javascript sdk and a…

DamienDoumer:
I added information about sending attachment via javascript sdk and ameliorated the C# example which seamed not totally correct. I was working with the graph api recently and I followed this documentation "JS and C# examples" to send attachments via mail, but I had difficulties sending a mail with attachments. 
So, I added to this doc some code and a little tips which helped me overcome the problem.

### DIFF
--- a/api-reference/v1.0/api/user-sendmail.md
+++ b/api-reference/v1.0/api/user-sendmail.md
@@ -6,7 +6,7 @@ localization_priority: Priority
 ms.prod: "microsoft-identity-platform"
 doc_type: apiPageType
 ---
-
+ 
 # Send mail
 
 Namespace: microsoft.graph
@@ -188,6 +188,9 @@ HTTP/1.1 202 Accepted
 ##### Request 3
 
 The next example creates a message with a file attachment and sends the message.
+
+> [!NOTE]
+> Using the javascript sdk, the attachment's "contentByte" property should contain Base64 encoded data gotten from your file. For example, if you read a file into an ArrayBuffer, you will have to encode it to Base64 before sending the message.
 
 
 # [HTTP](#tab/http)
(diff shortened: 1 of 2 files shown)


angelgolfer-ms:
@darrelmiller , @andrueastman , can you please review this PR for correctness? Thanks.

darrelmiller:
@DamienDoumer Thank you for the PR, and sorry for the lack of clarity here.  Your note about base64 encoding the content value is a good addition.  However, the sample code added will be overwritten the next time that we generate the snippet.  If you could modify the PR to only include the note that would be great.  
We will investigate if it is possible to annotate our metadata to recognize base64 encoded strings so that we can generate more helpful code in the future.

DamienDoumer:
Hello @darrelmiller  I updated the sample to only include the note, I think it's ok now.

angelgolfer-ms:
> Hello @darrelmiller I updated the sample to only include the note, I think it's ok now.

Hi @DamienDoumer , thank you for your PR. It's a good reminder that some customers may still not be aware that contentBytes should be base64-encoded. 
That, and because this encoding requirement applies to using REST, the JavaScript SDK, and any other SDKs, instead of adding a note at the API-topic level (e.g., sendMail API), I reformatted one of the two mentions of this requirement in the [fileAttachment](https://docs.microsoft.com/en-us/graph/api/resources/fileattachment?view=graph-rest-1.0) resource topic for emphasis, for both v1 and beta. You can see it in PR #7471 .

If we chose to add the note at the API-topic level, we would have to add that to a number of method topics as well (e.g., [add attachment for event](https://docs.microsoft.com/en-us/graph/api/event-post-attachments?view=graph-rest-1.0&tabs=http#example-file-attachment), [add attachment for message](https://docs.microsoft.com/en-us/graph/api/message-post-attachments?view=graph-rest-1.0&tabs=http#example-file-attachment) and for message-derived types), each for v1 and beta. For maintenance reasons, this is not a preferred approach.

I'd like to thank you again for bringing this up to our attention. If you don't mind, please close this PR. We look forward to your contribution in the future! :-)

PR #7414, opened 2020-03-17
Indexing API updates

jasonjoh:
- Deprecation of `externalFile`
- `externalItem` now supports more properties in update
- Added note about customizing search results page to find results using Microsoft Search UI


Lauragra:
@jasonjoh , I don't see an update to the changelog. Do we need one?

jasonjoh:
Oh yeah, we should. I’ll add it.

PR #7431, opened 2020-03-19
remove discoverysettings in json object

kasuken:


nkramer:
@kasuken , can you elaborate on what problem you're solving with this PR? Thanks.

kasuken:
> @kasuken , can you elaborate on what problem you're solving with this PR? Thanks.

If you pass the discoverysttings to the graph api, you obtain an error.
So, I removed that portion of json from the example.

PR #7479, opened 2020-03-24
mgt-teams-channel-picker new component

vogtn:
INTERNAL

Updates documentation with new component `mgt-teams-channel-picker` 

PR: 
https://github.com/microsoftgraph/microsoft-graph-toolkit/pull/249


concepts/toolkit/components/teams-channel-picker.md
```diff
@@ -0,0 +1,94 @@
...
+
+The selected channel section of the component renders the channel of the user's choice. Below is the method to manually set a `selectedChannel` through `selectChannelById()`. 
+
+![mgt-teams-channel-picker](./images/selected-channel.png)
+
+- `selectChannelsById()` accepts an array of Microsoft graph [Channel.id](https://docs.microsoft.com/graph/api/resources/users?view=graph-rest-1.0) to find associated Team and Channel details for that selection. The control currently allows for one selection or Channel per component:
+
+    ```javascript
+    // channelId = Microsoft Graph Channel, unique id property
+    document.querySelector('mgt-teams-channel-picker').selectChannelsById.push([channelId]);
```
nmetulev: this is weird, should this be:

```suggestion
    document.querySelector('mgt-teams-channel-picker').selectChannelsById(channelId);
```

And since we only support a single channel, it should not be an array

concepts/toolkit/components/teams-channel-picker.md
```diff
@@ -0,0 +1,94 @@
...
+The following example shows the `mgt-teams-channel-picker` component. Start searching for a channel or Team to see the results render.
+
+<iframe src="https://mgt.dev/iframe.html?id=components-mgt-teams-channel-picker--teams-channel-picker&source=docs" height="450"></iframe>
+
+[Open this example in mgt.dev](https://mgt.dev/?path=/story/components-mgt-teams-channel-picker--teams-channel-picker&source=docs)
+
+
+## Selected Channel and Team
+
+The selected channel section of the component renders the channel of the user's choice. Below is the method to manually set a `selectedChannel` through `selectChannelById()`.
```
nmetulev: This section should also include how do I get the selected channel and team, and how to listen to the event

vogtn:
> You are missing a section on properties/attributes and a section on events

From the original proposal: 

> Proposed attributes:
> 
> - selection-mode - enum for single or multiple, default multiple
> - team-id - id of a team to filter channels (default is none)
> 
> Proposed properties:
> - selected - array of selected channels. Adding channels to the array programaticly should update 
>  the UI. If selectionMode is single then the array can have only one channel.
> 
> I agree the **teamId** attribute filter should be added this will make the section necessary.  I originally removed the section since there wouldn't be any accessible properties otherwise. My concern is that our `selected` is of type `ChannelPickerItemState[]` which probably isn't very accessible.  However we can return these through `selecteditem()`. 
> 
> Events:
> 
> - selection changed

**MY proposal:** 

Proposed attributes:

- teamid -  id of a team to filter channels (default is none)

Proposed properties:

- selected - change the following ( is currently selected-channel): 
https://github.com/microsoftgraph/microsoft-graph-toolkit/blob/nivogt/team-channel-picker/src/components/mgt-teams-channel-picker/mgt-teams-channel-picker.ts#L186
 into **selected**, a private property. 

- selectedChannel - (public property) which returns:
https://github.com/microsoftgraph/microsoft-graph-toolkit/blob/nivogt/team-channel-picker/src/components/mgt-teams-channel-picker/mgt-teams-channel-picker.ts#L173
when updated. 

And follow up in the **Methods** section to explain the use of `selectedItem()` to get data for the channel or team. 

Open to ideas.

concepts/toolkit/components/teams-channel-picker.md
vogtn: Here is the new description: 
https://github.com/microsoftgraph/microsoft-graph-docs/pull/7479/commits/2359841207b3cf8981e37909c3547317bc49f9c2#diff-8d518e63ed64f9696523d40c015f0f57R27-R46

concepts/toolkit/components/teams-channel-picker.md
vogtn: Updated here:
https://github.com/microsoftgraph/microsoft-graph-docs/pull/7479/commits/2359841207b3cf8981e37909c3547317bc49f9c2#diff-8d518e63ed64f9696523d40c015f0f57R27-R46

PR #7497, opened 2020-03-26
iCalUId update

Harini84:
As described in blog post:
https://developer.microsoft.com/en-us/office/blogs/microsoft-graph-calendar-events-icaluid-update/

iCalUId value has been updated to return the correct value i.e. value that is unique across instances as per industry standard.

As iCalUId has been updated, UId that is available only in MS-Graph beta and been a source of confusion is no longer needed:
https://stackoverflow.com/questions/51377799/the-property-icaluid-does-not-exist-on-type-microsoft-outlookservices-event?rq=1

Graph beta documentation updated for the above changes.

+ @sochowdh


Harini84:
@angelgolfer-ms 
Gentle reminder to review this PR.

Thanks,
Harini

api-reference/beta/resources/event.md
```diff
@@ -111,6 +111,7 @@ Here is a JSON representation of the resource
 |createdDateTime|DateTimeOffset|The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`|
 |end|[DateTimeTimeZone](datetimetimezone.md)|The date, time, and time zone that the event ends. By default, the end time is in UTC.|
 |hasAttachments|Boolean|Set to true if the event has attachments.|
+|iCalUId|String|A unique identifier, as defined in [RFC5545](https://tools.ietf.org/html/rfc5545#section-3.8.4.7), that is shared by all instances of an event across different calendars. Read-only.|
```
angelgolfer-ms: Line 114: According to your blog post, it seems like this part of the description for uid still holds for iCalUid:
"For recurring events, the value is the same for the series master and all of its occurrences including exceptions."
Is that correct?
If so, please add that sentence to the description for iCalUId. I think that would make it more explicit, and easier for existing customers to realize it's primarily still the same as uid from the past, just renamed as iCalUId.

Harini84: Thank you Angela for rephrasing the definition. I have modified the definition accordingly.

Harini84:
> Hi @Harini84 ,
> Do I understand correctly the purpose of this PR is to rename in beta, the uid property as iCalUId? The functionality of uid and iCalUId is basically the same?
> 
> If so, that's great.
> On that assumption, uid will no longer exist as a property of the event entity. There are more than 7 topics to update and rename instances of "uid" - you can do a search in VS Code. I think there are 9 more files (7 reference topics, 2 concept topics) that have the uid property in their examples that you need to rename please:
> eventmessage-get.md
> group-get-event.md
> group-list-calendarview.md
> group-list-events.md
> group-post-events.md
> group-update-event.md
> group-update-thread.md
> outlook-calendar-meeting-proposals.md
> outlook-calendar-online-meetings.md

I identified 7 files which I had updated initially. Have made updates to all the above files listed. 

Thanks

Harini84:
@angelgolfer-ms 
Changes for review comments have been checked-in. Please let me know if any further updates needed.

PR #7501, opened 2020-03-26
Required Request body for PATCH is incorrect

JankeSkanke:
Patch does not accept the values version, isGlobalScript and highestAvailableVersion


Lauragra:
@davidmu1 , can you follow up on getting this updated in the Intune docs please?

PR #7509, opened 2020-03-26
Mezzo3

davidmu1:


concepts/query-parameters.md
```diff
@@ -330,6 +334,41 @@ Content-type: application/json
 
 To learn more about the People API, see [Get information about relevant people](./people-example.md#search-people).  
 
+### Using $search on directory object collections
+
+You can use the `$search` query parameter to restrict results based on a search criterion such as looking for words in strings delimited by spaces, casing, and character types (numbers and special characters). The `$search` parameter is only available for the **displayName** property. For example:
```
dkershaw10, line 339: Last statement seems wrong.
displayName on user and group
description on group

Are there other resource types supported?  Can we be explicit?

concepts/query-parameters.md
```diff
@@ -120,7 +126,7 @@ GET https://graph.microsoft.com/v1.0/me/drive/root?$expand=children($select=id,n
 
 ## filter parameter
 
-Use the `$filter` query parameter to retrieve just a subset of a collection. 
+Use the `$filter` query parameter to retrieve just a subset of a collection. The `$filter` query parameter can also be used to retrieve for relationships like member, memberOf, transitiveMember, and transitiveMemberOf.
```
dkershaw10, line 129: That's a weird way of putting it - "filter can be used to retrieve for relationships"?

The `$filter` query parameter can also be for relationships like members, memberOf, transitiveMembers, and transitiveMemberOf, to return a subset of the relationships. For example, get all the security groups I'm a member of.

BTW it's members and transitiveMembers (not the the singular).

concepts/changelog.md
```diff
@@ -85,13 +85,13 @@ For details about known issues with Microsoft Graph APIs, see [Known issues](kno
 
 ### Identity and access (Azure AD)
```
dkershaw10, line 86: Nothing about the improved query capability for some AAD resources?  No summary?

api-reference/beta/api/user-list.md
```diff
@@ -375,6 +333,183 @@ Content-type: application/json
...
+The following is an example of the request.
+
+<!-- {
+  "blockType": "request",
+  "name": "get_to_count"
+}-->
+```msgraph-interactive
+GET https://graph.microsoft.com/beta/users?$search="displayName:wa" OR "displayName:to"&$orderbydisplayName&$count=true
+ConsistencyLevel: eventual
+
```
dkershaw10, line 474: You are missing the close ticks for the code block.
It means the example doesn't render properly.

api-reference/beta/api/user-list-transitivememberof.md
```diff
@@ -54,39 +52,110 @@ Do not supply a request body for this method.
...
+#### Request
+
+The following is an example of the request.
+
+<!-- {
+  "blockType": "request",
+  "name": "get_count_only"
+}-->
+```msgraph-interactive
+GET https://graph.microsoft.com/beta/users/{id}/transitiveMemberOf/$/microsoft.graph.group/$count
```
dkershaw10, line 152: We support a more simplified form of this without the $
`GET https://graph.microsoft.com/beta/users/{id}/transitiveMemberOf/microsoft.graph.group/$count`

api-reference/beta/api/group-list.md
```diff
@@ -46,23 +48,28 @@ values, as shown in the following example:
 ```
 GET https://graph.microsoft.com/beta/groups?$orderby=displayName
 ```
+You can also use the `$count` and `$search` query parameters to limit the response. You can use `$search` on the **displayName** property. When items are added or updated for this resource, they are specially indexed for use with the `$count` and `$search` query parameters. There can be a slight delay between when an item is added or updated and when it is available in the index.
```
dkershaw10, line 51: Is it just displayName or is it description too?  In the query topic, it mentions description in one of the examples.

concepts/query-parameters.md
```diff
@@ -90,7 +90,13 @@ GET  https://graph.microsoft.com/v1.0/me/contacts?$count=true
 [Try in Graph Explorer](https://developer.microsoft.com/graph/graph-explorer?request=me/contacts?$count=true&method=GET&version=v1.0)
 
 
->**Note:** `$count` is not supported for collections of resources that derive from [directoryObject](/graph/api/resources/directoryobject?view=graph-rest-1.0) like collections of [users](/graph/api/resources/user?view=graph-rest-1.0) or [groups](/graph/api/resources/group?view=graph-rest-1.0).
+The `$count` query parameter is supported for these collections of resources and their relationships that derive from [directoryObject](/graph/api/resources/directoryobject?view=graph-rest-beta):
+- [application](https://docs.microsoft.com/graph/api/resources/application?view=graph-rest-beta)
+- [contact](https://docs.microsoft.com/graph/api/resources/contact?view=graph-rest-beta)
```
dkershaw10, line 95: should be organizational contact (orgcontact)

api-reference/beta/api/group-list-members.md
```diff
@@ -92,20 +103,166 @@ Content-type: application/json
...
+#### Request
+
+The following is an example of the request.
+
+<!-- {
+  "blockType": "request",
+  "name": "get_a_count"
+}-->
+```msgraph-interactive
+GET https://graph.microsoft.com/beta/groups/{id}/members?$count=true&$orderby=displayName&$filter=startswith(displayName, 'a')
```
dkershaw10, line 237: This query did not work for me, until I removed $orderby parameter.

api-reference/beta/api/application-list.md
```diff
@@ -16,36 +16,47 @@ Namespace: microsoft.graph
...
 
 ## Request headers
-| Name       | Type | Description|
-|:-----------|:------|:----------|
-| Authorization  | string  | Bearer {token}. Required.  |
+
+| Name | Description |
+|:---- |:----------- |
+| Authorization  | Bearer {token}. Required.  |
+| ConsistencyLevel | The value is always `eventual`. This header is required when using the `$search` or `$count` query parameter. It uses an index that may not be up-to-date with recent changes to the object. |
```
dkershaw10, line 44: Nit - the values elsewhere for request headers are not in ticks...  Maybe this should just say:
| ConsistencyLevel |  eventual. This header is required when using the `$search` or `$count` query parameter. It uses an index that may not be up-to-date with recent changes to the object. |

concepts/query-parameters.md
```diff
@@ -159,6 +165,7 @@ The following table shows some examples that use the `$filter` query parameter.
...
 | Get all unread mail in the signed-in user's Inbox. | [`https://graph.microsoft.com/v1.0/me/mailFolders/inbox/messages?$filter=isRead eq false`](https://developer.microsoft.com/graph/graph-explorer?request=me/mailFolders/inbox/messages?$filter=isRead+eq+false&method=GET&version=v1.0) 
 | List all Office 365 groups in an organization. | [`https://graph.microsoft.com/v1.0/groups?$filter=groupTypes/any(c:c+eq+'Unified')`](https://developer.microsoft.com/graph/graph-explorer?request=groups?$filter=groupTypes/any(c:c+eq+'Unified')&method=GET&version=v1.0) 
+| Use OData cast to get transitive membership in groups with a display name that starts with 'a' including a count of returned objects. | [https://graph.microsoft.com/beta/me/transitiveMemberOf/microsoft.graph.group?$count=true&$orderby=displayName&$filter=startswith(displayName, 'a')](https://developer.microsoft.com/graph/graph-explorer?request=me/transitiveMemberOf/microsoft.graph.group?$count=true&$orderby=displayName&$filter=startswith(displayName, 'a')) |
```
dkershaw10, line 168: The query itself doesn't render properly.
And in any case your link is incorrect.  You need to add "&method=GET&version=v1.0" on the end of the query.  Look at the other examples.
Also please remove the $orderby - I'm not convinced that works.

concepts/query-parameters.md
```diff
@@ -330,6 +335,41 @@ Content-type: application/json
 
 To learn more about the People API, see [Get information about relevant people](./people-example.md#search-people).  
 
+### Using $search on directory object collections
+
+You can use the `$search` query parameter to restrict results based on a search criterion such as looking for words in strings delimited by spaces, casing, and character types (numbers and special characters). The tokenized search support works only on the displayName and description fields. Any field can be put in `$search`, fields other than **displayName** and **description** defaults to `$filter` behavior. For example:
```
dkershaw10, line 340: defaults to `$filter` behavior.
->
defaults to `$filter` startswith behavior.

concepts/query-parameters.md
```diff
@@ -330,6 +335,41 @@ Content-type: application/json
...
+ 
+This looks for all mail-enabled groups with display names that look like "OneVideo". The results are restricted based on a logical conjunction (an "AND") of the `$filter` and the entire query in the `$search`. The search text is tokenized based on casing, but matches are performed in a case-insensitive manner. For example, "OneVideo" would be split into two input tokens "one" and "video", but matches properties in insensitive to case. 
+ 
+ 
+The syntax of search follows these rules: 
+ 
+- Generic format: $search="clause1" [AND | OR]  "[clauseX]". 
+- Any number of clauses is supported. Parentheses for precedence is also supported. 
+- The syntax for each clause is <property>:<text to search>. 
+- The property name must be specified in clause. Any property that can be used in `$filter` can also be used inside `$search`. Depending on the property, the search behavior is either "search" or "prefix match" if search is not supported on the property.
```
dkershaw10, line 356: "prefix match" -> "startswith"
That's the OData term.

**IMPORTANT** Just discovered this. Mezzo supports $filter on its own (with the $count and the ConsistencyLevel) - with exact and startswith match. So what you might say - that's already there.  **However** Mezzo has indexed many more properties than directory core.  That means more user and group properties can now be filtered - ones that customers have been asking for.  So we'll need to figure out how to weave this in. Karthik is going to give me a list of properties tomorrow.

api-reference/beta/api/application-list.md
```diff
@@ -16,36 +16,47 @@ Namespace: microsoft.graph
...
 
 ## Request headers
-| Name       | Type | Description|
-|:-----------|:------|:----------|
-| Authorization  | string  | Bearer {token}. Required.  |
+
+| Name | Description |
+|:---- |:----------- |
+| Authorization  | Bearer {token}. Required.  |
+| ConsistencyLevel | eventual. This header is required when using the `$search` or `$count` query parameter. It uses an index that may not be up-to-date with recent changes to the object. |
```
Licantrop0, line 44: Can we put "eventual" in a separate column called "value"? (also "Bearer {token}" from the previous line will be in the same column).

This should be updated in all files.

davidmu1: To be consistent with the rest of the API topics, we'll leave it as is. Thank you for the feedback.

Licantrop0: Hi David, sorry I wasn't clear.
I would like all the tables regarding request headers to have 3 columns and look like this:

Name | Value | Description
-- | -- | --
Authorization | Bearer {token} | Required.
ConsistencyLevel | eventual | This header and $count are required   when using $search, or when using $filter with   the $orderby query parameter. It uses an   index that may not be up-to-date with recent changes to the object.

Moreover, the table headers are already inconsistent (sometime they are: "name", "value", sometimes "Name", "Description", sometimes "Header", "Value")

davidmu1: I understand what you were saying, but that type of format change would need to be done throughout the documentation and this PR is not meant for that. I also understand that there are some inconsistencies already and don't want to add to them.

Licantrop0: I see, thank you for the explanation.
I’lll update after this merge then.

api-reference/beta/api/device-list-memberof.md
```diff
@@ -33,12 +33,14 @@ One of the following permissions is required to call this API. To learn more, in
 GET /devices/{id}/memberOf
 ```
 ## Optional query parameters
-This method supports the [OData Query Parameters](/graph/query_parameters) to help customize the response.
+
+This method supports the [OData Query Parameters](https://developer.microsoft.com/graph/docs/concepts/query_parameters) to help customize the response including `$search`, `$count`, and `$filter`. OData cast is also enabled, for example, you can cast to get just the directoryRoles the user is a member of. You can use `$search` on the **displayName** property. When items are added or updated for this resource, they are specially indexed for use with the `$count` and `$search` query parameters. There can be a slight delay between when an item is added or updated and when it is available in the index.
```
Lauragra, line 37: Global: @davidmu1 , please update all files to use the link I added to the previous topic. We should use sentence casing and a relative link for the URL. Please also added a comma after response and before including.

concepts/changelog.md
```diff
@@ -70,6 +70,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
...
.../li><li>[serviceprincipal-list-transitivememberof](/graph/api/serviceprincipal-list-transitivememberof?view=graph-rest-beta)</li><li>[serviceprincipal-list](/graph/api/serviceprincipal-list?view=graph-rest-beta)</li><li>[user-list-memberof](/graph/api/user-list-memberof?view=graph-rest-beta)</li><li>[user-list-transitivememberof](/graph/api/user-list-transitivememberof?view=graph-rest-beta)</li><li>[user-list](/graph/api/user-list?view=graph-rest-beta)</li></ul><br>Added examples for using OData cast to: <ul><li>[group-list-memberof](/graph/api/group-list-memberof?view=graph-rest-beta)</li><li>[group-list-members](/graph/api/group-list-members?view=graph-rest-beta)</li><li>[group-list-transitivememberof](/graph/api/group-list-transitivememberof?view=graph-rest-beta)</li><li>[serviceprincipal-list-memberof](/graph/api/serviceprincipal-list-memberof?view=graph-rest-beta)</li><li>[serviceprincipal-list-transitivememberof](/graph/api/serviceprincipal-list-transitivememberof?view=graph-rest-beta)</li><li>[user-list-memberof](/graph/api/user-list-memberof?view=graph-rest-beta)</li><li>[user-list-transitivememberof](/graph/api/user-list-transitivememberof?view=graph-rest-beta)</li></ul> |
```
Lauragra, line 73: Please move the changelog entry to May. We include the entry in the month in which the docs are published so that they are surfaced in the most recent what's new.

davidmu1:
@Lauragra - I think I corrected what you pointed out. Thanks

concepts/changelog.md
```diff
@@ -73,6 +72,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
...
```
Lauragra, line 75: It looks like we're using the file names for the link text; can we update the link text to use the friendly names for the methods?

PR #7514, opened 2020-03-26
adding media  facets image/video in remoteItem resource.

hijai:
Graph recent API would return in response media facets image or video based on query filter, those properties would be part of remoteItem facet or resource. Therefore, updating response model represented as a resource.


Lauragra:
@hijai , I assume that the new properties aren't live yet so setting the *do not merge* label. When the properties are publicly available, please set *ready to merge*.

hijai:
> @hijai , I assume that the new properties aren't live yet so setting the _do not merge_ label. When the properties are publicly available, please set _ready to merge_.

Hi Laura, properties were already there, Documentation was not updated and there was no code on service side to perform expected action when those properties are being passed. I have added that piece of code and here updating now documentation.

PR #7526, opened 2020-03-27
Updated TOC in beta and content

tommoser:
Moved Information Protection in TOC for beta and content to live under Security. Nested the existing items in Security one level deeper so that Security Graph and Information Protection are the two children under Security.


api-reference/beta/toc.yml
```diff
@@ -10569,6 +10553,23 @@
           href: api/tiindicator-deletetiindicatorsbyexternalid.md
       - name: Errors
         href: resources/security-error-codes.md
+      name: Information protection
```
Lauragra: Can you please make sure that the yml file is valid? This change somehow has made the Security node disappear from the TOC.

api-reference/beta/toc.yml
```diff
@@ -9015,21 +9015,6 @@
...
-            href: api/informationprotectionlabel-get.md
-          - name: Evaluate application
-            href: api/informationprotectionlabel-evaluateapplication.md
-          - name: Evaluate classification result
-            href: api/informationprotectionlabel-evaluateclassificationresults.md
-          - name: Evaluate removal
-            href: api/informationprotectionlabel-evaluateremoval.md
-          - name: Extract label
-            href: api/informationprotectionlabel-extractlabel.md
         - name: Threat assessment
```
Lauragra, line 9018: Should we move Threat assessment as well, or just Labels? With this change, we now have Information Protection nodes under both Identity and Access and Security.

PR #7537, opened 2020-03-30
Profile beta updates

kevinbellinger:
WIP for multiple content updates and additions related to Profile


kevinbellinger:
> You can please replace this part of the domain "M365x214355" by something more generic like "contoso".
> Is "IrenaK" also an email name for an [approved fictitious name](https://msgo.azurewebsites.net/add/document/guidelines/fictitious-names.html), like "MeganB" for "Megan Bowen"?

Irena Koren is one of the ficticious names from the CELA site. I'll update to something generic for the emails too.

api-reference/beta/api/profile-list-account.md
```diff
@@ -108,17 +108,35 @@ HTTP/1.1 200 OK
 Content-type: application/json
 
 {
```
angelgolfer-ms, line 110: @kevinbellinger , the beta CSDL metadata states that the account navigation property is a collection, not a single resource:
<NavigationProperty Name="account" Type="Collection(graph.userAccountInformation)" ContainsTarget="true"/>

Your example update shows the GET /me/profile/account returns a single resource.

Is the example correct, in which case will the beta CSDL change to reflect the same?

api-reference/beta/api/profile-get.md
```diff
@@ -99,7 +99,464 @@ HTTP/1.1 200 OK
 Content-type: application/json
 
 {
-  "id": "id-value"
+    "id": "profileId",
```
angelgolfer-ms, line 102: Line 102: Why is the value of the id property a fixed string "profileId"?
I signed into Graph Explorer with a demo tenant, and GET /me/profile also returned an id property of that fixed string.

Is that by design, that the id property always returns a fixed string regardless of the actual user? 
If so, how can an ID serve the purpose if its value is the same for every user?

Assuming that's not by design:
FYI, in the docs, in general it's best to provide real-looking values in the example response.

api-reference/beta/api/profile-get.md
```diff
@@ -19,11 +19,11 @@ Retrieve the properties and relationships of a [profile](../resources/profile.md
...
-| Delegated (work or school account)     | Not supported.                              |
-| Delegated (personal Microsoft account) | Not supported.                              |
-| Application                            | Not supported.                              |
+| Permission type                        | Permissions (from least to most privileged)                                      |
+|:---------------------------------------|:---------------------------------------------------------------------------------|
+| Delegated (work or school account)     | User.Read, User.ReadWrite, User.ReadBasic.All, User.Read.All, User.ReadWrite.All |
+| Delegated (personal Microsoft account) | User.Read, User.ReadWrite, User.ReadBasic.All, User.Read.All, User.ReadWrite.All |
+| Application                            | User.ReadBasic.All, User.Read.All, User.ReadWrite.All                            |
 
 ## HTTP request
```
angelgolfer-ms, line 28: Line 33:
**GLOBAL** for all the operation topics in the profile API - if you specify User.Read.All or User.ReadWrite.All as supported permissions for that operation, doesn't that mean the operation supports other users in the organization as well, not only the signed-in user?
That's how these permissions are [documented](https://docs.microsoft.com/en-us/graph/permissions-reference#user-permissions).

If the operation supports other users as well, then for the "HTTP request" syntax section, you would need to have two lines, one for '/me', and the other for a specified user.

For example, for GET profile, you'd have:
GET /me/profile
GET /users/{id | userPrincipalName}/profile

You can look at other topics, for example, [GET contact](https://review.docs.microsoft.com/en-us/graph/api/contact-get?view=graph-rest-beta&branch=master&tabs=http#http-request) for the signed-in user or specified user.

api-reference/beta/resources/useraccountinformation.md
```diff
@@ -46,15 +54,41 @@ The following is a JSON representation of the resource.
...
+  "keyProperty": "id"
 }-->
 
 ```json
 {
-  "ageGroup": "String",
-  "countryCode": "String",
-  "preferredLanguageTag": {"@odata.type": "microsoft.graph.localeInfo"},
-  "userPrincipalName": "String"
+  "allowedAudiences": "organization",
```
angelgolfer-ms, line 63: **GLOBAL**
For the "JSON representation" section of each entity topic:
You actually had it closer to what it's supposed to be - each property in the entity has its type defined.

What you were missing before was, you need to include the properties from the base type "itemFacet" as well. 
To do that, you just copy the lines under the "JSON representation" section of itemfacet.md and add those lines to the "JSON representation" section of each derived type. Those lines are:

  "allowedAudiences": "string",
  "createdBy": {"@odata.type": "microsoft.graph.identitySet"},
  "createdDateTime": "String (timestamp)",
  "id": "String (identifier)",
  "inference": {"@odata.type": "microsoft.graph.inferenceData"},
  "lastModifiedBy": {"@odata.type": "microsoft.graph.identitySet"},
  "lastModifiedDateTime": "String (timestamp)"

Can you please do that for each of the topics for the resource types that are derived from itemFacet?

api-reference/beta/api/profile-list-anniversaries.md
```diff
@@ -109,8 +109,8 @@ Content-type: application/json
 {
   "value": [
     {
-      "type": "type-value",
-      "date": "datetime-value"
```
angelgolfer-ms, line 113: Does this operation return the properties that are inherited from the itemFacet resource as well?
If it does, then please include those properties in the example response.

api-reference/beta/api/profile-list-anniversaries.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "List anniversaries"
```
angelgolfer-ms, line 2: This comment is about the topic personanniversary.md which is not included in this PR:
The "Methods" table in that topic only lists 3 operations - the GET, update, and delete. 
Please add the "list" and "create" operations to that table as well.

api-reference/beta/resources/educationalactivity.md
```diff
@@ -58,7 +60,7 @@ The following is a JSON representation of the resource.
   "endMonthYear": "String (timestamp)",
   "institution": {"@odata.type": "microsoft.graph.institutionData"},
   "program": {"@odata.type": "microsoft.graph.educationalActivityDetail"},
-  "startMonthYear": "String (timestamp)"
```
angelgolfer-ms, line 61: Lines 59-63:
Just wanted to make a note for this entity topic as an example - the "JSON representation" section here looks almost there, you've got the property types defined (and not showing property values per se), the only thing you're missing is the inherited properties from itemFacet.

api-reference/beta/api/profile-list-educationalactivities.md
```diff
@@ -110,34 +109,34 @@ Content-type: application/json
 {
   "value": [
     {
```
angelgolfer-ms, line 111: This is one operation I could try on Graph Explorer, so I can definitively comment that this example payload is missing the properties inherited from itemFacet. Just pointing this out as an example.

Per a **GLOBAL** comment in another topic in this PR, please make sure to include the inherited properties as appropriate, for the other operation examples.

api-reference/beta/api/profile-post-interests.md
```diff
@@ -65,11 +65,11 @@ Content-type: application/json
 
 {
   "categories": [
-    "categories-value"
+    "Sports Teams"
```
angelgolfer-ms, line 68: Line 16 (at the beginning of the topic): There's a typo with the link, it's now malformed, please correct.

api-reference/beta/resources/useraccountinformation.md
```diff
@@ -25,12 +25,20 @@ Inherits from [itemFacet](itemfacet.md).
 
 ## Properties
```
angelgolfer-ms, line 26: This comment is not for useraccountinformation.md, but for useraccountinformation-get.md that is not included in this PR, and, a number of other "Get" topics as well in the profile API.

**GLOBAL**
Where you have the "HTTP request" or example request specified like the following:
"GET /user/profile/account"

Please change that to:
"GET /users/{id}/profile/account"

kevinbellinger:
closing to address branch issues. will open with all updates

api-reference/beta/resources/useraccountinformation.md
angelgolfer-ms: Just to be complete, the other line that should be in the "HTTP request" section should be for the /me user. Since we usually put the /me syntax line first, that section should look like this:

GET /me/profile/account
GET /users/{id}/profile/account

PR #7694, opened 2020-04-09
Update group-post-groups.md

yyuank:


api-reference/v1.0/api/group-post-groups.md
```diff
@@ -174,7 +174,7 @@ Content-type: application/json
 
 ### Example 2: Create a group with owners and members
 
-The following example creates an Office 365 group with an owner and members specified.
+The following example creates an Office 365 group with an owner and members specified. Note that a maximum of 20 relationships, such as owners and members, can be added as part of group creation. You can subsequently add more members by using the [add member](group-post-members.md) API or JSON batching.
```
Lauragra, line 177: Please add the same update to the beta version of the topic that shows the same example. Thanks.

PR #7787, opened 2020-04-16
Clarify JSON of the profile resource, & need to use respective GET to get individual descriptive properties of user.

angelgolfer-ms:


api-reference/beta/api/profile-get.md
```diff
@@ -13,7 +13,9 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Retrieve the properties and relationships of a [profile](../resources/profile.md) object for a given user.
+Retrieve the properties of a [profile](../resources/profile.md) object for a given user.
+
```
angelgolfer-ms, line 17: @kevinbellinger , @darrelmiller , which is the intended behavior of GET /users/id/profile - should it return just the id property, or should it return the id property and all its nav properties?

If it's the latter, then @kevinbellinger , can you please update the example response in this profile-get.md topic?

api-reference/beta/api/profile-get.md
```diff
@@ -13,7 +13,9 @@ Namespace: microsoft.graph
...
+The **profile** resource exposes various rich properties that are descriptive of the user as [relationships](../resources/profile.md#relationships), for example, anniversaries and education activities. To get one of these navigation properties, use the corresponding GET method on that property. See the [methods](../resources/profile.md) exposed by **profile**.
```
angelgolfer-ms, line 18: Related questions/comments, in the context of Graph Explorer:

@kevinbellinger , @bettirosengugi , can you please track disambiguating the label for an existing sample query in Graph Explorer "GET my profile" which is actually GET /me? 

Presumably, if @kevinbellinger plans to expose GET /me/profile in Graph Explorer (which currently returns error for the sample tenant), his sample query should be labelled as "GET my profile" instead, and not the existing GET /me query.

bettirosengugi: @angelgolfer-ms what will GET /me be pointing to moving forward? now that profile is moving to GET /me/profile?

api-reference/beta/api/profile-get.md
kevinbellinger: It's covered in the other PR - once we merge that content it will have a much richer example.

api-reference/beta/api/profile-get.md
kevinbellinger: I think we solved this as well. @bettirosengugi was going to point GET my profile at /me/profile going forward.

PR #7857, opened 2020-04-21
Update attributes to add presence badge.

beth-panx:
Add attributes documentation.

PR: https://github.com/microsoftgraph/microsoft-graph-toolkit/pull/399


beth-panx:
@nmetulev I needed to update this with the new attribute. So just a reminder it needs be re-reviewed ha.

PR #7954, opened 2020-04-27
[Universal Print] Adding printIdentity and printUserIdentity references

braedenp-msft:
Adding `printIdentity` and `printUserIdentity` types and correcting references in the allowedUsers/Groups APIs.

### DIFF
--- a/api-reference/beta/api/printer-list-allowedgroups.md
+++ b/api-reference/beta/api/printer-list-allowedgroups.md
@@ -43,7 +43,7 @@ This method supports some of the OData query parameters to help customize the re
 ## Request body
 Do not supply a request body for this method.
 ## Response
-If successful, this method returns a `200 OK` response code and a collection of [identity](../resources/identity.md) objects in the response body.
+If successful, this method returns a `200 OK` response code and a collection of [printIdentity](../resources/printidentity.md) objects in the response body.
 ## Example
 ##### Request
 The following is an example of the request.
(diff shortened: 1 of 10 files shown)


braedenp-msft:
> @braedenp-msft , thanks for making this fix.
> Please also fix printer.md and printershare.md - their Methods table has the incorrect types in the Return Type column, for the related 2 "list" methods.
> With those fixes, we'd be good to publish.
> Thanks!

Good catch, thanks!

Fixed.

braedenp-msft:
> @braedenp-msft , can you include a changelog entry with this update please? Thanks!

These updates are just to reflect what has existed in the API all along (the documentation was incorrect before).

Should I still create a changelog entry?

Lauragra:
> > @braedenp-msft , can you include a changelog entry with this update please? Thanks!
> 
> These updates are just to reflect what has existed in the API all along (the documentation was incorrect before).
> 
> Should I still create a changelog entry?

Yes, we still want a changelog entry to call users' attention to the new docs. It can go in the table in the April entry.

braedenp-msft:
> > > @braedenp-msft , can you include a changelog entry with this update please? Thanks!
> > 
> > 
> > These updates are just to reflect what has existed in the API all along (the documentation was incorrect before).
> > Should I still create a changelog entry?
> 
> Yes, we still want a changelog entry to call users' attention to the new docs. It can go in the table in the April entry.

Done!

PR #7982, opened 2020-04-29
Push send channel message to v1

nkramer:


api-reference/v1.0/resources/chatmessage.md
```diff
@@ -0,0 +1,111 @@
...
+Namespace: microsoft.graph
+
+Represents an individual chat message within a [channel](./channel.md) or [chat](/graph/api/resources/chat?view=graph-rest-beta). The chat message can be a root chat message or part of a reply thread that is defined by the **replyToId** property in the chat message.
+
+## Methods
+
+| Method       | Return Type  |Description|
+|:---------------|:--------|:----------|
+|**Channel messages**| | |
+|[List channel chatMessage](/graph/api/channel-list-messages?view=graph-rest-beta) | [chatMessage](#chatmessage-resource-type) collection |List of all root chat messages in a channel.  |
```
Lauragra, line 21: This is the v1.0 file but the links are pointing to beta APIs. Please updated.

Lauragra:
@nkramer , please see my comment. Also, please update the v1.0 TOC to expose the new methods. In beta, we are listing these methods under Channel but not under Chat message. Should they appear in both places? Please update both TOC files if needed. Thanks.

api-reference/v1.0/resources/chatmessage.md
nkramer: Okay, I think I've fixed all the erroneous mentions of "beta". There are two intentional references to beta from the v1.0 docs, since they help explain the full feature, but if you feel strongly that the v1.0 docs need to be freestanding please go ahead and delete those sentances.

nkramer:
@Lauragra, I've fixed the beta references & added TOC entries.

PR #7983, opened 2020-04-29
filesfolder page and permission

nkramer:


nkramer:
Dup of another completed PR

PR #7993, opened 2020-04-29
- adds link to ODSP best practices to avoid throttling

baywet:


jthake-msft:
that does feel a bit buried as a "see also" as apposed to having a SharePoint section and saying that it is referenced in the sharepoint site...

baywet:
@jthake-msft what do you think now?

PR #8023, opened 2020-05-01
[Universal Print] Adding $orderby support to Printer and PrintJob lists

braedenp-msft:


braedenp-msft:
The related functionality is now available. Publish at your convenience :)

PR #8025, opened 2020-05-01
[Universal Print] Improving document upload API

braedenp-msft:
* Refactoring our document upload API to be more consistent with REST semantics and Microsoft Graph precedent

**We'd like these changes to be published on May 18 ahead of a Build demo.**


braedenp-msft:
This work has been postponed and we'll make a new PR when we're ready to publish.

PR #8089, opened 2020-05-07
Updated auth landing page, new schema

CelesteDG:


CelesteDG:
@Lauragra - JFYI. This is the PR for the updated schema/design for the landing page that's under the Auth section of the docs. Changing it now before that old page breaks by the June deadline. Thank you.

PR #8112, opened 2020-05-07
Adding new clientAppTypes to Conditional Access API

gapalani:


Lauragra:
@gapalani , this looks like a change that warrants a changelog entry; can you please add?

PR #8142, opened 2020-05-11
get followed sites by target user

holylander:
Just tells how to query graph api to retrieve the followed sites by a target user


Lauragra:
@learafa , please review and approve.

PR #8163, opened 2020-05-12

bhartono:
mark teamsapp.* permission as "Deprecated. Do Not Use." since it will be replaced with TeamsAppInstallation.*.
teamsapp.* permission eventually will be repurposed for POST /appcatalogs/teamsapps


Lauragra:
@bhartono , when are we adding the Teams app installation permissions? Do you need to mention the deprecation in the changelog?

bhartono:
> @bhartono , when are we adding the Teams app installation permissions? Do you need to mention the deprecation in the changelog?

the code change is done. I will create a new section for it sometime next week. Please merge, we would like to prevent people from using these teamsapp.* permissions any longer.

PR #8204, opened 2020-05-14
Raise an error if the stub folders are still present in the branches.

GraphDocsGen:
As we move automatic docs and changelog generation to the main repo, we will need to make sure that the PR authors do not commit the stub folders. This change adds a check to make sure that they have been deleted and raises an error if they still exist.


GraphDocsGen:
Disregard PR. Created it using the wrong account

PR #8274, opened 2020-05-18
address accessreview.md missing enum value

markwahl-msft:
While complex types should be moved to their own article, this PR is to just fix a missing enum in reccurenceType reported by a customer.


markwahl-msft:
#sign-off

PR #8277, opened 2020-05-18
Update group-list-owners.md

yyuank:


api-reference/v1.0/api/group-list-owners.md
```diff
@@ -11,7 +11,7 @@ doc_type: apiPageType
 
 Namespace: microsoft.graph
 
-Retrieve a list of the group's owners. The owners are a set of non-admin users who are allowed to modify the group object. 
+Retrieve a list of the group's owners. The owners are a set of non-admin users or service principals who are allowed to modify the group object. Currently service principals are not listed as group owners due to staged roll-out of service principals on Graph V1.0 endpoint.
```
Lauragra, line 14: How long will the staged rollout take? If it's a short amount of time, do we need to add and then remove this note?

yyuank: It will be a couple months depending on the apps team. Yes, we will need to add and then remove this note.

PR #8309, opened 2020-05-19
"spsite" resourceVisualization type

holylander:
I have noticed that there is also another type "spsite", which shows recent site data. However, since data is sorted by modifiedDate by default, instead of lastAccessedDateTime, unless you add the sorting mentioned, you will probably retrieve 0 results.

This type is quite important in order to receive recently visited sites.


holylander:
Any advice on how to go thru the validation? Thanks

holylander:
I will rebuilt this PR

PR #8313, opened 2020-05-19
update built-in tabs doc

laujan:


laujan:
image

gclark01:
Why were the OneNote steps removed?  The note above says its depreciated, meaning you are not allowed to add OneNote Tabs via Graph anymore?

VesaJuvonen:
Have to ask as this came up with numerous partners and customers: Seems that we just removed huge amounts of capabilities from the tab provisioning - is this a mistake or intentional change? Hopefully a mistake as otherwise it's a breaking change for any existing solutions.

Thanks for providing insights on this change @laujan and/or @Lauragra.

Lauragra:
> Have to ask as this came up with numerous partners and customers: Seems that we just removed huge amounts of capabilities from the tab provisioning - is this a mistake or intentional change? Hopefully a mistake as otherwise it's a breaking change for any existing solutions, impacting potentially countless of partner and customer solutions.
> 
> If this is an intentional change, let's ensure that we have documented reasoning, message and replacement solution in place to avoid larger impact on the ecosystem for automating operations in these areas.
> 
> Thanks for providing insights on this change @laujan and/or @Lauragra.

@nkramer as well.

BrianTJackett:
Following. I'm mobile at the moment. @VesaJuvonen Was this a removal from documentation or also feature deprecation (with or without notice)?

VesaJuvonen:
Based on the updates provided in this PR, all advance tab manipulations are no longer supported with Microsoft Graph API v1 - which then sooner or later will impact customers and partners who implemented their solutions based on previously available documentation for v1.... Currently code still works, but getting questions from partners if they can continue using old models which were documented... as this change implies that they could get broken without any further notice as it's no longer documented.

Key here is to provide clarity and clear messaging. What's supported and what's not supported. Typically that would be the documentation which certainly can be extended, but moving any previously shared docs/supported patterns would be a breaking change.

PR #8352, opened 2020-05-20
introduce delegated EntitlementManagement.Read.All perm

markwahl-msft:
add EntitlementManagement.Read.All as a delegated permission alongside of the existing EntitlementManagement.ReadWrite.All


markwahl-msft:
#sign-off

PR #8372, opened 2020-05-21
Add Permissions column

sunainamishra:
Hi Nick, La!

Added Permissions column to the following resources:
- Team
- Chat
- App

Kindly provide your valuable review comments.

Thanks,
Sunaina

### DIFF
--- a/chat.md
+++ b/chat.md
@@ -1,0 +1,88 @@
+---
+title: "chat resource type"
+description: "A chat is a collection of chatMessages between one or more participants."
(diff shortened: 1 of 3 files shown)


Lauragra:
@sunainamishra , @nkramer , why are we adding a permissions column to the resource topics instead of documenting the permissions in the API topics? If we were to add a permissions column, we would do that in the existing files rather than adding new files. I recommend closing this PR. Thanks.

PR #8381, opened 2020-05-21
Update reportroot-concept-overview.md

JeremyEsposito:
NOTE: Microsoft Graph API is not enabled in GCC High and DoD tenants. Confirmed by SPO PG and PM's.

### DIFF
--- a/concepts/reportroot-concept-overview.md
+++ b/concepts/reportroot-concept-overview.md
@@ -8,6 +8,7 @@ ms.custom: scenarios:getting-started
 ---
 
 # Microsoft Graph reports API overview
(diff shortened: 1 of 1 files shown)


concepts/reportroot-concept-overview.md
```diff
@@ -8,6 +8,7 @@ ms.custom: scenarios:getting-started
 ---
 
 # Microsoft Graph reports API overview
+NOTE: Microsoft Graph API is not enabled in GCC High and DoD tenants.
```
Lauragra, line 11: @pranoychaudhuri , please verify. These terms should be spelled out ad the note should be formatted correctly.

PR #8486, opened 2020-05-29
- adds support for graph webhooks in us gov national deployment

baywet:


concepts/deployments.md
```diff
@@ -69,7 +69,7 @@ The following Microsoft Graph features are generally available on the `/v1.0` en
 | SharePoint| ✔ | ✔ | ✔ |
 | Planner|✔ |✔ |✔ |
 | Reports  |➖| ✔ |➖|
-| Change notifications (webhooks)  | ➖|✔* |✔* |
+| Change notifications (webhooks)  | ✔* |✔* |✔* |
```
baywet, line 72:
```suggestion
| Change notifications (webhooks)  | ✔ |✔ |✔* |
```

baywet:
@arpitha-dhanapathi thanks for double checking. MC does support subscriptions to AAD entities according to the team (they are the ones who suggested me to fix it in internal conversation actually). BF is the only one that doesn't support it for the reasons you know.

PR #8487, opened 2020-05-29
- fixes #8454: adds missing available resources for group subscriptions

### LINKED ISSUE #8454: A subscription is returned for an unsupported resource?
I mistakenly assumed I could subscribe to **/users/**_{id}_**/memberOf** but did not realize this is not supported (or is it???) until I read again that page. However I was still getting a registration id and no error.
Some other inconsistent requests result in an exception like subscribing to any _created_ event for **/users/**_{id}_ - although this is not document that error makes sense to me.
So why not returning an exception as well for unsupported subscription resources?

---
#### Document Details

* ID: 33d14c58-7163-a7b7-6f84-c075d0f6870d
* Version Independent ID: 3710d019-8a67-59cc-75b6-bcecc106180d
* Content: [Use the Microsoft Graph API to get change notifications - Microsoft Graph v1.0](https://docs.microsoft.com/en-us/graph/api/resources/webhooks?view=graph-rest-1.0)
* Content Source: [api-reference/v1.0/resources/webhooks.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/v1.0/resources/webhooks.md)
* Product: ****
* GitHub Login: @baywet

baywet:


ChrisDXC:
So just to be clear, what is the expectation after this fix? Will **users/**_{id}_**/memberOf** support subscriptions or will it return an error if one attempts to subscribe to it.

baywet:
Subscribing to users/{id}/memberOf will return an error stating that resource path is not supported.

PR #8489, opened 2020-05-29
Doc bug

keylimesoda:
User.ReadWrite.All doesn't support app only hard delete because it requires a scope check to ensure the calling user is in a role which is allowed to permanently delete users.

Group.ReadWrite.All doesn't support hard deletion of Groups.


psignoret:
Looking to @krbain (or delegate) to review.

dkershaw10:
@krbain This seems like a service bug, rather than a doc bug.  Why aren't those permissions sufficient to delete users and groups?

dancazacu1:
This is still an issue and it seems like a service bug. Any updates?

PR #8510, opened 2020-06-01
- adds syncStateNotFound error to error list

baywet:


concepts/errors.md
```diff
@@ -126,6 +126,7 @@ prepared to handle any one of these errors.
 | **resourceModified**      | The resource being updated has changed since the caller last read it, usually an eTag mismatch.
 | **resyncRequired**        | The delta token is no longer valid, and the app must reset the sync state.
 | **serviceNotAvailable**   | The service is not available. Try the request again after a delay. There may be a Retry-After header. 
+| **syncStateNotFound**     | The sync state generation is not found. The delta token is expired and a full re-synchronisation should be performed.
```
davidmu1, line 129: Might be better to say "The delta token has expired and must be synchronized again." The word "fully" isn't needed unless there is a difference between full synchronization and partial synchronization.

baywet: This is the error message that's returned by the API. I think we should keep it identical so it shows up quickly when people search for it.

baywet:
```suggestion
| **syncStateNotFound**     | The sync state generation is not found. The delta token is expired and data must be synchronized again. 
```

baywet: Actually scratch that, only the first part is the error message from the response. Updated. Thanks.

PR #8518, opened 2020-06-02
Update teams-create-group-and-team.md

JarbasHorst:
Replaced Office 365 group with Microsoft 365 group


Lauragra:
We'll take care of this as part of a global update of the Office 365 terminology.

PR #8534, opened 2020-06-03
Update licensing terms

nkramer:


concepts/teams-licenses.md
```diff
@@ -0,0 +1,24 @@
+---
+title: "Licenses for subscribing to chat messages"
+description: "Licenses for subscribing to chat messages"
+author: "nkramer"
+localization_priority: Priority
+ms.prod: "microsoft-teams"
+---
+
+# Licenses for subscribing to chat messages
```
Lauragra, line 9: How are we planning to surface this article? In the concept TOC, under Teamwork? https://docs.microsoft.com/en-us/graph/teams-concept-overview

concepts/teams-licenses.md
```diff
@@ -0,0 +1,22 @@
...
+
+Creating chat message subscriptions with application permissions for `/teams/allMessages` and `/chats/allMessages` will soon require one of the following licenses:
+
```
Lauragra, line 12: We're saying "will soon" here, but in the previous topic we say "Starting in August 2020". Does that mean that the licenses are required now? If so, we should change "will soon" to "requires".

nkramer: No, it's not checked in yet -- will happen soon, though.

nkramer: I've had a TOC link, but the main path to get to it is from the create subscription page.

concepts/teams-licenses.md
nkramer: I've added a TOC link, but the main path to get to it is from the create subscription page.

PR #8557, opened 2020-06-03
Update table-post-columns.md

pixieofhugs:
@zengin Known Issue: Http snippet should be fixed: Id should be string not int


Lauragra:
@pixieofhugs , this PR is failing validation. API Doctor is looking for the value to be of type Int instead of type string. Can you please verify the change?

pixieofhugs:
Feel free to reject. I'll note that this one is unresloved and get more info on it later

PR #8595, opened 2020-06-05
Updating documentation to indicate app-only support

michalee:
Support for Get group conversation, List conversations, and List conversation threads.


api-reference/beta/api/conversation-list-threads.md
```diff
@@ -23,7 +23,7 @@ One of the following permissions is required to call this API. To learn more, in
 |:--------------------|:---------------------------------------------------------|
 |Delegated (work or school account) | Group.ReadWrite.All, Group.Read.All    |
```
angelgolfer-ms, line 24: @michalee , for each of lines 24 and 26, can you please swap the order of the 2 permissions, since the column header reads "(from least to most privileged)"?

api-reference/v1.0/api/conversation-list-threads.md
```diff
@@ -22,7 +22,7 @@ One of the following permissions is required to call this API. To learn more, in
 |:--------------------|:---------------------------------------------------------|
 |Delegated (work or school account) | Group.ReadWrite.All, Group.Read.All    |
 |Delegated (personal Microsoft account) | Not supported.    |
-|Application | Not supported. |
+|Application | Group.ReadWrite.All, Group.Read.All |
```
angelgolfer-ms, line 25: Same with this topic - for lines 23 and 25, can you please swap the order of the 2 permissions?

PR #8606, opened 2020-06-05
Corrected application permission for App Proxy APIs

japere:
Updated from Directory.ReadWrite.All to Not supported.


japere:
@dkershaw10 - I've added edits to the docs to correct the permissions for Application access.

PR #8664, opened 2020-06-09
FileServicesDocs-Update-v1.0

learafa:
Adding docs for phase 3 publishing beta Apis to v1.0


api-reference/v1.0/resources/pendingcontentupdate.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "pendingOperations"
```
chackman, line 2: This file is a duplicate of `pendingOperations.md`; I presume this file should instead be `pendingContentUpdate.md` from the beta branch.

PR #8697, opened 2020-06-11
Add note not to use creationOptions

BrianTJackett:
Add note not to use resourceProvisioningOptions (may be deprecated in future) but instead use createdByAppId.

### DIFF
--- a/api-reference/beta/resources/group.md
+++ b/api-reference/beta/resources/group.md
@@ -107,6 +107,8 @@ This resource supports:
 
 ## Properties
 
+> **Note:** The `resourceProvisioningOptions` property should not be used to determine the app that created a group and it may be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group. 
+
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
 |allowExternalSenders|Boolean| Indicates if people external to the organization can send messages to the group. Default value is **false**. <br><br>Returned only on $select. |
--- a/api-reference/beta/resources/group.md
+++ b/api-reference/beta/resources/group.md
@@ -107,7 +107,7 @@ This resource supports:
 
 ## Properties
 
-> **Note:** The `resourceProvisioningOptions` property should not be used to determine the app that created a group and it may be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group. 
+> **Note:** The `creationOptions` property should not be used to determine the app that created a group and it will be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group.  In many cases the value of `createdByAppId` will be null as it was introduced in April 2020.
 
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
--- a/api-reference/beta/resources/group.md
+++ b/api-reference/beta/resources/group.md
@@ -107,8 +107,6 @@ This resource supports:
 
 ## Properties
 
-> **Note:** In many cases the value of **createdByAppId** will be null as it was introduced in April 2020.
-
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
 |allowExternalSenders|Boolean| Indicates if people external to the organization can send messages to the group. Default value is **false**. <br><br>Returned only on $select. |
@@ -116,7 +114,7 @@ This resource supports:
 |assignedLicenses|[assignedLicense](assignedlicense.md) collection|The licenses that are assigned to the group. <br><br>Returned only on $select. Read-only.|
 |autoSubscribeNewMembers|Boolean|Indicates if new members added to the group will be auto-subscribed to receive email notifications. You can set this property in a PATCH request for the group; do not set it in the initial POST request that creates the group. Default value is **false**. <br><br>Returned only on $select.|
 |classification|String|Describes a classification for the group (such as low, medium or high business impact). Valid values for this property are defined by creating a ClassificationList [setting](directorysetting.md) value, based on the [template definition](directorysettingtemplate.md).<br><br>Returned by default.|
-|createdByAppId|String|App ID of the app used to create the group. Can be null for some groups. <br><br>Returned by default. Read-only. Supports $filter.|
+|createdByAppId|String|App ID of the app used to create the group. Can be null for some groups as this property was introduced in April 2020.<br><br>Returned by default. Read-only. Supports $filter.|
 |createdDateTime|DateTimeOffset| Timestamp of when the group was created. The value cannot be modified and is automatically populated when the group is created. The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`. <br><br>Returned by default. Read-only. |
 |deletedDateTime|DateTimeOffset| For some Azure Active Directory objects (user, group, application), if the object is deleted, it is first logically deleted, and this property is updated with the date and time when the object was deleted. Otherwise this property is null. If the object is restored, this property is updated to null. |
 |description|String|An optional description for the group. <br><br>Returned by default.|
--- a/api-reference/beta/resources/group.md
+++ b/api-reference/beta/resources/group.md
@@ -107,7 +107,7 @@ This resource supports:
 
 ## Properties
 
-> **Note:** The **creationOptions** property should not be used to determine the app that created a group and it will be deprecated.  Instead use the **createdByAppId** property for the App ID of the app used to create the group.  In many cases the value of **createdByAppId** will be null as it was introduced in April 2020.
+> **Note:** In many cases the value of **createdByAppId** will be null as it was introduced in April 2020.
 
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
--- a/api-reference/beta/resources/group.md
+++ b/api-reference/beta/resources/group.md
@@ -107,7 +107,7 @@ This resource supports:
 
 ## Properties
 
-> **Note:** The `creationOptions` property should not be used to determine the app that created a group and it will be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group.  In many cases the value of `createdByAppId` will be null as it was introduced in April 2020.
+> **Note:** The **creationOptions** property should not be used to determine the app that created a group and it will be deprecated.  Instead use the **createdByAppId** property for the App ID of the app used to create the group.  In many cases the value of **createdByAppId** will be null as it was introduced in April 2020.
 
 | Property	   | Type	|Description|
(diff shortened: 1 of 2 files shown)


api-reference/beta/resources/group.md
```diff
@@ -107,6 +107,8 @@ This resource supports:
 
 ## Properties
 
+> **Note:** The `resourceProvisioningOptions` property should not be used to determine the app that created a group and it may be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group.
```
dkershaw10, line 110: **Please fix this**
resourceProvisioningOptions is **NOT** going to be deprecated

You're not should probably state that in many cases this property will be null, as this property was only introduced in April 2020

You need to add (to the v1.0 docs) that **creationOptions** should not be used and will be deprecated in the future.

BrianTJackett: @dkershaw10 Thank you for the correction.  I was pulling details from the internal thread on this topic and referenced wrong property name.  I will make suggested edits for beta doc and addition on v1.0 doc.

yyuank: Confirming @dkershaw10 is correct. creationOptions is the property that will be deprecated in the future, not resourceProvisiningOptions

api-reference/beta/resources/group.md
```diff
@@ -107,6 +107,8 @@ This resource supports:
 
 ## Properties
 
+> **Note:** The `creationOptions` property should not be used to determine the app that created a group and it will be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group.  In many cases the value of `createdByAppId` will be null as it was introduced in April 2020.
+
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
 |allowExternalSenders|Boolean| Indicates if people external to the organization can send messages to the group. Default value is **false**. <br><br>Returned only on $select. |
```
dkershaw10, line 114: 1. Don't know if creationOptions is in /beta.
2. You should add a line item for the creationOptions property in the properties table.  It needs a note - something like "Do not use.  This property will be deprecated.  Use **resourceBehaviorOptions** instead."

api-reference/beta/resources/group.md
```diff
@@ -107,6 +107,8 @@ This resource supports:
 
 ## Properties
 
+> **Note:** The `creationOptions` property should not be used to determine the app that created a group and it will be deprecated at a future date.  Instead use the `createdByAppId` property for the App ID of the app used to create the group.  In many cases the value of `createdByAppId` will be null as it was introduced in April 2020.
```
dkershaw10, line 110: I think that the style guide says that properties should be in bold (not code ticks).

BrianTJackett: Yes, it does look like properties should be in bold in the [developer style guide](https://docs.microsoft.com/en-us/style-guide/developer-content/formatting-developer-text-elements).  I will update accordingly.

api-reference/beta/resources/group.md
BrianTJackett: I see creationOptions in /v1.0, but I do not see it in /beta any longer.  Will add to the /v1.0 properties table.  Do you want me to remove the mention of creationOptions on the beta doc?

api-reference/beta/resources/group.md
```diff
@@ -107,6 +107,8 @@ This resource supports:
 
 ## Properties
 
+> **Note:** The **creationOptions** property should not be used to determine the app that created a group and it will be deprecated.  Instead use the **createdByAppId** property for the App ID of the app used to create the group.  In many cases the value of **createdByAppId** will be null as it was introduced in April 2020.
```
angelgolfer-ms, line 110: It looks like creationOptions is not part of the group entity in v1 nor beta?
If that's the case, please don't add this note about not using creationOptions. 
If you still think the remaining for createdByAppId is worth calling out, please add that to the description column of the property. You can use a new paragraph within that description for clarity and readability.

api-reference/beta/resources/group.md
angelgolfer-ms: >> Do you want me to remove the mention of creationOptions on the beta doc?
Yes please remove such mentions.

api-reference/v1.0/resources/group.md
```diff
@@ -107,13 +107,17 @@ This resource supports:
 |[resetUnseenCount](../api/group-resetunseencount.md)|None|Reset the unseenCount to 0 of all the posts that the signed-in user has not seen since their last visit. Supported for only Office 365 groups.|
 
 ## Properties
+
+> **Note:** The **creationOptions** property should not be used to determine the app that created a group and it will be deprecated at a future date.
```
angelgolfer-ms, line 111: I searched in the v1 metadata and this property is not part of the group entity. Please do not mention it nor add it to the properties table.

BrianTJackett: @angelgolfer-ms I still see **creationOptions** on the v1.0 endpoint.  See below screenshot from Graph Explorer.   I do not see that property on the beta endpoint though.

image

dkershaw10: @angelgolfer-ms @BrianTJackett the Group entity type is marked as an open type (to support schema and open extensions).  The side effect is that if the master workload exposes any additional properties that are not in the public schema, those will show up.  This is by design for open types.

Unfortunately, a developer, without API review sign-off, or any approval, managed to sneak this property into the Directory Services schema, and lo and behold it shows up.  I want it removed, but it's been around for a while.  So we need to document that it shouldn't be used - and that's what Brian is doing (thanks Brian).

api-reference/beta/resources/group.md
dkershaw10: @BrianTJackett - remove this and I'll sign off.

BrianTJackett: I condensed the note about "null before April 2020" into the createdByAppId row in the properties table.

api-reference/v1.0/resources/group.md
angelgolfer-ms: I see, thanks for the context.
Has it been a common issue that customers use the creationOptions in v1 inappropriately?
Depending on @BrianTJackett 's confirmation:
(a) If it's a _hot_ issue and we want to call that out for customers, then please add the note to the top section, just before the "Permissions" section title (most visibility).
(b) If it's _not_ a hot issue, I'd suggest just keeping the info to the description column of the creationOptions property.

Whether to call an issue out as a note should take into consideration whether a lot of customers would run into the issue, whether there are few or many properties, the length of the topic,... and whether the overall topic is something a customer can reasonably scan and take in. 

I have one more question about this subject for v1 - the description for createdOptions says:
"Do not use. This property will be deprecated. Use resourceBehaviorOptions instead."

Customers would be lost if they don't find resourceBehaviorOptions - what is resourceBehaviorOptions? Is it in the metadata, is it in the docs?






 This is because the group topic is so huge, whether to call issues out should take into consideration is a matter of relativity
(b) If it's worth calling this out separately but it's not a hot pain point, please add the content after the "Properties" table. You can add it as an appropriately titled H3 section for discoverability and uniformity, as there's another H3 after the Properties table, for a similar purpose to qualify some other properties. This is preferred to (c).
(c) If this is not a common pain point but you do want to call it out as a note, then as an alternative to (b), please add the note after the H3 section [Group visibility options](https://review.docs.microsoft.com/en-us/graph/api/resources/group?view=graph-rest-1.0&branch=pr-en-us-8697#group-visibility-options).
(d) If we're adding this for clarity as the property should not be used, then I think just qualifying that in the creationOptions description is sufficient.

The reason we shouldn't add the note before the Properties table is for consistency. First, no other Graph topic does that, second, a consistent look and feel in a reference topic lets customers find the info where they expect it, and not to have to wade through unexpected sections of extra info. So FYI, in general, please add any "rich information" about an entity to the top section before the first H2 section; this has been our long standing guidance. We try to limit the top rich info to half a page (anything needing more should go to appropriately titled concept topics). 

On a related note, FYI, we include H3's under the H2 "Properties" section, to give the customer a chance to read about specific properties first, and if there's more to add, such as info involving multiple properties, or lengthy info about a property that's too long to fit in a table cell, that goes into H3 sections.

dkershaw10: Agree - we should remove the note.
resourceBehaviorOptions is in /beta.  Not sure if it's in v1 or when it will get there. @yyuank

api-reference/v1.0/resources/group.md
```diff
@@ -107,13 +107,17 @@ This resource supports:
...
 |autoSubscribeNewMembers|Boolean|Indicates if new members added to the group will be auto-subscribed to receive email notifications. You can set this property in a PATCH request for the group; do not set it in the initial POST request that creates the group. Default value is **false**. <br><br>Returned only on $select.|
 |classification|String|Describes a classification for the group (such as low, medium or high business impact). Valid values for this property are defined by creating a ClassificationList [setting](groupsetting.md) value, based on the [template definition](groupsettingtemplate.md).<br><br>Returned by default.|
 |createdDateTime|DateTimeOffset| Timestamp of when the group was created. The value cannot be modified and is automatically populated when the group is created. The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`. <br><br>Returned by default. Read-only. |
+|creationOptions|String|Do not use. This property will be deprecated. Use **resourceBehaviorOptions** instead.|
```
angelgolfer-ms, line 120: Please provide more context what **resourceBehaviorOptions** is. Is it a property of group? As long as we (Microsoft) is defining that property, that it should be documented somewhere.
Customers/I don't see it documented though... :-(

api-reference/v1.0/resources/group.md
BrianTJackett: @angelgolfer-ms Good points on style and consistency regarding.

I don't have a large pool of customer / partner feedback if this is a _hot_ issue, but I have seen or heard from at least 5-10+ people who have blogged or tweeted out using creationOptions inappropriately.

I will need to look into resourceBehaviorOptions if it is included in v1 unless @yyuank can confirm. Holding off any further edits until confirm this.

BrianTJackett: I got delayed with other work, taking a look at this again.  I do see resourceBehaviorOptions in both v1.0 and beta.

@dkershaw10 or @yyuank is it safe to copy the description of resourceBehaviorOptions from the beta doc and add to v1.0 doc?  I'm can't verify if the functionality is the same or not.

api-reference/v1.0/resources/group.md
BrianTJackett: resourceBehaviorOptions is only documented in the beta doc at this point.  In separate comment asking if ok to copy description from beta doc to the v1.0 doc.

PR #8699, opened 2020-06-11
add entitlement management settings singleton

markwahl-msft:
Add entitlementManagement/settings to beta.


api-reference/beta/resources/entitlementmanagementsettings.md
```diff
@@ -0,0 +1,65 @@
...
+
+  ],
+  "@odata.type": "microsoft.graph.entitlementManagementSettings",
+  "baseType": "",
+  "keyProperty": ""
+}-->
+
+```json
+{
+  "externalUserLifecycleAction": "None",
```
Lauragra, line 52: We don't use example values in the JSON representation; this should just show the type, i.e. Int.

markwahl-msft: Changed the first property from the example to just String.  I tried using the number type for the second property but the apidoctor couldn't validate it, so left as is.

api-reference/beta/api/entitlementmanagementsettings-get.md
```diff
@@ -0,0 +1,101 @@
...
+
+<!-- { "blockType": "ignored" } -->
+
+```http
+GET /identityGovernance/entitlementManagement/settings
+```
+
+## Optional query parameters
+
+None.
```
Lauragra, line 38: To be clear, this should use the boilerplate text: "This method does not support the OData query parameters to customize the response." Assuming that's what you mean by none - that query parameters aren't supported?

api-reference/beta/api/entitlementmanagementsettings-get.md
```diff
@@ -0,0 +1,101 @@
...
+
+If successful, this method returns a `200 OK` response code and the requested [entitlementManagementSettings](../resources/entitlementManagementSettings.md) object in the response body.
+
+## Examples
+
+### Request
+
+The following is an example of the request.
+
+# [HTTP](#tab/http)
```
Lauragra, line 60: Remove this - it's duplicative because the only example is HTTP. This will get added via the snippet generation process.

api-reference/beta/api/entitlementmanagementsettings-update.md
```diff
@@ -0,0 +1,87 @@
...
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+PATCH /identityGovernance/entitlementManagement/settings
+```
+## Request headers
+| Name         | Description |
+|:-------------|:------------|
+| Authorization | Bearer \{token\}. Required. |
```
Lauragra, line 36: Add the Content-type header to the table.

api-reference/beta/api/entitlementmanagementsettings-get.md
markwahl-msft: Updated to the boilerplate text.

api-reference/beta/api/entitlementmanagementsettings-get.md
markwahl-msft: Removed

api-reference/beta/api/entitlementmanagementsettings-update.md
markwahl-msft: Added.

PR #8717, opened 2020-06-12
Add v1.0 documentation for Azure AD named locations and security defaults API

abhishekkp09:


api-reference/v1.0/toc.yml
```diff
@@ -3773,6 +3815,13 @@ items:
             href: api/tokenlifetimepolicy-delete.md
           - name: List applied to applications
             href: api/tokenlifetimepolicy-list-appliesto.md
+        - name: Security defaults policy
```
davidmu1, line 3818: What is @dkershaw10 thinking about the location of this set of APIs? There has been some discussion about having all policies listed under a Policies node.

abhishekkp09: It looks like they are already listed under a Policies node.

videor:
@Lauragra ready to merge.

api-reference/v1.0/api/identitysecuritydefaultsenforcementpolicy-get.md
```diff
@@ -0,0 +1,100 @@
+---
```
Lauragra, line 1: This and the following API topic are not included in the TOC. Please add them to the appropriate location.

abhishekkp09:
> Please make sure that all method topics are exposed in the v1.0 TOC. Thanks.

Sorry I had added it at first but it got removed during rebase. Added it back.

PR #8724, opened 2020-06-14
Permissions for type Application changed to 'Sites.Manage.All'

nmboris:


Lauragra:
@JeremyKelley , please review and approve. If the change is correct, we need to apply it to the beta version of the topic as well.

JeremyKelley (approved):
Catching up on old PRs, this looks fine to me.

PR #8797, opened 2020-06-19
Clarify how to retrieve a call record id

stephenjust:
Because there is no API currently available to list callRecords, we need to clarify how customers would get the id required for reading a call record


api-reference/beta/api/callrecords-callrecord-get.md
```diff
@@ -15,6 +15,11 @@ Namespace: microsoft.graph.callRecords
 
 Retrieve the properties and relationships of a [callRecord](../resources/callrecords-callrecord.md) object.
 
+There are two options to retrive the **id** value required for reading a **callRecord**:
+
+* Create a [webhook subscription](/graph/api/resources/webhooks?view=graph-rest-beta) to the **communications/callRecords** resource
```
angelgolfer-ms, line 20: Line 20:
Can you please rephrase it to say:
```
Subscribe to change notifications to the `/communications/callRecords` endpoint.
```
Please link "change notifications" to the same topic you have it in line 20, and end the bullet point with a period.

But just so I understand, doesn't that mean, unless a particular callRecord has changed, the app won't get that callRecord included in a change notification, and thus won't get the id value, right?

api-reference/beta/api/callrecords-callrecord-get.md
```diff
@@ -15,6 +15,11 @@ Namespace: microsoft.graph.callRecords
...
+* Read the **callChainId** property of a [call](../resources/call.md) (note that the record will not be available until the associated call is complete)
```
angelgolfer-ms, line 21: Line 21: 
```
Use the **callChainId** property of a [call](../resources/call.md). The call record is available only after the associated call is completed.
```

api-reference/beta/api/callrecords-callrecord-get.md
```diff
@@ -15,6 +15,11 @@ Namespace: microsoft.graph.callRecords
 
 Retrieve the properties and relationships of a [callRecord](../resources/callrecords-callrecord.md) object.
 
+There are two options to retrive the **id** value required for reading a **callRecord**:
```
angelgolfer-ms, line 18: Line 18:
"There are two ways to get the id of a callRecord:"
Per [Microsoft Voice](https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28361), we should use a relaxed tone and every day words :-)

api-reference/v1.0/api/callrecords-callrecord-get.md
```diff
@@ -13,6 +13,11 @@ Namespace: microsoft.graph.callRecords
 
 Retrieve the properties and relationships of a [callRecord](../resources/callrecords-callrecord.md) object.
 
+There are two options to retrive the **id** value required for reading a **callRecord**:
```
angelgolfer-ms, line 16: Please apply the above comments to the v1 topic as well. Thanks!

api-reference/beta/api/callrecords-callrecord-get.md
stephenjust: I've made the suggested changes.

Yes, your understanding is correct. Since our "List" API is not yet available, customers can only determine a call record id either by programmatically creating the call via the calling APIs, or by subscribing to change notifications and listening for ids. It's definitely not ideal, but this was the way we had to sequence things to get the data to customers faster.

api-reference/v1.0/api/callrecords-callrecord-get.md
stephenjust: Done, thanks!

PR #8868, opened 2020-06-25
Update connectorgroup.md

japere:
Updated description that this property relates to a preview feature.

### DIFF
--- a/api-reference/beta/resources/connectorgroup.md
+++ b/api-reference/beta/resources/connectorgroup.md
@@ -38,7 +38,7 @@ After a connector group is created, you can add or move connectors to the connec
 |id|string| Unique identifier for this connectorGroup. Read-only. |
 |isDefault|boolean| Indicates if the connectorGroup is the default connectorGroup. Only a single connector group can be the default connectorGroup and this is pre-set by the system. Read-only. |
 |name|string| The name associated with the connectorGroup. |
-|region|string| The region the connectorGroup is assigned to and will optimize traffic for. This region can only be set if **no connectors or applications** are assigned to the connectorGroup. The regions available include: North America, Europe, Australia, Asia, and India. Possible values are: `nam`, `eur`, `aus`, `asia`, `ind`.|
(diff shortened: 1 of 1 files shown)


api-reference/beta/resources/connectorgroup.md
```diff
@@ -38,7 +38,7 @@ After a connector group is created, you can add or move connectors to the connec
 |id|string| Unique identifier for this connectorGroup. Read-only. |
 |isDefault|boolean| Indicates if the connectorGroup is the default connectorGroup. Only a single connector group can be the default connectorGroup and this is pre-set by the system. Read-only. |
 |name|string| The name associated with the connectorGroup. |
-|region|string| The region the connectorGroup is assigned to and will optimize traffic for. This region can only be set if **no connectors or applications** are assigned to the connectorGroup. The regions available include: North America, Europe, Australia, Asia, and India. Possible values are: `nam`, `eur`, `aus`, `asia`, `ind`.|
+|region|string| The region the connectorGroup is assigned to and will optimize traffic for. This region can only be set if **no connectors or applications** are assigned to the connectorGroup. The regions available include: North America, Europe, Australia, Asia, and India. Possible values are: `nam`, `eur`, `aus`, `asia`, `ind`. This functionality is currently in preview. |
```
Lauragra, line 41: This is a beta topic, so how is having this property in preview different than this being a preview API?

japere: Unlike the other properties that are attached to GA features in the product this setting relates to something still in private preview and has limited functionality. If this is already expected since it is in beta and not necessary, I can remove adding this note it.

Lauragra: By private preview, do you mean that this property should not be publicly visible and you're using exposure control? If not, and it's expected that the public can view and use this property, then you don't need the preview note.

japere: Yes, it's not public yet and we only have a small set of customers testing it out and providing feedback, So this note was more so to set expectations on it.

Lauragra: I'm not sure I follow because the property does appear in the metadata. If it's not public, we wouldn't include it in the public docs. Is there some functionality that is in private preview?

japere: So this specific property that customers can use to assign the "region" property is in private preview right now (not officially released yet). Is the right approach to hide it from docs until we publicly release it?

PR #8908, opened 2020-06-29
- adds clarification for the term tenant

baywet:


concepts/throttling.md
```diff
@@ -80,6 +80,9 @@ Microsoft Graph allows you to access data in [multiple services](overview-major-
 > [!NOTE]
 > The specific limits described here are subject to change.
 
+> [!NOTE]
+> In the current section the term "tenant" refers to the Microsoft 365 organization where the application is installed. This tenant can be the same as the the one where the application was created in the case of a single tenant application, or it can be different in the case of a [multi-tenant application](/azure/active-directory/develop/setup-multi-tenant-app).
```
davidmu1, line 84: Not sure that 'current' is appropriate here. I think "In this section..." might be better.

baywet:
```suggestion
> In this section the term "tenant" refers to the Microsoft 365 organization where the application is installed. This tenant can be the same as the the one where the application was created in the case of a single tenant application, or it can be different in the case of a [multi-tenant application](/azure/active-directory/develop/setup-multi-tenant-app).
```

PR #8952, opened 2020-07-02
- removes changenotification sequencenumber that was introduced by error

baywet:
@angelgolfer-ms after some more internal reviews it looks like I've accidently introduced a property that never existed at the first place (POC that never went public). I'm doing the work to get this removed from schema ASAP.
I've already reviewed with @dkershaw10 the fact that this property never existed for customers and that it's ok to "discretely" remove it.


angelgolfer-ms:
> Please leave out updating the what's new, otherwise all looks good to me, thanks!

After doing so, this PR is good to be published, IMHO.

baywet:
@angelgolfer-ms reverted! thanks for the quick review!

PR #8955, opened 2020-07-02
[Universal Print] Adding new permissions to the permissions reference and changelog

braedenp-msft:
* Printer.ReadWrite.All
* PrintJob.ReadBasic.All
* PrintJob.Read.All
* PrintJob.ReadWriteBasic.All
* PrintJob.ReadWrite.All


concepts/permissions-reference.md
```diff
@@ -1856,6 +1856,20 @@ Threat assessment permissions are valid only on work or school accounts.
 
 ---
 
+## Universal Print permissions
```
angelgolfer-ms, line 1859: Line 1859:
I suggest using "Cloud printing" as the H2 section title.
That maps to the label we use in the TOC, to facilitate customer recognition.
It also starts with a 'c' that bubbles your permission higher up in the long right-side TOC in the permissions-reference.md topic.

PR #8956, opened 2020-07-02
Restructure API TOCs

davidmu1:


davidmu1:
@Lauragra  and @dkershaw10 - I would like to go ahead and publish the restructuring of the Applications and Identity and access nodes, but I have a question about Information protection. I thought there were discussions and a PR to move this set of APIs out of the Identity and access space. Did I imagine that? or was there a decision to not do it? Thanks, David

Lauragra:
> @Lauragra and @dkershaw10 - I would like to go ahead and publish the restructuring of the Applications and Identity and access nodes, but I have a question about Information protection. I thought there were discussions and a PR to move this set of APIs out of the Identity and access space. Did I imagine that? or was there a decision to not do it? Thanks, David

Yes, we should move Information protection into the Security node. I'm not sure who owns that content because we have Preeti listed as the owner in the topics. As part of your restructuring, can you also move that content into Security, in beta and v1.0? Thanks.

PR #8967, opened 2020-07-04
Update opentypeextension-get.md

gavinbarron:
The current state of the example documentation is causing confusion and developer issues when trying to use the `$expand=extensions($filter=<filter_expression>)` syntax

Adding a note to call out the expectation that the equals symbol needs to be encoded and reference to the specific error message given in this instance to allow for ease for searchability when trying to resolve these issues.

Ref: https://stackoverflow.com/questions/62686415/receiving-400s-and-500s-when-attempting-to-get-singlevalueextendedproperties
https://stackoverflow.com/questions/62475512/openextensions-find-and-retrieve-events-with-a-given-extension

### DIFF
--- a/api-reference/v1.0/api/opentypeextension-get.md
+++ b/api-reference/v1.0/api/opentypeextension-get.md
@@ -75,7 +75,8 @@ GET /groups/{Id}/threads/{Id}/posts/{Id}?$expand=extensions($filter=id eq '{exte
 GET /users/{Id|userPrincipalName}/messages/{Id}?$expand=extensions($filter=id eq '{extensionId}')
 GET /users/{Id|userPrincipalName}/contacts/{Id}?$expand=extensions($filter=id eq '{extensionId}')
 ```
-
+>**Note:** that equals signs inside the $expand extensions expression need to be uri encoded, thus sections like `$expand=extensions($filter=id eq '{extensionId}')` should be transmitted as `$expand=extensions($filter%3Did eq '{extensionId}')`  
+Failing to correctly encode the = symbol will result in an HTTP 400 with the message: Parsing OData Select and Expand failed: Found an unbalanced bracket expression.
 
 For the device, group, organization, and user resource types, you must also use a `$select` parameter to include
 the **id** property and any other properties you want from the resource instance:


angelgolfer-ms:
@gavinbarron , thank you for your contribution!
@svpsiva , can you confirm if the reported behavior is by design or if it is a bug to be fixed please?
If it's the latter, we'll need to follow up on the fix, and the appropriate docs surfaced in a different way, linking to its description in the known issues topic. Thanks.

angelgolfer-ms:
@gavinbarron , thank you for your patience!
FYI I was curious why that second '=' in a query like the following needs encoding but not the subsequent open and close parentheses, space, and single quote:
```
GET /users/{Id|userPrincipalName}/messages/{Id}?$expand=extensions($filter=id eq '{extensionId}')
```

Upon internal investigation by my peer engineering team (@MIchaelMainer ), we confirm the following:

- In Graph Explorer, the expected behavior is that no URL encoding should be needed, not for that '=' nor characters listed in [RFC 3968 section 2.2](https://tools.ietf.org/html/rfc3986#section-2.2). Graph Explorer is designed to handle these characters.
- There is currently a bug in Graph Explorer such that encoding that '=' becomes necessary.
- We have assigned an engineer to fix this problem in the near future.

We do appreciate your time looking into a work around meanwhile, sharing this with the community, and raising an alert. Because this behavior will change back to the expected soon, we will not be adding the note to the docs at this point.

Thank you for your continued support.

gavinbarron:
I hope that when you're referring to Graph Explorer you're talking about the tool that is parsing the URL on the server and not the Graph Explorer browser based tool offered here: https://developer.microsoft.com/en-us/graph/graph-explorer

I'm asking as I encountered this behavior in my own application which was calling into Microsoft Graph.

angelgolfer-ms:
@MIchaelMainer , do you mind getting clarification whether we're addressing the same problem here, per @gavinbarron 's last comment?

If further investigation turns out this is a separate issue, @gavinbarron , can you change your PR to document the issue in the [known issues](https://docs.microsoft.com/en-us/graph/known-issues) topic, under a new section titled "Open extensions" instead?
I will be offline for the next few days. Will check back later.

gavinbarron:
@angelgolfer-ms or @MIchaelMainer do you have any updates as to if this needs to be moved to a known issue for Open Extensions or if there's a fix underway for the underlying URL parsing here?

angelgolfer-ms:
Hi @gavinbarron , I suggest you move the

Issue #8798, opened 2020-06-19
Endpoint no longer works.


RayGHeld:
This endpoint no longer works but instead, returns an "Request_InvalidRequestUrl" message.  Switching to "https://graph.microsoft.com/beta/legacy/policies" does work however, there is no documentation related to that new endpoint.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 14d632e7-1811-aaaf-89ca-0184b334997b
* Version Independent ID: bd3f530e-5746-94c3-6a6f-ededd46b5773
* Content: [List Policies - Microsoft Graph beta](https://docs.microsoft.com/en-us/graph/api/policy-list?view=graph-rest-beta)
* Content Source: [api-reference/beta/api/policy-list.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/policy-list.md)
* Product: **microsoft-identity-platform**
* Technology: **microsoft-graph**
* GitHub Login: @dkershaw10
* Microsoft Alias: **MSGraphDocsVteam**

dkershaw10:
@RayGHeld While this document has not been removed, it's not what is shown in the TOC.  Please start here: https://docs.microsoft.com/graph/api/resources/policy-overview?view=graph-rest-beta.  This is also available in v1.0.

I'll sync with the doc team to figure out whether we should remove these old docs.

hpsin:
Thanks @dkershaw10 .  I think this keeps cropping up because the problematic pages still have the top search engine hit for searches like "MSGraph list policy" or "Delete policy ms graph".  The faster we can get these removed and redirected, the better.  Not sure how removing the page will impact our search result rankings - whichever one you wanted us to find doesn't show in the first page of results, so I'm worried that removing these old pages will just mean users get no hits when they search for this.

dkershaw10:
This no longer shows up when I try these searches.  So @hpsin I think this is fixed.

hpsin:
Same here - thanks!

PR #8968, opened 2020-07-05
Removed old policy topics from beta and added redirects

dkershaw10:
Fix for #8798


api-reference/beta/api/application-delete-tokenissuancepolicies.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Remove tokenIssuancePolicy"
```
Lauragra, line 2: These three new API topics need to be surfaced in the beta TOC.

dkershaw10: Yeah - you are correct.  Will fix.  I only discovered they were missing when removing the old topics.

dkershaw10:
@Lauragra Any reason this wasn't merged?

Lauragra:
> @Lauragra Any reason this wasn't merged?

Yes, the reason is that builds were taking 30-45 mins per PR again yesterday and I didn't have enough time to get each PR merged in sequentially before I had to go to bed. :| If this keeps happening we might need to look into a better way to manage the growing volume of PRs. I will try to get the remaining PRs merged in throughout the day today.

dkershaw10:
Thanks Laura!

PR #8980, opened 2020-07-07
[Universal Print] Clarify that capabilities are not returned by GET /print/shares

braedenp-msft:


api-reference/beta/api/print-list-shares.md
```diff
@@ -15,6 +15,8 @@ Namespace: microsoft.graph
 
 Retrieve a list of **printerShares**.
 
+To see a list of each printer share's capabilities, include the optional `$select=capabilities` query parameter.
```
Lauragra, line 18: Let's move this information to the Optional query parameters section of the topic.

PR #9017, opened 2020-07-08
Add Information protection policy permissions

BrianTJackett:
Add section for "Information protection policy permissions" using permissions and descriptions from the Azure AD portal.  No remarks added at this time, but using sample queries to highlight the difference in delegated vs. application endpoints available.


concepts/permissions-reference.md
```diff
@@ -930,6 +930,37 @@ For more complex scenarios involving multiple permissions, see [Permission scena
...
+|:-----------------------------|:-----------------------------------------|:-----------------|:-----------------|
+| _InformationProtectionPolicy.Read.All_ |   Read all published labels and label policies for an organization | Allows an app to read published sensitivity labels and label policy settings for the entire organization or a specific user, without a signed in user. | Yes |
+
+### Example usage
+
+#### Delegated
+
+* _InformationProtectionPolicy.Read_: Read labels available to the signed-in user (`GET /me/informationprotection/policy/labels`)
+* _InformationProtectionPolicy.Read_: Read labels available to the signed-in user (`GET /users/{id | user-principal-name}/informationprotection/policy/labels`)
+
```
angelgolfer-ms, line 953: @Jackson-Woods is actually the owner of this topic, so including him to comment.

Lines 951-957:
From the perspective of ease of scanning and normal formatting, I'd suggest the following:
```
### Example usage
#### Delegated
Get labels available to the signed-in user:
* `GET /me/informationprotection/policy/labels`
* `GET /users/{id | user-principal-name}/informationprotection/policy/labels`

#### Application
* Get labels available to a specific user: 
  `GET /users/{id | user-principal-name}/informationprotection/policy/labels`
* Get labels available to the organization:
  `GET /informationprotection/policy/labels`
```

angelgolfer-ms: I understand there are other sections in this very long topic that are putting the actual examples (the request query URL) in parentheses, and that's actually not an appropriate use of parentheses.
In general, we only put content in parentheses if it's supplementary, of secondary importance. In an "Example usage" section, I see the opposite - the request query URLs are the primary content.

BrianTJackett: Yes, I had copied the "Example usage" formatting and syntax from the "Identity risky user permissions" section above as a prior example.

Since I am not the API owner I would prefer to take out the Example Usage section and let the API owner address adding that if they wish.  Main goal with this PR is to get the permissions documented.  I'll remove the examples from PR for now.

PR #9041, opened 2020-07-09
Yuayuan long running operation update

grangeryy:
This is a replacement of PR 8101(8101 met some pulishing issue due to uploading files directly in online). I'm updating for beta first since this API is already in Beta version. Will update for v1.0 once later.


api-reference/beta/resources/workbookoperation.md
```diff
@@ -0,0 +1,67 @@
...
+
+The status of certain long-running workbook operation.
+
+
+## Properties
+
+| Property     | Type        | Description |
+|:-------------|:------------|:------------|
+|id|String| The operation id. Read-only.|
+|status|[workbookOperationStatus](workbookoperationstatus.md)| The current status of the operation. Possible values are: `notStarted`, `running`, `succeeded`, `failed`.|
```
Lauragra, line 20: For enums, set the Type to string and provide the values in the Description. Please remove the enums file from the PR.

api-reference/beta/resources/workbookoperationstatus.md
```diff
@@ -0,0 +1,23 @@
+---
```
Lauragra, line 1: See the guidance for documenting enums, here: https://msgo.azurewebsites.net/add/document/guidelines/enumerations.html

api-reference/beta/resources/workbookoperationerror.md
```diff
@@ -0,0 +1,51 @@
+---
+title: "workbookOperationError resource type"
+description: "workbookOperationError resource type"
```
Lauragra, line 3: For the description, please use the summary text rather than repeat the title.

api-reference/beta/resources/workbookoperation.md
grangeryy: Got you. I've removed the workbookoperationstatus file and modified the Type as "String"

api-reference/beta/resources/workbookoperationerror.md
grangeryy: ACK.

grangeryy:
> Please update the beta TOC file to surface the new workbook operation resource and GET operation. Please also add an entry to the changelog for the new APIs.

@Lauragra, we don't need a seperate item for TOC, since the start point of this new API is the same as  exsiting create session, so we provide an example in the workbook-createsession.md. The API entry is in workbook.md->relationship->operation, similar as other app such as OneNote :https://docs.microsoft.com/en-us/graph/api/resources/onenote?view=graph-rest-1.0 (onenote->relationship->operation)

PR #9048, opened 2020-07-09
Add docs to delete registered owners and users from a device

michaelrm97:
Docs for deleting registered owners and users for devices was missing


CelesteDG:
@davidmu1 - Hi David, do you have any ideas why API Doctor might be failing here?

CelesteDG:
Warnings appearing. These are expected. Merging.

PR #9054, opened 2020-07-09
Add acquireAccessToken

ArvindHarinder1:


ArvindHarinder1:
@davidmu1 / @Lauragra I have this pretty much done but wasn't sure about the TOC. Any suggestions on placement? whether it can just be at the same level as applications/synchronization/schema or it needs to be applications /synchronization/schema/secrets/secrets

PR #9073, opened 2020-07-13
Update printjob-cancelprintjob.md

JarbasHorst:
Fixed article title


Lauragra:
This is the title convention that we use for actions. Thanks.

JarbasHorst:
Ok. I changed it to have the same naming convention as the other print articles. (e. g.: [Get print](https://docs.microsoft.com/graph/api/printjob-get?view=graph-rest-beta))

PR #9105, opened 2020-07-15
Files - add media, source properties, & recordings special folder

MarcMroz:


kevklam:
@MarcMroz Sorry I didn't see this earlier. Are MediaSource/DriveItemSource exposed through MSGraph, or only when directly calling SPO? I don't see them in the service metadata as described: https://graph.microsoft.com/beta/$metadata

PR #9134, opened 2020-07-16
- adds escaping requirement for webhooks endpoints

baywet:


angelgolfer-ms:
Incorporated the addition to this PR (listed below) to PR 9117. 
" As this endpoint could be used by malicious actors for cross-site scripting type of attacks, you must escape HTML/JavaScript before returning the value. Microsoft Graph will never send any value containing HTML or JavaScript code."

PR #9144, opened 2020-07-17
Create workbook-error-codes.md

grangeryy:
This is the error code for excel graph. Later on, we will update and publish the error handling instrusctions in a seperate PR.

### DIFF
--- a/concepts/workbook-error-codes.md
+++ b/concepts/workbook-error-codes.md
@@ -68,6 +68,7 @@ The following are some additional errors that your app might encounter within th
 | **unsupportedOperation**         | The operation being attempted is not supported.
 | **unsupportedWorkbook**         | The request failed. The workbook contains unsupported features or exceeds the size limit.
 
+>**Note:** The innerError object might recursively contain more deeper innErerror objects with additional, more specific error codes. These deeper inneError codes are intented for the developer to read.
 <!-- {
   "type": "#page.annotation",
   "description": "Workbook error code and message",
--- a/concepts/workbook-error-codes.md
+++ b/concepts/workbook-error-codes.md
@@ -6,86 +6,66 @@ localization_priority: Normal
 ms.prod: "excel"
 ---
 
-# Error object for Excel APIs in Microsoft Graph
 
-This article describes error codes that are returned by the Excel APIs in Microsoft Graph when a request sent through the API fails.
+This article describes error codes that are returned by the Excel APIs in Microsoft Graph when a request sent through the API fails. See [Errors](/graph/errors.md)for more details about error responses and resource types in Microsoft Graph.
 
-The error object includes the following properties:
-
-```json
-{
-  "code": "string",
-  "message": "string",
-  "innerError": { "@odata.type": "odata.error" }
-}
-```
-
-| Property name  | Value                  | Description                                                                                             |
-|:---------------|:-----------------------|:-----------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


concepts/workbook-error-codes.md
```diff
@@ -0,0 +1,93 @@
+---
+title: "Error codes for Excel APIs in Microsoft Graph"
+description: "Error codes and messages for Excel APIs in Microsoft Graph"
+author: "grangeryy"
+localization_priority: Normal
+ms.prod: "excel"
+---
+
+# Error codes for Excel APIs in Microsoft Graph
```
joezhao-msft, line 9: This section is almost the same as the first part in https://docs.microsoft.com/en-us/graph/errors, as Excel Graph follows the same error response format in Microsoft Graph, do we need to introduce the error response/object again here? Or we could just have a simple mentioning and add a link to Microsoft Graph error page.

concepts/workbook-error-codes.md
```diff
@@ -0,0 +1,93 @@
...
+|:---------------|:-----------------------|:-----------------------------------------------------------------------------------------------------------|
+| **code**       | string                 | An error code string for the error that occurred                                                            |
+| **message**    | string                 | A developer ready message about the error that occurred. |
+| **innerError** | error object           | Optional. Additional error objects that may be more specific than the top level error.|
+
+The `innerError` object might recursively contain more `innerError` objects with additional, more specific error codes.
+
+## Top-level error code and message
+
+Below we listed current top-level error codes and messages. The `message` property contains an error message intended for the developer to read. Error messages are not localized and shouldn't be displayed directly to the user. When handling errors, your code should not key off of `message` values because they may change at any time, and they often contain dynamic information specific to the failed request. The service might add new error codes at any time.
```
joezhao-msft, line 41: is this the standard description for first level error code in Microsoft Graph?

joezhao-msft: do we need to mention the mapping between first-level/top-level error code and status code?

concepts/workbook-error-codes.md
```diff
@@ -0,0 +1,93 @@
...
+| **methodNotAllowed**        |The HTTP method in the request is not allowed on the resource.
+| **conflict**          | The current state conflicts with what the request expects.
+| **requestSizeExceeded**       | The request size exceeds the maximum limit.
+| **tooManyRequests**     | The app or user has been throttled.
+| **internalServerError**            | An internal server error occurred while processing the request.
+| **notImplemented**          | The requested feature isn’t implemented.
+| **serviceUnavailable**      | The service is not available. Please try your request again. 
+| **gatewayTimeout**        | The server, while acting as a proxy, did not receive a timely response from the upstream server in order to complete the request.
+
+## Detailed error codes
```
joezhao-msft, line 58: we need to add those error codes returned by API to this section as well.

grangeryy: ACK. Will update and commit the changes later

concepts/workbook-error-codes.md
grangeryy: I've removed some dup content but remain the part of error object introduction to make it more clear for users when they reference this page.

concepts/workbook-error-codes.md
grangeryy: For this docs, I removed the top-level, second-level to avoid confusing. I'd introduce them in the error handling docs.

PR #9160, opened 2020-07-17
Added additional permissions for default templates

jasonjoh:
Tricky to explain this I suppose, since theoretically you wouldn't need them if you overrode the template to omit the `mgt-people` or `mgt-person` controls. But out of the box, with an mgt-agenda, you get prompted for `user.readbasic.all`, `people.read`, and `contacts.read`.


nmetulev:
I like the addition, it makes a ton of sense to include those additional permissions

PR #9176, opened 2020-07-20
Long running operation -async createsession update for 1.0

grangeryy:
This is for v1.0 update for https://github.com/microsoftgraph/microsoft-graph-docs/pull/9041. No TOC change needed.

### DIFF
--- a/api-reference/v1.0/api/workbook-createsession.md
+++ b/api-reference/v1.0/api/workbook-createsession.md
@@ -119,7 +119,7 @@ Content-length: 52
 #### Request
 
 ```http
-POST https://graph.microsoft.com/beta/me/drive/items/{drive-item-id}/workbook/worksheets({id})/createSession
+POST https://graph.microsoft.com/v1.0/me/drive/items/{drive-item-id}/workbook/worksheets({id})/createSession
 Prefer: respond-async
 Content-type: application/json
 {
--- a/api-reference/v1.0/api/workbook-createsession.md
+++ b/api-reference/v1.0/api/workbook-createsession.md
@@ -1,17 +1,17 @@
 ---
-title: "Create Session"
-description: "Use this API to create a new workbook session. "
+title: "Create session"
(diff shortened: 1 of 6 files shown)


grangeryy:
Thank you very much @Lauragra for reviewing! The same as beta version, the entry of this new API locates at existing api-reference/v1.0/api/workbook-createsession.md. In this case, there will be two patterns of create session, one is the exisitng API, the other is this new long running operarion pattern. So we don't need a separate TOC for it.

Lauragra:
> Thank you very much @Lauragra for reviewing! The same as beta version, the entry of this new API locates at existing api-reference/v1.0/api/workbook-createsession.md. In this case, there will be two patterns of create session, one is the exisitng API, the other is this new long running operarion pattern. So we don't need a separate TOC for it.

As far as I can tell, the only way that users can discover the operation topic is to click workbookOperation in the Relationships table in workbook.md, and then click the Get link in the Methods table in workbookOperation. Neither the resource nor the operation topic are exposed in the TOC in beta and v1.0. Are you saying that that is by design? It seems that these won't be discoverable to users.

grangeryy:
> Thank you very much @Lauragra for reviewing! The same as beta version, the entry of this new API locates at existing api-reference/v1.0/api/workbook-createsession.md. In this case, there will be two patterns of create session, one is the exisitng API, the other is this new long running operarion pattern. So we don't need a separate TOC for it.
> 
> As far as I can tell, the only way that users can discover the operation topic is to click workbookOperation in the Relationships table in workbook.md, and then click the Get link in the Methods table in workbookOperation. Neither the resource nor the operation topic are exposed in the TOC in beta and v1.0. Are you saying that that is by design? It seems that these won't be discoverable to users.

@Lauragra, this is by design since the new API is a addition to existing create session API. To have better visibility, we are constructing the best practice section to introduce when to use this new API in PR #9227: https://github.com/microsoftgraph/microsoft-graph-docs/pull/9227

PR #9189, opened 2020-07-20
Update call.md

hrazi:
Added Transcription Info. Can you please check if this is correct.

### DIFF
--- a/api-reference/beta/api/application-post-calls.md
+++ b/api-reference/beta/api/application-post-calls.md
@@ -192,8 +192,9 @@ Content-Type: application/json
   "chatInfo": null,
   "meetingInfo": null,
   "meetingCapability": null,
+  "transcriptionInfo": null,
   "toneInfo": null
-}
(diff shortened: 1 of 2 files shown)


Lauragra:
@hrazi , based on this error, it looks like you need to update the response example:  
Error: /api/application-post-calls.md:
      Missing properties: response was missing these required properties: transcriptionInfo

  Error: /api/application-post-calls.md:
      Missing properties: response was missing these required properties: transcriptionInfo

hrazi:
@Lauragra fixed errors. please let me know if I need to make any other changes.

Lauragra:
> @Lauragra fixed errors. please let me know if I need to make any other changes.

We're still getting the same API Doctor error. Please make sure that you've updated all the relevant response examples that should return that property.

lramosvea:
In the live metadata, I see the property is called transcription, not transcriptionInfo.

PR #9200, opened 2020-07-21
Add new Microsoft 365 usage report: Microsoft 365 Apps usage

bjggzxb:


bjggzxb:
I saw one validation failed with below message, is there any instruction on resolving this?
> for PR "Made the title and location of any error-related section consistent.": add "conflicting", remove "ready to merge"

api-reference/beta/api/reportroot-getm365appplatformusercounts.md
```diff
@@ -0,0 +1,194 @@
...
+doc_type: apiPageType
+---
+
+# reportRoot: getM365AppPlatformUserCounts
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Enables you to understand the trend of active users across all apps for each platform – Windows, Mac, Web, and Mobile in your organization.
```
Lauragra, line 16: Can we have a description that conveys what this API call does - i.e. "Get a report that provides the trend of active users...". That is, the report enables them to understand the active user trend, not the API.

api-reference/beta/api/reportroot-getm365appplatformusercounts.md
```diff
@@ -0,0 +1,196 @@
...
+
+This method supports the `$format` [OData query parameter](/graph/query-parameters) to customize the response. The default output type is text/csv. However, if you want to specify the output type, you can use the OData `$format` query parameter set to text/csv or application/json.
+
+## Request headers
+
+| Name          | Description               |
+| :------------ | :------------------------ |
+| Authorization | Bearer {token}. Required. |
+
+## Response
```
Lauragra, line 58: Global: We're missing the ## Request body section. Please include it with the text "Do not supply a request body with this method."

api-reference/beta/api/reportroot-getm365appplatformusercounts.md
```diff
@@ -0,0 +1,196 @@
...
+
+The following is an example that outputs CSV.
+
+#### Request
+
+The following is an example of the request to get the **content** property.
+
+
+<!-- {
+  "blockType": "ignored",
```
Lauragra, line 96: Global: Why are we setting the request examples to Ignored?

api-reference/beta/api/reportroot-getm365appplatformusercounts.md
bjggzxb: Makes sense, I've followed this pattern and updated the description for all three APIs

api-reference/beta/api/reportroot-getm365appplatformusercounts.md
bjggzxb: Added this section under ## Request headers

api-reference/beta/api/reportroot-getm365appplatformusercounts.md
bjggzxb: To be honest, I don't known the meaning of this field, and I just followed the existing doc like [getemailactivitycounts](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/reportroot-getemailactivitycounts.md).

Should we change it to another value?

Lauragra: The values should be request and response. For details, see this Wiki topic: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/66013/Set-up-for-API-Doctor

bjggzxb: Updated all the code blocks to `request` and `response`

bjggzxb: Reverted blobType value for Http request section to `ignored`

api-reference/beta/resources/intune-shared-report.md
```diff
@@ -19,6 +19,7 @@ Returns the content appropriate for the context, including:
 
 - Device Configuration profile history reports.
 - Enrollment failure reports.
+- Microsoft 365 usage reports
```
Lauragra, line 22: @bjggzxb , did you mean to update the intune-specific report topic? @davidmu1 , how can we ensure that this update is preserved?

bjggzxb: Even though this doc's name indicates it is intune specific, I think the `report` resource is shared because there is only one `report` complex type in [Graph's OData metadata.](https://graph.microsoft.com/beta/$metadata)

```xml
<ComplexType Name="report">
    <Property Name="content" Type="Edm.Stream"/>
</ComplexType>
```

So my understanding is, there should be only one page for this resource.

Lauragra: @bjggzxb , the Intune docs are generated a little differently, so we'll need to make sure that this content update doesn't get overwritten with the next Intune doc drop. Can you please work with Zhuo Shan to make sure that they integrate this update into their code? Thanks.

bjggzxb: Thanks Laura, will work with Intune to address this.

PR #9209, opened 2020-07-21
MGT scoped component docs

shweaver-MSFT:


concepts/toolkit/scoped.md
```diff
@@ -0,0 +1,24 @@
...
+# Using scoped components in the Microsoft Graph Toolkit
+
+Use scoped components to prevent name collisions.
+
+Typically components can be loaded all together with the mgt-loader, or individually referenced with an import statement:
+```
+import "@microsoft/mgt/dist/es6/components/mgt-person"
+```
+
+However, when running in certain environments, such as SharePoint, registering multiple instances of the same component can cause errors in the element registry.
```
nmetulev, line 17: Clarify the scenario where this would be the case. For example, if two webparts are using the toolkit, component registration can collide since they are running in the same context.

PR #9227, opened 2020-07-23
Create workbook-best-practice.md

grangeryy:


concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+  "values": [[“east”, “pear”, 4]]
+}
+```
+
+The successful response looks like:
+
+```http
+HTTP/1.1 200 OK
+Content-type: application/json
+Content-length: “42”
```
joezhao-msft, line 61: nit: it seems that the double quotes is not aligned with others.

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+
+```http
+HTTP/1.1 204 No Content
+```
+
+For more details, see [Create session](/graph/api/workbook-createsession?view=graph-rest-1.0) and [Close session](/graph/api/workbook-closesession?view=graph-rest-1.0).
+
+## Working with APIs that may take a long time to complete
+
+You may notice some API responses require indeterminate time to complete, for example, open a workbook with large size. To resolve this issue, we provide the long running operation pattern. By using this pattern, you do not need to wait until the action is complete before returning a response or worry about the timeout for the request.
```
joezhao-msft, line 117: the intention to introduce LRO is to avoid timeout issue. Not sure if the first statement is the good or not.

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+
+Currently, Excel Graph has enabled long running operation pattern for session creation API. Here are the steps:
+
+1. Adds a header of `Prefer: respond-async` in the request to indicate it as a long running operation when creating a session.
+2. In long running operation pattern, it will return a `202 Accepted` response along with a Location header to retrieve operation status. Otherwise, if session creation succeeds in several seconds, it will return a `201 Created` response directly instead of going to long running operation pattern.
```
joezhao-msft, line 122: in "otherwise" case, it should be "if session creation completes in serveral seconds" since it might be succeeded or failed (and then it will returned as regular CreateSession)

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+```http
+HTTP/1.1 202 Accepted
+Location: https://graph.microsoft.com/v1.0/me/drive/items/{drive-item-id}/workbook/operations/{operation-id}
+Content-type: application/json
+
+{
+}
+```
+
+In some cases, if the creation succeeds directly in seconds, it won't enter long running operation pattern, instead it returns a `201 Created` response like following.
```
joezhao-msft, line 150: we might also need to mention the failed case for createSession here.

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+HTTP/1.1 200 OK
+Content-type: application/json
+{
+    "id": {operation-id},
+    "status": "succeeded",
+    "resourceLocation": "https://graph.microsoft.com/v1.0/me/drive/items/{drive-item-id}/workbook/sessionInfoResource(key='{key}')
+}
+```
+
+Case of `failed`: the operation fails.
```
joezhao-msft, line 198: should we mention that the error object below is aligned with our error codes (and add a link to the page)?

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+Content-type: application/json
+Content-length: 52
+
+{
+  "id": "id-value",
+  "persistChanges": true
+}
+```
+
+### Poll status of the long running create session
```
joezhao-msft, line 163: we could also mention that the suggested interval to poll status is 30s, while the max internal should be no more than 4min.

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,240 @@
...
+      "innerError": {
+        "code": "GenericFileOpenError",
+        "message": "The workbook cannot be opened."
+      }
+    }
+  }
+}
+```
+
+### Acquire session information
```
joezhao-msft, line 221: we could also mention that the 3rd step depends on the initial request, there is be no need to aquire the result if the inital request doesn't return a reponse body.

concepts/workbook-best-practice.md
grangeryy: fixed

concepts/workbook-best-practice.md
grangeryy: Modified a bit on the statement.

concepts/workbook-best-practice.md
grangeryy: Good suggestion. I've modified the sentence.

concepts/workbook-best-practice.md
grangeryy: ACK

concepts/workbook-best-practice.md
grangeryy: Provided a link after the failed case.

concepts/workbook-best-practice.md
grangeryy: Got you. I've added in the poll status part

concepts/workbook-best-practice.md
grangeryy: ACK. Modified in the acquire section

concepts/workbook-best-practice.md
```diff
@@ -0,0 +1,266 @@
...
+        "message": "The workbook cannot be opened."
+      }
+    }
+  }
+}
+
+
+### Poll status of the long running create session
+
+In long running operation pattern, you can get creation status with specified location with request similar as following. The suggested interval to poll status is around 30 seconds and the maximum internal should be no more than 4 minutes.
```
joezhao-msft, line 187: get create status => retrieve operation status?

PR #9256, opened 2020-07-24
Adding docs for delta on sites and listItems

learafa:


api-reference/beta/api/listitem-delta.md
```diff
@@ -0,0 +1,303 @@
+---
+author: learafa
+description: "This method allows your app to track changes to a list and its items over time."
+title: "sites: delta"
```
JeremyKelley: The title property should be listItem: delta right?

api-reference/beta/api/listitem-delta.md
```diff
@@ -0,0 +1,303 @@
...
+
+**Note:** you should only delete a folder locally if it is empty after syncing all the changes.
+
+## Permissions
+
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | Sites.Read.All, Sites.ReadWrite.All    |
```
JeremyKelley, line 36: Are Sites.Manage.All and Sites.FullControl.All valid as well?

api-reference/beta/api/listitem-delta.md
```diff
@@ -0,0 +1,303 @@
...
+
+```http
+GET /sites/{siteId}/lists/{listId}/items/delta
+```
+
+## Query Parameters
+
+| Name   | Value  | Description                                                                                                                          |
+|:-------|:-------|:-------------------------------------------------------------------------------------------------------------------------------------|
+| token  | string | Optional. If unspecified, enumerates the hierarchy's current state. If `latest`, returns empty response with latest delta token. If a previous delta token, returns new state since that token.
```
JeremyKelley, line 52: I might say "since that token was generated" but that's minor.

JeremyKelley (changes requested):
Left a few things we should update.  I didn't see the changes for Sites delta, is that in a separate PR?

api-reference/beta/api/sites-delta.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Permissions
+
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | Not Supported  |
+|Delegated (personal Microsoft account) | Not Supported   |
+|Application | Sites.Read.All, Sites.ReadWrite.All |
```
JeremyKelley, line 34: Should also check the full control and manage permissions for this.

JeremyKelley (changes requested):

PR #9292, opened 2020-07-28
toc restructure 2

davidmu1:


api-reference/beta/resources/report-identity-access.md
```diff
@@ -0,0 +1,47 @@
...
+Get a better understanding of how users in your organization use Azure AD capabilities, such as self-service password rest and multi-factor authentication (MFA). You can determine which authentication methods are more successful for your organization, what types of errors end users are running into, and what campaign you need to run to help your end users adopt the use of self-service password rest and MFA. For more information, see the [authentication methods usage report API](/graph/api/resources/applicationsigninsummary?view=graph-rest-beta).
+
+## Next steps
+
+Reports resources and APIs can open up new ways for you to engage with users and manage their experiences with Microsoft Graph. To learn more:
+
+- Drill down on the methods and properties of the resources most helpful to your scenario.
+- Try the API in the [Graph Explorer](https://developer.microsoft.com/graph/graph-explorer).
+
+Need more ideas? See [how some of our partners are using Microsoft Graph](https://developer.microsoft.com/graph/graph/examples#partners).
```
Lauragra, line 47: None of the partners are using reporting scenarios currently. Do you still want to link to the partners page? Also, we have an updated URL: https://developer.microsoft.com/en-us/graph/partners

davidmu1: Thanks Laura. Removed the link.

api-reference/beta/toc.yml
```diff
@@ -10940,8 +10896,33 @@ items:
             href: api/multivaluelegacyextendedproperty-get.md
     - name: Reports
       items:
+      - name: Azure AD activity reports
```
Lauragra, line 10899: Just to verify, is this the structure that you want for the beta TOC?
image

davidmu1: Yes, I thought it was consistent with what is there.

Lauragra: Okay, I didn't know if the Azure AD node should be placed under Identity and access, rather than as a peer.

davidmu1: I guess I don't feel strongly either way. If you think they should be moved under Identity and Access I'm fine with that. They were under the Identity and Access node and Dan suggested they would be better in under reports. Maybe Azure AD activity reports isn't the right title, just was looking for something more inclusive than Audit logs.

PR #9309, opened 2020-07-29
New BitLocker recovery key API documentation

hafowler:


hafowler:
@Lauragra - For TOC, I put the BitLocker recovery key in Identity and access>Directory management. However, conceptually, BitLocker recovery key doesn't fall under Directory management. Some guidance as to where BitLocker recovery key in TOC is greatly appreciated.

Other options we considered: Devices & apps>Corporate management>Device management

Lauragra:
> Hi Hannah! I actually don't have perms to merge to master. Maybe @angelgolfer-ms (I assume that's Angela) can help?

Hi, Angela looped me in on this PR and I will help out on it when I'm back on Thursday. Thanks.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+}
+-->
+``` http
+GET /bitlocker/recoveryKeys/{bitlockerRecoveryKeyId}
+```
+
+## Optional query parameters
+This method supports the follwoing OData query parameters to help customize the response. For general information, see [OData query parameters](/graph/query-parameters).
+
+### OData query parameters
```
Lauragra, line 50: Why do we have a separate H3 and table just to list a single supported query parameter? Suggest updating the existing boilerplate text as "This method supports the `$select` OData query parameter to return the **key** property...."

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+---
+
+# Get bitlockerRecoveryKey
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Read the properties and relationships of a [bitlockerRecoveryKey](../resources/bitlockerrecoverykey.md) object. Does not return the key property by default.
+
+One of the following roles must be assigned to call this API: <br>
```
angelgolfer-ms, line 17: Hi @hafowler , just browsing and noticed this, thought I'd leave you a comment:
Line 17: "One of the following roles must be assigned to call this API:"
I have 2 comments for that sentence: 
(a) In general, we try to avoid passive voice in docs; please use active voice so that customers are clear who does what .
(b) Can you explain more clearly who needs to assign whom one of the listed roles?
If I may make a guess - 

If that's the case, I would suggest moving lines 17-24 to below line 36, in the "Permissions" section. You can add a sentence to the effect of:
```
> **Note**: For delegated permissions to allow apps to get bitLockerRecoveryKey resources on behalf of the signed-in user, the tenant administrator must have assigned the user one of the following roles, or, the user is the registered owner of the device that the BitLocker recovery key was originally backed up from."

{list of roles from lines 18-24}
```

And, can you please do the same to the bitlockerrecoverykey-list.md topic as well?

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+## HTTP request
+
+<!-- {
+  "blockType": "ignored"
+}
+-->
+``` http
+GET /bitlocker/recoveryKeys/{bitlockerRecoveryKeyId}
+```
+
```
angelgolfer-ms, line 46: Since getting the key property returned is a major use case of this API, I would suggest you elevate that by documenting the syntax in the "HTTP request" section as well. You'd have two syntax lines for this API.
To do that, preceding line 39, please add something like:
"To get the specified BitLocker key without retrieving its recovery key:"

Then to add a second syntax line, after line 46, insert:
"To get the specified key including its recovery key:

``` http
GET /bitlocker/recoveryKeys/{bitlockerRecoveryKeyId}?$select=key
```

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+## Optional query parameters
+This method supports the follwoing OData query parameters to help customize the response. For general information, see [OData query parameters](/graph/query-parameters).
+
+### OData query parameters
+|Parameter|Description|Example|
+|:---|:---|:---|
+|[$select](/graph/query-parameters#select-parameter)|Use to return the 'key' property. Requires BitLocker.Read.All permission.|`/recoveryKeys/{bitlockerRecoveryKeyId}?$select=key`|
```
angelgolfer-ms, line 53: Line 53:
For the example, instead of trying to show some pseudo example with a partial query URL, I think it would be more helpful if you can replace that by a link to example 2. You can find an example in the [GET open extension](https://docs.microsoft.com/en-us/graph/api/opentypeextension-get?view=graph-rest-1.0&tabs=http#optional-query-parameters) topic.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+ms.prod: "microsoft-identity-platform"
+doc_type: apiPageType
+---
+
+# Get bitlockerRecoveryKey
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Read the properties and relationships of a [bitlockerRecoveryKey](../resources/bitlockerrecoverykey.md) object. Does not return the key property by default.
```
angelgolfer-ms, line 15: Line 15:
As I read this topic, I see some significant info that can be better discovered if you just roll them all up in the top section here, before the first H2 section "Permissions".

You can do something like:
```
Read the properties and relationships of a [bitlockerRecoveryKey](../resources/bitlockerrecoverykey.md) object. 

By default, this operation does not return the **key** property which represents the recovery key. 

To include the **key** property in the response, use `$select`. 

Including the $select query parameter triggers Azure AD auditing the operation and generating an audit log. You can find this log in Azure AD audit logs under the KeyManagement category.
```

One additional comment is, can you add a link to for this phrase "find this log in Azure AD audit logs under the KeyManagement category"? I can't figure where to find this, or find more info about this KeyManagement category.

Please also note the bolding of properties like the key property, and enclosing tick marks around $select.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+## Optional query parameters
+This method supports the follwoing OData query parameters to help customize the response. For general information, see [OData query parameters](/graph/query-parameters).
+
+### OData query parameters
+|Parameter|Description|Example|
+|:---|:---|:---|
+|[$select](/graph/query-parameters#select-parameter)|Use to return the 'key' property. Requires BitLocker.Read.All permission.|`/recoveryKeys/{bitlockerRecoveryKeyId}?$select=key`|
+> NOTE: When the $select query parameter is included in the API call, the call is audited, and an audit log is generated. You can find this log in Azure Active Directory audit logs under the KeyManagement category.
```
angelgolfer-ms, line 54: Line 54: I think this special info would be better discovered if you put it up top in the context of using $select for this operation.
That space before the first H2 section is where we enter useful conceptual info about the operation (or resource, if this was a resource type topic). See my comment above around line 15.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+## Request headers
+|Name|Description|
+|:---|:---|
+|Authorization|Bearer {token}. Required.|
+|ocp-client-name|Name of the client application performing the API call. Required.|
+|ocp-client-version|Version of the client application performing the API call. Required.|
+
```
angelgolfer-ms, line 63: Lines 61 and 62:
If these 2 request headers are required, then you should include them in each of example 1 and example 2, as part of the corresponding request section.
You can see [example 3](https://docs.microsoft.com/en-us/graph/api/message-get?view=graph-rest-1.0&tabs=http#example-3) of the GET message topic - the example request query includes a Prefer header.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+### Example 1: Get BitLocker key without 'key' property by specifying key id
+
+#### Request
+<!-- {
+  "blockType": "request",
+  "name": "get_bitlockerrecoverykey"
+}
+-->
+``` http
+GET https://graph.microsoft.com/beta/bitlocker/recoveryKeys/{bitlockerRecoveryKeyId}
```
angelgolfer-ms, line 82: Line 82 and 118:
I strongly recommend that in place of {bitlockerRecoveryKeyId} in the example query URL, you specify the corresponding GUID key ID, for each of the two example requests.
That is, for the first example:
GET https://graph.microsoft.com/beta/bitlocker/recoveryKeys/b465e4e8-e4e8-b465-e8e4-65b4e8e465b4

To properly support API Doctor, please add the following line after line 77 for the first example:
"sampleKeys": ["b465e4e8-e4e8-b465-e8e4-65b4e8e465b4"],

Do the similar for the second example.

I recommend this because I had found the "recoveryKeys" segment, and terminology used around the "BitLocker key", "recovery key", and "BitLocker recovery key" are easily confused. Showing what {bitlockerRecoveryKeyId} can be like (like a GUID) helps customers.

api-reference/beta/resources/bitlockerrecoverykey.md
```diff
@@ -0,0 +1,58 @@
...
+doc_type: resourcePageType
+---
+
+# bitlockerRecoveryKey resource type
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Collectively represents a stored BitLocker recovery key and contains the actual key. Inherits from [entity](../resources/entity.md).
```
angelgolfer-ms, line 16: Line 16:
What does "Collectively" here mean?
Would it be fine if we just start with "Represents a stored BitLocker key that contains the actual recovery key. "?
I use the term "BitLocker key" and "actual recovery key" based on how you describe the id property (line 31) and key property (line 32).

I also suggest removing "Inherits from entity." I think all the entities in the Microsoft Graph API inherit from the entity resource, so there's no special reason we have to specify it here :-)

api-reference/beta/toc.yml
```diff
@@ -9483,6 +9483,13 @@ items:
             href: api/administrativeunit-get-scopedrolemembers.md
           - name: Remove a scoped-role administrator
             href: api/administrativeunit-delete-scopedrolemembers.md
+        - name: BitLocker recovery key (preview)
+          href: resources/bitlockerrecoverykey.md
+          items:
+          - name: Get BitLocker recovery key
```
angelgolfer-ms, line 9489: Line 9489: You can just say "Get" and not repeat "BitLocker recovery key".
Same with line 9491.
This is because the children of this node (line 9486, which is labelled "BitLocker recovery key") all have a verb and the same object "BitLocker recovery key".

concepts/changelog.md
```diff
@@ -123,6 +123,8 @@ Adding PSTN and direct routing call reports to Microsoft Graph (beta).
 | Addition        | v1.0        | Added [delta query](delta-query-overview.md) capability to [servicePrincipal](/graph/api/serviceprincipal-delta). |
 | Addition | beta | Added the [connectedOrganization](/graph/api/resources/connectedorganization?view=graph-rest-beta) to entitlement management.  |
 | Addition        | v1.0        | Added [delta query](delta-query-overview.md) capability for [oauth2PermissionGrant](/graph/api/oauth2Permissiongrant-delta). |
+| Addition | beta | Added the [bitlockerRecoveryKey](/graph/api/resources/bitlockerRecoveryKey?view=graph-rest-beta) resource type|
```
angelgolfer-ms, line 126: Lines 126 and 127: Can you end each description with a period?

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type|Permissions (from most to least privileged)|
+|:---|:---|
+|Delegated (work or school account)|BitLocker.ReadBasic.All, BitLocker.Read.All|
+|Delegated (personal Microsoft account)|Not supported|
+|Application|Not supported|
+
+## HTTP request
+
```
angelgolfer-ms, line 38: This comment applies to both this operation and the "list" operation:
Since the api.md file describes that there is a "recoveryKeys" root segment for all BitLocker operations, I think you should document the bitLocker entity, which has recoveryKeys as a navigation property. That would fully explain why the get and list operation topics show syntax with segments "/bitlocker/recoveryKeys/".

Was there a reason why you didn't include the bitLocker resource type topic in this PR?

api-reference/beta/api/bitlockerrecoverykey-list.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Optional query parameters
+This method supports the following OData query parameters to help customize the response. For general information, see [OData query parameters](/graph/query-parameters).
+
+The response may also contain an `odata.nextLink`, which you can use to page through the result set. See [Paging Microsoft Graph data](/graph/paging) for more information.
+
+### OData query parameters
+|Parameter|Description|Example|
+|:---|:---|:---|
+|[$filter](/graph/query-parameters#filter-parameter)|Filters results by device|`/recoveryKeys?$filter=deviceId eq {deviceId}`|
```
angelgolfer-ms, line 54: Line 54:
The api.md file says one can filter by the id or deviceId property. Can you confirm and clarify if necessary?

api-reference/beta/api/bitlockerrecoverykey-get.md
davidmu1: To be consistent with most GET documentation, I suggest using the word "Retrieve" rather than "Read", as in "Retrieve the properties..."

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+## Response
+
+If successful, this method returns a `200 OK` response code and a [bitlockerRecoveryKey](../resources/bitlockerrecoverykey.md) object in the response body.
+
+## Examples
+
+### Example 1: Get BitLocker key without 'key' property by specifying key id
+
+#### Request
```
davidmu1, line 75: Add this sentence:

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+  "blockType": "request",
+  "name": "get_bitlockerrecoverykey"
+}
+-->
+``` http
+GET https://graph.microsoft.com/beta/bitlocker/recoveryKeys/{bitlockerRecoveryKeyId}
+```
+
+
+#### Response
```
davidmu1, line 86: Add this sentence:

The following is an example of the response.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+-->
+``` http
+HTTP/1.1 200 OK
+Content-type: application/json
+
+{
+  "value": {
+    "@odata.type": "#microsoft.graph.bitlockerRecoveryKey",
+    "id": "b465e4e8-e4e8-b465-e8e4-65b4e8e465b4",
+    "createdDateTime": "String (timestamp)",
```
davidmu1, line 102: These values should show something realistic looking from a response.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+    "volumeType": "String",
+    "deviceId": "String"
+  }
+}
+```
+
+### Example 2: Get BitLocker key with 'key' property by specifying key id
+
+#### Request
```
davidmu1, line 111: Add this sentence:

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+  "blockType": "request",
+  "name": "get_bitlockerrecoverykey"
+}
+-->
+``` http
+GET https://graph.microsoft.com/beta/bitlocker/recoveryKeys/{bitlockerRecoveryKeyId}?$select=key
+```
+
+
+#### Response
```
davidmu1, line 122: Add this sentence:

The following is an example of the response.

api-reference/beta/api/bitlockerrecoverykey-get.md
```diff
@@ -0,0 +1,144 @@
...
+``` http
+HTTP/1.1 200 OK
+Content-type: application/json
+
+{
+  "value": {
+    "@odata.type": "#microsoft.graph.bitlockerRecoveryKey",
+    "id": "b465e4e8-e4e8-b465-e8e4-65b4e8e465b4",
+    "createdDateTime": "String (timestamp)",
+    "volumeType": "String",
```
davidmu1, line 139: Add realistic values.

api-reference/beta/api/bitlockerrecoverykey-get.md
hafowler: I did not include the BitLocker resource type as it is not accessible via the MS Graph API, and we though it would further confuse customers if we did include it.

api-reference/beta/api/bitlockerrecoverykey-list.md
hafowler: You can filter by the deviceID. To get a specific key based on the key ID, you have to include it in the request. I will make that clearer in the updated PR.

api-reference/beta/api/bitlockerrecoverykey-get.md
hafowler: @Lauragra What are your thoughts on this?

hafowler:
@Lauragra I think that I have addressed all feedback requested. Please let me know if I have not taken the correct steps to get this merged.

api-reference/beta/api/bitlockerrecoverykey-get.md
Lauragra: What does "not accessible" mean? Is it defined in the metadata or hidden via exposure control?

hafowler: See Teams message. But essentially by "not accessible" I mean that the only customer facing information that users see on the BitLocker entity is that it is a root segment of recovery keys. Aside from that, there is no further information, and no operations are supported on the bitlocker entity. We felt that including documentation on this would confuse customers, as it would be publicly surfacing something that is inoperable.

api-reference/beta/api/bitlockerrecoverykey-list.md
```diff
@@ -0,0 +1,172 @@
...
+title: "List bitlockerRecoveryKeys"
+description: "Get a list of the bitlockerRecoveryKey objects and their properties."
+author: "hafowler"
+localization_priority: Normal
+ms.prod: "microsoft-identity-platform"
+doc_type: apiPageType
+---
+
+# List bitlockerRecoveryKeys
+Namespace: microsoft.graph
```
angelgolfer-ms, line 11: @hafowler , please rename the file of the list operation as bitlocker-list-recoverykeys.md, and fix any links to it.
FYI, this is based on how your API is defined in the CSDL, and on the filename conventions used in Microsoft Graph.
If the stub topic generator generated the files for this PR, that would be the name of the file too. Having consistent filenaming is important especially for full topic generation when the generator can reliably determine filenames so as to generate links to them in the docs.
CDK topic: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55061/File-naming

angelgolfer-ms: Line 10: Please also change the topic title to "List recoveryKeys".
The second term is the name of the navigation property. Please do not pluralize the resource name (adding an 's' to the resource bitLockerRecoveryKey).
This would also be how the stub topic generator generates the topic title for a "list" operation topic - using the nav property name.

api-reference/beta/api/bitlockerrecoverykey-get.md
angelgolfer-ms: Hi @hafowler , FYI in Microsoft Graph, every resource that's defined in the CSDL should be documented. We maintain the expectations for customers that an API in the CSDL <-> the API is available for access <-> API is documented.

Based on this PR and just the list and get operations you're exposing at this point, the bitLocker entity would have an entity file bitlocker.md, exposing 1 operation, the "list" operation, and 1 nav property, recoveryKeys. The entity needs to exist because it is the parent resource for recoveryKeys. Without defining this parent resource, there would be no "list" operation to get all the recovery keys for the tenant. Where would the term "recoveryKeys" come from? 

If you like, you can take a look at the entity topic todo.md (https://docs.microsoft.com/en-us/graph/api/resources/todo?view=graph-rest-beta) and the "list task lists" topic (todo-list-lists.md) (https://docs.microsoft.com/en-us/graph/api/todo-list-lists?view=graph-rest-beta). The todo entity is used in much the same way for their "list" and "create" operations, as your bitLocker entity for your "list" operation.

api-reference/beta/toc.yml
```diff
@@ -9544,6 +9544,13 @@ items:
             href: api/administrativeunit-get-scopedrolemembers.md
           - name: Remove a scoped-role administrator
             href: api/administrativeunit-delete-scopedrolemembers.md
+        - name: BitLocker recovery key (preview)
+          href: resources/bitlockerrecoverykey.md
+          items:
+          - name: Get
```
angelgolfer-ms, line 9550: Line 9548:
Please remove this href line. 
Instead, please add 2 lines after line 9549:
(Indent the way as the existing line 9550)
- name: BitLocker recovery key
- href: resources/bitlockerrecoverykey.md

api-reference/beta/toc.yml
```diff
@@ -9544,6 +9544,13 @@ items:
...
+            href: api/bitlockerrecoverykey-get.md
```
angelgolfer-ms, line 9551: Please flip the order of "list" and "get" in the TOC, to be consistent with the order of CRUD topics in other parts of the TOC.
Related CDK topic: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55080/ToC-and-topic-title

api-reference/beta/api/bitlockerrecoverykey-list.md
angelgolfer-ms: Reminder: when you change the topic title, please also update the Methods table in bitlockerrecoverykey.md.

angelgolfer-ms (changes requested):
@hafowler , Laura brought to my attention that you had some questions about my earlier comment regarding documenting the bitlocker entity. I added some more explanation. 
I also added comments for the "list" operation and the TOC. Please take a look. Those are simple to incorporate. Please ping me on Teams if you have more questions.

concepts/changelog.md
```diff
@@ -236,6 +236,9 @@ Adding PSTN and direct routing call reports to Microsoft Graph (beta).
 | Addition | v1.0 | Added the [identitySecurityDefaultsEnforcementPolicy API](/graph/api/resources/identitysecuritydefaultsenforcementpolicy?view=graph-rest-1.0), which represents Azure Active Directory security defaults policy.|
 | Addition        | v1.0        | Added [delta query](delta-query-overview.md) capability to [servicePrincipal](/graph/api/serviceprincipal-delta). |
 | Addition | beta | Added the [connectedOrganization](/graph/api/resources/connectedorganization?view=graph-rest-beta) to entitlement management.  |
+| Addition        | v1.0        | Added [delta query](delta-query-overview.md) capability for [oauth2PermissionGrant](/graph/api/oauth2Permissiongrant-delta).
+| Addition | beta | Added the [bitlockerRecoveryKey](/graph/api/resources/bitlockerRecoveryKey?view=graph-rest-beta) resource type.|
```
angelgolfer-ms, line 240: @hafowler , can you please move row 240 (your changelog item) from the July section to the September section? 
We do go by when the docs are eventually published, for the benefit of customers to discover the new API.

api-reference/beta/api/bitlocker-list-recoverykeys.md
```diff
@@ -0,0 +1,172 @@
+---
+title: "List bitlockerRecoveryKeys"
+description: "Get a list of the bitlockerRecoveryKey objects and their properties."
+author: "hafowler"
+localization_priority: Normal
+ms.prod: "microsoft-identity-platform"
+doc_type: apiPageType
+---
+
+# List bitlockerRecoveryKeys
```
angelgolfer-ms, line 10: I think the topic title should reflect the name of the nav property, not pluralizing the actual resource name.
Like you have it in the Methods table in bitlocker.md, "List recoverKeys", based on recoveryKeys being the name of the nav property.

angelgolfer-ms (changes requested):
@hafowler , just two comments, one on the list operation topic title, and the changelog item location.

PR #9338, opened 2020-08-01
Add documentation for sending inline images in ChatMessage

RamjotSingh:


api-reference/beta/api/channel-post-message.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Create chatMessage in a channel
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Create a new [chatMessage](../resources/chatmessage.md) in the specified [channel](../resources/channel.md).
+
+> **Note**: We don't recommend that you use this API for data migration. It does not have the throughput necessary for a typical migration.
```
Lauragra, line 18: This is two Notes in a row, which is repetitive for the reader. Can we make this first note regular text?

api-reference/beta/api/channel-post-message.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Request headers
+
+| Name          | Description   |
+|:--------------|:--------------|
+| Authorization | Bearer {code} |
+
+## Request body
+
+In the request body, supply a JSON representation of a [message](../resources/chatmessage.md) object. Only the body property is mandatory, other properties are optional.
```
Lauragra, line 48: Should we refer to it as a chatMessage?

api-reference/beta/api/chat-post-message.md
```diff
@@ -1,19 +1,23 @@
...
+description: "Send a new chatMessage in a chat."
 localization_priority: Normal
-author: "clearab"
+author: "RamjotSingh"
 ms.prod: "microsoft-teams"
 doc_type: "apiPageType"
 ---
 
-# Create chatMessage
+# Send chatMessage in a chat
```
Lauragra, line 10: Why is the title send and not create?

api-reference/beta/api/chatmessage-post.md
```diff
@@ -0,0 +1,544 @@
...
+doc_type: apiPageType
+---
+
+# Send chatMessage in a channel or a chat
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Create a new [chatMessage](../resources/chatmessage.md) in the specified [channel](../resources/channel.md) or a [chat](../resources/chat.md).
```
Lauragra, line 16: Do you need to clarify when to use this API vs. the previous to create a chatMessage in a chat?

api-reference/beta/api/chatmessage-post.md
```diff
@@ -0,0 +1,544 @@
...
+### Permissions for Chat
+| Permission type                        | Permissions (from least to most privileged) |
+|:---------------------------------------|:--------------------------------------------|
+| Delegated (work or school account)     | ChatMessage.Send, Chat.ReadWrite |
+| Delegated (personal Microsoft account) | Not supported. |
+| Application                            | Not supported. |
+
+## HTTP request
+
+### Sending message in a channel
```
Lauragra, line 42: Why are we covering these three scenarios in a single API topic in addition to the separate topics for each scenario? Instead of reference topics that overlap one another, can you cover the gap in a conceptual topic?

api-reference/beta/api/chatmessage-post.md
```diff
@@ -0,0 +1,544 @@
...
+
+## Response
+
+If successful, this method returns a `201 Created` response code and a new [chatMessage](../resources/chatmessage.md) object in the response body.
+
+## Examples
+
+In the following examples, the URL can use the [HTTP syntax](#http-request) described to [send a message to a chat](chat-post-message.md), [send a message to a channel](channel-post-message.md), or [send reply to a channel](channel-post-messagereply.md).
+
+### Example 1: Hello World
```
Lauragra, line 84: Please provide more descriptive titles for these examples. I would expect that we would have examples for each of the three syntaxes provided in this topic. These examples seem like scenarios that you would cover in a conceptual how to article rather than the API reference.

api-reference/beta/api/channel-post-message.md
RamjotSingh: Sure I can make the change.

api-reference/beta/api/chat-post-message.md
RamjotSingh: Coz it's not really 'create'. Creation would be like drafting a message and leaving it. This is more like POSTing a message and hence the 'Send'. This is in line with other places too.

api-reference/beta/api/chatmessage-post.md
RamjotSingh: There are 2 places where you can create a message, chat and channel. What I have done with this PR is 3 docs

1. Send message in a chat
2. Send message in a channel
3. Send message

1 and 2 give you 1 example and the url paths while 3 gives you lost more examples. 1 and 2 both lead to 3 and in turn 3 refers to 1 and 2. This allows us to have one place with samples and not duplicate them in 2 places.

api-reference/beta/api/chatmessage-post.md
RamjotSingh: There are individual docs for those scenarios, this is one central doc under the entity itself which talks briefly about all 3 and gives more samples. Just like the comment above, the idea is to not have same samples repeated in 2 or 3 places. Instead the individual API docs list 1 super basic sample and this central doc is giving you more samples.

*Why is this not a conceptual topic?*
It is not really big enough to be one. This is just me making it easy to update samples in one place and have people use them instead of same sample in 3 places.

api-reference/beta/api/chatmessage-post.md
RamjotSingh: As discussed offline. Resolving.

PR #9351, opened 2020-08-03
Pushed cancel media processing doc to v1

frankpeng7:
Pushed cancel media processing doc to v1


frankpeng7:
I actually have a question about the code snippets. For example, `../includes/snippets/csharp/call-cancelmediaprocessing-csharp-snippets.md` I'm not sure it exists yet but when I looked at other snippets they all had a comment `description: "Automatically generated file. DO NOT MODIFY"` at the beginning. So I'm curious how are those snippets generated?

frankpeng7:
Looks like some other file is broken
```

**Comment by frankpeng7** [comment]
Some other file contains a wrong path
```
Error: /api-reference/beta/resources/useridentity.md:
        Exception processing links.
      System.ArgumentException: Link Definition was null. Link text: !INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)
```
which broke one of the auto checks

**Comment by frankpeng7** [comment]
> this is direct copy from the beta version ?

@giftbowen yes but I removed `/app` path in request example and the banner that warns that path is deprecated.

> should we add the remarks from the mpass team in the description section?

What do you mean by remarks from the mpass team?

File: api-reference/v1.0/api/call-cancelmediaprocessing.md [hunk 1]
```diff
@@ -0,0 +1,154 @@
...
+        "@odata.id": "/communications/calls/57DAB8B1894C409AB240BD8BEAE78896/operations/0FE0623FD62842EDB4BD8AC290072CC5",
+        "@odata.etag": "W/\"54451\"",
+        "id": "0fe0623f-d628-42ed-b4bd-8ac290072cc5",
+        "clientContext": "d45324c1-fcb5-430a-902c-f20af696537c",
+        "status": "failed",
+        "resultInfo": {
+          "@odata.type": "#microsoft.graph.resultInfo",
+          "code": 400,
+          "subcode": 8508,
+          "message": "Action falied, the operation was cancelled."
```
jsandoval-msft, line 131: typo: "falied" should be "failed".

jsandoval-msft:
Found and noted one type.  Also, I'm a bit confused by the example for "Operation canceled for recordResponse". The example indicates the operation failed, but I'm assuming the cancel media processing succeeded, and the notification of failure is for the recordResponse operation correct? If so, we should simply mote it to be more clear.

api-reference/v1.0/api/call-cancelmediaprocessing.md
frankpeng7: will fix in next revision

frankpeng7:
@jsandoval-msft I fixed the typo. As for the 2nd example I think you're correct. Any idea how we can make it more clear?

api-reference/v1.0/api/call-cancelmediaprocessing.md
```diff
@@ -0,0 +1,154 @@
...
+description: "Cancels media processing for any in-progress PlayPrompt or RecordResponse operations."
+author: "ananmishr"
+localization_priority: Normal
+ms.prod: "cloud-communications"
+doc_type: apiPageType
+---
+
+# call: cancelMediaProcessing
+
+Namespace: microsoft.graph
```
giftbowen, line 12: add the remark provided by mpaas team here?

PR #9393, opened 2020-08-06
Create workbook-error-handling.md

grangeryy:
This will provide detailed error handling suggestions when encournter errors in Excel graph.


concepts/workbook-error-handling.md
```diff
@@ -0,0 +1,115 @@
...
+    "code": <Second-level error code>,
+    "message":<Second-level error message>,
+    "innerError" (optional):{ "@odata.type": "odata.error" }
+  }
+}
+```
+
+## Steps to handle error response
+
+To better handle errors occurs in Excel graph, here are the general steps we suggest to follow.
```
joezhao-msft, line 66: nit: occurs => occur or occuring?

concepts/workbook-error-handling.md
```diff
@@ -0,0 +1,115 @@
...
+
+### 1. Determine whether it is a long running operation error
+
+Before handling an error, the first step is to determine whether the error response is from long running operation pattern or regular pattern. A long running operation error will return a `200 OK` HTTP status code and `failed` operation status in response body, while regular error response will return corresponding HTTP error status code.
+
+### 2. Parse second-level error code
```
joezhao-msft, line 72: Should we add RichAPI error codes here as well?

concepts/workbook-error-handling.md
```diff
@@ -0,0 +1,115 @@
...
+| **notFoundUncategorized**             | The Graph client is not expected to resend the failed request.
+| **notImplementedUncategorized**            | The Graph client is not expected to resend the failed request.
+| **payloadTooLargeUncategorized**              | The Graph client is not expected to resend the failed request.
+| **serviceUnavailableUncategorized**      | The Graph client is not expected to resend the failed request.
+| **tooManyRequestsUncategorized**             | The Graph client is not expected to resend the failed request.
+| **transientFailure**           | The Graph client is not expected to resend the failed request.
+| **unauthorizedUncategorized**         | The Graph client is not expected to resend the failed request.
+| **unsupportedWorkbook**         | The Graph client is not expected to resend the failed request until the unsupported factors are removed.
+
+### 3. Parse top-level error code
```
joezhao-msft, line 101: should we mention that our top-level error codes is bound to status code?

concepts/workbook-error-handling.md
```diff
@@ -0,0 +1,115 @@
...
+
+```json
+{
+  "code": "string",
+  "message": "string",
+  "innerError": { "@odata.type": "odata.error" }
+}
+```
+
+The `innerError` object might recursively contain more `innerError` objects with additional, more specific error codes. For example, error object may contain more detailed error information in the second-level error code and message below.
```
joezhao-msft, line 50: Microsoft Graph documentation doesn't have the naming of "top-level"/"second-level" error code, should we align with them?

concepts/workbook-error-handling.md
grangeryy: modified.

concepts/workbook-error-handling.md
grangeryy: OK. Modified.

concepts/workbook-error-handling.md
grangeryy: I think it is fine since we provide a sample structure at the beginning and it could be easy to understand.

concepts/workbook-error-handling.md
grangeryy: Added this part based on discussion

concepts/toc.yml
```diff
@@ -364,10 +364,15 @@ items:
             href: excel-update-range-format.md
           - name: Write data to a workbook
             href: excel-write-to-workbook.md
-          - name: Best practices
-            href: workbook-best-practice.md
-          - name: Error codes
-            href: workbook-error-codes.md
+          - name: Resources
```
Lauragra, line 367: FYI, I removed this Resources node earlier, for a couple of reasons. We already have a Resources node in the TOC, so that can get confusing. It seems that the number of topics in this node is small enough that another child node isn't really warranted. Also, the deeper the TOC, the less discoverable the topics are (SEO). Can we remove this?

PR #9403, opened 2020-08-06
Renamed changelog areas to match taxonomy

dkershaw10:
Best effort to rename 3rd level sections so that conversion to JSON yields better results


Lauragra:
I removed the parentheticals in that file at least twice before; hopefully they won't come back this time....

PR #9469, opened 2020-08-12
Move POST /teams to v1.0

nkramer:
Everything here is cut-and-paste from the beta docs.


api-reference/v1.0/includes/snippets/csharp/convert-team-from-group-csharp-snippets.md
```diff
@@ -0,0 +1,60 @@
+---
+description: "Automatically generated file. DO NOT MODIFY"
```
Lauragra, line 2: Why is this PR adding these snippets? Generally the snippets are added during the weekly snippet autogeneration process.

laujan: Hi @Lauragra. I'm helping Nick out with this PR. Should I remove all of the snippet files?

Lauragra: Yes, it looks like they are removed now so we're good to merge. Thanks.

PR #9487, opened 2020-08-12
Update user-post-users.md

RideMo:
I made suggested edits to the JSON - in the first user create POST, the original example did not set the password "forceChangePasswordNextSignIn" to false, and it was also missing the "passwordPolicies": "DisablePasswordExpiration"   element
------- 
cc: @krbain


Lauragra:
@RideMo , can you make the change to the beta topic as well? Thanks.

PR #9501, opened 2020-08-13
Explicit `userAgentApplication` passed to mgt

MaJaHa95:
Docs for PR: https://github.com/microsoftgraph/microsoft-graph-toolkit/pull/576

Probably needs some wordsmithing. It's late, I'm tired. But this should hit the key points as a draft.

I wasn't sure on the best way to communicate the polymorphic nature of `MsalConfig`. It seems weird to me having the two interface declarations with the same name, but I didn't want to make up names either.

I also wasn't sure whether this repo supports `[!Note]` syntax. It didn't preview, and I can't find examples in the Graph docs, but one of those felt appropriate for the "be sure to understand collisions" paragraph.

### DIFF
--- a/concepts/toolkit/providers/msal.md
+++ b/concepts/toolkit/providers/msal.md
@@ -47,7 +47,11 @@ import {UserAgentApplication} from "msal";
 Providers.globalProvider = new MsalProvider(config: MsalConfig);
 ```
 
-where MsalConfig is:
+There are two configuration options for the `MsalProvider` constructor parameter:
+
+#### Provide a `clientId` to create a new `UserAgentApplication`
+
+This option makes sense when Graph Toolkit is responsible for all authentication in your application.
 
 ```ts
 interface MsalConfig {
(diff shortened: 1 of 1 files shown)


concepts/toolkit/providers/msal.md
```diff
@@ -61,7 +65,20 @@ interface MsalConfig {
 }
 ```
 
-You must provide a `clientId` (to create a new `UserAgentApplication`).
+#### Pass an existing `UserAgentApplication` in the `userAgentApplication` property.
```
shweaver-MSFT, line 68: How about this?

#### Configure the provider with an existing `UserAgentApplication`

MaJaHa95:
@shweaver-MSFT Agreed about the interface names. I'm fine with `MsalUserAgentApplicationConfig` and `MsalConfig` since they line up with the exposed contract, which does seem fairly important to my more rested brain.

My hesitation last night stemmed from `MsalConfig`, and the fact that its name isn't more descriptive now. I left it unchanged to avoid a breaking change, but it feels weird to say "you can either pass an `MsalUserAgentApplicationConfig`, or a `MsalConfig`." Especially when the latter actually has _more_ functionality. I wish we could call it `MsalClientIdConfig` or something, but I don't really think it's worth introducing a type alias just for that, so here we are.

concepts/toolkit/providers/msal.md
MaJaHa95: Agreed. I tried to name the property in the header, to keep some consistency with the original ("provide a `clientId` to create..."), but I think it's more than intuitive that someone could quickly glance at the interface and immediately know how to do it, and this is much cleaner.

MaJaHa95:
What if I flip it around and do something more like [the MSAL docs themselves](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-js-sso), and provide a code sample for each? Then we don't have to name the interfaces at all, and we'd have the added advantage of providing something that's, for better or worse, copy-pastable.

>#### Provide a clientId to create a new UserAgentApplication
>This option makes sense when Graph Toolkit is responsible for all authentication in your application.
>
> ```ts
> import {Providers, MsalProvider, LoginType} from '@microsoft/mgt'
> import {UserAgentApplication} from "msal";
> 
> const config = {
>     clientId: "11111111-1111-1111-111111111111",
>     authority: "https://login.microsoftonline.com/common",
>     redirectUri: "https://localhost:3001",
>     scopes: [ 'user.read' ],
>     loginType: LoginType.Redirect,
>     loginHint: preferred_username,
>     options: {
>         // Pass additional MSAL configuration
>         cacheLocation: "localStorage"
>     }
> };
> 
> Providers.globalProvider = new MsalProvider(config);
> ```

And another for the new stuff.

It'd be a bit more verbose than what's already there, but maybe more clear? It just doesn't make it clear what's optional versus required, which is unfortunate.

shweaver-MSFT (approved):

PR #9532, opened 2020-08-14
Add `uniquename` property to application resource

sureshja:


sureshja:
@davidmu1 & @CelesteDG - You already reviewed this as part of the MSA developer support PR. I have moved `uniqueName` support as a separate PR to independently roll this out. The text I used here already includes your suggestion.

davidmu1:
@sureshja - I think there are a couple more things that need to be added to the PR. I believe the uniqueName property should be added to the JSON representation and a changelog entry should be added. Also, will this property be returned on GET call? If so, it should probably be added to those articles too.

api-reference/beta/resources/application.md
```diff
@@ -86,6 +86,7 @@ This resource supports using [delta query](/graph/delta-query-overview) to track
...
 | tokenEncryptionKeyId       | String                                                         | Specifies the keyId of a public key from the keyCredentials collection. When configured, Azure AD encrypts all the tokens it emits by using the key this property points to. The application code that receives the encrypted token must use the matching private key to decrypt the token before it can be used for the signed-in user.                                                                                                                                                                                                                     |
+| uniqueName | String | The unique identifier for the application that's assigned to it when it's created through Azure Resource Manager. Immutable. Read-only. |
```
CelesteDG, line 89:
```suggestion
| uniqueName | String | The unique identifier that's assigned to an application when the application is created through Azure Resource Manager. Immutable. Read-only. |
```

api-reference/v1.0/resources/application.md
```diff
@@ -79,6 +79,7 @@ This resource supports using [delta query](/graph/delta-query-overview) to track
 | signInAudience | String | Specifies what Microsoft accounts are supported for the current application. Supported values are:<ul><li>`AzureADMyOrg`: Users with a Microsoft work or school account in my organization’s Azure AD tenant (i.e. single tenant)</li><li>`AzureADMultipleOrgs`: Users with a Microsoft work or school account in any organization’s Azure AD tenant (i.e. multi-tenant)</li> <li>`AzureADandPersonalMicrosoftAccount`: Users with a personal Microsoft account, or a work or school account in any organization’s Azure AD tenant</li></ul> |
 | tags |String collection| Custom strings that can be used to categorize and identify the application. Not nullable.|
 | tokenEncryptionKeyId |String|Specifies the keyId of a public key from the keyCredentials collection. When configured, Azure AD encrypts all the tokens it emits by using the key this property points to. The application code that receives the encrypted token must use the matching private key to decrypt the token before it can be used for the signed-in user.|
+| uniqueName | String | The unique identifier for the application that's assigned to it when it's created through Azure Resource Manager. Immutable. Read-only. |
```
CelesteDG, line 82:
```suggestion
| uniqueName | String | The unique identifier that's assigned to an application when the application is created through Azure Resource Manager. Immutable. Read-only. |
```

lramosvea:
Added the schema live label since I can see the uniquename property in the live metadata for beta. However, I didn't see this property in the v1.0 live metadata.

PR #9543, opened 2020-08-17
Update teamsapp-publish.md

JarbasHorst:
Fixed http response example


Lauragra:
@JarbasHorst , I don't see a change in this PR?

PR #9552, opened 2020-08-18
Update directoryaudit-list.md

SarahBar:
"createdDateTime" is not a valid query parameter. Changing example to show "activityDateTime" which is the correct syntax.

### DIFF
--- a/api-reference/v1.0/api/directoryaudit-list.md
+++ b/api-reference/v1.0/api/directoryaudit-list.md
@@ -37,7 +37,7 @@ This method supports the following OData query parameters to help customize the 
 
 | Parameter                                                       | Description                                                                   | Example                                                                     |
 | :--------------------------------------------------------- | :---------------------------------------------------------------------------- | :-------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


Lauragra:
@SarahBar , please update the beta version of the topic as well. Thanks.

SarahBar:
> @SarahBar , please update the beta version of the topic as well. Thanks.

@Lauragra , I don't see anywhere to update the beta version of the topic. The word "beta" does not appear anywhere in the text of the topic, nor in the URL. Can you either make the change or explain what you mean in more detail?

Lauragra:
> > @SarahBar , please update the beta version of the topic as well. Thanks.
> 
> @Lauragra , I don't see anywhere to update the beta version of the topic. The word "beta" does not appear anywhere in the text of the topic, nor in the URL. Can you either make the change or explain what you mean in more detail?

Yep, there's a /beta folder within the api-reference folder that includes the files for the beta reference. Because APIs ship to beta before they ship to v1.0, any API docs in the v1.0 folder have a comparable topic in the beta folder, and they should be kept in sync (with the exception of changes applied to beta that aren't yet in v1.0). This is the location of the beta version of the topic you updated in this PR: https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/directoryaudit-list.md

Issue #3877, opened 2019-01-28
Provide an example of a 429 response from Graph for supported resources


somehowadev:
Like the [create contact](https://docs.microsoft.com/en-us/graph/api/user-post-contacts?view=graph-rest-1.0) documentation, it'd be helpful if an example of a 429 response was actually shown in the documentation.

Thanks,
Z

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8df32e5f-8ccb-0c93-28d9-02392f9c7256
* Version Independent ID: 4aadd084-fb61-e89b-f63f-3b71622636b7
* Content: [Microsoft Graph throttling guidance - Microsoft Graph](https://docs.microsoft.com/en-us/graph/throttling)
* Content Source: [concepts/throttling.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/concepts/throttling.md)
* GitHub Login: @piotrci
* Microsoft Alias: **jthake-msft**

daverant:
Adding to this request. Graph seems to be returning invalid retry-after header values. Fiddler traces show I'm receiving 429 Too Many Requests error codes from batched requests sent to the v1.0/$batch endpoint, but the Retry-After header I'm receiving are all decimals (0.249, 0.124 etc). HTTP spec says positive integers or timestamps are valid, but I'm receiving a decimal. Not sure how to handle it so some documentation would be useful! I get an exception when calling

`new HttpResponseMessage().Headers.Add("Retry-After","0.256");`

but absolutely fine when I call

`new HttpResponseMessage().Headers.Add("Retry-After","120");`

as an example of a valid Retry-After header.

![Batch Response](https://pbs.twimg.com/media/DxwoZVlWoAAthrW.jpg)

daverant:
> Adding to this request. Graph seems to be returning invalid retry-after header values. Fiddler traces show I'm receiving 429 Too Many Requests error codes from batched requests sent to the v1.0/$batch endpoint, but the Retry-After header I'm receiving are all decimals (0.249, 0.124 etc). HTTP spec says positive integers or timestamps are valid, but I'm receiving a decimal. Not sure how to handle it so some documentation would be useful! I get an exception when calling
> 
> `new HttpResponseMessage().Headers.Add("Retry-After","0.256");`
> 
> but absolutely fine when I call
> 
[...]

PR #9559, opened 2020-08-18
- fixes: #3877 adds a throttling response sample

baywet:
fixes: #3877 adds a throttling response sample

### DIFF
--- a/concepts/throttling.md
+++ b/concepts/throttling.md
@@ -31,7 +31,7 @@ The most common causes of throttling of clients include:
 
 ## Sample response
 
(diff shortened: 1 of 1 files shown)


concepts/throttling.md
```diff
@@ -29,6 +29,30 @@ The most common causes of throttling of clients include:
 - A large number of requests across all applications in a tenant.
 - A large number of requests from a particular application across all tenants.
 
+## Sample response
+
+Whenever the throttling threshold is exceeded, Microsoft Graph will respond with a response similar to this one.
```
davidmu1, line 34:
```suggestion
Whenever the throttling threshold is exceeded, Microsoft Graph responds with a response similar to this one.
```

PR #9560, opened 2020-08-18
- fixes #4694: documents batching in combinatino with throttling

### LINKED ISSUE #4694: Batch Request Throttling Behaviour
This page does not cover the requests batching behaviour. 

As far as I can understand, because one batch request is multiplexed to multiple requests, the behaviour is similar to the one explained in the page, but:

* How long should we wait when a batch request is throttled? (e.g. the sum of all Retry-After reply headers? Double of the maximum Retry-After response headers?)
* The batch request itself, that always returns 200, affects the throttling behaviour?
* How should we manage batch request that continue to fail indefinitely? 

Thanks.

---
#### Document Details

* ID: 8df32e5f-8ccb-0c93-28d9-02392f9c7256
* Version Independent ID: 4aadd084-fb61-e89b-f63f-3b71622636b7
* Content: [Microsoft Graph throttling guidance - Microsoft Graph](https://docs.microsoft.com/en-us/graph/throttling)
* Content Source: [concepts/throttling.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/concepts/throttling.md)
* GitHub Login: @piotrci

baywet:
fixes #4694: documents batching in combinatino with throttling

### DIFF
--- a/concepts/throttling.md
+++ b/concepts/throttling.md
@@ -59,6 +59,12 @@ Programming patterns like continuously polling a resource to check for updates a
 >[!NOTE]
 >[Best practices for discovering files and detecting changes at scale](https://docs.microsoft.com/onedrive/developer/rest-api/concepts/scan-guidance?view=odsp-graph-online) describes best practices in details.
 
+## Throttling and batching
+
+[JSON batching](./json-batching.md) allows you to optimize your application by combining multiple requests into a single JSON object. Requests in a batch are evaluated individually against throttling limits and should any request be exceed limits, it will fail individually with a `status` of `429` and an error similar to the one provided above. The batch itself will fail with status code `424` (Failed Dependency). It is possible multiple requests get throttled in a single batch. You should retry each failed request from the batch after the value provided in the `retry-after` response header from the JSON content. You may retry all the failed requests in a new batch after the longest `retry-after` value.
+
+If SDKs retry throttled requests automatically when they are not batched, throttled requests that were part of a batch are not retried automatically.
+
 ## Service-specific limits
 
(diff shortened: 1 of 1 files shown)


concepts/throttling.md
```diff
@@ -59,6 +59,12 @@ Programming patterns like continuously polling a resource to check for updates a
 >[!NOTE]
 >[Best practices for discovering files and detecting changes at scale](https://docs.microsoft.com/onedrive/developer/rest-api/concepts/scan-guidance?view=odsp-graph-online) describes best practices in details.
 
+## Throttling and batching
+
+[JSON batching](./json-batching.md) allows you to optimize your application by combining multiple requests into a single JSON object. Requests in a batch are evaluated individually against throttling limits and should any request be exceed limits, it will fail individually with a `status` of `429` and an error similar to the one provided above. The batch itself will fail with status code `424` (Failed Dependency). It is possible multiple requests get throttled in a single batch. You should retry each failed request from the batch after the value provided in the `retry-after` response header from the JSON content. You may retry all the failed requests in a new batch after the longest `retry-after` value.
```
davidmu1, line 64:
```suggestion
[JSON batching](./json-batching.md) allows you to optimize your application by combining multiple requests into a single JSON object. Requests in a batch are evaluated individually against throttling limits and if any request exceeds the limits, it fails with a `status` of `429` and an error similar to the one provided above. The batch itself fails with a status code of `424` (Failed Dependency). It is possible for multiple requests to be throttled in a single batch. You should retry each failed request from the batch using the value provided in the `retry-after` response header from the JSON content. You may retry all the failed requests in a new batch after the longest `retry-after` value.
```

PR #9572, opened 2020-08-19
[Permissions] Adding our ShortNotes scopes to list of permissions

Corvids:
Adding documentation for our ShortNotes scopes

### DIFF
--- a/concepts/changelog.md
+++ b/concepts/changelog.md
@@ -19,12 +19,6 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | :-------------- | :------------ | :--------------------------------------- |
 | Addition | beta | Added the **lobbyBypassSettings**, **isEntryExitAnnounced**, and **allowedPresenters** property to the [onlineMeeting](/graph/api/resources/onlinemeeting?view=graph-rest-beta) entity.|
 
-### Scopes
-
-| **Change type** | **Version**   | **Description**                          |
-| :-------------- | :------------ | :--------------------------------------- |
-| Addition | v1.0 | The following permission scopes have been added:<ul><li>ShortNotes.Read</li><li>ShortNotes.ReadWrite</li><li>ShortNotes.Read.All</li><li>ShortNotes.ReadWrite.All</li></ul>See the [permissions reference](/graph/permissions-reference#short-notes-permissions) for details. |
-
 ### Devices and apps | Cloud printing
 
 | **Change type** | **Version**   | **Description**                          |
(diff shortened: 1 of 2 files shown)


concepts/permissions-reference.md
```diff
@@ -1577,6 +1577,34 @@ Security permissions are valid only on work or school accounts.
 
 ---
 
+## Short Notes permissions
```
sasuppal, line 1580: We should mark the permissions as `private preview`. See ChatMessage as an example

Corvids:
@Lauragra could we get a "ready to merge" tag added to this PR if it looks okay to merge? Thank you!

concepts/changelog.md
```diff
@@ -17,6 +17,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | **Change type** | **Version**   | **Description**                          |
 | :-------------- | :------------ | :--------------------------------------- |
 | Addition | Beta | Added the [passwordSingleSignOnSettings](/graph/api/resources/passwordsinglesignonsettings?view=graph-rest-beta) property to the [servicePrincipal](/graph/api/resources/serviceprincipal?view=graph-rest-beta) resource.
+| Addition | v1.0 | The following permission scopes have been added:<ul><li>ShortNotes.Read</li><li>ShortNotes.ReadWrite</li><li>ShortNotes.Read.All</li><li>ShortNotes.ReadWrite.All</li></ul>See the [permissions reference](/graph/permissions-reference#short-notes-permissions) for details. |
```
Lauragra, line 20: Please move the changelog entry to September.

Lauragra:
> @Lauragra could we get a "ready to merge" tag added to this PR if it looks okay to merge? Thank you!

If you're a contributor to the repo, you can set the label. Please set the label when the permissions are available in the Azure Portal. We're documenting these permissions because they're visible to users, right? They don't currently apply to any APIs? Also, please see my comment - we should add the changelog entry to September instead of August. Thanks.

concepts/changelog.md
Corvids: Done. I've moved the changes under September 2020 > Scopes to reflect the other items in the changelog.

Corvids:
> > @Lauragra could we get a "ready to merge" tag added to this PR if it looks okay to merge? Thank you!
> 
> If you're a contributor to the repo, you can set the label. Please set the label when the permissions are available in the Azure Portal. We're documenting these permissions because they're visible to users, right? They don't currently apply to any APIs? Also, please see my comment - we should add the changelog entry to September instead of August. Thanks.

Thank you, @Lauragra. I've made the change to add the entry under September. I'm happy to set the label but I don't see anything about Azure Portal nor can I update the tags on this Github page.

These scopes don't currently apply to any other APIs and they should be visible to users.

concepts/changelog.md
Lauragra: I'm actually going to suggest that we remove the changelog entry, because the permissions are in private preview and they don't apply to any of the workloads that are currently available in Graph. I'll just remove the changelog entry.

PR #9605, opened 2020-08-21
Add userPurpose API to Graph

abheek-das:
Adding userPurpose property to mailboxSettings resource to Microsoft Graph v1.0


api-reference/v1.0/resources/userpurpose.md
```diff
@@ -0,0 +1,64 @@
...
+|Property|Type|Description|
+|:---------------|:--------|:----------|
+|value|[userPurpose](#userpurpose-values)|Represents the user's recipient or mailbox type in Exchange Online. Possible values are: `unknown`, `user`, `linked`, `shared`, `room`, `equipment`, and `others`. See the next section for more information.|
+
+### userPurpose values
+|Member|Description|
+|:---------------|:--------|
+|unknown|No information found about the mailbox.|
+|user|A user account with a mailbox in the local forest.|
+|linked|A mailbox linked to a user account in another forest.|
```
angelgolfer-ms, line 28: Line 27 and 28:
@abheek-das , can you educate and confirm for me if the term "forest" applies to mailbox servers that are Exchange Online? 
All the search results I have read in the last hour only talk about "forests" in Exchange Servers (on-prem). 
I would imagine for line 27, "user" would have to refer to a user mailbox on Exchange Online? 
So if you confirm that the term "forest" can apply to either on-prem or online servers, that I guess that description is fine.

As for linked mailboxes, I can only find docs that describe them to be on an Exchange resource forest (on-prem):
https://docs.microsoft.com/en-us/Exchange/recipients/linked-mailboxes?view=exchserver-2019

I suppose even if "linked" mailboxes exist only on-prem servers, that description is ok. I would suggest we provide a link like this:
```
See more information about [linked mailboxes](https://docs.microsoft.com/Exchange/recipients/linked-mailboxes?view=exchserver-2019).
```

api-reference/v1.0/resources/userpurpose.md
```diff
@@ -0,0 +1,64 @@
...
+|shared|A mailbox shared by two or more user accounts.|
+|room|A mailbox representing a conference room.|
```
angelgolfer-ms, line 30: Line 29:
I suggest for "shared", we provide a link to this end-user topic (which links to the IT pro topic): https://docs.microsoft.com/en-us/exchange/collaboration-exo/shared-mailboxes

api-reference/v1.0/resources/userpurpose.md
```diff
@@ -0,0 +1,64 @@
...
+|equipment|A mailbox representing a piece of equipment.|
```
angelgolfer-ms, line 31: Line 31:
Perhaps you can add to the end of the description: "such as a camera or touch-screen whiteboard."

api-reference/v1.0/resources/userpurpose.md
```diff
@@ -0,0 +1,64 @@
...
+## Properties
+|Property|Type|Description|
+|:---------------|:--------|:----------|
+|value|[userPurpose](#userpurpose-values)|Represents the user's recipient or mailbox type in Exchange Online. Possible values are: `unknown`, `user`, `linked`, `shared`, `room`, `equipment`, and `others`. See the next section for more information.|
+
+### userPurpose values
+|Member|Description|
+|:---------------|:--------|
+|unknown|No information found about the mailbox.|
+|user|A user account with a mailbox in the local forest.|
```
angelgolfer-ms, line 27: Line 27:
For "user", perhaps the first part of the description should be "A user mailbox..." rather than "A user account with a mailbox..."?
The account to mailbox mapping is many-to-one, right (a mailbox supports one or more accounts). Since we don't drill down to the account level for the other enum values like room or equipment, it's not necessary to relate to "account" for user? Or am I missing something?
On the other hand, for an unfamiliar customer, the "account with a mailbox" part may lead to the question whether the user account can be associated with more than one mailbox; that would essentially change the many-to-one mapping of accounts-to-mailbox.

concepts/changelog.md
```diff
@@ -157,6 +157,13 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | Addition | v1.0 | Added the following methods to the [channel](/graph/api/resources/channel) resource: [List channel members](/graph/api/channel-list-members), [Get channel member](/graph/api/channel-get-members), [Add channel member](/graph/api/channel-post-members),  [Update channel member's role](/graph/api/channel-update-members), [Remove channel member](/graph/api/channel-delete-members).|
 
 
+### Users | Outlook settings
+
+| **Change type** | **Version**   | **Description**                          |
+| :-------------- | :------------ | :--------------------------------------- |
+|Addition | v1.0 | Added the new **userPurpose** property to [mailboxSettings](/graph/api/resources/mailboxsettings?view=graph-rest-v1.0). See [userPurpose resource type](/graph/api/resources/userPurpose?view=graph-rest-beta) for information on the supported use cases.|
+|Addition | v1.0 | Added new [userPurpose](/graph/api/resources/userPurpose?view=graph-rest-v1.0) enum type.|
+
```
angelgolfer-ms, line 166: Hi @abheek-das , there are 4 "suggestions" that OPS is returning for lines 20, 164 (x2) and 165. It's about the current way to use this type of links called site-relative link. In particular, you now have to add one more query parameter "preserve-view=true" to each link, IF you use the ?view query parameter. OPS help here: https://review.docs.microsoft.com/en-us/help/contribute/links-how-to?branch=master#links-to-specific-versions

Also, in line 164, the second link should be going to v1, not beta?

Then FYI, because your links in these lines are going to v1, technically you don't have to use the ?view query parameter at all. The default behavior is going to v1 anyway. So for these 3 links going to v1, you can drop all query parameters. Then you can fix the "suggestions" too; later on OPS will make these warnings, and then later on, they will be errors. So dropping the ?view parameter now is appropriate.

PR #9611, opened 2020-08-21
Added guidance on using beta with SDKs

### LINKED ISSUE #8912: No mention of Beta nuget package
I think it would be a good idea to mention that the user will need to install the Microdoft.Graph.Beta nuget in order to access some of these functions.

For example, I was trying to find the '.Authentication' property on  IUserRequest builder in order to do a password reset, and I only found out about the beta package from a Stack Overflow answer. A note at the top level would have saved me some time.

---
#### Document Details

* ID: 28f291cb-b55b-4b55-c42a-cec61e5c4384
* Version Independent ID: afc8208d-990b-098f-0508-fcd5745d35e6
* Content: [Microsoft Graph beta endpoint reference - Microsoft Graph beta](https://docs.microsoft.com/en-us/graph/api/overview?toc=./ref/toc.json&view=graph-rest-beta)
* Content Source: [api-reference/beta/overview.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/overview.md)
* GitHub Login: @angelgolfer-ms

jasonjoh:
Fixes #8912


concepts/sdks/use-beta.md
```diff
@@ -0,0 +1,69 @@
+---
+title: "Use the Microsoft Graph SDKs with the beta API"
```
Lauragra, line 2: Do we need to surface this topic in the SDKs section of the TOC?

jasonjoh: Oh yeah, forgot about that. 😫

PR #9617, opened 2020-08-23
Updated iCalUId documentation in v1

Bhupesh12Hub:
@angelgolfer-ms Please review the PR for iCalUID definition update in v1.0.

Thanks,
Bhupesh


concepts/changelog.md
```diff
@@ -22,6 +22,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
 
 | **Change type** | **Version**   | **Description**                          |
 | :-------------- | :------------ | :--------------------------------------- |
+| Change | v1.0 | Updated the description for **iCalUId** property to the [event](/graph/api/resources/event?view=graph-rest-1.0) entity. |
```
angelgolfer-ms, line 25: @Harini84 , is this doc update just reflecting how the iCalUId property has actually always  behaved, that is, correcting the docs?
Seems like from the beta description, this property will eventually go away, so it's not likely that at this point your team would make a change to the property behavior (in v1 too).

If that's the case, perhaps we should say "Corrected the property description to reflect the behavior of the **iCalUId** property of event...", just to be clear.

api-reference/v1.0/resources/event.md
```diff
@@ -69,7 +69,7 @@ by providing a [delta](../api/event-delta.md) function.
 |createdDateTime|DateTimeOffset|The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: `'2014-01-01T00:00:00Z'`|
 |end|[dateTimeTimeZone](datetimetimezone.md)|The date, time, and time zone that the event ends. By default, the end time is in UTC.|
 |hasAttachments|Boolean|Set to true if the event has attachments.|
-|iCalUId|String|A unique identifier that is shared by all instances of an event across different calendars. Read-only.|
+|iCalUId|String|A unique identifier across calendars which is different for each instance in a series. Read-only.|
```
angelgolfer-ms, line 72: Line 72:
I guess the phrase you're replacing with is from the uid property description in the beta docs?
I'd suggest this:
"A unique identifier for an event across calendars. This ID is different for each occurrence in a recurring series."

Reasons:
- The "which" comes right after "calendars", which can be ambiguous especially for localizers, whether the "which" phrase applies to the identifier to to calendar.
- In the beta docs, the entire description for uid includes "recurring events" and "occurrences". These are the appropriate terms to use when documenting recurring events/ series. In this case, please apply these terms on first reference.

concepts/changelog.md
Harini84: @angelgolfer-ms 
Yes, the documentation is being updated to reflect the behavior. We took the initiative to eventually deprecate but considering all the feedback we got clients have lot of dependency on this property and hence cannot be deprecated.
We need to restore the iCalUId in MS-Graph beta which we plan to do in upcoming sprints.

Hope this helps.

api-reference/v1.0/resources/event.md
Bhupesh12Hub: Updated

Bhupesh12Hub:
@angelgolfer-ms I have updated the file, please review.

concepts/changelog.md
angelgolfer-ms: Thanks for the clarification and heads-up, @Harini84 !

angelgolfer-ms: @Bhupesh12Hub ,
For the changelog item, can you make two changes:
- Please move the description to the September section of the changelog.
- Please change the description per my last comment.

Thanks.

Bhupesh12Hub:
@angelgolfer-ms 

Files has been updated as per the comment. Please review.

Thanks,
Bhupesh

angelgolfer-ms (approved):
@Bhupesh12Hub , please add "the" before "event" in the changelog description. We're then good to publish.

concepts/changelog.md
```diff
@@ -17,6 +17,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
 
 | **Change type** | **Version**   | **Description**                          |
 | :-------------- | :------------ | :--------------------------------------- |
+| Change | v1.0 | Corrected the property description to reflect the behavior of the **iCalUId** property of the [event](/graph/api/resources/event?view=graph-rest-1.0) entity. |
```
Lauragra, line 20: This is just an update to the documentation, not the actual API, so we don't need to include it in the changelog. I'll go ahead and remove this entry and then publish the update.

PR #9622, opened 2020-08-24
Updated descriptions of ids

dkershaw10:
Updated a couple of descriptions to clarify that IDs aren't always GUIDs and that developers should treat them as opaque identifiers.

### DIFF
--- a/api-reference/beta/resources/directoryobject.md
+++ b/api-reference/beta/resources/directoryobject.md
@@ -36,7 +36,7 @@ This resource supports:
 
 | Property   | Type |Description|
 |:---------------|:--------|:----------|
-|id|String|A Guid that is the unique identifier for the object; for example, 12345678-9abc-def0-1234-56789abcde12. Key. Not nullable. Read-only.|
+|id|String|The unique identifier for the object. For example, 12345678-9abc-def0-1234-56789abcde. NOTE: **id** is often but not exclusively in the form of a GUID; treat it as an opaque identifier and do not rely on it being a GUID. Key. Not nullable. Read-only.|
 
 ## Relationships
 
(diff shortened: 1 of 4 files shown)


api-reference/beta/resources/directoryobject.md
```diff
@@ -36,7 +36,7 @@ This resource supports:
 
 | Property   | Type |Description|
 |:---------------|:--------|:----------|
-|id|String|A Guid that is the unique identifier for the object; for example, 12345678-9abc-def0-1234-56789abcde12. Key. Not nullable. Read-only.|
+|id|String|The unique identifier for the object. For example, 12345678-9abc-def0-1234-56789abcde. NOTE: **id** is often but not exclusively in the form of a GUID; treat it as an opaque identifier and do not rely on it being a GUID. Key. Not nullable. Read-only.|
```
davidmu1, line 39:
```suggestion
|id|String|The unique identifier for the object. For example, 12345678-9abc-def0-1234-56789abcde. The value of the **id** property is often but not exclusively in the form of a GUID; treat it as an opaque identifier and do not rely on it being a GUID. Key. Not nullable. Read-only.|
```

api-reference/v1.0/resources/directoryobject.md
```diff
@@ -30,7 +30,7 @@ Represents an Azure Active Directory object. The **directoryObject** type is the
 
 | Property   | Type |Description|
 |:---------------|:--------|:----------|
-|id|String|A Guid that is the unique identifier for the object; for example, 12345678-9abc-def0-1234-56789abcde. Key. Not nullable. Read-only.|
+|id|String|The unique identifier for the object. For example, 12345678-9abc-def0-1234-56789abcde. NOTE: **id** is often but not exclusively in the form of a GUID; treat it as an opaque identifier and do not rely on it being a GUID. Key. Not nullable. Read-only.|
```
davidmu1, line 33:
```suggestion
|id|String|The unique identifier for the object. For example, 12345678-9abc-def0-1234-56789abcde. The value of the **id** property is often but not exclusively in the form of a GUID; treat it as an opaque identifier and do not rely on it being a GUID. Key. Not nullable. Read-only.|
```

PR #9653, opened 2020-08-25
Updating mail property of user as writable.

atriagrawal:
As part of External Identities project, the user's "mail" property was made writable. This change is to update the public docs with this information.


api-reference/beta/api/user-update.md
```diff
@@ -63,6 +63,7 @@ In the request body, supply the values for relevant fields that should be update
 |identities|[objectIdentity](../resources/objectidentity.md) collection| Represents the identities that can be used to sign in to this user account. An identity can be provided by Microsoft, by organizations, or by social identity providers such as Facebook, Google, and Microsoft, and tied to a user account. Any update to **identities** will replace the entire collection and you must supply the userPrincipalName **signInType** identity in the collection.|
 |interests|String collection|A list for the user to describe their interests.|
 |jobTitle|String|The user’s job title.|
+|mail|String|The SMTP address for the user, for example, "jeff@contoso.onmicrosoft.com". <br><br>Returned by default. Supports $filter.|
```
eketo-msft, line 66: Should we mention somewhere that the update will also reflect in the proxyAddresses collection?  Also, what is the error scenario when mail can't be updated?  I thought there was a scenario where it would error, like non-B2B or something?

eketo-msft:
Can `mail` be set on create as well?

api-reference/beta/api/user-update.md
atriagrawal: Sure, I will update the doc to include proxyAddresses change as well. This change was implemented for B2X, so it is not specific for B2B. The only case I know of where it would fail is where user is dirsync enabled and proxyAddresses is an on-prem owned property.

eketo-msft: Sounds good.  I'm not sure we need to call out when it would fail as long as the error message is prescriptive.

atriagrawal:
> Can `mail` be set on create as well?
Yes, it can be.

> Can you also specify mail on create requests?
I looked at the create user document, and it only talks about the required fields. For creating users, I believe user.md would be used to check what all fields could be sent during create.

eketo-msft:
> > Can `mail` be set on create as well?
> > Yes, it can be.
> 
> > Can you also specify mail on create requests?
> > I looked at the create user document, and it only talks about the required fields. For creating users, I believe user.md would be used to check what all fields could be sent during create.

Great, if its not required then lets not add it to the sample and user.md is fine.  Thanks for confirming!

api-reference/beta/api/user-update.md
atriagrawal: Yes, the error message is very clear. Here is an example:

`{"error":{"code":"directory_error","message":"Unable to update the specified properties for on-premises mastered Directory Sync objects or objects currently undergoing migration.","response":"error_bad_directory_request","target":"ProvisionKey","clientrequestid":"0fe727d1-d70a-421f-aeda-83d1a8be50a3","time":"07-31-2020 16:15:14Z","innererror":{"trace":null,"context":null}}}`

agrabhi, line 66: Should we say primary SMTP address? Or there could be only one? 
Do we need to link to any doc which can explain this field in directory?

api-reference/v1.0/resources/user.md
```diff
@@ -146,7 +146,7 @@ This resource supports:
 |lastPasswordChangeDateTime| DateTimeOffset | The time when this Azure AD user last changed their password. The date and time information uses ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 would look like this: '2014-01-01T00:00:00Z'|
 |legalAgeGroupClassification|String| Used by enterprise applications to determine the legal age group of the user. This property is read-only and calculated based on `ageGroup` and `consentProvidedForMinor` properties. Allowed values: `null`, `minorWithOutParentalConsent`, `minorWithParentalConsent`, `minorNoParentalConsentRequired`, `notAdult` and `adult`. Refer to the [legal age group property definitions](#legal-age-group-property-definitions) for further information.)|
 |licenseAssignmentStates|[licenseAssignmentState](licenseassignmentstate.md) collection|State of license assignments for this user. Read-only.|
-|mail|String|The SMTP address for the user, for example, "jeff@contoso.onmicrosoft.com". Read-Only. Supports $filter.|
```
agrabhi, line 149: should we also update the userPrincipalName format which is auto generated for social users in AAD?

api-reference/beta/api/user-update.md
atriagrawal: There could only be one Primary SMTP address and the caps "SMTP" denotes primary SMTP address.

api-reference/v1.0/resources/user.md
atriagrawal: Yes, we can update that but needs to be done as a separate change.

atriagrawal:
> This looks like an API update that warrants a changelog entry. If so, please include that update with this PR.

Could you please point me to some documentation on what needs to be updated?

eketo-msft:
> > This looks like an API update that warrants a changelog entry. If so, please include that update with this PR.
> 
> @Lauragra Could you please point me to some documentation on what needs to be updated?

https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55053/Changelog has details

atriagrawal:
> > > This looks like an API update that warrants a changelog entry. If so, please include that update with this PR.
> > 
> > 
> > @Lauragra Could you please point me to some documentation on what needs to be updated?
> 
> https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55053/Changelog has details

Thanks @eketo-msft. @Lauragra, the documentation says that we need to
`provide a CSDL file or schema publication branch (such as the candidate-beta branch ) that defines the set of APIs you're adding.`

For this change there was no change to CSDL required. In such cases how do we generate ChangeLog entry?

eketo-msft:
> > > > This looks like an API update that warrants a changelog entry. If so, please include that update with this PR.
> > > 
> > > 
> > > @Lauragra Could you please point me to some documentation on what needs to be updated?
> > 
> > 
> > https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55053/Changelog has details
> 
> Thanks @eketo-msft. @Lauragra, the documentation says that we need to
> `provide a CSDL file or schema publication branch (such as the candidate-beta branch ) that defines the set of APIs you're adding.`
> 
> For this change there was no change to CSDL required. In such cases how do we generate ChangeLog entry?

Just manually edit the changelog.md to add to the table and state something along the lines of 'Updated the mail property on the user resource to be writeable'.

atriagrawal:
> Just manually edit the changelog.md to add to the table and state something along the lines of 'Updated the mail property on the user resource to be writeable'.

Done. Thanks!

PR #9665, opened 2020-08-26
Update tablerow-update.md

grangeryy:
the old doc is using the API in the wrong way. Update the correct method.


api-reference/v1.0/api/tablerow-update.md
```diff
@@ -53,13 +54,12 @@ Here is an example of the request.
   "name": "update_tablerow"
 }-->
 ```http
-PATCH https://graph.microsoft.com/v1.0/me/drive/items/{id}/workbook/tables/{id|name}/rows/{index}
+PATCH https://graph.microsoft.com/v1.0/me/drive/items/{drive-item-id}/workbook/tables('1')/rows/itemAt(index=0)
```
Lauragra, line 57: We are getting a malformed function parameter error on this request example. Should it be (index=`1`)?

Lauragra: Also, do you need to document the parameters? See the guidance here: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55072/Parameters

Lauragra, line 57: I don't think the syntax here is correct/accurate. It doesn't match the syntax in the HTTP request section. After /tables, there should be an id, is that what "(`1`)" is? Is itemAt a parameter, and if so, it should be in parentheses.

Lauragra:
@grangeryy , do you still want to merge this PR? If so, please look into the validation error.

Lauragra:
@grangeryy , do you still want to merge this PR? If so, please look into the validation error.

Lauragra:
@grangeryy , do you still want this PR merged? I'm setting it to do not merge for now until the validation failure is resolved.

PR #9689, opened 2020-08-27
User created date fix

JohnMaina:
Updating documentation after an update to a graph API v1.0.


changelog/Microsoft.DirectoryServices.json
```diff
@@ -1,5 +1,41 @@
 ﻿{
   "changelog": [
+    {
+      "ChangeList": [
```
Lauragra, line 4: This change is not part of this PR.

changelog/Microsoft.DirectoryServices.json
```diff
@@ -1,5 +1,41 @@
...
+      ],
+      "Id": "e1580140-e7a5-11ea-a541-8f56854a98ac",
+      "Cloud": "prd",
+      "Version": "v1.0",
+      "CreatedDateTime": "2020-08-26T14:10:22.448Z",
+      "WorkloadArea": "Identity and access (Azure AD)",
+      "SubArea": null
+    },
+    {
+      "ChangeList": [
```
Lauragra, line 22: We need to update the changelog.md file for this change. The filterable changelog that consumes the JSON snippets is not live on the site yet.

changelog/Microsoft.DirectoryServices.json
JohnMaina: I removed it in a new commit

changelog/Microsoft.DirectoryServices.json
JohnMaina: I have updated changelog.md.

The user resource was updated to add the property as well. The docs had been updated earlier that's why you can't see the change in this PR.

api-reference/v1.0/api/user-get.md
```diff
@@ -165,7 +165,7 @@ If you need a different property set, you can use the OData `$select` query para
 ##### Request
 <!-- { "blockType": "ignored" } -->
 ```http
-GET https://graph.microsoft.com/v1.0/users/{id | userPrincipalName}?$select=displayName,givenName,postalCode
+GET https://graph.microsoft.com/v1.0/users/{id | userPrincipalName}?$select=displayName,givenName,postalCode,createdDateTime
```
Lauragra, line 168: This request example is missing from the beta user-get topic. The beta topic should be consistent with v1.0. Please add this example to the beta topic as well.

concepts/changelog.md
```diff
@@ -18,6 +18,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | **Change type** | **Version** | **Description**                          |
 | :-------------- | :---------- | :--------------------------------------- |
 | Change | beta and v1.0 | Updated the mail property on the [user](/graph/api/resources/user) entity to be writeable. |
+| Addition | v1.0 |Added the **createdDateTime** property to the [user](/graph/api/resources/user) resource.|
```
Lauragra, line 21: Actually, the property already existed in user.md, you updated the GET topic to reflect it. In that case, do we need this changelog entry? Also, do you need to update the LIST topic to also show the property?

JohnMaina: In the docs, the property existed in user.md in both beta and v1.0 versions, but in ms graph schema, the property existed in beta only.   (was reported as bug by users) Also, in the docs, there s a changelog that added it to beta but not changelog for v1.0. That's why I added this changelog.

There was no change in GET topic code or schema. I only added this property to an existing example. We don't need to update LIST topic. 

So the actual change was ADDING THE PROPERTY TO v1.0 in ms graph schema.

api-reference/v1.0/api/user-get.md
JohnMaina: I added the example

api-reference/beta/api/user-get.md
```diff
@@ -174,6 +174,30 @@ Content-length: 491
 }
 ```
 
+### Example 3: Users request using $select
+
+If you need a different property set, you can use the OData `$select` query parameter. For example, to return _displayName_, _givenName_, and _postalCode_, you would use the add the following to your query `$select=displayName,givenName,postalCode`
+
+##### Request
```
Lauragra, line 182: Why  are we ignoring instead of validating these examples?

JohnMaina: I am not sure why the person who designed the example originally made it that way. I only extended the example to add createdDateTime

PR #9717, opened 2020-08-31
[Universal Print] Adding clarifications around print job redirection

braedenp-msft:


braedenp-msft:
@Lauragra This small change is ready for review 😊

concepts/universal-print-concept-overview.md
```diff
@@ -79,6 +79,8 @@ Follow these steps to enable pull printing:
 
 7. When the user selects one or more jobs to print, your application can [redirect the print job(s)](/graph/api/printjob-redirect?view=graph-rest-beta) to the physical printer and the job will start printing! The redirect call will only succeed if there is a [printTask](/graph/api/resources/printtask?view=graph-rest-beta) in `processing` state on the associated printer started by a trigger that this app created in step 4. The task will automatically be set to `completed` state after redirecting it.
 
+   **NOTE: Paused print jobs that are not redirected within 2 days will be deleted.**
```
nilakhan, line 82: @braedenp-msft missed adding this before approving,
Please also add that we will be standardizing it to default job behavior in future, which is (Incomplete jobs get aborted after 3 days, and all jobs get deleted after 10 days)

concepts/universal-print-concept-overview.md
```diff
@@ -79,12 +79,17 @@ Follow these steps to enable pull printing:
 
 7. When the user selects one or more jobs to print, your application can [redirect the print job(s)](/graph/api/printjob-redirect?view=graph-rest-beta) to the physical printer and the job will start printing! The redirect call will only succeed if there is a [printTask](/graph/api/resources/printtask?view=graph-rest-beta) in `processing` state on the associated printer started by a trigger that this app created in step 4. The task will automatically be set to `completed` state after redirecting it.
```
Lauragra, line 80: This doesn't need to block publishing, but this procedure might be more detailed than warranted for the concept overview topic; you might want to think about a separate how to article for it?

concepts/universal-print-concept-overview.md
```diff
@@ -79,12 +79,17 @@ Follow these steps to enable pull printing:
...
+   >**NOTE:** Paused print jobs that are not redirected within 2 days will be deleted.
+
 ## API reference
 Looking for the API reference for this service?
 
 - [Universal Print API in Microsoft Graph beta](/graph/api/resources/print?view=graph-rest-beta)
 
+## Feedback
+
+We'd love to hear your feedback about the Universal Print APIs. Provide your suggestions on [UserVoice](https://microsoftgraph.uservoice.com/forums/920506-microsoft-graph-feature-requests).
```
Lauragra, line 91: Making a separate section for feedback, since it's a little different than a See also, which is content that provides more information about the subject. Also, I don't see a category for Universal Print on the UserVoice site.

concepts/universal-print-concept-overview.md
braedenp-msft: Actually @angelgolfer-ms sent me an email awhile back (that I still owe a response to) asking for the same thing.

I'll hopefully get some time this month to do that.

concepts/universal-print-concept-overview.md
braedenp-msft: Fixed! We're trying to migrate there but we're using the Tech Community forums for now.

concepts/universal-print-concept-overview.md
braedenp-msft: I think that's too much information especially since most of it doesn't apply right now. Let's add that when we're actually behaving that way.

PR #9719, opened 2020-08-31
fix for git issue 9084

davidmu1:


api-reference/beta/api/user-get.md
```diff
@@ -26,6 +26,8 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account) | User.Read, User.ReadWrite    |
 |Application | User.Read.All, User.ReadWrite.All, Directory.Read.All, Directory.ReadWrite.All |
 
+Calling the `/me` endpoint requires a signed-in user and therefore a delegated permission. Application permissions are not supported when using the `/me` endpoint.
```
Lauragra, line 29: This topic doesn't show the /me syntax in the HTTP syntax section, like we have in the v1.0 topic. Add that syntax?

PR #9768, opened 2020-09-02
"requires review" is only in beta at the moment

joeyglocke:


Lauragra:
@laujan , can you please review these PRs to remove the requiresReview parameter?

laujan (approved):

PR #9795, opened 2020-09-04
Update workbook-error-codes.md

grangeryy:


concepts/workbook-error-codes.md
```diff
@@ -12,16 +12,16 @@ This article describes error codes that are returned by the Workbooks and charts
 
 ## Error code
 
-The following table lists the current error codes and messages. The service might add new error codes at any time.
+The following table lists the current error codes and messages. The service might add new error codes at any time. 
 
-| Code                      | Message
+| Error Code                       | Error Message
```
joezhao-msft, line 17: can you add an item with name of badGateway here? which is for status code of 502 so that others will not be confused when you mention 502/badGateWay later.

concepts/workbook-error-handling.md
```diff
@@ -124,6 +124,10 @@ If the error code you encounter is not in the second-level list or the top-level
 ### 5. Error recovery cooldown
 
 For some of the responses in the regular pattern, a recovery cooldown duration in seconds might be provided via a `Retry-After` header. When a recovery cooldown duration is present, the Microsoft Graph client is not expected to send any followup requests before the specified duration passes.
+
+## Special case handling
+
+In persist or non-persist session mode, we recommend to recreate session if you encounter `503/serviceUnavailable` or `502/badGateway` status code.
```
joezhao-msft, line 130: should it just be Sessionful requests? and add the link of Sessionful introduction to it.

concepts/workbook-error-codes.md
grangeryy: ACK

concepts/workbook-error-handling.md
grangeryy: Modified accordingly

grangeryy:
@Lauragra, we made some changes based on the guideline for graph errors: "The top-level code value will be standardized to be camel cased versions of the HTTP reason phrase". The related link here: [link](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). THX!

Lauragra (approved):

PR #9802, opened 2020-09-04
Update azuread overview to link to new tutorial

markwahl-msft:
The overview didn't include entitlement management, a neighbor feature to access reviews in Azure AD Premium P2, and also now there's also a new tutorial for illustrating how to use entitlement management,

### DIFF
--- a/concepts/azuread-identity-access-management-concept-overview.md
+++ b/concepts/azuread-identity-access-management-concept-overview.md
@@ -71,6 +71,11 @@ You can get secure access to resources by activating administrative privileges o
 
 You can configure access reviews of group memberships and application access. [Access reviews](/graph/api/resources/accessreviews-root?view=graph-rest-beta) is featured in Azure AD Premium P2.
 
+### Establish access request workflows (preview)
+
+The tutorial on how to [create an access package using Microsoft Graph APIs](tutorial-access-package-api.md) illustrates how to use [entitlement management](/graph/api/resources/entitlementmanagement-root?view=graph-rest-beta) to create a package of resources that users can self-service request.
+
+
 ## API reference
 
 Looking for the API reference for this service?


davidmu1 (approved):

concepts/azuread-identity-access-management-concept-overview.md
```diff
@@ -71,6 +71,11 @@ You can get secure access to resources by activating administrative privileges o
 
 You can configure access reviews of group memberships and application access. [Access reviews](/graph/api/resources/accessreviews-root?view=graph-rest-beta) is featured in Azure AD Premium P2.
 
+### Establish access request workflows (preview)
+
```
angelgolfer-ms, line 75: @markwahl-msft , @davidmu1 , does the following overlook anything pertinent at this "Learn" level:
```
[Create access packages](tutorial-access-package-api.md) for users to self-serve access needs to groups, applications, and SharePoint sites.
```

angelgolfer-ms: I rephrased the original because this conceptual overview is meant to list at a high level the value of the identity and access API. Seems to me the focus should be the value of the functionality and therefore should start with the action - what users can do using the API.
Per [Microsoft Voice](https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28939), we're also supposed to trim all unnecessary words, like "You can" in the preceding H3 sections, hence I didn't start with that phrase :-)

angelgolfer-ms (changes requested):
Hi @markwahl-msft , since this update goes to the more important conceptual overview, do you mind rephrasing the sentence to put focus on the value (what task customers can get done), rather than on the existence of a tutorial that shows how to achieve the value? We can implicitly blend in the tutorial with a link to it.

PR #9812, opened 2020-09-05
add tabs in chats (1149676)

laujan:


Lauragra:
@laujan , what are these files with no content? For new APIs, you'll want to use the stub generator to generate the topics with the required sections and boilerplate text that you can then author.

PR #9826, opened 2020-09-08
Update team.md

JarbasHorst:
As far as I know, the new property "isMembershipLimitedToOwners" is only available in Microsoft Teams for Education.


Lauragra:
Closing this PR because we don't really refer to Teams for Education this way. We're following up internally on whether this topic should note that an EDU license is required. Thanks.

JarbasHorst:
As a developer, having a note would be helpful! Thanks for your feedback @Lauragra

PR #9857, opened 2020-09-10
Updated Prefer header delimeter

JonathanHuss:
Prefer header description incorrectly listed semi-colons (;) as the delimiter.  Updated to commas (,) instead.


api-reference/beta/api/driveitem-delta.md
```diff
@@ -279,6 +279,15 @@ Content-type: application/json
     | Create/Modify | n/a |
     | Delete | `ctag`, `size` |
 
+## Scanning permissions hierarchies
```
Lauragra, line 282: This is a non-standard H2 heading for reference content and therefore will break the docs generation process that we're currently developing. I suggest moving this content into a concept topic about change tracking for drive items.

PR #9910, opened 2020-09-14
Remove Participant API document for beta and V1.0

jackry6350:


api-reference/beta/api/participant-delete.md
```diff
@@ -0,0 +1,83 @@
...
+localization_priority: Normal
+ms.prod: "cloud-communications"
+doc_type: apiPageType
+---
+
+# Delete participant
+
+> **Important:** APIs under the /beta version in Microsoft Graph are in preview and are subject to change. Use of these APIs in production applications is not supported.
+
+Delete a specific participant in the call.
```
jsandoval-msft, line 16: I'd like a add a short into here about the API. Something to the effect of:
"In some situations it is appropriate for an application to remove a participant from an active call, via the Delete participant API. This action can be done either before or after the participant answers the call. When an active caller is removed via Delete participant, they are immediately dropped from the call with no pre or post-removal notification."
Is this the sort of thing we would provided in these documents?

frankpeng7:
Notes: this PR is from internal instead of external which the bot suggested

frankpeng7:
@jackry6350 you can remove the tag `do not merge` and add `ready to merge` once you get signoff from Josey

frankpeng7 (changes requested):
@jackry6350 You need to add a changelog entry in `microsoft-graph-docs/concepts/changelog.md` (make sure you merge latest master before adding anything to avoid conflict)

frankpeng7 (approved):

jackry6350:
@Lauragra , can you help to take a quick look at this doc update and merge it if no big concerns? Thanks!

Lauragra (approved):

PR #9982, opened 2020-09-18
Add PowerShell SDK docs

jasonjoh:


concepts/toc.yml
```diff
@@ -515,6 +515,16 @@ items:
             href: sdks/large-file-upload.md
           - name: Use beta API
             href: sdks/use-beta.md
+      - name: Use PowerShell
```
Lauragra, line 518: Right now we have a "Use SDKs" node and we're adding a "Use PowerShell" node that is about using the PowerShell SDK. Should we nest this node under "Use SDKs"? Or at a minimum should the title of this node be "Use the PowerShell SDK"?

jasonjoh: No, we want User PowerShell. We call it the PS SDK because they've decided to name it that, but it isn't really an SDK. The concern is PS users wouldn't figure out to expand the "Use SDKs" node to find the PS docs.

jasonjoh: I've moved it under the Use SDKs per our discussion.

PR #9990, opened 2020-09-21
Fix JSON sample for HTTP request

BrianTJackett:
Invalid JSON with additional comma (,) included after "role" value.


BrianTJackett:
JavaScript sample also contains invalid JSON.  Unsure if snippet will be auto-updated or not.  Will check again after automation job runs.

PR #9994, opened 2020-09-22
2020-Microsoft-Search-API-Update

lanhongvp:


nmoreau:
Hi @Lauragra  and @angelgolfer-ms  - This PR is ready to be merged. Minor updates / clarifications / Polish.
Thanks.

api-reference/beta/resources/search-api-overview.md
```diff
@@ -93,6 +93,11 @@ The **listItem** and **externalItem** entities are the only supported entities t
 
 If the **fields** specified in the request are not present in the schema, they will not be returned in the response. Invalid fields in the request are silently ignored.
 
+If you do not specify any **fields** in the request, then you will get the default set of properties for all types. For extended properties, **listItem** and **externalItem** behave differently when no **fields** are passed in the request:
+
+- **listItem** will not return any field.
```
angelgolfer-ms, line 98: Line 98: 
For listItem resources from SharePoint, do you mean, if the query request doesn't specify any value for fields in the request body, then only default fields and no custom fields would be returned?

If so, please change line 98 to:
"**listItem** will not return any custom fields."

I would expect that to be the behavior based on the fact that the **fields** property is optional in the query request body, and its description currently says: 
"Contains the fields to be returned for each resource object specified in entityTypes, allowing customization of the fields returned by default otherwise..."

angelgolfer-ms: BTW, since you have the PR open, can you please fix the spelling error "earch" in the description for the **fields** property, in searchrequest.md? Thanks.

concepts/search-concept-custom-types.md
```diff
@@ -26,7 +26,7 @@ TODOSEARCHAPI - Bug 1653398
 
 - The **entityTypes** property as `externalItem`.
 
-- The **fields** property to include the fields in the external item to retrieve.
+- The **fields** property to include the fields in the external item to retrieve. Note that if you do not include any **fields** in the request, the response will contain all the fields marked as *retrievable* in the Graph Connectors search schema for that particular connection.
```
angelgolfer-ms, line 29: Line 29:
There's a branding problem - we should not publicly refer to Microsoft Graph as just "Graph".
However, looking at the context of the entire topic, I don't think we necessarily have to loop in the full term "Microsoft Graph connectors" here, as the earlier parts of the topic has already been talking about connectors.
Please change the part of the sentence to:
"... marked as *retrievable* in the data schema specified for that [external connection](/graph/api/resources/externalconnection?view=graph-rest-beta&preserve-view=true)."

concepts/search-concept-files.md
```diff
@@ -413,6 +413,9 @@ Content-type: application/json
   ]
 }
 ```
+## Known limitations
+
+When searching for **drive**, you need to include in the **queryString** a term contained in the name of the document library. Query `*` is not supported and will not return all available drives.
```
angelgolfer-ms, line 418: Line 418:
Please change the last sentence:
"Querying `*` is not supported and does not return all..."
We're supposed to use present tense in most cases, to reflect that this is the current behavior.

angelgolfer-ms (changes requested):
Please see and address a few comments, thanks!

api-reference/beta/resources/search-api-overview.md
nmoreau: Fixed and clarified.
Updated searchrequest.md?

concepts/search-concept-custom-types.md
nmoreau: Updated but did not refer to the external Connection since this would only be valid for custom connectors.
Here the connection can be an OOB connector

nmoreau: Fixed

Lauragra (approved):

PR #10023, opened 2020-09-23
Small bug fixing in Permissions section

elmakhmu:
Privacy settings don't support app calls, only delegated (work or school acc) permissions, thus even if an admin will create a new app and grant all required permissions, it will be impossible to tune setting via third-party app.
@angelgolfer-ms


concepts/insights-customize-item-insights-privacy.md
```diff
@@ -19,7 +19,7 @@ While existing apps could continue to use **officeGraphInsights**, these apps sh
...
 
-These settings provide flexibility for administrators to use Azure AD tools and disable item insights for only members of the specified group and not necessarily across the organization. Configure each of these properties by [updating item insights settings](/graph/api/iteminsightssettings-update?view=graph-rest-beta) in an app, PowerShell, or other applications with due permissions.
+These settings provide flexibility for administrators to use Azure AD tools and disable item insights for only members of the specified group and not necessarily across the organization. Configure each of these properties by [updating item insights settings](/graph/api/iteminsightssettings-update?view=graph-rest-beta) in Graph Explorer or PowerShell with due permissions.
```
Lauragra, line 22: Is this suggesting that the developer uses Graph Explorer to update their production environment? We should verify that this is an approved workflow.

angelgolfer-ms: @elmakhmu , by "Privacy settings don't support app calls, only delegated (work or school acc) permissions,... " do you mean PATCH /organization/{organizationId}/settings/itemInsights (iteminsightssettings-update.md) works only for delegated permissions but not application permissions?

Or do you mean that PATCH operation actually cannot be called in an app, regardless of the permissions?

elmakhmu: @angelgolfer-ms Actually we've never supported only app permissions for our settings (I mean, no background process or daemon without signed in user can read or configure that settings), so it was just a typo in docs, we are copying the same as profileCardProperty behavior.

angelgolfer-ms: With @elmakhmu 's clarification that delegated permissions still work for apps to do  GET and PATCH  /organization/{organizationId}/settings/itemInsights, we can keep the original description in line 22: 

"Configure each of these properties by [updating item insights settings](/graph/api/iteminsightssettings-update?view=graph-rest-beta) in an app,..."

We also communicated that in general, even though users with global administrator roles signing into Graph Explorer  can achieve GET and PATCH as well, in the docs, we generally recommend customers to create their own apps to make the REST API calls instead. Graph Explorer is intended to be a learning and prototyping tool.

PR #10149, opened 2020-10-02
Update listitem-get.md

JarbasHorst:
Added a new application permission "Sites.Manage.All" that is required if the SharePoint list has content approval settings turned on. Otherwise, Graph won't retrieve those list items that have an approval status other than Approved. This permission won't be required if list items are retrieved using delegated permissions! I've also written a [blog post](https://www.devjhorst.com/2020/09/quick-tip-avoid-limitations-when.html) about this topic that might be useful during the review process since it contains more information about this issue!

### DIFF
--- a/api-reference/beta/api/listitem-get.md
+++ b/api-reference/beta/api/listitem-get.md
@@ -28,7 +28,7 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account) | Not supported.    |
 |Application | Sites.Read.All, Sites.ReadWrite.All, Sites.Manage.All |
 
-> **Note**: The application permission _Sites.Manage.All_ will be required if the SharePoint list has content approval settings turned on. Otherwise, Graph won't retrieve those list items that have an approval status other than Approved.
+> **Note**: The application permission Sites.Manage.All is required if the SharePoint list has content approval settings turned on. Otherwise, Microsoft Graph won't retrieve list items that have an approval status other than Approved.
 
 ## HTTP request
 
--- a/api-reference/beta/api/listitem-get.md
+++ b/api-reference/beta/api/listitem-get.md
@@ -24,9 +24,11 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type      | Permissions (from least to most privileged)              |
 |:--------------------|:---------------------------------------------------------|
-|Delegated (work or school account) | Sites.Read.All, Sites.ReadWrite.All    |
+|Delegated (work or school account) | Sites.Read.All, Sites.ReadWrite.All |
 |Delegated (personal Microsoft account) | Not supported.    |
-|Application | Sites.Read.All, Sites.ReadWrite.All |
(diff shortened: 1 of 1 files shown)


BrianTJackett:
@JarbasHorst A few questions so that we capture all of the use cases at once if possible.

1. Does this also apply to listitem-list endpoint?
1. Does this apply to delegated permissions (another user who isn't the author / approver of item) also or only application permissions?
1. Can you make this change (and above ones if necessary) to the v1.0 version of doc as well?

JarbasHorst:
Hello @BrianTJackett 

I'll have a look at the questions 1 and 2. In the meanwhile, I've already authored a PR for v1.0 version of that doc: #10153

JarbasHorst:
Hello @BrianTJackett 

here are my findings:


User | ADD Permission | API | Behavior
-- | -- | -- | --
User who has created list item | Sites.ReadWrite.All (Delegated) | lists{0}/items | Correct response
User who has created list item | Sites.ReadWrite.All (Delegated) | lists{0}/items/{1} | Correct response
User who hasn’t created the list   item, but has access to list | Sites.ReadWrite.All (Delegated) | lists{0}/items | Except those items that have an   approval status other than Approved
User who hasn’t created the list   item, but has access to list | Sites.ReadWrite.All (Delegated) | lists{0}/items/{1} | Exception       System.Exception: {       "error": {         "code": "itemNotFound",         "message": "The specified list was not found",         "innerError":   {           "date": "2020-10-04T09:43:44",           "request-id":   "15fb88e0-5add-43ff-8e56-5c848e7b7e6f",           "client-request-id": "15fb88e0-5add-43ff-8e56-5c848e7b7e6f"         }       }   }
impersonated | Sites.ReadWrite.All (Application) | lists{0}/items | Except those items that have an   approval status other than Approved
impersonated | Sites.ReadWrite.All (Application) | lists{0}/items/{1} | Exception       System.Exception: {       "error": {         "code": "itemNotFound",         "message": "The specified list was not found",         "innerError":   {           "date": "2020-10-04T09:43:10",           "request-id":   "6de5c1d3-a3c5-4977-a129-588e379b7e93",           "client-request-id":   "6de5c1d3-a3c5-4977-a129-588e379b7e93"         }       }   }
User who has created list item | Sites.Manage.All (Delegated) | lists{0}/items | Correct response
User who has created list item | Sites.Manage.All (Delegated) | lists{0}/items/{1} | Correct response
User who hasn’t created the list   item, but has access to list | Sites.Manage.All (Delegated) | lists{0}/items | Correct response
User who hasn’t created the list   item, but has access to list | Sites.Manage.All (Delegated) | lists{0}/items/{1} | Correct response
impersonated | Sites.Manage.All (Application) | lists{0}/items | Correct response
impersonated | Sites.Manage.All (Application) | lists{0}/items/{1} | Correct response

So, the issue applies to the listitem-list endpoint as well as to delegated permissions (another user who isn't the author).

Can I update the listitem-list docs as well?

I hope this helps, but please let me know if you need anything else or if you have any other questions.

Cheers
Jarbas Horst

PR #10153, opened 2020-10-02
Update listitem-get.md

JarbasHorst:
Added a new application permission "Sites.Manage.All" that is required if the SharePoint list has content approval settings turned on. Otherwise, Graph won't retrieve those list items that have an approval status other than Approved. This permission won't be required if list items are retrieved using delegated permissions! I've also written a [blog post](https://www.devjhorst.com/2020/09/quick-tip-avoid-limitations-when.html) about this topic that might be useful during the review process since it contains more information about this issue!


JarbasHorst:
Please, @Lauragra don't merge as we are still reviewing the impact of this issue. Ping @BrianTJackett

PR #10157, opened 2020-10-02
Property name should be data not datakey

kevinblumenfeld:
```csharp
RSACryptoServiceProvider rsaProvider = ...;        
byte[] encryptedSymmetricKey = Convert.FromBase64String(<value from dataKey property>);
```
should be
```csharp
RSACryptoServiceProvider rsaProvider = ...;        
byte[] encryptedSymmetricKey = Convert.FromBase64String(<value from data property>);
```


kevinblumenfeld:
The property is correct.

PR #10159, opened 2020-10-02
Document generator preparation

kanchm:
- Format the existing documents with documentation generator
- Only spacing and table formatting applied
- No content should be affected by this change
- Treatment applied only to beta/v1.0 api/resources only

Subsequent update will inject content from schema updates.


wandeg (approved):

wandeg:
Looks like one of the files has an error.

api-reference/beta/api/accesspackage-delete.md
```diff
@@ -63,21 +64,24 @@ The following is an example of the request.
 ```http
 DELETE https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackages/{id}
 ```
+
 # [C#](#tab/csharp)
+
```
darrelmiller, line 69: @kanchm Is this empty line necessary?  We will need to verify if our Snippet Generation mechanism doesn't depend on the [!include being on the next line.

angelgolfer-ms: @andrueastman confirmed that the snippet generator currently assumes there are no other lines between a language-tab-line and an "!INCLUDE" line.
So lines like 74, 79, will break snippet generation.

PR #10163, opened 2020-10-04
Fix tags for some titles

dingyx99:
Titles of authentication providers contains an `<a>` tag, which is not correctly closed, causing the following description included.


dingyx99:
@darrelmiller @MIchaelMainer

Lauragra:
Good catch, thanks!

PR #10186, opened 2020-10-05
fix for git issue 9503

davidmu1:


davidmu1:
$search is no supported in the SDKs yet.

PR #10212, opened 2020-10-07
Add preserve-view=true to versioned links in changelog

psignoret:
Avoids the extremely verbose list of suggestions whenever there's a change to changelog. Assumes there versioned links in changelog were intentionally versioned.


Lauragra:
Thanks for helping to get rid of all those!

PR #10221, opened 2020-10-07
Fix create teams sample

NickSevens:
The sample to create a Team using `POST /v1.0/Teams` with Application Permissions had the wrong syntax, resulting in an error when executing it.
This change has the correct syntax.
See #9641 

Additionally: removed duplicate HTTP sample blocks which prevented switching tabs in some examples.


Lauragra:
@laujan , please review. Given that this is an update to a request example, we should remove the SDK snippets until they can be regenerated. If approved, @NickSevens , we need to apply this fix to the topic in /beta as well. Thanks.

NickSevens:
@Lauragra I've updated the `beta` docs as well

NickSevens:
@laujan @Lauragra merge conflicts have been resolved.

laujan (approved):
Per Wajeed:
Hi la januari- I just verified that change is correct. 
With current code snippet request is failing with 400 Bad Request. Request worked perfectly fine with PR changes.

PR #10244, opened 2020-10-09

nmoreau:
Correction on Delegated Permission for Search which was wrong.
The correct permission is ExternalItem.Read.All for delegated.

### DIFF
--- a/concepts/permissions-reference.md
+++ b/concepts/permissions-reference.md
@@ -1533,26 +1533,26 @@ For more complex scenarios involving multiple permissions, see [Permission scena
 
 |   Permission    |  Display String   |  Description | Admin Consent Required | Microsoft Account supported |
 |:----------------|:------------------|:-------------|:-----------------------|:--------------|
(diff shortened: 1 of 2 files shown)


concepts/permissions-reference.md
```diff
@@ -1533,26 +1533,26 @@ For more complex scenarios involving multiple permissions, see [Permission scena
 
 |   Permission    |  Display String   |  Description | Admin Consent Required | Microsoft Account supported |
 |:----------------|:------------------|:-------------|:-----------------------|:--------------|
-| _ExternalItem.ReadWrite.All_ | Read and Write external data | Allows an app to write external data into the indexing API. | Yes | No |
+| _ExternalItem.ReadWrite.All_ | Read and Write external data | Allows an app to write external with Microsoft Graph connectors indexing API. | Yes | No |
```
Lauragra, line 1536: Is a word missing here, i.e. should it be "external **data** with the..."?

nmoreau: Fixed in the next commit

PR #10268, opened 2020-10-12
noting that riskEventTypes has been deprecated

ebasseri:


Lauragra:
@khotz , please review and approve. If we need this note in the beta topic, we should add it to the v1.0 topic as well.

PR #10274, opened 2020-10-12
Document callRecord throttling policies

stephenjust:
The callRecord resource will be adding some new throttling policies, which should be reflected in documentation.


stephenjust:
It's not super clear what the best way is to express these limits. I'm open to suggestions for how we can make this more consistent.

baywet (changes requested):
thank you for including me as a reviewer. Can you format it as a table to match the other workloads please?

baywet:
@stephenjust QQ: are you implementing these throttling policies using the configuration at the AGS level? If yes, you don't need to document anything, I have a tool that periodically picks-up the configuration and spits markdown.

stephenjust:
@baywet Yes, I am configuring this using AGS's built-in throttling. In that case should I abandon this PR?

PR #10312, opened 2020-10-16
Yw/participant info with platform

yizhenww:


Lauragra (changes requested):
Looks good; do you need to update any response examples to include the new property?

yizhenww:
I added an example to the get participant api.
it's an optional property so I didn't add it to all examples containing participantInfo

Lauragra (approved):

PR #10319, opened 2020-10-16
Update Directory Audit $filter Example to Reflect Valid Attribute

amorrowbellarmine:
The filter example is using an attribute that doesn't exists in the directory audit logs. `createdDateTime` exists in the sign-in logs, but the equivalent in the directory audit logs is `activityDateTime`


Lauragra:
@amorrowbellarmine , please sign the license/cla. Thanks.

amorrowbellarmine:
CLA signed. I thought I had already done that, since I've made similar submissions to other Microsoft Docs repos.

PR #10357, opened 2020-10-20
fix: Invalid trailing comma in JSON

hedgecox:
JSON representation of the resource had a trailing comma causing an invalid JSON error when copied directly from docs.


hedgecox:
#10357

hedgecox:
#10356

PR #10374, opened 2020-10-21
Jingha add cloud pc permissions

timefrozen19:
Add delegated cloud pc permissions: CloudPC.Read.All and CloudPC.ReadWrite.All


Jackson-Woods:
@timefrozen19 Can you make sure the description strings in the table match up with the "official" description strings for the permissions? I think they need to add the "on your behalf"/"on behalf of the signed in user" verbiage.

Is there any new API reference content for these APIs where these permissions should be listed?

timefrozen19:
> @timefrozen19 Can you make sure the description strings in the table match up with the "official" description strings for the permissions? I think they need to add the "on your behalf"/"on behalf of the signed in user" verbiage.
> 
> Is there any new API reference content for these APIs where these permissions should be listed?


@Jackson-Woods Updated the doc to include on behalf of the signed-in user.
Yes there will be new API reference content that will list these permissions. We are preparing the API reference content PR now. There are no existing API reference content that need to list these permissions.

PR #10390, opened 2020-10-22
fix for git issue 9319

davidmu1:


Lauragra:
@davidmu1 , just checking, how does this fix the example? We generally don't set the request examples to ignored, so that API Doctor can validate them.

davidmu1:
@Lauragra - unfortunately we have many instances where the SDK is not in sync with the REST definitions. Because the SDK examples are automatically generated, there isn't a way to correct them until the SDK is updated. The only way to handle this that I know of is to set the request to ignore so that the incorrect example isn't added to the page.

PR #10394, opened 2020-10-23
Push meeting option documents to v1.0

frankpeng7:
This moves #9630 to v1.0


Lauragra (approved):

frankpeng7:
Hi @Lauragra sorry for the last minute notice but we just realize the "role" property on meetingParticipantInfo is still subject to Graph CSDL deployment so set the tag back to "do not merge"

PR #8965, opened 2020-07-03
feature/lifecycle-notifications-v1

baywet:
closes #7487

Note: we've extended the concept to other workloads/resources

PR #10399, opened 2020-10-23
bugfix/lifecycle-notifications-adjustments

baywet:
adjustments for the late comments in #8965


concepts/changelog.md
```diff
@@ -29,7 +29,7 @@ For a summary of the value of these API changes, as well as recent tools, compon
 
 | **Change type** | **Version**   | **Description**                          |
 | :-------------- | :------------ | :--------------------------------------- |
-| Addition | beta and v1.0 | Added [lifecycle notifications](./webhooks-lifecycle.md).|
+| Addition | v1.0 | Added [lifecycle notifications](./webhooks-lifecycle.md) for personal [chatMessage][/graph/api/resources/chatmessage?view=graph-rest-1.0], [contact][/graph/api/resources/contact?view=graph-rest-1.0], [event][/graph/api/resources/event?view=graph-rest-1.0], and [message][/graph/api/resources/message?view=graph-rest-1.0].|
```
angelgolfer-ms, line 32: @baywet , does chatMessage in beta support lifecycle notifications as well? 
If so, is beta support also new?
If beta support is new, then we should add that as a separate table row for the beta version.

angelgolfer-ms: Also, the links are not properly formatted with curly brackets, please fix.

concepts/webhooks-lifecycle.md
```diff
@@ -27,6 +27,15 @@ An app subscribing to change notifications should listen to the `subscriptionRem
 
 To receive lifecycle notifications, you can use the existing **notificationUrl** endpoint that already receives change notifications, or you can register a separate **lifecycleNotificationUrl** to receive `subscriptionRemoved` and `missed` lifecycle notifications in a separate endpoint.
 
+Lifecycle notifications are emitted for subscriptions created on these resource types:
```
angelgolfer-ms, line 30: Line 30:
Please replace "emitted" by "supported", to use a more commonly used term, per Microsoft Voice: https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=29173

angelgolfer-ms (changes requested):
@baywet , thank you so much for patching up with this PR!
Just three minor comments, please take a look. 
Apart from those issues, the PR looks good to be merged.

Lauragra (approved):

concepts/changelog.md
baywet: @Lauragra thanks for fixing the links.
@angelgolfer-ms thanks for the feedback. lifecycle notifications has been in beta for chatMessages, events, messages and contacts for almost a year now. There should have been a changelog entry created by Peter back then. The only changes we're introducing at this point is the GA of that capability on those resources.

concepts/webhooks-lifecycle.md
baywet: @Lauragra thanks for making the change!

concepts/changelog.md
angelgolfer-ms: Ok, I see where our disconnect is - I searched the changelog, it only mentions in Dec. 2019, chatMessage started supporting notifications with resource data. There's no separate mention of lifecycle notifications for chatMessage.

angelgolfer-ms (approved):
Thanks for doc updates and clarification.

PR #10436, opened 2020-10-27
[MVP4] 2020-Microsoft-Search-API-v1.0

lanhongvp:
Microsoft search API v1.0 shipping update


api-reference/v1.0/resources/searchresult.md
```diff
@@ -1,15 +1,15 @@
 ---
-author: JeremyKelley
-ms.author: JeremyKelley
+author: lanhong
```
nmoreau, line 2: I think this file should not be modified.

PR #10529, opened 2020-11-04
Update

damiura:
CompanyName is not supported now as following.
https://stackoverflow.com/questions/47091291/filtering-users-by-companyname

I also verified '$filter=companyName' on MS Graph Explorer.
As Dan said on StackOverFlow, I got an unsupported response from MS Graph.


Lauragra:
Thanks.

PR #10537, opened 2020-11-05
Update webhooks.md

VoiceOfSoftware:
Feel free to re-word my attempt here. I spent days looking for some fancy JWT decryption library, in order to 'decode' the validationToken. But all you really mean is that you want me to reflect back the *exact* same string that was sent in, no parsing or decoding required! I find the word 'decode' leads to misunderstanding.


Lauragra:
Thanks.

PR #10570, opened 2020-11-06
Intune 2011 beta

davidmu1:


Lauragra:
@davidmu1 , does this PR include an update to the Intune.json file for the changelog entries? I can't tell in the PR because it has so many files.

davidmu1:
I added the changelog json.

davidmu1:
That's goofy. I assume you mean this line in the file: "CreatedDateTime": "2020-11-16T00:00:00", Done.

PR #10571, opened 2020-11-06
2020 10 26 mahage e discovery - Custodian

mahage-msft:
Adding Custodian API docs and supporting APIs


api-reference/beta/api/custodian-activate.md
```diff
@@ -0,0 +1,78 @@
...
+### Request
+
+<!-- {
+  "blockType": "request",
+  "name": "custodian_activate"
+}
+-->
+
+``` http
+POST https://graph.microsoft.com/beta/compliance/ediscovery/cases/{ediscoveryCaseId}/custodians/{custodianId}/activate
```
Lauragra, line 65: Global: In general, it's helpful to provide "real" IDs and values in request and response examples.

api-reference/beta/api/custodian-create.md
```diff
@@ -0,0 +1,108 @@
+---
+title: "Create custodian"
```
Lauragra, line 2: The file naming convention for Create operations is to use -post. Did you rename this file, or did the stub generator output this file name?

mahage-msft: These were from the stub generator.  Oddly enough, there were always two files for each of the methods that were generally the same, but a little different.  One was created to be referenced from the parent object and the other from the object itself.  Since that isn't going to make any sense, here's an example :).

_Custodains.md_ had a reference to a file called _custodian-post-sitesources.md_
_sitesources.md_ had a references to a file called _sitesource-create.md_

Because the two files were almost identical (other than one was written with plurals and the other was singular) and we don't support bulk creation, I opted for the latter and changed the reference in the parent (in this case sitesource-create.md)

Will change filenames to -post.

api-reference/beta/api/custodian-activate.md
mahage-msft: Will add GUIDs to examples.

api-reference/beta/api/custodian-create.md
mahage-msft: Fixed.

api-reference/beta/api/custodian-activate.md
mahage-msft: Fixed.

api-reference/beta/api/custodian-delete.md
```diff
@@ -0,0 +1,77 @@
+---
+title: "Delete custodian"
```
ashwkum, line 2: Delete custodian is not supported.

api-reference/beta/api/custodian-delete.md
```diff
@@ -0,0 +1,77 @@
...
+description: "Delete a custodian object."
+author: "mahage-msft"
+localization_priority: Normal
+ms.prod: "compliance"
+doc_type: apiPageType
+---
+
```
ashwkum, line 9: create custodian md file is missing. Can you check ?

mahage-msft: It's there - api-reference/beta/api/custodian-post.md

api-reference/beta/api/custodian-delete.md
mahage-msft: Removed.

api-reference/beta/api/sitesource-list.md
```diff
@@ -0,0 +1,101 @@
+---
+title: "List siteSources"
```
Lauragra, line 2: We should make this file name plural. Did the tool output this file name?

mahage-msft: Yes, the tool output this filename.  This is how things are a bit strange.  I got two stubs for this object which were almost identical.  One was named custodian-list-sitesources.md, the other named sitesource-list.md.  I opted to use the latter since this will be referenced from objects other than custodian.

api-reference/beta/api/custodian-activate.md
```diff
@@ -0,0 +1,78 @@
+---
+title: "custodian: activate"
+description: "Reactivate a custodian in a case."
+author: "mahage-msft"
+localization_priority: Normal
+ms.prod: "compliance"
```
Lauragra, line 6: Global: Should we be tagging this content with "ediscovery", or is compliance sufficient? Ediscovery is the specific subservice under the Compliance service node.

api-reference/beta/resources/sitesource.md
```diff
@@ -0,0 +1,64 @@
...
+doc_type: resourcePageType
+---
+
+# siteSource resource type
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Data source entity for SharePoint sites - This is the container for a custodian's site.
```
Lauragra, line 16: Can you help make this description more clear? Is this a site associated with a custodian?

api-reference/beta/resources/unifiedgroupsource.md
```diff
@@ -0,0 +1,82 @@
...
+doc_type: resourcePageType
+---
+
+# unifiedGroupSource resource type
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Data source entity for Groups - This is the container for a custodian's group
```
Lauragra, line 16: Please clarify this description as well.

api-reference/beta/api/custodian-activate.md
mahage-msft: You, Angela and I had a discussion about this when I first published docs.  I don't recall all of the details, but I think it was along the lines of eDiscovery being a feature in a feature (compliance).  Your call on this one.  I can pull in all of the other topics and modify if we want to go that way.

PR #10614, opened 2020-11-10
Update usersettings.md

elmakhmu:
adding new itemInsightsEnabled param to user object


Lauragra (changes requested):
Please include an update to the changelog.md and JSON files to mention the addition of the new property.

elmakhmu:
this issue will be merged soon. please, do not close or delete.

elmakhmu:
can be deleted, since we changed API shape and it requires new docs changing, which is happening in separate PR

PR #10653, opened 2020-11-12
Various fixes (deployments, orgcontacts...)

dkershaw10:


davidmu1 (approved):

eketo (approved):
Approved the orgContact changes.

eketo-msft:
> Approved the orgContact changes.

Woops, wrong Github account.

api-reference/beta/api/schemaextension-update.md
```diff
@@ -33,7 +33,14 @@ One of the following permissions is required to call this API. To learn more, in
 |Application | Not supported. |
 
 > [!NOTE]
-> Additionally for the delegated flow, the signed-in user can only update schemaExtensions they own (where the **owner** property of the schemaExtension is the `appId` of an application the signed-in user owns).
+> Additionally for the delegated flow, the signed-in user can only update schemaExtensions they own (where the **owner** property of the schemaExtension is the `appId` of an application the signed-in user owns). This allows a signed-in user to make updates through other applications they don't own, like [Microsoft Graph Explorer](https://aks.ms/ge).
+
+
+> [!NOTE]
+> When using Graph Explorer to update a schemaExtension, you'll need to specify the **owner** property in the PATCH request body. For more information, see the [Extensions](/graph/known-issues#extensions) section in [Known issues with Microsoft Graph](/graph/known-issues).
```
Lauragra, line 40: @bettirosengugi , is Graph Explorer supported for use in production environments? I'm not sure if we should be conveying this in the documentation.

api-reference/beta/resources/schemaextension.md
```diff
@@ -44,7 +44,7 @@ See the [schema extension example](/graph/extensibility-schema-groups) to learn
...
-|owner|String|The `appId` of the application that is the owner of the schema extension. This property can be supplied on creation, to set the owner.  If not supplied, then the calling application's `appId` will be set as the owner. In either case, the signed-in user must be the owner of the application. Once set, this property is read-only and cannot be changed.|
+|owner|String| The `appId` of the application that is the owner of the schema extension. This property can be supplied on creation, to set the owner.  If not supplied, then the calling application's `appId` will be set as the owner. In either case, the signed-in user must be the owner of the application. So, for example, if creating a new schema extension definition using Graph Explorer, you **must** supply the owner property. Once set, this property is read-only and cannot be changed. Once set, this property is read-only and cannot be changed.|
```
Lauragra, line 47: I'm not sure if our docs should be implying that Graph Explorer is an approved tool for production use vs. a learning tool for trying APIs in sandbox environments.

api-reference/v1.0/api/schemaextension-update.md
```diff
@@ -30,7 +30,11 @@ One of the following permissions is required to call this API. To learn more, in
 |Application | Not supported. |
 
 > [!NOTE]
-> Additionally for the delegated flow, the signed-in user can only update schemaExtensions they own (where the **owner** property of the schemaExtension is the `appId` of an application the signed-in user owns).
+> Additionally for the delegated flow, the signed-in user can only update schemaExtensions they own (where the **owner** property of the schemaExtension is the `appId` of an application the signed-in user owns). This allows a signed-in user to make updates through other applications they don't own, like [Microsoft Graph Explorer](https://aks.ms/ge).
+
+
+> [!NOTE]
```
Lauragra, line 36: Same question applies here.

api-reference/v1.0/resources/schemaextension.md
```diff
@@ -41,7 +41,7 @@ See the [schema extension example](/graph/extensibility-schema-groups) to learn
...
-|owner|String|The `appId` of the application that is the owner of the schema extension. This property can be supplied on creation, to set the owner.  If not supplied, then the calling application's `appId` will be set as the owner. In either case, the signed-in user must be the owner of the application. Once set, this property is read-only and cannot be changed.| 
+|owner|String|The `appId` of the application that is the owner of the schema extension. This property can be supplied on creation, to set the owner.  If not supplied, then the calling application's `appId` will be set as the owner. In either case, the signed-in user must be the owner of the application. So, for example, if creating a new schema extension definition using Graph Explorer, you **must** supply the owner property. Once set, this property is read-only and cannot be changed. Once set, this property is read-only and cannot be changed.|
```
Lauragra, line 44: Same comment as per the beta topic.

concepts/known-issues.md
```diff
@@ -204,6 +204,10 @@ You must first create the resource instance and then do a `PATCH` to that instan
 
 Directory resources, such as **device**, **group** and **user**, currently limit the total number of schema extension property values that can be set on a resource instance, to 100.
 
+### Updating a schemaExtension definition using Microsoft Graph Explorer
+
+When using `PATCH` to update a schemaExtension using Graph Explorer, you must specify the **owner** property and set it to its current `appid` value (which will need to be an `appId` of an application that you own).
```
Lauragra, line 209: Same question re: Graph Explorer.

Lauragra (changes requested):
I would like us to clarify whether we recommend Graph Explorer as a tool for use in production environments. I was under the impression that we don't. Let's let Bettirose confirm.

api-reference/beta/resources/schemaextension.md
dkershaw10: Will fix the repeated sentence.  Thanks.
Customers keep asking this question, which is why I've added this info.  Until we add a capability in the Azure Portal to edit schema extension definitions, GE would be the only way to do this without creating your own custom app.  You could also use PowerShell I suppose.  Again - how is this different from any other client application call Microsoft Graph?  Why are we making a distinction here?  The customer can choose what client apps they use for whatever purpose suits them.  If not GE, then don't grant it the permissions it needs to access Graph (or schema extension), or disable it for your tenant.

Also why does this language (as opposed to other places we mention Graph Explorer) suggest production usage here?

concepts/known-issues.md
dkershaw10: Actually this doesn't just need to be Graph Explorer.  This could apply to any app (like PowerShell too).

api-reference/beta/resources/schemaextension.md
dkershaw10: Fixed repeated sentence

api-reference/beta/api/orgcontact-update.md
```diff
@@ -59,10 +59,11 @@ In the request body, supply the values for relevant fields that should be update
 
 ## Response
 
-If successful, this method returns a `200 OK` response code and updated [orgContact](../resources/orgcontact.md) object in the response body.
+If successful, this method returns a `204 No Content` response code.
+
 ## Example
-##### Request
-Here is an example of the request.
+#####equest
```
angelgolfer-ms, line 65: Line 65: Typo here, please fix.

api-reference/beta/api/schemaextension-update.md
```diff
@@ -33,7 +33,14 @@ One of the following permissions is required to call this API. To learn more, in
 |Application | Not supported. |
 
 > [!NOTE]
```
angelgolfer-ms, line 35: Hi @dkershaw10 ,
Can I make an unsolicited contribution? I found the messaging around "own" and "owner" a bit confusing, and further compounded by what owner can be set to. Since we can't change the owner property name or its behavior, let's try to document it more clearly.
I'd suggest moving the content for the two notes back to the top before the "Permissions" section, because there's already a paragraph about who can update an extension: "Only the app that created a schema extension..."

I'll just present the entire section, so you can see the flow is clearer:

```
Update properties in the definition of the specified [schemaExtension](../resources/schemaextension.md).

The update applies to all the resources that are included in the **targetTypes** property of the extension. These resources are among the [supporting resource types](/graph/extensibility-overview#supported-resources).

Only the app that created a schema extension (owner app) can make additive updates to the extension when the extension is in the **InDevelopment** or **Available** status. That means the app cannot remove custom properties or target resource types from the definition. The app can, however, change the description of the extension.

For delegated flow, the signed-in user can update a schema extension as long as the **owner** property of the extension is set to the **appId** of an application the signed-in user owns. That application can be the one that initially created the extension, or some other application owned by the signed-in user. 

This criteria for the **owner** property allows a signed-in user to make updates through other applications they don't own, such as Microsoft Graph Explorer. When using Graph Explorer to update a **schemaExtension** resource, include the **owner** property in the PATCH request body. For more information, see the [Extensions](/graph/known-issues#extensions) section in [Known issues with Microsoft Graph](/graph/known-issues).
```

angelgolfer-ms (commented):
@dkershaw10 , please take a look at a comment I made here: https://github.com/microsoftgraph/microsoft-graph-docs/pull/10653/files#r525826261
I hope that would make the term "own", and setting the owner property a bit clearer.

api-reference/beta/api/orgcontact-update.md
dkershaw10: I'll make the change but it's pretty pointless.  I'm removing this topic from the TOC and the resource topic as updates to this object are not permitted.

api-reference/beta/api/schemaextension-update.md
dkershaw10: @angelgolfer-ms  Updates made.  Thanks for the feedback

api-reference/beta/api/orgcontact-update.md
dkershaw10: Fixed

api-reference/beta/resources/schemaextension.md
dkershaw10: @Lauragra - I don't think use of GE against production environments is an issue.  It's just like any other client app that needs to be approved.  No matter what you say here, that choice is up to admins.
That said - can you clarify exactly what it is about the language used that suggests "Graph Explorer is an approved tool for production use vs. a learning tool for trying APIs in sandbox environments"?
cc: @bettirosengugi

bettirosengugi:
> I would like us to clarify whether we recommend Graph Explorer as a tool for use in production environments. I was under the impression that we don't. Let's let Bettirose confirm.

Graph Explorer can be used for both prod and demo accounts

api-reference/beta/resources/schemaextension.md
bettirosengugi: Graph Explorer can be used for both prod and demo accounts depending on the type of user, learners use the demo accounts while expert developers use it with their prod accounts to experiment.

PR #10659, opened 2020-11-12
Labeling new access review api documentation as Preview

isabelleatmsft:
- removing "Legacy" from old access review apis
- labeling new access review apis as "Preview"

### DIFF
--- a/api-reference/beta/resources/accessreviews-root.md
+++ b/api-reference/beta/resources/accessreviews-root.md
@@ -1,5 +1,5 @@
 ---
-title: "Azure AD access reviews"
+title: "Azure AD access reviews (all resources)"
 description: "You can use Azure AD access reviews to configure one-time or recurring access reviews for attestation of user's access rights. This documentation serves the legacy APIs."
 localization_priority: Normal
 author: "markwahl-msft"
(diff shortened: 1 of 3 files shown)


markwahl-msft (approved):

davidmu1 (approved):

Lauragra:
@isabelleatmsft , @davidmu1 , is it confusing to have two nodes, both in beta, with one labeled (Preview)? What is the state of the non-preview Access review node? i.e. is it deprecated? Is the one that's in Preview in a different state than beta? Let's make sure that we're clearly communicating the right thing to customers.

isabelleatmsft:
> @isabelleatmsft , @davidmu1 , is it confusing to have two nodes, both in beta, with one labeled (Preview)? What is the state of the non-preview Access review node? i.e. is it deprecated? Is the one that's in Preview in a different state than beta? Let's make sure that we're clearly communicating the right thing to customers.

@Lauragra 
non-preview access review: this is the original access review apis, but our engineering team wanted to redo them to support more scenarios in the future, hence the preview node. non-preview still supports all resources we can do reviews on.

preview access review: this is our new apis that we want to move our users to using, but the catch is that it only supports groups as a resource. they are being built up to support the rest of our resources soon.

so it doesn't yet make sense to call the non-preview node deprecated, because to review resources other than groups, that's what customers have to use. do you have any naming suggestions?

Lauragra:
> > @isabelleatmsft , @davidmu1 , is it confusing to have two nodes, both in beta, with one labeled (Preview)? What is the state of the non-preview Access review node? i.e. is it deprecated? Is the one that's in Preview in a different state than beta? Let's make sure that we're clearly communicating the right thing to customers.
> 
> @Lauragra
> non-preview access review: this is the original access review apis, but our engineering team wanted to redo them to support more scenarios in the future, hence the preview node. non-preview still supports all resources we can do reviews on.
> 
> preview access review: this is our new apis that we want to move our users to using, but the catch is that it only supports groups as a resource. they are being built up to support the rest of our resources soon.
> 
> so it doesn't yet make sense to call the non-preview node deprecated, because to review resources other than groups, that's what customers have to use. do you have any naming suggestions?
> 
> @markwahl-msft FYI if you want to chime in here :)

It seems like we need a label for the non-preview access review APIs, which are technically also "preview" because they're in beta only. You already removed "Legacy", but that might be the closest term. Also, please remove the complex types from the TOC as part of this PR.

Lauragra (approved):

PR #10719, opened 2020-11-17
Add registerdevice user action

elcarlosadrian:
Add urn:user:registerdevice to the list of supported user actions in Beta


videor (commented):
This looks okay. There is no schema changes. So error codes for cloud apps / user action selection and everything else remains the same. Incorrect strings will also apply as it part of same schema. Looks good.

Issue #21, opened 2015-11-20
Cannot retrieve group/unseenCount


waldekmastykarz:
I've just noticed that although the **unseenCount** property is specified in Group's schema, it's not returned as a part of the `/v1.0/directoryObjects/<id>` request: either when requesting the complete group information or only the **unseenCount** property using the **$select** parameter. If this is a known limitation it would be helpful if it was mentioned in the documentation.

sumurthy:
@dkershaw10 @VenkatAyyadevara-MSFT @rohitag1

dkershaw10:
Filed an internal bug about unseenCount

Lauragra:
Closing this issue for now. Feel free to reopen if you need anything else. Thanks.

Issue #20, opened 2015-11-20
/joinedGroups endpoint returns 500 error


waldekmastykarz:
If you try to call the `/v1.0/me/joinedGroups` endpoint you get a 500 Internal Server Error. Given that this endpoint is included in the `/v1.0/me` metadata you would expect it to work. If it's a known issue it should be listed as known issue in the documentation including the right way to retrieve user's groups.

sumurthy:
@VenkatAyyadevara-MSFT @rohitag1

martellaj:
@waldekmastykarz, can you tell me where you see /joinedGroups as part of the /v1.0/me metadata? I think `/v1.0/me/memberOf`is what you're looking for.

waldekmastykarz:
It's listed in the metadata if you request `/v1.0/me` with the `Accept: application/json;odata.metadata=full` header

martellaj:
I passed this along to the dev team and will let you know when I hear something. Thanks very much for the feedback!

dkershaw10:
We filed an internal bug about /joinedGroups being visible in the full metadata response on /v1.0

Lauragra:
It looks like we can close this issue. Feel free to reopen if you need anything else. Thanks.

PR #10729, opened 2020-11-18
Update enum types in printerProcessingStateDetail

tomsato-ms:


api-reference/beta/resources/printerstatus.md
```diff
@@ -27,36 +27,50 @@ Represents the processing status of the printer, including any errors.
...
 |:---|:---|:---|
 |unknown|0|The processing state reported by the printer is unknown.|
-|idle|1|The printer is idle and ready to accept new print jobs.|
-|processing|2|The printer is currently processing a print job and will process any pending jobs upon completion.|
-|stopped|3|The printer encountered an issue (for example, ran out of paper in the active tray) and cannot continue the current print job until the issue is addressed. See the **details** value(s) or the **description** value for more information.|
-|unknownFutureValue|4|Evolvable enumeration sentinel value. Do not use.|
+|idle|3|The printer is idle and ready to accept new print jobs.|
+|processing|4|The printer is currently processing a print job and will process any pending jobs upon completion.|
+|stopped|5|The printer encountered an issue (for example, ran out of paper in the active tray) and cannot continue the current print job until the issue is addressed. See the **details** value(s) or the **description** value for more information.|
+|unknownFutureValue|6|Evolvable enumeration sentinel value. Do not use.|
```
braedenp-msft, line 33: These (and the changes to `printerProcessingDetail`) are breaking changes, so we'll need to follow the deprecation process. Are these part of the ongoing GA API scrub?

tomsato-ms: For `printerProcessingState`, the change in numbers associated with the enum is not a breaking change because these numbers are never used. Even after this change kicks in, the number will not be used because we have our own mapping of the enum values in the backend. We also don't return these numbers to our partners, we only return the values in string. The change is only so that the enum number values matches that of the RFC.

The changes in `printerProcessingDetail` could be a breaking change because for example printers with detail of "disconnected" will now show as "shutdown" because in the backend we always stored these detail as "shutdown" but before returning these values to graph, we converted "shutdown" to "disconnected" which is a non-existing enum value in the RFC. This could break partners that are dependent on these specific details.

I'm not sure what GA API scrub is, is there a way to notify customer of these changes?

braedenp-msft: It's a breaking change because of the MS Graph SDKs that get generated. Even though we transmit text over the wire, various SDKs are generated using the numeric values we set in the CSDL.

See [PrinterProcessingStateDetail](https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/blob/master/src/Microsoft.Graph/Generated/model/PrinterProcessingStateDetail.cs) in the .NET SDK, for example.

We also need to avoid moving `unknownFutureValue` around in the enum (we can only add things after it). It can only be moved to the end of the list when we move the API to v1.0.

See [this documentation](https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/54970/Evolvable-enums) which explains how the evolvable enum pattern works.

tomsato-ms: I see, I wasn't thinking of the SDKs. 
To make it less breaking, I will revert the changes on `printerProcessingState`. 
For `printerProcessingDetail` I will add 16 new values to the end of the list, and update these existing values:
(rename) outputAlmostFull -> outputAreaAlmostFull
(remove) disconnected
(remove) queueFull

When you say avoid moving `unknownFutureValue`, does it mean to leave the `unknownFutureValue` as #20, and add the new values starting from #21?

tomsato-ms: I've made the change to not change the existing enum orders, and only remove/update the ones being changed, and add the new values in place of `unknownFutureValue` from #20. The Service code is already checked in, it's already affecting customers now. Resolving this comment as I need to get the new SDK out as soon as possible.

braedenp-msft: Yep, you should add new values starting at `21`.

I don't see the new commit.

tomsato-ms: I've actually added the new values starting from `20` in the CSDL changes which is going to be deployed this week. After talking with Nishant, we decided that adding to `20` should be good since it is already a breaking change where we need customers to update the SDK.

The new commit is https://github.com/microsoftgraph/microsoft-graph-docs/pull/10729/commits/2b85b6073161fcd2bc39f4f50f6dd12c8eeec6d7
If you check the 'Files Changed' tab, the order should be kept for the existing enums.

braedenp-msft: Angela had previously asked us to stick to the proper deprecation process even though we're in beta, so let's make sure we try to avoid publishing breaking changes in the future without going through the process.

We should tag Angela and Laura and let them know about the upcoming breaking change when flipping this PR to "content review".

We should also avoid moving the sentinel value within a particular Graph version. The evolvable enum pattern protects clients that **don't** know about new values by positioning them after the last known value ("unknownFutureValue").

By moving the sentinel, clients now have a (newly) valid value in the enum position that was previously the sentinel.

tomsato-ms: Ok, I will make sure to follow the proper deprecation process next time.
I will add Angela and Laura to the PR when I mark the PR as "content review" as well.
Thanks

concepts/changelog.md
```diff
@@ -28,6 +28,12 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | Addition | beta | Added [Get the installed app in team](/graph/api/team-get-installedapps?view=graph-rest-beta&preserve-view=true), [Get installed app for user](/graph/api/userteamwork-get-installedapps?view=graph-rest-beta&preserve-view=true)|
 | Addition | beta | Added [Get chat between user and app](/graph/api/userscopeteamsappinstallation-get-chat?view=graph-rest-beta&preserve-view=true) method to the new [userScopeTeamsAppInstallation](/graph/api/userScopeTeamsAppInstallation?view=graph-rest-beta&preserve-view=true) resource. |
 
+### Devices and apps | Cloud printing
```
Lauragra, line 31: The H3 sections in the changelog are alphabetized by month, so please put this entry in the correct alphabetical order. Please also include an update to the PrintService JSON file in the changelog folder, for the filterable changelog experience.

tomsato-ms: Ok. I've updated the changelog.md to order the H3 sections in alphabetical order.
I've also updated the PrintService JSON file.

Lauragra (approved):

concepts/changelog.md
```diff
@@ -30,6 +30,12 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | :-------------- | :------------ | :--------------------------------------- |
 | Addition | beta | Added the [custodian](/graph/api/resources/custodian?view=graph-rest-beta&preserve-view=true), [unifiedGroupSource](/graph/api/resources/unifiedgroupsource?view=graph-rest-beta&preserve-view=true), and [siteSource](/graph/api/resources/sitesource?view=graph-rest-beta&preserve-view=true) resources and associated methods. |
 
+### Devices and apps | Cloud printing
+
+| **Change type** | **Version**   | **Description**                          |
+|:---|:---|:---|
+| Change | beta | Updated enum types in [printerProcessingStateDetail](graph/api/resources/printerstatus?view=graph-rest-beta&preserve-view=true#printerprocessingstatedetail-values) resource. |
```
Lauragra, line 37: Please fix this link - you probably need to move the section link to before the version parameter.

tomsato-ms: do you mean like this? graph/api/resources/printerstatus#printerprocessingstatedetail-values?view=graph-rest-beta&preserve-view=true

I tested it manually, but having the section link after version parameter routes me to the right section
https://docs.microsoft.com/en-us/graph/api/resources/printerstatus?view=graph-rest-beta&preserve-view=true#printerprocessingstatedetail-values

whereas having the section link before version parameter does not
https://docs.microsoft.com/en-us/graph/api/resources/printerstatus#printerprocessingstatedetail-values?view=graph-rest-beta&preserve-view=true

tomsato-ms: So I think it's fine to keep the section link after the version parameter

Lauragra: @tomsato-ms , the section link doesn't work across content sets. If you test it in the changelog on the review site, it's a broken link: https://review.docs.microsoft.com/en-us/graph/changelog?branch=pr-en-us-10729. Can you please try to fix? Or, we can just have the link go to the resource topic and not the section.

tomsato-ms: I see, it seems like the link wasn't working because I didn't have a slash / in front of the link path.
I've submitted a commit to fix this.
Thanks

PR #10802, opened 2020-11-24
add description for CR timeout behavior

yizhenww:
previous doc was missing timeout description for specific scenario, adding it in to prevent confusion.


Lauragra (changes requested):
Please update the same topic in /beta.

Lauragra (approved):

PR #10833, opened 2020-11-25
update permission-update to be more specific on what fields and permission types can be modified

k-tsoi:


JeremyKelley (approved):
Looks good to me.

PR #10843, opened 2020-11-26
changes for recurrence properties in V1

Bhupesh12Hub:
@angelgolfer-ms , please review the PR.


api-reference/v1.0/api/event-get.md
```diff
@@ -100,9 +100,53 @@ The first example gets the specified event. It specifies the following:
 }-->
 
 ```msgraph-interactive
```
angelgolfer-ms, line 102: @Bhupesh12Hub , I think you should keep the original first example which is the basic case, getting a single event by simply specifying its ID. Recurring meetings with these 3 new properties you're GA-ing deserves a more focused example, not convoluting the basic case of getting a single event not as part of a recurring series. 
Then create a separate example that is basically example 4 in the event-get.md topic in beta.

angelgolfer-ms: In the separate example, don't forget to mention you're getting an event which is a series master. 
Also your example request should show this query parameter:
$expand=exceptionOccurrences
Without that, what you show in the example response doesn't match the request.

api-reference/v1.0/resources/event.md
```diff
@@ -64,10 +64,12 @@ by providing a [delta](../api/event-delta.md) function.
 |attendees|[attendee](attendee.md) collection|The collection of attendees for the event.|
 |body|[itemBody](itembody.md)|The body of the message associated with the event. It can be in HTML or text format.|
 |bodyPreview|String|The preview of the message associated with the event. It is in text format.|
+|cancelledOccurrences|String collection|Contains **occurrenceId** property values of cancelled instances in a recurring series, if the event is the series master. Instances in a recurring series that are cancelled are called cancelledOccurences.<br><br>Returned only on $select in a [Get](../api/event-get.md) operation which specifies the id of a series master event (that is, the seriesMasterId property value).|
```
angelgolfer-ms, line 67: Line 67:
Can you please correct the misspelling "cancelledOccurences"? Should be "cancelledOccurrences".
That misspelling is in both the v1 and beta topics.

concepts/changelog.md
```diff
@@ -13,6 +13,11 @@ For a summary of the value of these API changes, as well as recent tools, compon
```
angelgolfer-ms, line 13: GLOBAL:
Please be aware that we are releasing a new changelog experience. For the time being, we're updating both the changelog.md and the corresponding changelog json file for your workload (in your case, that's the Microsoft.Exchange.json file). The changelog json file is in the changelog folder of your branch.
I suggest you search for "cancelledOccurrences" in Microsoft.Exchange.json, see how the ChangeList is specified when the properties were added to beta, and add a similar ChangeList for the v1 version.

CDK topic: https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55075/Graph-doc-stub-generator-tool?anchor=update-the-json-changelog

angelgolfer-ms (changes requested):
Please see my comments, thanks!

api-reference/v1.0/resources/event.md
Bhupesh12Hub: Done

concepts/changelog.md
Bhupesh12Hub: Done

api-reference/v1.0/api/event-get.md
Bhupesh12Hub: Need more clarification, because while performing the above action API doctor is failing.

angelgolfer-ms: @Bhupesh12Hub , what I mean is, in event-get.md in v1:
(1) Revert example 1 to what it is on the live site. 
(2) Add a third example. That example should be essentially the same as [example 4](https://docs.microsoft.com/en-us/graph/api/event-get?view=graph-rest-beta&tabs=http#example-4-expand-a-series-master-event) in event-get.md in beta.

Reasons are in my first comment.

changelog/Microsoft.Exchange.json
```diff
@@ -1,5 +1,23 @@
...
+          "ChangedApiName": "cancelledOccurrences,exceptionOccurrences,occurrenceId",
+          "ChangeType": "Addition",
+          "Description": "Added the **cancelledOccurrences**, **exceptionOccurrences**, and **occurrenceId** properties to the [event](https://docs.microsoft.com/en-us/graph/api/resources/event?view=graph-rest-1.0) entity.",
+          "Target": "event"
+        }
+      ],
+      "Id": "8a32aa94-ac63-42d3-b323-5312bfc80a9d",
+      "Cloud": "prd",
+      "Version": "v1.0",
+      "CreatedDateTime": "2020-08-01T00:00:00",
```
angelgolfer-ms, line 17: Line 17:
Please update the date to reflect more closely when this will get published. The accuracy to the month matters!

changelog/Microsoft.Exchange.json
```diff
@@ -1,5 +1,23 @@
 {
   "changelog": [
+    {
+      "ChangeList": [
+        {
+          "Id": "8a32aa94-ac63-42d3-b323-5312bfc80a9d",
```
angelgolfer-ms, line 6: Lines 6 and 14:
These 2 GUID values are the same as the ones in lines 42 and 50.
Please use a generator to come up with another GUID and assign the Id in lines 6 and 14 to that GUID.

changelog/Microsoft.Exchange.json
```diff
@@ -1,5 +1,23 @@
...
+          "ApiChange": "Property",
+          "ChangedApiName": "cancelledOccurrences,exceptionOccurrences,occurrenceId",
+          "ChangeType": "Addition",
+          "Description": "Added the **cancelledOccurrences**, **exceptionOccurrences**, and **occurrenceId** properties to the [event](https://docs.microsoft.com/en-us/graph/api/resources/event?view=graph-rest-1.0) entity.",
```
angelgolfer-ms, line 10: Lines 8 and 10:
Even though this approach is technically supported at this point, the owner of the new changelog experience explained that in order for future filtering (which is the reason for the new changelog) to work as designed, technically ChangedApiName should specify only one new property at a time.
That means you'd have 3 changelist records, instead of one, like now.
To give a consistent experience for customers, and let the feature owner determine if there's cause for changing the filtering design, I'd suggest you follow that and break this up to 3 separate changelist records.

concepts/changelog.md
```diff
@@ -13,6 +13,11 @@ For a summary of the value of these API changes, as well as recent tools, compon
 
 ## December 2020
 
+### Calendar
+
+| **Change type** | **Version**   | **Description**                          |
+| :-------------- | :------------ | :--------------------------------------- |
+| Addition | v1 | Added the **cancelledOccurrences**, **exceptionOccurrences**, and **occurrenceId** properties to the [event](/graph/api/resources/event?view=graph-rest-1.0&preserve-view=true) entity.
```
angelgolfer-ms, line 20: Since this is the last day to publish this year and you still need to fix the example in event-get.md, please update changelog.md to surface this change under January 2021.

angelgolfer-ms (changes requested):
@Bhupesh12Hub , I responded to your question and gave a few more comments. 
Please take a look. 
Looking back, this is a simple PR, in the future, we should try to tighten the iterations. Feel free to message me on Teams. I do get many Github notifications each day, so relying on Github alone to communicate is not efficient.

PR #10919, opened 2020-12-04
Adding replyURLIndex and default

luleonpla:


luleonpla:
Please keep this open

lramosvea:
I added the schema live label since I can see the defaultRedirectUri property for application in the live metadata. However, I didn't see the new resources in the live metadata.

PR #10925, opened 2020-12-04
Added enum to Microsoft AAD

wachuga:


Lauragra:
@davidmu1 , please review. Are we missing a value in the enum type definition?

angelgolfer-ms:
@davidmu1 , please track down if and where the enum itself is documented. It's not in master right now.

davidmu1:
@wachuga - can you please provide more information about this change? @Lauragra and @angelgolfer-ms , currently this enumeration is not defined in the beta documentation. As is, unless more information is provided, I suggest closing the request.

davidmu1 (commented):
The enumeration can't be found in the repository. This change is not currently accurate without more context.

PR #10949, opened 2020-12-07
Adding Hidden Folder Support to mailFolders Entity in Beta

abheek-das:
Adding new mailFolder property: "isHidden"


api-reference/beta/resources/mailfolder.md
```diff
@@ -77,6 +77,7 @@ Well-known names work regardless of the locale of the user's mailbox, so the abo
 |childFolderCount|Int32|The number of immediate child mailFolders in the current mailFolder.|
 |displayName|String|The mailFolder's display name.|
 |id|String|The mailFolder's unique identifier.|
+|isHidden|Boolean|Indicates whether the mailFolder is hidden.|
 |parentFolderId|String|The unique identifier for the mailFolder's parent mailFolder.|
```
angelgolfer-ms, line 81: Line 80:
FYI I added one more point about this property and then link to the H3 section below.

api-reference/beta/resources/mailfolder.md
```diff
@@ -93,6 +94,13 @@ https://outlook.office.com/api/beta/me/folders/inbox/messages?$count=true&$filte
 
 Mail folders in Outlook can contain more than one type of items, for example, the Inbox can contain meeting request items which are distinct from mail items. `TotalItemCount` and `UnreadItemCount` include items in a mail folder irrespective of their item types.
 
+### Hidden Mail Folders
```
angelgolfer-ms, line 97: Line 97:
Please remember, all section titles should use sentence case.
Microsoft Writing Style Guide: https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=33685

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -171,7 +193,77 @@ Content-type: application/json
...
+  "type": "#page.annotation",
+  "description": "List mailFolders",
+  "keywords": "",
+  "section": "documentation",
+  "tocPath": "",
+  "suppressions": [
+  ]
+}
+-->
+
```
angelgolfer-ms, line 216: Lines 203-215:
BTW, each topic should have only one set of these lines at the end of the topic. I am deleting these ones.

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -44,7 +55,10 @@ Do not supply a request body for this method.
 ## Response
 
 If successful, this method returns a `200 OK` response code and collection of [mailFolder](../resources/mailfolder.md) objects in the response body.
-## Example
+## Examples
+
+### Example 1
```
angelgolfer-ms, line 60: Line 60:
Now that we added an example 2, it's helpful to add a short description to the H3 section title. You can take a look at my change to this line.
BTW, if it's necessary to provide more info, you can do so in an additional paragraph. Usually that should be right after the H3 before the H4 "Request", unless you have specific text to go under both the H4 "Request" and H4 "Response" sections.

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -30,8 +30,19 @@ One of the following permissions is required to call this API. To learn more, in
 GET /me/mailFolders
 GET /users/{id | userPrincipalName}/mailFolders
 ```
+
+To optionally also include **hidden** mail folders in the response:
```
angelgolfer-ms, line 34: Line 34:
You might notice that I removed "also" in this sentence because "include" implies "this is not all but is part of the response". I can imagine you included "also" for emphasis, but Microsoft Voice says trim every redundant word :-)

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -30,8 +30,19 @@ One of the following permissions is required to call this API. To learn more, in
...
+<!-- { "blockType": "ignored" } -->
+```http
+GET /me/mailFolders/?includeHiddenFolders=true
+GET /users/{id | userPrincipalName}/mailFolders/?includeHiddenFolders=true
+```
+
 ## Optional query parameters
```
angelgolfer-ms, line 41: Lines 41-45:
Please take notice that I followed the CDK (https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/55072/Parameters?anchor=documenting-odata-system-query-options) and applied a more consistent structure to this section, since this operation supports both query parameters and OData system query options. Another example that has a similar structure for query parameters: [list calendarView](https://docs.microsoft.com/en-us/graph/api/user-list-calendarview?view=graph-rest-1.0&tabs=http#query-parameters).

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -171,7 +193,77 @@ Content-type: application/json
...
+  "tocPath": "",
+  "suppressions": [
+  ]
+}
+-->
+
+### Example 2
+
+##### Request
+The next example uses `includeHiddenFolders` query parameter to get a list of mail folders that also includes the **hidden** mail folders.
```
angelgolfer-ms, line 220: Line 220:
Please take a look that I have moved this paragraph before the H4 "Request" section (I corrected it from H5 to H4, the heading level should go continuous). I added a sentence to explicitly point out which returned folder is hidden and its isHidden property.

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -171,7 +193,77 @@ Content-type: application/json
...
+```http
+HTTP/1.1 200 OK
+Content-type: application/json
+Content-length: 232
+
+{
+    "@odata.context": "https://graph.microsoft.com/beta/$metadata#users('68ca8ec0-11f8-456b-a785-70d9936650d5')/mailFolders",
+    "value": [
+        {
+            "id": "AAMkADg3NTY5MDg4LWMzYmQtNDQzNi05OTgwLWAAA=",
```
angelgolfer-ms, line 249: @abheek-das , question about the example response - how does a user get a mailbox without the Inbox?
I think a number of the folders cannot be deleted?
I changed line 233 to make that a bit more obvious.

api-reference/beta/resources/mailfolder.md
```diff
@@ -93,6 +94,13 @@ https://outlook.office.com/api/beta/me/folders/inbox/messages?$count=true&$filte
...
+By default, a [list mailFolders](../api/user-list-mailfolders.md) returns only those mail folders which are not hidden. To optionally also include the **hidden mail folders** in the result, the query parameter `includeHiddenFolders=true` needs to be included in the API call. The `isHidden` property indicates whether the mail folder is hidden.
```
angelgolfer-ms, line 98: Line 98: 
I moved this paragraph to the end of this section. 
This is a resource topic; I think it's helpful to list the info about the isHidden property first, and then get to the list operation.

api-reference/beta/resources/mailfolder.md
```diff
@@ -93,6 +94,13 @@ https://outlook.office.com/api/beta/me/folders/inbox/messages?$count=true&$filte
 
 Mail folders in Outlook can contain more than one type of items, for example, the Inbox can contain meeting request items which are distinct from mail items. `TotalItemCount` and `UnreadItemCount` include items in a mail folder irrespective of their item types.
 
+### Hidden Mail Folders
+By default, a [list mailFolders](../api/user-list-mailfolders.md) returns only those mail folders which are not hidden. To optionally also include the **hidden mail folders** in the result, the query parameter `includeHiddenFolders=true` needs to be included in the API call. The `isHidden` property indicates whether the mail folder is hidden. 
+
+**Note**: The `isHidden` property can only be set once when [creating a new mailFolder](../api/user-post-mailfolders.md), but cannot be updated using a PATCH. To change the `isHidden` status of a folder, delete the existing folder and create a new one with the desired value.
```
angelgolfer-ms, line 100: Line 100:
Please note that I changed the first sentence to use active voice rather than passive. We generally avoid passive voice in tech writing, so that we don't inadvertently make it unclear who/what is the subject that does a task.

concepts/changelog.md
```diff
@@ -58,6 +58,12 @@ For a summary of the value of these API changes, as well as recent tools, compon
 | Addition | v1.0 | Added the [authorizationPolicy](/graph/api/resources/authorizationpolicy?view=graph-rest-1.0&preserve-view=true) resource type. |
 | Addition | beta | Added [email authentication method configuration](/graph/api/resources/emailauthenticationmethodconfiguration?view=graph-rest-beta) to the [authentication methods policy](/graph/api/resources/authenticationmethodspolicies-overview) API.|
 
+### Mail
+
+| **Change type** | **Version**   | **Description**                          |
+| :-------------- | :------------ | :--------------------------------------- |
+| Addition | beta | Added the **isHidden** property and **includeHiddenFolders** query parameter to the [mailFolder](/graph/api/resources/mailfolder?view=graph-rest-beta&preserve-view=true) entity. |
```
angelgolfer-ms, line 65: Line 65:
Technically, you're adding the query parameter to the "list" operation, right?
I broke up the changelog row to two rows.

angelgolfer-ms: Didn't mention this above:
In general, you apply bolding to API element names, like property names, and apply the tick marks to programmability terms like parameter names, or general programmability terms like `true`, `false`.

changelog/Microsoft.Exchange.json
```diff
@@ -1,5 +1,23 @@
...
   "changelog": [
+    {
+      "ChangeList": [
+        {
+          "Id": "ce3c6f2b-b3e3-4efe-aa1a-046ed1b62e3c",
+          "ApiChange": "Property",
+          "ChangedApiName": "mailFolder",
+          "ChangeType": "Addition",
+          "Description": "Added _isHidden_ property to **mailFolder** entity.",
+          "Target": ""
```
angelgolfer-ms, line 11: FYI, please take a look at my commit:
Line 8:
ChangedApiName in this case should be the property you're adding.
Line 10:
Since the description will appear on a web page (https://developer.microsoft.com/en-us/graph/changelog), we need to include the appropriate formatting (bolding the property name) and linking (link to the mailFolder topic).
Line 11:
Target in this case should be the parent resource name.

api-reference/beta/api/user-list-mailfolders.md
```diff
@@ -30,8 +30,19 @@ One of the following permissions is required to call this API. To learn more, in
 GET /me/mailFolders
 GET /users/{id | userPrincipalName}/mailFolders
 ```
```
angelgolfer-ms, line 32: Please note that after line 17 at the top part of the topic, I added info about the default behavior of this operation which is not including hidden folders. I then mentioned the use of the query parameter, for fuller context.

angelgolfer-ms (approved):
@abheek-das , please take a look at the commit I did to directly incorporate my feedback into your PR. I did so in the interest of time, as there's only a week left before most folks start their break for the holiday, and there are a number of PRs in the queue to review and publish.
I also tried to add comments to call out what I changed and for what reasons, hoping that you will be able to apply similar guidelines in future PRs.
Please let me know if you have any questions.
When the API is available to most external customers, please set the PR to Ready to merge. You can also message me offline to let me know the timeline.

Again, your PR came pretty well thought out and executed, thank you for paying that high level of attention!

api-reference/beta/api/user-list-mailfolders.md
abheek-das: Thanks for the guidelines - this actually makes sense to me! Will keep in mind going forward.

changelog/Microsoft.Exchange.json
abheek-das: @angelgolfer-ms Thanks for the updates! Since the field name was "ChangedApiName", I though we need to mention the API/Entity that's affected and not the property that was added. Is their a doc which provides guidance on the meaning of each of the fields in the changelog/Microsoft.Exchange.json.

abheek-das:
Thank you @angelgolfer-ms  for not only help review the PR, but also making the needed changes :) I will keep the guidelines you mentioned in the comments in mind and they would be super helpful for me going forward! I just have one question regarding the _changelog.json_, which I have mentioned above. As suggested, I will update the PR tag, once the Beta roll out is completed at our end.

PR #10966, opened 2020-12-08
APIs for tab CRUD in chats

AkJo:
This PR deals with introduction of tab CRUD APIs for chats, and fixes inconsistencies in the tab crud APIs for channels.

(Note: This PR is based on another pending PR branch. That is the reason for the extra files in this PR).


api-reference/beta/api/chat-patch-tabs.md
```diff
@@ -0,0 +1,112 @@
+---
+title: "Update tab in chat"
```
Lauragra, line 2: Does the tool output file names with patch? The standard convention is to use update in the file names.

api-reference/beta/api/chat-patch-tabs.md
```diff
@@ -0,0 +1,112 @@
...
+|:---------------|:--------|
+| Authorization  | Bearer {token}. Required.  |
+| Content-Type  | application/json  |
+
+## Request body
+In the request body, supply a JSON representation of [tab](../resources/teamstab.md) object.
+
+## Response
+
+If successful, this method returns a `200 OK` response code.
```
Lauragra, line 46: Update to state that it also returns an updated teamstab object, per the example.

api-reference/beta/api/chat-patch-tabs.md
AkJo: Leaving it based as is based on discussion.

api-reference/beta/api/chat-patch-tabs.md
AkJo: Fixed.

PR #10980, opened 2020-12-08
Permissionstest2

davidmu1:


api-reference/v1.0/api/subscription-post-subscriptions.md
```diff
@@ -15,55 +15,133 @@ Subscribes a listener application to receive change notifications when the reque
...
+|:--------------- |:---------------------- |:----------------- |
+| Delegated (work or school account) | Not supported. | Not supported. |
+| Delegated (personal Microsoft account) | Not supported. | Not supported. |
+| Application | Chat.Read.All | Not supported. |
+
+### Contact
+
+| Permission type | Recommended permission | Other permissions |
+|:--------------- |:---------------------- |:----------------- |
+| Delegated (work or school account) | Contacts.Read | Not supported. |
```
Jackson-Woods, line 56: Contacts.ReadWrite doesn't apply under "other permissions"?

api-reference/v1.0/api/attachment-get.md
```diff
@@ -39,15 +39,32 @@ For an item attachment that is a [contact](../resources/contact.md), [event](../
...
 
-* If accessing attachments in messages: Mail.Read.
-* If accessing attachments in events: Calendars.Read.
-* If accessing attachments in group posts: Group.Read.All.
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. If necessary, you can consider other permissions that have a higher privilege for your specific scenario. For more information, see [Calendars permissions](/graph/permissions-reference#calendars-permissions), [Group permissions](/graph/permissions-reference#group-permissions), and [Mail permissions](/graph/permissions-reference#mail-permissions).
 
-<!--
-* If accessing attachments in group events or posts: Group.Read.All.
--->
+### Messages
```
angelgolfer-ms, line 45: In the case of documenting the GET operation on attachment objects, this treatment lends room for misunderstanding. I think it's partly because there's more than one vector of variance when documenting attachments - first, there are the attachment types (file, item, reference). Then there are the types of item attachments (contact, event, message). Then, there are the objects that can take on attachments (event, message, Outlook task, group post).

So, for instance, when a customer scans the new section titles in lines 45, 53, and 61, what should the customer assume these names that represent resources in Graph mean? The customer is likely not most familiar with the API (and that's why he/she is looking at the docs). Does the section title "Messages" mean, the customer is getting an attachment that's a message, or does it mean getting an attachment off a message...? But wait, if it's the latter, what type of attachment is that, and does it matter to the permissions what type of attachment it is?

I think in an extreme case of attachments that involves that many vectors, which is atypical, this permissions approach makes the customer think too hard and, makes room for misunderstanding. But it's good you picked this example. 

I can see similar confusion can arise if we apply this approach to the create operation for attachments as well.

api-reference/v1.0/api/opentypeextension-post-opentypeextension.md
```diff
@@ -17,21 +17,95 @@ Create an open extension ([openTypeExtension](../resources/opentypeextension.md)
...
-| [group](../resources/group.md) | Group.ReadWrite.All | Not supported | Group.ReadWrite.All |
-| [group event](../resources/event.md) | Group.ReadWrite.All | Not supported | Not supported |
-| [group post](../resources/post.md) | Group.ReadWrite.All | Not supported | Group.ReadWrite.All |
-| [message](../resources/message.md) | Mail.ReadWrite | Mail.ReadWrite | Mail.ReadWrite | 
-| [organization](../resources/organization.md) | Organization.ReadWrite.All | Not supported | Organization.ReadWrite.All |
-| [personal contact](../resources/contact.md) | Contacts.ReadWrite | Contacts.ReadWrite | Contacts.ReadWrite |
-| [user](../resources/user.md) | User.ReadWrite | User.ReadWrite | User.ReadWrite.All |
-| [task](../resources/todotask.md) | Tasks.ReadWrite | Tasks.ReadWrite | Tasks.ReadWrite.All |
-| [tasklist](../resources/todotasklist.md)  | Tasks.ReadWrite | Tasks.ReadWrite | Tasks.ReadWrite.All |
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. For more information, see [Permissions](/graph/permissions-reference).
```
angelgolfer-ms, line 20: In this example of a "combo" topic (one topic supporting an operation that is supported by a number of resources, 1:many mapping), again, I think we skipped explicitly point out "what are the supported resources?"

This was actually one important piece of info that the existing design explicitly points out, in the replaced line 22, first column heading "Supported resource".

Otherwise, if I am new to this feature open extensions, what am I to make out of these H3 section titles "device", "event", "group", and so forth?
When designing the existing approach, I wanted to avoid having to introduce the supported resources  - first in the top section, explicitly listing them, then again in the Permissions section - to keep the topic short, understandable, and scannable.

api-reference/v1.0/api/attachment-get.md
davidmu1: I think the information you suggest here doesn't exist in the current version of the page. This is the information in the section:

If accessing attachments in messages: Mail.Read
If accessing attachments in events: Calendars.Read
If accessing attachments in Outlook tasks: Tasks.Read
If accessing attachments in group posts: Group.Read.All

All my proposal does is move these into tables. Please provide your suggested proposal.

api-reference/v1.0/api/subscription-post-subscriptions.md
davidmu1: The current page does not list any other permissions for contacts.

api-reference/v1.0/api/opentypeextension-post-opentypeextension.md
davidmu1: You are to think  that the section headings and their associated tables are the resources that have permissions for the API. Would a sentence in the intro stating that the sub-sections are supported resources make it clearer?

api-reference/v1.0/api/attachment-get.md
davidmu1: The current page doesn't include the information that you are referring to. Here is the contents of the current section:

If accessing attachments in messages: Mail.Read
If accessing attachments in events: Calendars.Read
If accessing attachments in Outlook tasks: Tasks.Read
If accessing attachments in group posts: Group.Read.All

All I did was move the existing information into tables. Please provide an example of what you are think the section should be.

api-reference/v1.0/api/subscription-post-subscriptions.md
davidmu1: The current page doesn't include other permissions for contacts.

api-reference/v1.0/api/opentypeextension-post-opentypeextension.md
dkershaw10: +1 for what David says.  I think it's clear enough and would definitely be the case with an additional sentence in the intro.
And, if you are new to extensions, is this really the first topic we expect the reader to land on?

api-reference/v1.0/api/attachment-get.md
angelgolfer-ms: It might be clearer and take less time if we talk about this, but we need to sync up a meeting, so let me try.

So as I mentioned, there are several vectors of variance:
(1) "first, there are the attachment types (file, item, reference)" - that's in lines 19-21.
(2) "Then there are the types of item attachments (contact, event, message)." - that's part of line 20 in parentheses.
(3) "Then, there are the objects that can take on attachments (event, message, Outlook task, group post)." - that's lines 14-15.

api-reference/v1.0/api/opentypeextension-post-opentypeextension.md
angelgolfer-ms: Ok, in my PR #11477 , I added a sentence in the top section for each of the CRUD topics for openTypeExtension to explicitly point out the Permissions table shows the list of supporting resources.

@dkershaw10 , in general, there is a chance that a customer who's not as familiar with this API would start in the create operation topic. Many customers ignore our TOC and use search, even when they're already in the docs. Say if they are aware of this feature but want to verify how to create an extension, searching for "microsoft graph create extension" indeed returns this topic as the first search result.

api-reference/v1.0/api/attachment-get.md
angelgolfer-ms: But you made me realize the existing content is not as clear. I made some clarification in my work-in-progress PR #11477  and it's really important that a combo topic like this gets customers going in a breeze. That's the whole point about APIs in a "graph", customers can use the same approach with an API on different resources without the cognitive load of learning something different for each resource.

angelgolfer-ms: @davidmu1 , FYI, I merged PR #11477 into master yesterday. Basically I am advocating keeping the combo topics for the APIs for attachment, openTypeExtension, and subscription using one Permissions table that shows only the permissions with least privileges for each of the supporting resources. The multiple perms tables in a combo topic can be overwhelming. This defeats the purpose of providing the same endpoint, same API that supports multiple workloads, so that it's easier for developers.

Besides, the reality (I can go through the data from the recent dev survey if you like) is even the advanced devs still find auth the top hurdle, meaning just listing all supported perms (for most of our current topics) hasn't been sufficient anyway. They'd still need to look up the auth docs, and perhaps seek help.

Hence I'm advocating keeping the existing combo topics with the similar single perm table approach. Take a look at the preamble.

In the future when there's full topic generation, there will likely be few to no new combo topics anyway. So this issue won't apply of creating an overwhelming Permissions section won't apply. 

I thank you again for pointing out the l

api-reference/v1.0/api/opentypeextension-post-opentypeextension.md
dkershaw10: Not being funny, but I still think this is pretty obvious.  If you've searched for open extension, you probably know that this allows you to add data to resources in Graph.  Which resources? Well we show them in topic.  IMHO we are overthinking this.

api-reference/v1.0/api/subscription-post-subscriptions.md
```diff
@@ -15,55 +15,133 @@ Subscribes a listener application to receive change notifications when the reque
...
+
+### User
+
+| Permission type | Recommended permission | Other permissions |
+|:--------------- |:---------------------- |:----------------- |
+| Delegated (work or school account) | User.Read.All | Not supported. |
+| Delegated (personal Microsoft account) | User.Read.All | Not supported. |
+| Application | User.Read.All | Not supported. |
+
+In addition to selecting permissions, the following information should be considered:
```
Lauragra, line 132: This content does not apply only to the User resource. For clarity and discoverability, the bullets should be placed under the applicable resource H3.

api-reference/v1.0/api/attachment-get.md
```diff
@@ -39,15 +39,32 @@ For an item attachment that is a [contact](../resources/contact.md), [event](../
...
+
+| Permission type | Recommended permission | Other permissions |
```
Lauragra, line 47: What is the pattern for when we include two columns - recommended, and other - for the permissions tables? It doesn't look like other topics in this PR have that column.

api-reference/v1.0/api/attachment-get.md
```diff
@@ -39,15 +39,32 @@ For an item attachment that is a [contact](../resources/contact.md), [event](../
 Attempting to get the `$value` of a reference attachment returns HTTP 405.
 
 ## Permissions
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
 
-* If accessing attachments in messages: Mail.Read.
-* If accessing attachments in events: Calendars.Read.
-* If accessing attachments in group posts: Group.Read.All.
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. If necessary, you can consider other permissions that have a higher privilege for your specific scenario. For more information, see [Calendars permissions](/graph/permissions-reference#calendars-permissions), [Group permissions](/graph/permissions-reference#group-permissions), and [Mail permissions](/graph/permissions-reference#mail-permissions).
```
Lauragra, line 43: We'll need to adjust the boilerplate text if we plan to list both Recommended and Other permissions. Probably you can change the link text from "recommended permissions" to "permissions". Also, the For more information part would need to be hand-authored; is the intent that this would be a TODO in the stub generator output? Including the link to the corresponding permissions section seems like a high amount of overhead for contributors and risky if we end up refactoring the Permissions reference in the future.

api-reference/v1.0/api/attachment-get.md
Lauragra, line 47: Instead of labeling the second column "Recommended" permissions, it might be more meaningful to just call it "Least privileged". Then in the boilerplate text preceding the table, you can make it more concise: "Choose one of the following [permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. We recommend that you use the least privileged permission, unless it's necessary for your scenario to use a higher privileged permission."

api-reference/v1.0/api/subscription-post-subscriptions.md
```diff
@@ -15,55 +15,133 @@ Subscribes a listener application to receive change notifications when the reque
...
-If a chat thread is initiated by a user outside the tenant, that chat thread is not owned by the tenant, and does not create change notifications.
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. If necessary, you can consider other permissions that have a higher privilege for your specific scenario. For more information, see [Permissions](/graph/permissions-reference).
+
+### CallRecord (/communications/callRecords)
+
+| Permission type | Recommended permission | Other permissions |
+|:--------------- |:---------------------- |:----------------- |
+| Delegated (work or school account) | Not supported. | Not supported. |
+| Delegated (personal Microsoft account) | Not supported. | Not supported. |
+| Application | CallRecords.Read.All | Not supported. |
```
Lauragra, line 26: It seems odd to have a "recommended" permission and then have Not supported for Other permissions. Not supported conveys a different meaning than "No more privileged permissions". So for the Other column, if there isn't one but the scenario itself is supported, it should say "None".

api-reference/v1.0/api/subscription-post-subscriptions.md
```diff
@@ -15,55 +15,133 @@ Subscribes a listener application to receive change notifications when the reque
...
+
+### ChatMessage (/teams/{id}/channels/{id}/messages)
```
Lauragra, line 28: The high number of H3s in this topic makes for a lot of scrolling before readers get to other important content on the page. I wonder if we can consolidate the H3s by having just one H3 for chatMessge, for example, with several tables for the different syntaxes, and the same for driveItem and so on.

api-reference/v1.0/api/subscription-post-subscriptions.md
```diff
@@ -15,55 +15,133 @@ Subscribes a listener application to receive change notifications when the reque
...
-
-> **Note:** `/teams/getAllMessages` and `/chats/getAllMessages` are available to users that have the
-[required licenses](https://aka.ms/teams-changenotification-licenses).
-
-> **Note:** `/chats/getAllMessages` only returns messages from chats owned by the tenant. 
-If a chat thread is initiated by a user outside the tenant, that chat thread is not owned by the tenant, and does not create change notifications.
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. If necessary, you can consider other permissions that have a higher privilege for your specific scenario. For more information, see [Permissions](/graph/permissions-reference).
+
+### CallRecord (/communications/callRecords)
```
Lauragra, line 20: Global: Resource names are lower camel case, so you'll want the H3s to be callRecord, chatMessage, etc.

api-reference/v1.0/api/user-post-approleassignments.md
```diff
@@ -19,13 +19,13 @@ Use this API to assign an app role to a user. To grant an app role assignment to
...
 
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. For more information, see [Application permissions](/graph/permissions-reference#application-permissions).
 
-|Permission type      | Permissions (from least to most privileged)              |
-|:--------------------|:---------------------------------------------------------|
-|Delegated (work or school account) | AppRoleAssignment.ReadWrite.All, Directory.AccessAsUser.All    |
-|Delegated (personal Microsoft account) | Not supported.    |
-|Application | AppRoleAssignment.ReadWrite.All |
+| Permission type | Recommended permission |
```
Lauragra, line 24: This is an example of a table that doesn't include the "Other" column. We should use the same pattern in all topics; otherwise it's ambiguous to users. We should have a plan for how we will refactor all the existing permissions tables, that hopefully is not manual. What is that plan?

api-reference/v1.0/api/user-post-approleassignments.md
```diff
@@ -19,13 +19,13 @@ Use this API to assign an app role to a user. To grant an app role assignment to
 
 ## Permissions
 
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. For more information, see [Application permissions](/graph/permissions-reference#application-permissions).
```
Lauragra, line 22: This boilerplate text doesn't seem quite right when there is only one permission to choose. If the intent is to only list the least-privileged permission and not have an Other column for additional permissions, the boilerplate text should reflect that: "The following table lists the least privileged permissions required to call this method. For other permissions, see...". I'm not a fan of the manual work required here to deep link to the section of the Permissions reference, though.

api-reference/v1.0/api/calendar-list-events.md
```diff
@@ -17,12 +17,24 @@ To get expanded event instances, you can [get the calendar view](calendar-list-c
...
 
 ## Permissions
-Depending on the type of calendar that the events are in and the permission type (delegated or application) requested, one of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
 
-| Calendar | Delegated (work or school account) | Delegated (personal Microsoft account) | Application |
-|:-----|:-----|:-----|:-----|
-| user calendar | Calendars.Read, Calendars.ReadWrite | Calendars.Read, Calendars.ReadWrite | Calendars.Read, Calendars.ReadWrite |
-| group calendar | Group.Read.All, Group.ReadWrite.All | Not supported. | Not supported. |
+Choose one of the following [recommended permissions](/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. The recommended permissions use the least privilege necessary for calling the method. If necessary, you can consider other permissions that have a higher privilege for your specific scenario. For more information, see [Calendars permissions](/graph/permissions-reference#calendars-permissions) and [Group permissions](/graph/permissions-reference#group-permissions).
```
angelgolfer-ms, line 21: Line 21 and GLOBAL for topics that need to list permissions for up to two scenarios or resources (like the calendar topics that apply to the user calendar and group calendar scenarios):

I'd suggest tweaking the boilerplate to the following:

Choose one of the following [permissions](https://docs.microsoft.com/en-us/graph/auth/auth-concepts#microsoft-graph-permissions) to call this method. Use the least privileged permission that your app requires. To learn more, including [taking caution](https://docs.microsoft.com/en-us/graph/auth/auth-concepts#best-practices-for-requesting-permissions) before choosing more privileged permissions, search for the following permissions in [Permissions](https://docs.microsoft.com/en-us/graph/permissions-reference).

And the permission table headers, for example:

| Permission type | Least privileged permission | Higher privileged permissions |
|:--------------- |:---------------------- |:----------------- |
| Delegated (work or school account) | Calendars.Read | Calendars.ReadWrite |
...

angelgolfer-ms: I found that the first link text in line 21 "recommended permissions", that points to a section that just describes general Graph permissions info, doesn't quite match what I had expected for the "recommended" part of the link text. Basically the destination content doesn't "recommend" anything. 
The next sentence starts with "The recommended permissions use..." which refers to a concept different from what's brought up in the first sentence.
So I feel we might as well cut out the "recommended" in this boilerplate, and let the customer look up the specified permissions and decide which to use.
That approach would also make it simpler to not have to taylor the lead-in text for each topic.

angelgolfer-ms: One more comment - the column header "Other permissions" doesn't convey that permissions are ordered from least to most privileges. I think authors and customers need to be aware of that ordering, so that customers can make the appropriate choice for their app. Hopefully at least the more experienced customers can just make a decision just from the permissions table without necessarily going through the other auth topics.

PR #10986, opened 2020-12-09
Teams: Team Membership API - Bulk and filtering

AkJo:
Added function for adding members to a team in bulk.
Also added filtering example for list team membership API.


api-reference/beta/api/team-list-members.md
```diff
@@ -112,7 +157,100 @@ Content-Type: application/json
...
+    ]
+}
+```
+
+### Example 4: List only those members who are *owners* of the team
+
+The following example shows a request to find all the members who are have *owner* role attached to them.
+
+> [!NOTE]
+> There are some known issues with this functionality.
```
Lauragra, line 224: We should describe the known issues in the Known issues topic and link to it from here.

api-reference/beta/api/team-members-add.md
```diff
@@ -0,0 +1,206 @@
+---
```
Lauragra, line 1: To follow the standard naming convention, this needs to be conversationmember-add, and the title will be conversationMember: add (singular).

Lauragra (changes requested):
I think this PR will be good to go after we settle on how to handle the add action.

api-reference/beta/api/team-list-members.md
AkJo: We are being vague over here on purpose (because the range of issues are varied). Adding more details will lead to unnecessary questions for us. 

The main objective of the note is that people can experiment with that request example but should not take a dependency on it yet.

We are debating whether to keep that specific functionality or not based on how many people use this and the problems they might encounter .

AkJo: And we do want the note to say that the issues are still "being investigated" (and have not been finalized in anyway).

AkJo: Added a known issue link. Hope the link is working.

AkJo:
> I think this PR will be good to go after we settle on how to handle the add action.

@Lauragra 

Renaming to conversationmember-add.md is conflicting with the already existing conversationmember-add.md which caters to an entirely different API.

For now, I am renaming this to conversationmembers-add.md, and maintaining parity with chats-getallmessages.md and channels-getallmessages.md. Lets get this in.

(From a pedantic perspective, ideally, the filename for the action should have been conversationmember-collection-add.md or collection-conversationmember-add.md; and the title would have followed suit.)

api-reference/beta/api/team-members-add.md
AkJo: Renaming to conversationmember-add.md is conflicting with the already existing conversationmember-add.md which caters to an entirely different API.

For now, I am renaming this to conversationmembers-add.md, and maintaining parity with chats-getallmessages.md and channels-getallmessages.md. Lets get this in.

(From a pedantic perspective, ideally, the filename for the action should have been conversationmember-collection-add.md or collection-conversationmember-add.md; and the title would have followed suit.)

PR #11006, opened 2020-12-11
Updated toc to remove entries not needed in the main Table of Content

AkJo:


api-reference/v1.0/toc.yml
```diff
@@ -5058,32 +5058,6 @@ items:
         items:
         - name: Get chat between user and app
           href: api/userscopeteamsappinstallation-get-chat.md
-      - name: Conversation member
```
Lauragra, line 5061: Why are we deleting this section from the TOC? These v1.0 topics are no longer exposed in the TOC and won't be discoverable.

AkJo: As discussed offline, most of these files will get deprecated or are already superseded by alternate files.

api-reference/v1.0/toc.yml
```diff
@@ -5100,41 +5074,47 @@ items:
...
-          href: api/channel-list-tabs.md
-        - name: Get tab in channel
-          href: api/channel-get-tabs.md
-        - name: Add tab to cannel
-          href: api/channel-post-tabs.md
-        - name: Update tab in channel
-          href: api/channel-patch-tabs.md
-        - name: Remove tab from channel
-          href: api/channel-delete-tabs.md
+        - name: In channel
```
Lauragra, line 5106: Do we need this extra level of the TOC when there's only one category? Can the TOC entry be "Tab in channel" or just "Tab"?

PR #11017, opened 2020-12-11
add update print job configuration page

tomsato-ms:
Adding API doc for allowing customers to update print job configuration property.


api-reference/beta/api/printjob-update-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Update print job configuration
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Update the [configuration](../resources/printjobconfiguration.md) property of a [print job](../resources/printjob.md).
+
+Update print job configuration API requires the print job to be in a held state by registering a task trigger for the printer. For details about how to register a task trigger, see [Extending Universal Print to support pull printing](/graph/universal-print-concept-overview#extending-universal-print-to-support-pull-printing).
```
braedenp-msft, line 18: Since the rest of the sentence is in natural language, it would be easier to read as 
> Updating a print job configuration requires..."

api-reference/beta/api/printjob-update-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+| Name          | Description   |
+|:--------------|:--------------|
+| Authorization | Bearer {token}. Required. |
+
+## Request body
+In the request body, supply the values of the relevant [printJobConfiguration](../resources/printjobconfiguration.md) fields. Existing properties that are not included in the request body will maintain their previous values.
+
+| Property     | Type        | Description |
+|:-------------|:------------|:------------|
+|configuration|[printJobConfiguration](../resources/printjobconfiguration.md)|Updated configuration of print job.|
```
braedenp-msft, line 46: This should be a list of `printJobConfiguration` properties, right?

api-reference/beta/api/printjob-update-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Response
+If successful, this method returns a `204 NoContent` response code.
```
braedenp-msft, line 49: `204 No Content`

tomsato-ms: Thanks. Fixed

api-reference/beta/api/printjob-update-configuration.md
tomsato-ms: Sounds good. Fixed the wording.

api-reference/beta/api/printjob-update-configuration.md
tomsato-ms: No just one printJobConfiguration property is being passed here, just like in the Request example below, because a single print job only contains one configuration.

braedenp-msft: The current table suggests the client should pass an object that looks like:
```
{
    "configuration": {
        "feedOrientation": "longEdgeFirst",
        ...
    }
}
```

Since the PATCH URL already references the `configuration` property, we don't need that container. The passed object should look like the one in the example below, so we should either list the `printJobConfiguration` properties out in the table or omit the table entirely and rely on the below example.

tomsato-ms: I see, thanks. I've updated this section and removed the table entirely.

braedenp-msft: Awesome, the rest looks great! Nice work 😊

braedenp-msft (approved):

Lauragra (approved):

PR #11077, opened 2020-12-18
Fix extra lines in http blocks

andrueastman:
This PR fixes some http examples with extra newlines after the startline.

The http headers should appear immediately after the startline (second line) and an empty line used to separate the headers section and the body section. More info [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages#HTTP_Requests).

This fix will also prevent the breaking of the master branch on future versions of ApiDoctor and it will also enable snippet generation for the affected bodies.


andrueastman:
Closing this in favour of #11078

PR #11078, opened 2020-12-18
Remove unwanted new line from requests in affected files

millicentachieng:
This PR fixes some http examples with extra newlines after the startline.

The http headers should appear immediately after the startline (second line) and an empty line used to separate the headers section and the body section. More info here.

This fix will also prevent the breaking of the master branch on future versions of ApiDoctor and it will also enable snippet generation for the affected bodies.


andrueastman (approved):

zengin (approved):

zengin (approved):

angelgolfer-ms (approved):
Thanks for seeking out and fixing the offending topics, @millicentachieng !

PR #11123, opened 2020-12-24
Update search-concept-aggregation.md

ZChristine:
Enrich aggregation filter format for the scenario that multiple values for the same filter are required


api-reference/beta/resources/search-api-overview.md
```diff
@@ -34,7 +34,7 @@ Search requests run on behalf of the user. Search results are scoped to enforce
 |[Use KQL in query terms](#keyword-query-language-kql-support) | **query** |
 |[Sort search results](#sort-search-results)| **sort** |
 |[Refine results using aggregations](#refine-results-using-aggregations)| **aggregations** |
-|[Search across Graph connectors](/graph/search-concept-custom-types)| **contentSources** |
+|[Search Graph connectors](/graph/search-concept-custom-types)| **contentSources** |
```
Lauragra, line 37: This title doesn't match the title of the link target. We also appear to have the same link in this topic with slightly different link titles. If we want to keep both links, please change the link text to "Search custom types imported using connectors".

api-reference/beta/resources/searchrequest.md
```diff
@@ -72,7 +72,7 @@ The following is a JSON representation of the resource.
 - Search [mail messages](/graph/search-concept-messages)
 - Search [calendar events](/graph/search-concept-events)
 - Search content in SharePoint and OneDrive ([files, lists and sites](/graph/search-concept-files))
-- Search [custom types (Graph Connectors)](/graph/search-concept-custom-types) data
+- Search [Graph Connectors)](/graph/search-concept-custom-types) data
```
Lauragra, line 75: Same comment here - the link title should be "Search custom types imported using connectors".

Lauragra (changes requested):
Please update the link text per my comments.

api-reference/beta/resources/search-api-overview.md
nmoreau: Fixed

Lauragra (approved):

PR #11171, opened 2021-01-04
Update people-picker.md

HiltonGiesenow:
removed duplicate "properties", and also changed selectionMode to say "items" instead of "users" because the person might be selecting a group


Lauragra:
Thanks!

PR #9085, opened 2020-07-13
MSA developer support to call app API

sureshja:

PR #11194, opened 2021-01-06
MSA developer support to call app API v1.0

sureshja:
The MSA developer support is now GA. Changes in this payload are exactly same as the one we pushed to /beta (https://github.com/microsoftgraph/microsoft-graph-docs/pull/9085/files).


Lauragra (changes requested):
Please add an update to the changelog.md file and DirectoryServices JSON file for this update.

Lauragra (approved):

PR #11195, opened 2021-01-06
Missing basetype causes issues for client library

MIchaelMainer:
The base type gets set on the generated models. BaseType is set empty in the docs. This causes an issues in the tools that generate the client libraries. 

@snlraju-msft @jasonbro @markwahl-msft @raprakasMSFT @sureshja

Changes to be committed:
	modified:   api-reference/beta/resources/applicationserviceprincipal.md
	modified:   api-reference/beta/resources/approvalsettings.md
	modified:   api-reference/beta/resources/approvalstage.md
	modified:   api-reference/beta/resources/assignmentreviewsettings.md
	modified:   api-reference/beta/resources/expirationpattern.md
	modified:   api-reference/beta/resources/externalitemcontent.md
	modified:   api-reference/beta/resources/regionalformatoverrides.md
	modified:   api-reference/beta/resources/requestorsettings.md
	modified:   api-reference/beta/resources/requestschedule.md
	modified:   api-reference/beta/resources/termsexpiration.md
	modified:   api-reference/beta/resources/userset.md

@zengin This is something we could handle in ApiDoctor by ignoring empty base type.

### DIFF
--- a/api-reference/beta/resources/applicationserviceprincipal.md
+++ b/api-reference/beta/resources/applicationserviceprincipal.md
@@ -40,7 +40,6 @@ The following is a JSON representation of the resource.
 
   ],
   "@odata.type": "microsoft.graph.applicationServicePrincipal",
-  "baseType": "",
   "keyProperty": "id"
 }-->
 
--- a/api-reference/beta/resources/approvalsettings.md
+++ b/api-reference/beta/resources/approvalsettings.md
@@ -34,8 +34,7 @@ The following is a JSON representation of the request approval settings property
   "optionalProperties": [
 
   ],
-  "@odata.type": "microsoft.graph.approvalSettings",
-  "baseType": ""
+  "@odata.type": "microsoft.graph.approvalSettings"
 }-->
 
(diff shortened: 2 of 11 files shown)


angelgolfer-ms (approved):
Per our Teams chat, the current CDK guidance for the baseType attribute is not to set it if there isn't a parent type for the resource (entity or complex type). I will request @millicentachieng to confirm if this guidance is still aligned with API Doctor's expectations. If it is, then we should request API Doctor to check for violation, and the stub topic generator to not set baseType to "" for resources without parent types. For our info, there are currently 155 files (including the 11 included in this PR that are caught by your tool) in the master branch that have baseType set to "".
I can approve this PR for the time being, per your request to unblock the tool that generates the client libraries. We'll keep in mind that API Doctor dependencies may cause us to revisit this issue.

Lauragra (approved):

millicentachieng:
I'll be updating API Doctor to validate empty base types. We will be having a new release by the first week of February.

PR #11208, opened 2021-01-06
Describing new licensing mechanism for MGDC

ajacks-msft:
Beginning February 1, Microsoft Graph data connect will be able to bill based on usage through Azure Billing.


tlenig:
I assume we will wait to deploy this until 2/1?

ajacks-msft:
> I assume we will wait to deploy this until 2/1?

We should give advance warning of an upcoming billing/licensing change, so we should deploy asap. Verify it is clear that the change is taking effect on 2/1 though.

tlenig:
> > I assume we will wait to deploy this until 2/1?
> 
> We should give advance warning of an upcoming billing/licensing change, so we should deploy asap. Verify it is clear that the change is taking effect on 2/1 though.

I'm not sure we should remove the information about the MGdc license then, as we have an ISV (PeopleMatter) with instances of it. Is the expectation that the MGdc license will no longer work on 2/1?

tlenig (approved):

Lauragra (approved):

PR #11237, opened 2021-01-08
Add outOfOfficeSettings for presence

elvinyang-msft:


api-reference/beta/resources/outOfOfficeSettings.md
```diff
@@ -0,0 +1,40 @@
...
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Represents out-of-office settings for a user in [presence](presence.md).
+
+## Properties
+
+| Property            | Type    | Description                                                                    |
+|:--------------------|:--------|:-------------------------------------------------------------------------------|
+| message           | String  | The out-of-office message that the user configured on Outlook client -> "Automatic Replies (Out of Office) or Teams client -> "Schedule out of office" |
```
jsandoval-msft, line 22: Is there a character limit here?

jsandoval-msft (approved):

api-reference/beta/resources/outOfOfficeSettings.md
```diff
@@ -0,0 +1,40 @@
...
+doc_type: resourcePageType
+---
+
+# outOfOfficeSettings resource type
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Represents the out of office settings related to the [presence](presence.md) of a user.
```
Lauragra, line 16: In the presence topic we state that this is only supported for Teams. Does that apply to this property as well?

api-reference/beta/resources/outOfOfficeSettings.md
elvinyang-msft: OOF message can be set on both Teams and Outlook and eventually they call the same API.
There is a length limit on Teams but not on Outlook, and also limit is not talked about on the presence feature document. So I don't think we should list anything either.

api-reference/beta/resources/outOfOfficeSettings.md
elvinyang-msft: We have the note `This resource is currently only supported for Microsoft Teams users.`  on presence is because presence is a navigation property of a user (users/<id>/presence). So the note is to reduce ambiguity that it's not supported for any user.
For the case of `outOfOfficeSettings` resource since it is already a member of presence, and it is not directly under user, I think we do not need a note.

PR #11285, opened 2021-01-14
Update insights-customize-item-insights-privacy.md

elmakhmu:
1. added block describing necessity of calling profile beta.
2. added block describing how to connect MSGraph with more permissions.
3.  deleted "Using earlier versions of the PowerShell module" section. as practice shows admins keep updated sdk module.


elmakhmu:
@Lauragra , @angelgolfer-ms I've added some clarifications in the section "How to configure item insights setting via PowerShell" (we found out with one tenant that current version is not super clear for understanding). Could you, please, have a look on it and if everything is ok - publish to prod. Many thanks in advance!

Lauragra (approved):

PR #11320, opened 2021-01-17
Update group.md non-implemented List properties

Ahmad-K-A:
hideFromOutlookClients and hideFromAddressLists are not exposed on the list-all-groups API. I am proposing this change to clear customer confusion which has been evident on several cases. The properties are only retrievable on the GET /groups/{ID} API. Please adjust the wording as necessary.


Lauragra:
@yyuank , please review and approve. If we accept this change, we need to make the same update to the file in the /beta folder. Thanks.

yyuank (approved):

PR #11329, opened 2021-01-18
Update user-list-transitivememberof.md

JarbasHorst:
Removed numbers 893 and 588 as they don't make any sense

### DIFF
--- a/api-reference/beta/api/user-list-transitivememberof.md
+++ b/api-reference/beta/api/user-list-transitivememberof.md
@@ -155,9 +155,6 @@ HTTP/1.1 200 OK
 Content-type: text/plain
 ```
 
-893
-
-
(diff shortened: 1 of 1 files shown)


api-reference/beta/api/user-list-transitivememberof.md
```diff
@@ -155,9 +155,6 @@ HTTP/1.1 200 OK
 Content-type: text/plain
 ```
 
-893
```
Lauragra, line 158: In this case, the example is returning a count of group members, so having a number in the response might be expected?

JarbasHorst: Thanks! Now I got it. In this case, I'd suggest us moving the response values to the formatted area as it is more clear.

PR #11331, opened 2021-01-18
Update attachment-get.md

abheek-das:
Update attachment-get.md to add new property "attachments" for nested attachments. No Graph schema change, only update of sample response to add the missing properties.


angelgolfer-ms (commented):
Hi @abheek-das ,
Thanks for opening this PR. Can you please first clarify the following about example 3:
(1) Example 3 was set up with this scenario: Message A has an attachment which is message B. Get the properties of A and B. Is that approach shown on the [live site](https://docs.microsoft.com/en-us/graph/api/attachment-get?view=graph-rest-1.0&tabs=http#example-3-expand-and-get-the-properties-of-the-item-attached-to-a-message) still correct (other than missing the bodyPreview and internetMessageHeaders properties)?

(2) If (1) is still correct (other than the missing properties), then is the purpose to show what happens, if message B also has a file attachment X, and that same $expand example request would expand both B and X?

abheek-das:
Hello @angelgolfer-ms ,
For (1): Yes the scenario is the same and the request is unchanged. I checked that all the missing properties are added in the PR.
For (2): Yes that's exactly correct - with the $expand the same example request would expand both B and X (if it exists). Hence, the name nested attachments. You can view up 30 levels of nested attachments using this approach.

Please let me know if you need more context! Thanks!

angelgolfer-ms (changes requested):
Hi @abheek-das ,
Thanks for the confirmation for the purpose of this PR.

For clarity, can you please add your example as example 4, and restore example 3 to the simpler case? So example 3 would show expanding an item attachment (a message attached to another message), and example 4 would show expanding the item attachment (a message) together with its file attachment. I recall customers asking about the simpler case (example 3) before we documented it, so it would be useful to keep that separate from the nesting case (example 4).

I know that would be a bit of overhead in addition to what you originally intended (building on top of example 3). I also wanted to try out spec-ing the description of example 4 and see the experience, so I went ahead and added the sub-section title for example 4, and fixed all the related links including links to the new example 4 and pre-existing links to examples 5-8, shifting them by 1 as examples 6-9 respectively.

So your task is to restore example 3 to what it used to be (but add the bodyPreview and internetMessageHeaders properties), insert your example as example 4, under the example 4 sub-section I already added to each of the v1 and beta attachment-get.md topic.

Also, per our communication some weeks ago, I understand you're basically now the PM owner for the mail-related APIs that used to be owned by Siva, so I went ahead to change the Author name from Siva to your Github name in the 3 MD files currently in this PR. At your earliest convenience, please complete that update of the author field in other doc files :-)

Just to be clear, can you please take a look at my last 2 commits ([commit 1](https://github.com/microsoftgraph/microsoft-graph-docs/pull/11331/commits/c616ff0208db38a9ddc34d7fd9093a41fa745d26) and [commit 2](https://github.com/microsoftgraph/microsoft-graph-docs/pull/11331/commits/ae2ac9785fdb129a7dfdfd7f9adfb90e82082801), so that you know exactly what I have done and where you should please insert your example (in example 4)?

Thanks.

abheek-das:
Hi @angelgolfer-ms ,

Thanks for the feedback. I've:
1. Reviewed your last 2 commits. They look great!
2. Added the response for the newly inserted Example 4, where the original attached item has another attachment of pdf format. The description of the example looks good as well.

Also, acknowledging the request for author name update in the mail API graph pages. I had a draft PR where I think I had made this change, will dig it up, ,review it and float it for review soon :)

Thanks again for helping make the changes!

api-reference/beta/api/attachment-get.md
```diff
@@ -395,7 +399,101 @@ Content-type: application/json
 }
 ```
 
-### Example 4: Get the properties of a reference attachment
+### Example 4: Expand and get the properties of an item attached to a message, including any attachment to the item
+#### Request
```
angelgolfer-ms, line 403: After line 405:
You actually needed to add the request again, for 2 reasons: 
(1) Clarity
(2) API Doctor - surprised that API Doctor didn't hickup! I recall definitely it would have before - requests and responses need to be paired, I learned that a harder way but got the API Doctor creator Ryan (who's now back in our group!) to point that out to me.

I added the request back to both the v1 and beta topics.

Also, in line 476, I suppose the ID that follows attachments should be consistent with that ID in the request? I changed it. It's also truncated like the other IDs, for readability.

angelgolfer-ms (approved):
I made a couple more changes, please see my last [commit ](https://github.com/microsoftgraph/microsoft-graph-docs/pull/11331/commits/0a4e7c181c2ba2559d47b7452a35d39810e27e27).

Next week as soon as you confirm that the API is available in v1 WW, please follow the process and set the PR to Ready to merge. Laura can then merge and publish the PR.

api-reference/beta/api/attachment-get.md
```diff
@@ -19,24 +19,24 @@ Read the properties, relationships, or raw contents of an attachment that is att
 
 An attachment can be one of the following types:
 
-* A file. Programmatically, this is a [fileAttachment](../resources/fileattachment.md) resource.
-* An Outlook item (contact, event or message). Programmatically, an item attachment is an [itemAttachment](../resources/itemattachment.md) resource. You can use `$expand` to further get the properties of that item. See an [example](#request-2) below.
-* A link to a file stored in the cloud. Programmatically, this is a [referenceAttachment](../resources/referenceattachment.md) resource.
+* A file. Programmatically, this is a [fileAttachment](../resources/fileattachment.md) resource. See an [example](#example-1-get-the-properties-of-a-file-attachment) below.
```
Lauragra, line 22: Global: Please avoid using the term "below" to refer to something later in the current article. In line 22, you can be more specific and simply say "See example 1." In line 24, you can change the text to "See example 5." Please remove all occurrences of below. For reference, see this guidance: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/b/below.

api-reference/beta/api/attachment-get.md
```diff
@@ -19,24 +19,24 @@ Read the properties, relationships, or raw contents of an attachment that is att
...
+For a file attachment, the content type is based on its original content type. See an [example](#example-6-get-the-raw-contents-of-a-file-attachment-on-a-message) below.
 
 For an item attachment that is a [contact](../resources/contact.md), [event](../resources/event.md), or [message](../resources/message.md), the raw contents returned is in MIME format.
 
 | Item attachment type  | Raw contents returned |
 |:-----------|:----------|
-| **contact** | [vCard](http://www.faqs.org/rfcs/rfc2426.html) MIME format. See [example](#example-6-get-the-mime-raw-contents-of-a-contact-attachment-on-a-message). |
-| **event** | iCal MIME format. See [example](#example-7-get-the-mime-raw-contents-of-an-event-attachment-on-a-message). |
-| **message** | MIME format. See [example](#example-8-get-the-mime-raw-contents-of-a-meeting-invitation-item-attachment-on-a-message). |
+| **contact** | [vCard](http://www.faqs.org/rfcs/rfc2426.html) MIME format. See [example](#example-7-get-the-mime-raw-contents-of-a-contact-attachment-on-a-message). |
```
Lauragra, line 37: It will help to guide the reader and also help non-sighted readers if you include the example number in the link text. Otherwise there are many links on the page with the same generic link title "example".

api-reference/beta/api/attachment-get.md
abheek-das: @Lauragra Removed all occurrences of below in the context of meaning later from the article in both beta & v1.0 version of the article.

api-reference/beta/api/attachment-get.md
abheek-das: Add the example number explicitly in all links as suggested.

Lauragra (approved):

PR #11354, opened 2021-01-20
Update externalconnection-delete.md

KoenZomers:
Deleting a custom connector through Microsoft Graph results in a HTTP 202 Accepted and not a 204 No Content


Lauragra:
If we update the response example, we also need to update the Response section in the article. @snlraju-msft , please take a look. Thanks.

KoenZomers:
@Lauragra Good catch. I've updated the Response section as well to make mention of the 202

Screenshot of Postman showing the 202 response:
image

PR #11366, opened 2021-01-21
Removed Preview tags from Graph connector articles

rsamai:


concepts/search-concept-overview.md
```diff
@@ -20,7 +19,7 @@ Microsoft Search is an enterprise search engine that delivers productivity gains
 
 The Microsoft Search API provides one unified search endpoint that you can use to [query](/graph/api/search-query) data in the Microsoft cloud - messages and events in Outlook mailboxes, and files on OneDrive and SharePoint - that Microsoft Search already indexes.
 
-### Include custom external data in search experience (preview)
+### Include custom external data in search experience
```
Lauragra, line 22: We use (preview) to indicate that APIs are only in beta and have not been implemented in v1.0. Is it the case that the search connector (Index) APIs are or will be available in v1.0?

rsamai: APIs are still in beta. Let me change it back!

Lauragra (approved):

PR #11373, opened 2021-01-21
Documenting future possibilities of consumption fees

nkramer:


nkramer:
FWIW the verbiage is identical to wording we used to have in the docs.

mishakogan:
@nkramer / @Lauragra "In the future, Microsoft may require you or your customers to pay additional fees based on the amount of data accessed through the API." - this seems very ambiguous - when is this "future" coming? Can documentation be updated to be more specific?

PR #11401, opened 2021-01-23
Clarified minimum 24-month deprecation period is for GA APIs only.

angelgolfer-ms:


mikepizzo (approved):
These look to me like good clarifications -- thanks Angela!

PR #11435, opened 2021-01-27
Update app permission for get printer endpoint

jahsu-MSFT:


api-reference/beta/api/printer-get.md
```diff
@@ -24,7 +24,7 @@ To use the Universal Print service, the user or app's tenant must have an active
 |:---------------|:--------------------------------------------|
 |Delegated (work or school account)| Printer.Read.All, Printer.ReadWrite.All, Printer.FullControl.All |
 |Delegated (personal Microsoft account)|Not Supported.|
-|Application| Not Supported. |
+|Application| Printer.Read.All, Printer.ReadWrite.All |
```
nilakhan, line 28: Also, in printdocument-get-file.md
We need to add this: 
"In addition to the following permissions, the user or app's tenant must have an active Universal Print subscription and have a permission that grants [Get printer](printer-get.md) access."

By replacing this: 
In addition to the following permissions, the user or app's tenant must have an active Universal Print subscription and have a permission that grants [Get printer](printer-get.md) access in case of user's access tokens and either the Printer.Read.All or Printer.ReadWrite.All application permission in case of app-only access tokens.

nilakhan (changes requested):

nilakhan (approved):

Lauragra (approved):

PR #11440, opened 2021-01-27
2021 01 search updates - Typo

nmoreau:


Lauragra:
@nmoreau , what is your intent with this PR? It looks like it includes over 100  generated stubs? Did you use the stub generator tool against a CSDL? Please ensure that the PR includes only the files that you intend to update, and remove the stubs folders. Also, please don't use the Upload files button and instead commit your files via git; otherwise it will fail the line endings validation check. Thanks.

nmoreau:
This is definetly not intended.
I think you merged what I needed so we should be good.
I deleted many other old branch to clean up
Thanks

PR #11452, opened 2021-01-28
adding app perm for EM

markwahl-msft:
Add EntitlementManagement.Read.All and ReadWrite.All app only perms for Azure AD entitlement management APIs


davidmu1 (approved):

changelog/Microsoft.DirectoryServices.json
```diff
@@ -1,5 +1,23 @@
 {
   "changelog": [
+    {
+      "ChangeList": [
+        {
+          "Id": "157d7d86-bf1d-4658-9a13-33d1aec9da4c",
+          "ApiChange": "Resource",
+          "ChangedApiName": "EntitlementManagement.ReadWrite.All,EntitlementManagement.Read.All",
```
Lauragra, line 8: I think what we need to do here is add one changelist item for each resource - accessPackage, etc. Use the resource name in the ChangedApiName and Target, and the description will be "Added the... permissions." ChangedApiName and Target are single-valued and should be the name of an API (not a document, for example).

markwahl-msft: Updated, thanks.

PR #11477, opened 2021-01-30
Addressed higher privileges in combo topics. Added best practices to request least privileged permissions.

angelgolfer-ms:


includes/auth-use-least-privileged.md
```diff
@@ -0,0 +1,10 @@
+---
+author: "jackson-woods"
+localization_priority: Priority
+ms.topic: include
+---
+
+<!-- markdownlint-disable MD041-->
+
+> [!Note]
```
angelgolfer-ms, line 9: Incorporated Jackson's feedback to this on Teams from 1/31.

PR #11532, opened 2021-02-05
[CloudPC]Update boilerplate text

AshleyYangSZ:
Our Marketing and Sales team don’t like the reference to contact the GBBs. So we need to remove it to prevent more further News on our project.


angelgolfer-ms (approved):
Per Bhavya's confirmation in email, since in practice, customers can’t get any info from the GBBs, there’s no point to mislead customers with that sentence. That basically extends the customers' route to find more info to use these APIs but with no results. We'll discuss offline how we should clarify and message the purpose of these APIs in its current stage.

Lauragra (approved):

PR #11535, opened 2021-02-05
RSC permissions patch

nkramer:


api-reference/v1.0/api/channel-get-message.md
```diff
@@ -20,7 +20,7 @@ One of the following permissions is required to call this API. To learn more, in
 |---------|-------------|
 |Delegated (work or school account)| ChannelMessage.Read.All |
 |Delegated (personal Microsoft account)|Not supported.|
-|Application| Not supported. |
+|Application| ChannelMessage.Read.Group* |
```
angelgolfer-ms, line 23: @nkramer , is it true that when such a resource-specific consent (RSC) permission is listed as an application permission, unlike other Graph application permissions, it's the team owner and not the admin who needs to consent?

That question is based on the [Teams endpoint topic](https://docs.microsoft.com/en-us/microsoftteams/platform/graph-api/rsc/resource-specific-consent) you link to, which says:
" The resource-specific consent (RSC) permissions model enables team owners to grant consent for an application to access and/or modify a team's data."

If that is true for all the resource-specific permissions used as application permissions, then I think it's necessary to state that in this section within the Graph docs: https://docs.microsoft.com/en-us/graph/permissions-reference?view=graph-rest-beta#teams-resource-specific-consent-permissions

Please clarify.

PR #11538, opened 2021-02-05
Removed curly apostrophe

MIchaelMainer:
@zengin This should cleanup the code comments


zengin:
@MichaelMainer, I opened https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/424 to prevent this from happening again.

zengin (approved):

Lauragra:
@davidmu1 , FYI, this affects a number of the Intune topics; can we get it addressed upstream? @MIchaelMainer , what are the consequences of the curly apostrophe, does it affect the snippet generator? I can merge this, but for the Intune files, they'll get reintroduced until we fix at the source because those files are generated.

zengin:
@Lauragra I have unblocked ourselves by fixing the encoding issue in our code.

The reflection of this was that description text such as
```
Defender’s actions
```
was showing up as
```
Defender?s actions
```
in inline documentation of the SDK.

From now on, this is not a big issue on our end as we can at least show one form of apostrophe, but it would be nice if the docs were consistent in terms of punctuation.

Because our issue is resolved and the PR will be overwritten by Intune's auto generation, I am closing it.

PR #11560, opened 2021-02-07
Remove unused suppressions

millicentachieng:


millicentachieng:
The one error in beta docs should go away after API Doctor version update

angelgolfer-ms (approved):
@millicentachieng , well it does look like those suppressed error messages are no longer applicable, as API Doctor doesn't return them anymore!
The few cases that were missing @odata.type in the JSON preceding an example response - I added the error message to the CDK. Hopefully that would further help partners, if they would check out the CDK!

millicentachieng:
Thanks @angelgolfer-ms for adding the information to the CDK. The API Doctor update has been done, this PR can now be merged.

PR #11564, opened 2021-02-07
fixed typos

ganesh-sanap:
fixed typos in article.


Lauragra:
Thanks for catching the typos, @ganesh-sanap! FYI, it's not our style to introduce examples with a colon so I restored the periods.

PR #11590, opened 2021-02-09
Remove ms.author metadata property

FaithOmbongi:


davidmu1 (approved):

Lauragra:
@FaithOmbongi , I'm going to merge this anyway but I don't know why we're getting those warnings about ms.author missing, since most of our topics don't have that. I don't think it's a required attribute (it shouldn't be).

PR #11616, opened 2021-02-10
Update print job update documentation

nilakhan:


tomsato-ms (approved):

api-reference/beta/api/printjob-update-configuration.md
```diff
@@ -15,7 +15,7 @@ Namespace: microsoft.graph
 
 Update the [configuration](../resources/printjobconfiguration.md) property of a [print job](../resources/printjob.md).
 
-Updating a print job configuration will only succeed if there is a [printTask](../resources/printTask.md) in `processing` state on the associated print job, started by a trigger that requesting app created. For details about how to register a task trigger, see [Extending Universal Print to support pull printing](/graph/universal-print-concept-overview#extending-universal-print-to-support-pull-printing).
+Updating a print job configuration will only succeed if a [printTask](../resources/printTask.md) in a `processing` state, started by a trigger that the requesting app created, is associated with the print job. For details about how to register a task trigger, see [Extending Universal Print to support pull printing](/graph/universal-print-concept-overview#extending-universal-print-to-support-pull-printing).
```
Lauragra, line 18: This topic should be removed; we don't create separate method topics for performing an operation on a property of a resource. We should remove this file in favor of the new printjob-update file that you're adding here.

api-reference/beta/api/printjob-update.md
```diff
@@ -0,0 +1,166 @@
...
+```http
+PATCH /print/printers/{id}/jobs/{id}
+```
+## Request headers
+| Name          | Description   |
+|:--------------|:--------------|
+| Authorization | Bearer {token}. Required. |
+
+## Request body
+In the request body, supply the values of the relevant [printJob](../resources/printjob.md) fields. Existing properties that are not included in the request body will maintain their previous values.
```
Lauragra, line 42: You might want to mention here that they can only include the configuration property?

Lauragra (changes requested):
Please see comments. Also, does this update merit a mention in the changelog?

api-reference/beta/api/printjob-update-configuration.md
nilakhan: In general, we don't enable specific property updates on a resource.
If we don't publish this topic, how can we communicate to partners that they can update the "configuration" property as well?

Lauragra: That is covered in the Update printJob topic.

Lauragra:
@nilakhan , can you please revert the update to changelog.md? It has a merge conflict and we don't need that file.

Lauragra (approved):

PR #11624, opened 2021-02-11
Site Collection TermStore and displaying parent site of TermGroup

vishriv:


api-reference/beta/resources/termstore-group.md
```diff
@@ -36,6 +36,7 @@ Inherits from [entity](../resources/entity.md).
 | id                   | string             | Unique identifier of group. Read-Only.
 | displayName          | string             | Name of group.
 | scope                | string              | Returns type of group. Possible values are 'global', 'system' and 'siteCollection'.
+| parentSiteId         | string             | Id of the parent site of this group.
```
Lauragra, line 39: When you add a new property or make any API-level change, you need to include an update to the changelog. For details, see: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103153/Changelog

api-reference/beta/api/termstore-store-get-group.md
```diff
@@ -103,6 +103,61 @@ Content-Type: application/json
   "displayName": "myGroup"  
 }
 ```
+### Request
```
Lauragra, line 106: Is your intent here to add a second example? In this case, when there is more than one example, we add an H3 heading with the title "Example 1: {description}", and so on. Also, new examples should not include the SDK snippet links; those get generated via the snippet generation process, which runs on the master branch weekly.

concepts/changelog.md
```diff
@@ -13,6 +13,13 @@ For a summary of the value of these API changes, as well as recent tools, compon
 
 ## February 2021
 
+### Sites and lists | Taxonomy
```
Lauragra, line 16: The changelog.md file is deprecated. Please add an entry to the Microsoft.SharePoint.json file in the /changelog folder. We no longer update the MD file.

vishriv: Done. And removed chanelog.md file as well

lramosvea:
Added the schema live label since I can see the termStore nav property for Site in the live metadata.

Lauragra (approved):

PR #11660, opened 2021-02-15
Add appRoleAssignments relationship to user & group resources

FaithOmbongi:


psignoret:
@FaithOmbongi Thanks for fixing this! Could you make the same update on the doc for the group entity (it's also missing the appRoleAssignments relationship)?

davidmu1 (approved):

PR #11667, opened 2021-02-16
Clarification on app role assignment relationships

psignoret:


api-reference/beta/resources/serviceprincipal.md
```diff
@@ -4,6 +4,7 @@ description: "Represents an instance of an application in a directory. Inherits
 localization_priority: Priority
 doc_type: resourcePageType
 ms.prod: "applications"
+ms.author: "sureshja"
```
Lauragra, line 7: We don't generally use ms.author in the graph docs; we just use the author field to list the GitHub name. This field for the internal Microsoft alias, which we generally don't want to expose publicly. Is there a reason to add this attribute?

PR #11678, opened 2021-02-17
Update accesspackageresourceenvironment.md

millicentachieng:
We are about to flag empty baseType as an error in new API Doctor version


angelgolfer-ms (approved):
Thanks, @millicentachieng !

Lauragra (approved):

PR #11682, opened 2021-02-17
Clarify supported errorCodes and errorCategories

ArvindHarinder1:


ArvindHarinder1:
No changes to the API are being made here. Just clarifying the definition of some of the properties.

PR #11695, opened 2021-02-18
Changed author and prod for notifications and subscriptions

davidmu1:


api-reference/beta/api/locatedriskevent-get.md
```diff
@@ -3,8 +3,8 @@ title: "Get locatedRiskEvent"
 description: "Retrieve the properties and relationships of a locatedriskevent object."
 localization_priority: Normal
 doc_type: apiPageType
-ms.prod: ""
-author: "davidmu1"
+ms.prod: "microsoft-identity-platform"
```
Lauragra, line 6: @davidmu1 , FYI, the value we use is "identity-and-access". Also, do you need to tag with a subservice value?

PR #11701, opened 2021-02-18
update to v1 and addtokensigningcert

luleonpla:


kenwith:
@davidmu1 - I believe publishing is down right now. There is a new action that was added and the functionality released to v1.0 so those updates are in this PR. If you have time would you be able to provide us any guidance on if this looks good for publish?

PR #11714, opened 2021-02-19
Added PATCH method to onlineMeeting v1.0

frankpeng7:
Added PATCH method to onlineMeeting v1.0 and updated author to PM's alias


frankpeng7:
Meant TOC instead of POC in the commit message

frankpeng7:
@Lauragra @angelgolfer-ms can I get some review on this one?

api-reference/beta/api/onlinemeeting-update.md
```diff
@@ -50,22 +52,40 @@ PATCH /users/{userId}/onlineMeetings/{meetingId}
...
+| participants         | [meetingParticipants](../resources/meetingparticipants.md)   | The participants associated with the online meeting. This includes the organizer and the attendees.                                            |
+| isEntryExitAnnounced | Boolean                                                      | Whether or not to announce when callers join or leave.                                                                                         |
+| lobbyBypassSettings  | [lobbyBypassSettings](../resources/lobbyBypassSettings.md)   | Specifies which participants can bypass the meeting lobby.                                                                                     |
+| allowedPresenters    | onlineMeetingPresenters                                      | Specifies who can be a presenter in a meeting. Possible values are everyone, organization, roleIsPresenter, organizer, and unknownFutureValue. |
+
+**Notes:**
+- The **startDateTime** and **endDateTime** must appear in pairs.
```
angelgolfer-ms, line 69: Line 69:
Do you mean if you update one, you should specify the other as well?
I'm going to assume so in my comment above after line 56. 
Please clarify if that's not what you mean.

frankpeng7: L#69 was moved from original L#53 and I think you got it right.
btw I'm not seeing any comment above after L#56

api-reference/beta/api/onlinemeeting-update.md
```diff
@@ -50,22 +52,40 @@ PATCH /users/{userId}/onlineMeetings/{meetingId}
 | Content-type  | application/json. Required. |
 
 ## Request body
-In the request body, supply a JSON representation of the [onlineMeeting](../resources/onlinemeeting.md) object. Only the **startDateTime**, **endDateTime**, **participants**, and **subject** properties can be modified. The **startDateTime** and **endDateTime** must appear in pairs.
+In the request body, supply the values for relevant properties that need updating.  
+
```
angelgolfer-ms, line 56: Lines 68-71:
I suggest integrating the info in the notes for the corresponding property, under the Description column as a separate paragraph for discoverability.
Alternatively, if you want to call them out outside of the Description column, then list the info in a separate paragraphs after line 55. 
This is because tables tend to visually pull attention away from normal text. When customers scan, they may not notice content right below a table.

Something like this, for the second choice:
The next table lists the properties that can be updated. In the request body, include only the properties that need updating, with the following exceptions:
- Adjusting the start or end date/time of an online meeting always requires both **startDateTime** and **endDateTime** properties in the request body. 
- Adjusting the **attendees** field of the **participants** property, such as adding or removing an attendee to the meeting, always requires the full list of attendees in the request body.

api-reference/beta/api/onlinemeeting-update.md
angelgolfer-ms: Because I was still composing the comment :-) Now you see it.

angelgolfer-ms: To reduce confusion for you, you may wait till I submit review comments for the PR before looking at them? Sometimes I may go back to adjust my comments before submitting the review... so not to waste your time. :-)

api-reference/beta/api/onlinemeeting-update.md
frankpeng7: Thanks for the advice @angelgolfer-ms. I'd prefer calling them out in a separate paragraph and have updated the PR.

api-reference/v1.0/api/onlinemeeting-update.md
```diff
@@ -0,0 +1,241 @@
...
+> - `meetingId` is the **id** of an [onlineMeeting](../resources/onlinemeeting.md) object.
+
+## Request headers
+| Name          | Description                 |
+| :------------ | :-------------------------- |
+| Authorization | Bearer {token}. Required.   |
+| Content-type  | application/json. Required. |
+
+## Request body
+In the request body, supply the values for relevant properties that need updating.
```
angelgolfer-ms: Request body section:
Please make the similar updates as commented in the beta version of this topic.

api-reference/v1.0/api/onlinemeeting-update.md
```diff
@@ -0,0 +1,244 @@
...
+| subject              | String                                                       | The subject of the online meeting.                                                                                                             |
+| participants         | [meetingParticipants](../resources/meetingparticipants.md)   | The participants associated with the online meeting. This includes the organizer and the attendees.                                            |
+| isEntryExitAnnounced | Boolean                                                      | Whether or not to announce when callers join or leave.                                                                                         |
+| lobbyBypassSettings  | [lobbyBypassSettings](../resources/lobbyBypassSettings.md)   | Specifies which participants can bypass the meeting lobby.                                                                                     |
+| allowedPresenters    | onlineMeetingPresenters                                      | Specifies who can be a presenter in a meeting. Possible values are everyone, organization, roleIsPresenter, organizer, and unknownFutureValue. |
+
+**Notes:**
```
angelgolfer-ms, line 69: Lines 69-72:
Please remove these lines since you already mentioned this above the table.

api-reference/beta/api/onlinemeeting-update.md
```diff
@@ -50,22 +52,37 @@ PATCH /users/{userId}/onlineMeetings/{meetingId}
...
-<!-- {
-  "blockType": "request",
-  "name": "patch_onlinemeeting_request"
-}-->
+#### Request
+
+> **Note:** The meeting ID has been truncated for readability.
+
+The following request uses a user token.
+<!-- { "blockType": "ignored" } -->
```
angelgolfer-ms, line 83: Hey @frankpeng7 ,
Is there a specific reason you think it's necessary for the examples in this topic to each show calling with delegated permission and with application permission?
Let me know if there is, and that I am missing it.
From my perspective: many of the Graph APIs support both delegated and application permissions. The Permissions section is supposed to indicate that; the Examples section doesn't call that out. So that's one inconsistency.
But the bigger loss is disabling API Doctor (setting "blockType" to "ignored") for the examples. API Doctor is supposed to work in request/response pairs, so one would have to disable API Doctor when there are 2 example requests mapping to 1 example response.
Perhaps if you feel so inclined, you can choose one example to use delegated permission, and the other to use application permission, and state that in the example section title, line 76 and 159 respectively? Then for each example, you need to just show the request syntax for the kind of permission you state in the example title, keep a 1:1 mapping for the example request and response, and enable API Doctor?

api-reference/beta/api/onlinemeeting-update.md
```diff
@@ -50,22 +52,37 @@ PATCH /users/{userId}/onlineMeetings/{meetingId}
...
 ```http
-PATCH https://graph.microsoft.com/beta/me/onlineMeetings/{id}
+PATCH https://graph.microsoft.com/v1.0/me/onlineMeetings/MSpkYzE3Njc0Yy04MWQ5LTRhZGItYmZi
 Content-Type: application/json
```
angelgolfer-ms, line 86: Line 85:
Please change the endpoint version to beta.

api-reference/beta/api/onlinemeeting-update.md
```diff
@@ -144,6 +156,73 @@ Content-Type: application/json
...
 ```
 
+#### Example 2: Update the lobbyBypassSettings
+> **Note:** The meeting ID has been truncated for readability.
+
+The following request uses a user token.
+```http
+PATCH https://graph.microsoft.com/v1.0/me/onlineMeetings/MSpkYzE3Njc0Yy04MWQ5LTRhZGItYmZi
```
angelgolfer-ms, line 166: Line 165 or 178:
Please change the endpoint version to beta.

api-reference/v1.0/api/onlinemeeting-update.md
angelgolfer-ms, line 69: Lines 69-72:
Please delete these lines as the ideas are in lines 55 and 56.

api-reference/v1.0/api/onlinemeeting-update.md
```diff
@@ -0,0 +1,244 @@
...
+- Provide a full list of attendees to update `participants.attendees`. For example, when adding a new attendee, please provide
+  a collection including all existing attendees and the new one to be added.
+
+
+## Response
+If successful, this method returns a `200 OK` response code and an [onlineMeeting](../resources/onlinemeeting.md) object in the response body.
+
+## Examples
+
+### Example 1: Update the startDateTime, endDateTime and subject
```
angelgolfer-ms, line 80: Please see my comments in the beta topic regarding showing only 1 example request in each example.

changelog/Microsoft.Skype.Calling.json
```diff
@@ -1,5 +1,23 @@
 {
   "changelog": [
+    {
+      "ChangeList": [
+        {
+          "Id": "3a98db4e-7246-4b4c-b97c-feff3f93ce42",
+          "ApiChange": "Method",
+          "ChangedApiName": "onlineMeeting",
+          "ChangeType": "Addition",
+          "Description": "Added PATCH method to the [onlineMeeting](https://docs.microsoft.com/en-us/graph/api/resources/onlinemeeting?view=graph-rest-v1.0) resource.",
```
angelgolfer-ms, line 10: Line 10:
It would be useful to provide a link to the "PATCH method" as well, to point to the PATCH operation topic.

api-reference/beta/api/onlinemeeting-update.md
```diff
@@ -50,22 +52,37 @@ PATCH /users/{userId}/onlineMeetings/{meetingId}
 | Content-type  | application/json. Required. |
 
 ## Request body
-In the request body, supply a JSON representation of the [onlineMeeting](../resources/onlinemeeting.md) object. Only the **startDateTime**, **endDateTime**, **participants**, and **subject** properties can be modified. The **startDateTime** and **endDateTime** must appear in pairs.
+The table below lists the properties that can be updated. In the request body, include only the properties that need updating, with the following exceptions:
+
+- Adjusting the start or end date/time of an online meeting always requires both **startDateTime** and **endDateTime** properties in the request body.
+- Adjusting the **attendees** field of the **participants** property, such as adding or removing an attendee to the meeting, always requires the full list of attendees in the request body.
+
+### Properties that allows updating
```
angelgolfer-ms, line 60: Line 60:
I'd recommend removing this custom section title. Technically, a table in the "Request body" section of a PATCH operation topic typically exists and hence listing a table is not custom.
That's to set and follow the same expectations in the PATCH operation topics for other entities as well. Those topics don't use a H3 custom section title, as the table typically exists in PATCH topics.

api-reference/v1.0/api/onlinemeeting-update.md
```diff
@@ -0,0 +1,244 @@
...
+| Authorization | Bearer {token}. Required.   |
+| Content-type  | application/json. Required. |
+
+## Request body
+The table below lists the properties that can be updated. In the request body, include only the properties that need updating, with the following exceptions:
+
+- Adjusting the start or end date/time of an online meeting always requires both **startDateTime** and **endDateTime** properties in the request body.
+- Adjusting the **attendees** field of the **participants** property, such as adding or removing an attendee to the meeting, always requires the full list of attendees in the request body.
+
+### Properties that allows updating
```
angelgolfer-ms, line 58: Line 58:
I'd recommend removing this custom section title.

api-reference/v1.0/resources/onlinemeeting.md
```diff
@@ -19,8 +19,9 @@ Contains information about a meeting, including the URL used to join a meeting,
...
-| [Delete onlineMeeting](../api/onlinemeeting-delete.md)             | None                              | Delete an online meeting.                                                                                    |
-| [Create or get onlineMeeting](../api/onlinemeeting-createorget.md) | [onlineMeeting](onlinemeeting.md) | Create an online meeting with a custom, external ID. If the meeting already exists, retrieve its properties. |
+| [Update](../api/onlinemeeting-update.md)                           | [onlineMeeting](onlinemeeting.md) | Update the properties of an **onlineMeeting** object. |
+| [Delete onlineMeeting](../api/onlinemeeting-delete.md)             | None                              | Delete an **onlineMeeting** object.                                                                                    |
+| [Create or get onlineMeeting](../api/onlinemeeting-createorget.md) | [onlineMeeting](onlinemeeting.md) | Create an **onlineMeeting** object with a custom, external ID. If the meeting already exists, retrieve its properties. |
```
angelgolfer-ms, line 24: Line 24:
For parity and better experience, would you mind moving the row in line 24 to after line 21? Similar to the beta topic?

angelgolfer-ms (changes requested):
Just a few relatively minor comments inline in the topics.
And one more that's starting to pose as a usability and consistency issue - would you mind reorganizing the Properties table in the onlinemeeting.md topic, and the table in the "Request body" section that shows updateable properties, to use alphabetical order?

If you take a look at the branches generated by the stub topic generator, and existing entities like user, group, event, message, and so on, properties in a properties table should be in alphabetical order.
Thanks!

api-reference/beta/api/onlinemeeting-update.md
frankpeng7: ok removed app token request sample

api-reference/v1.0/resources/onlinemeeting.md
frankpeng7: Actually quite the opposite. I can update beta table order to align with this one. CreateOrGet is an action and should come after the basic CRUD operations - unless they are required to be in alphabetically order. Please let me know if that's the case.

frankpeng7:
> Just a few relatively minor comments inline in the topics.
> And one more that's starting to pose as a usability and consistency issue - would you mind reorganizing the Properties table in the onlinemeeting.md topic, and the table in the "Request body" section that shows updateable properties, to use alphabetical order?
> 
> If you take a look at the branches generated by the stub topic generator, and existing entities like user, group, event, message, and so on, properties in a properties table should be in alphabetical order.
> Thanks!

@angelgolfer-ms done. Please take a look at the only unresolved comment and let me know if any changes are advised. Thanks.

api-reference/v1.0/resources/onlinemeeting.md
angelgolfer-ms: Hi @frankpeng7 , I see what you're saying, and the general guidance is to list the CRUD operations first (most entities have some CRUD), followed by any actions or functions in alphabetical order. When I first saw that ordering in the beta TOC months ago, I was also taken aback... 
IMHO now, to optimize your case for customers looking into a "create" operation in the Methods table, having "create or get" in the vicinity (one row down) from the "create" operation would make the choices ("create" vs "create or get") more discoverable. Especially because most entities do not support a single-call method "create or get", customers who are not aware of this action may not browse through the Methods table to see this method. The similar considerations apply to the TOC ordering, under the "Online meeting" parent node.

There are a few entities that support a PUT operation for "create or update/replace" (yes these are operations, not actions :-)). In those cases, we list them higher up in the CRUD ordering like a "create" operation, so that customers notice them when they're looking into a "create" use case. We wouldn't list such an operation after a PATCH operation, if one exists.

One other thought to share - the general guidance is good for stub topic generation and automation based on the CSDL. There are limits for such generation because the CSDL doesn't contain certain level of info, and generators do not parse the semantics of an action description, so where we can still manually customize and optimize on the customer's discoverability, we can consider doing that. :-)

api-reference/v1.0/toc.yml
```diff
@@ -1214,10 +1214,12 @@ items:
         items:
         - name: Create
```
angelgolfer-ms, line 1215: @frankpeng7 , please order the TOC the way you have it in the Methods table in onlinemeeting.md.

api-reference/beta/toc.yml
```diff
@@ -1422,14 +1422,14 @@ items:
         items:
         - name: Create
```
angelgolfer-ms, line 1423: @frankpeng7 , please order the TOC the way you have it in the Methods table in onlinemeeting.md.

angelgolfer-ms (changes requested):
@frankpeng7 , just 2 more comments in the toc.yml files, and we can wrap up the content review for this PR.
For our records in case anyone looks back at this PR in the future: 
We exchanged our POVs over how to order "create or get", whether to keep that method after "Create" and "Get" as in the current beta version, vs "Create", "Get", "Update", "Delete", "Create or get". The first optimizes on discoverability for customers. The second follows the existing guidance for the majority of entities that support the usual CRUD operations without related methods like "Create or get". 
In a limited set of a total of 5 methods, choosing one or the other may not make a big difference, so I am happy to leave it at that for now. We'll update the order if we get any feedback from internal or external customers in the future. Information architecture and TOC evolve.

angelgolfer-ms (approved):
In light of the OPS publishing tool problem, please continue to follow up with Laura about merging this PR.

PR #11723, opened 2021-02-22
Documentation: Removing reference to user/calendarGroup

kanchm:
Schema does not have `calendarGroup` defined on `user`.

Validated using graph explorer where `me/calendargroup` resolves to open property in `DirectoryServices` instead of `Exchange`.
Affects both `v1.0` and `beta`.


kanchm:
@Lauragra one more PR review to clean up mismatch with schema.

angelgolfer-ms (commented):
@Harini84 , can you please confirm if there isn't a concept of a "default calendar group" in the calendar API?
The syntax lines that Kanchan removed in this PR that refer to "/users/{id}/calendarGroup" were generated possibly around 2015.

@kanchm , thanks for the PR. Once Harini confirms negative, the 50 files related to the calendar API would be good to go.
As for the changes you made to the 5 files related to organizational branding properties and published resource, the changes look right to me based on the CSDL. But just so you know, this was the first time I had seen these few operations.

kanchm:
@angelgolfer-ms we have approval from Ashwin and Harini over email that the change looks good.

angelgolfer-ms (approved):
Thanks to @Harini84 for confirming the non-support for users/id/calendargroup.

PR #11753, opened 2021-02-24
Update examples displayName

BarryShehadeh:


angelgolfer-ms (approved):
The sample app name replacements look fine. Thanks.

PR #11787, opened 2021-02-28
Removing Intune Auditing and Intune Enrollment from V1.0

kanchm:
Workloads do not exist in v1.0.


kanchm:
@Lauragra checking who to include from Intune to look this over. I have messaged some folks internally but they don't show up in the github accounts list.

Lauragra:
@davidmu1 , can you work with the Intune team to make these updates? Thanks.

kanchm:
From: Zhuo Shan
Subject: RE: Intune documentation discrepancies

Hi Kanchan, 

I can confirm that Intune Auditing and Enrollment are not in v1.0. So the docs should be removed to V1.0 document.

Zhuo

PR #11788, opened 2021-02-28
EducationSubmission: Remove Recall and Release actions

kanchm:
Operations do not exist in the /beta schema.


angelgolfer-ms (approved):
Looks like the product team might have changed their minds and replaced "recall" by "unsubmit", and "release" by "return", but forgot to remove the topics for recall and release. The descriptions look very similar to their respective replacement action topics.

Separate from this PR, I noticed that the CSDL specifies a return type of educationSubmission for these actions, but the corresponding topics all say the response doesn't include a body. I'll check offline with you on that.

PR #11790, opened 2021-02-28
Remove user/calendarGroup

kanchm:
User entity does not have calendarGroup property.
These were missed in the original PR. Duplicate examples are being removed here.


angelgolfer-ms (approved):
Thanks for the updates.

PR #11791, opened 2021-02-28
Use user/contactFolders (instead of contactFolder)

kanchm:
Fixed URLs in samples.


angelgolfer-ms (approved):
Looks good to me, thanks!

PR #11793, opened 2021-02-28
Dynamics: Remove trialBalance

kanchm:
Beta schema does not have this entity.
Fixing URL in create-customerpayment request.


angelgolfer-ms (approved):
Updates look good to me.

PR #11794, opened 2021-03-01
Call: Remove updateMetadata

kanchm:
Call entity does not have updateMetadata in /beta.


kanchm:
@Lauragra @angelgolfer-ms should I have removed:
- [ ] Snippets
- [ ] Change log entries

concepts/changelog.md
```diff
@@ -3231,7 +3231,7 @@ Debut of the places API which provides rich details for locations in apps.
...
-| Addition        | Beta        | Added APIs for applications to manage and participate in calls and meetings, including the ability to [share content](/graph/api/call-changescreensharingrole?view=graph-rest-beta&preserve-view=true), [mute and unmute itself](/graph/api/call-unmute?view=graph-rest-beta&preserve-view=true), and [update the metadata associated with a call](/graph/api/call-updatemetadata?view=graph-rest-beta&preserve-view=true). |
```
angelgolfer-ms, line 3234: @kanchm , is there a way you or your team can verify if they actually ever released an updateMetadata method?
If they never did, then yes we should remove such record in the changelog, including Microsoft.Skype.Calling.json.

On the other hand, if they had released the method but subsequently removed it from the beta endpoint, then we can leave the changelog.md and changelog-archive.md alone. The proper thing to do is to contact Josey Sandoval and have someone on their team add back a record for the method removal for the month that happened :-|

angelgolfer-ms (commented):
@kanchm , I have one comment about updating the changelog files. I provided the corresponding contact person's name you can follow up with please.

Other updates look good in this PR. Thanks!

concepts/changelog.md
kanchm: Zheng will confirm soon. This was not made public. So the changes here should be good to go.

zhengni-ms: Yes it is removed in Aug 2019. @yizhenww, please help adding removal entry in Aug 2019 instead of remove the original addition entry for UpdateMetadata.

yizhenww: @angelgolfer-ms i reverted the changelog, and added an new entry with today's date to the json changelog for calling, as advised by Laura. Let me know if it looks ok. thanks

PR #11797, opened 2021-03-01
Outlook task group sample URLs

kanchm:
Navigation is through me/outlook/task* instead of me/task*.


angelgolfer-ms (approved):
The updates look good, thanks.
The topics themselves have other consistency issues but we wouldn't bother as this API set is deprecated.

PR #11822, opened 2021-03-02
Update userteamwork-sendactivitynotification.md

JarbasHorst:
Fixed broken example
Fixes: #11817


JarbasHorst:
PR onhold because of MSFT feedback

PR #11841, opened 2021-03-03
Excel: Use DELETE verb instead of POST on /delete

kanchm:
Only exception is for Range. POST /delete is preserved in that.


angelgolfer-ms (approved):
Thanks for the fixes. They look good to me; there's currently only one delete action which is for workbookRange, and so I assume all the supported delete methods should be operations. (Note that file name is old: range-delete.md, created before they renamed all the Excel entities preceding them with "workbook".)

PR #11843, opened 2021-03-04
Drive: Navigate through items collection

kanchm:
Fix navigation through items. Drive is not a collection.
Replacing "drive/{item-id}" with "drive/items/{item-id}".


angelgolfer-ms (approved):
Thanks for the updates. They look good to me.

PR #11844, opened 2021-03-04
Excel: Use valid path to workbook

kanchm:
Provide a proper path to workbook.
Replaces /workbook with /me/drive/root/workbook.


api-reference/beta/api/chart-delete.md
```diff
@@ -26,7 +26,7 @@ One of the following permissions is required to call this API. To learn more, in
 ## HTTP request
 <!-- { "blockType": "ignored" } -->
 ```http
-POST /workbook/worksheets/{id|name}/charts/{name}/delete
+POST /me/drive/root/workbook/worksheets/{id|name}/charts/{name}/delete
```
angelgolfer-ms, line 29: GLOBAL:
Since this is the "HTTP request" section, the supported syntax should be complete. According to the first section of the Excel API [overview](https://docs.microsoft.com/en-us/graph/api/resources/excel?view=graph-rest-beta), there are two main formats to identify a workbook:
https://graph.microsoft.com/{version}/me/drive/items/{id}/workbook/
https://graph.microsoft.com/{version}/me/drive/root:/{item-path}:/workbook/

Seems like we should include the first format as well... I'll let @grangeryy confirm.

grangeryy (approved):

joezhao-msft (approved):

PR #11856, opened 2021-03-05
Skip attempting to create PRs for subsequent commits to the snippet branch

### LINKED ISSUE #11178: Github Action for snippet PRs should succeed for subsequent pushes to the same branch
Currently Github Action for snippet generation is triggered for every push and it tries to create a new PR using Github API. If there is already an open PR, that causes Action to fail with the following message:

> A pull request already exists for microsoftgraph:refs/heads/snippet-generation/35531.

e.g. https://github.com/microsoftgraph/microsoft-graph-docs/pull/11107/checks?check_run_id=1646394667

Then the failure shows up as a merge blocker, which is a false signal. We should be resilient for this scenario and we should silently ignore PR creation request if there is already a PR.

zengin:
Fixes #11178

- Added a check on number of commits that the snippets branch is ahead of master. If the number of commits is 1, we generate the PR, otherwise we skip.

Tested the changes in my own fork.

### DIFF
--- a/.github/workflows/snippet-generation-pr.yml
+++ b/.github/workflows/snippet-generation-pr.yml
@@ -19,6 +19,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 10 # number of commits to fetch so that we can query history
 
     # Create a pull request [0]
     - name: Create PR using the GitHub REST API via hub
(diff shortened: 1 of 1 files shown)


baywet (approved):
If this works as expected, can you generalize it to all the generation repos please ? (service libs)

PR #11885, opened 2021-03-07
Update get-a-count-endswith-java-snippets.md

gerardv:
In order to use advanced queries, you need to include the $count=true query parameter. See https://developer.microsoft.com/en-us/identity/blogs/microsoft-graph-advanced-queries-for-directory-objects-are-now-generally-available/ for more information.


andrueastman:
Hi @gerardv.
Thanks for your feedback.
This file is however autogenerated and merging this PR would result to it being eventually overwritten by the snippet generator. 
I have created an issue with the snippet generator to track this so that it can be fixed. See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/497

PR #11926, opened 2021-03-10
Fix inconsistencies in content type api documentation

swapnil1993:


api-reference/beta/api/contenttype-addcopy.md
```diff
@@ -80,11 +80,30 @@ Content-Type: application/json
...
+        "group": "Display Template Content Types",
+        "hidden": false,
+        "name": "JavaScript Display Template",
+        "readOnly": false,
+        "sealed": false
+    }
+}
 ```
 
 [site]: ../resources/site.md
```
angelgolfer-ms, line 109: This comment is actually for the file displayed next - the POST operation topic to create a contentType resource.
According to the current CSDL on the beta endpoint, site and list each has a navigation property called columns, that's a collection of contentType resources. Is it true that you can create a contentType resource for a site, and similarly for a list?

Assuming they're both true, then you would have 2 files that create contentType, with the filenames site-post-contenttypes.md and list-post-contenttypes.md. These filenames follow the pattern parententityname-post-navigationpropertyname.md, as described in the table row for "API (create)" in this topic: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103161/File-naming. For each of the files, you can use the topic title "Create contentType for site", "Create contentType for list".

Each of these topics should appear in the Methods table of the corresponding parent type (i.e., site resource topic and list resource topic). And both topics should also appear in the Methods table of the contentType resource topic.

angelgolfer-ms: Technically, the similar filenaming pattern would apply for the "list" operation too:  site-list-contenttypes.md and list-list-contenttypes.md. These filenames follow the pattern parententityname-list-navigationpropertyname.md. I see that currently, the topic contentype-list.md defines the syntax for both the site and list cases.

angelgolfer-ms: Perhaps you can confirm the supported syntax for the "create" and "list" operations, and I'll then confirm the actual files we should have.
And then I can confirm where the operations (especially the "create" and "list" operation topics) should be exposed in the TOC, for contentType and for columnDefinition.

api-reference/beta/api/contenttype-addcopy.md
```diff
@@ -80,11 +80,30 @@ Content-Type: application/json
 ### Response
 
 
-<!-- { "blockType": "response" } -->
```
angelgolfer-ms, line 83: Please also update line 59, the "Response" section. It should have the correct response code and response type.

api-reference/beta/resources/contentType.md
```diff
@@ -15,6 +15,22 @@ Namespace: microsoft.graph
...
 
+## Methods
+|Method|Return type|Description|
+|:---|:---|:---|
+|[List contentTypes](../api/contenttype-list.md)|[contentType](../resources/contenttype.md) collection|Get a list of the [contentType](../resources/contenttype.md) objects and their properties.|
+|[Create contentType](../api/contenttype-create.md)|[contentType](../resources/contenttype.md)|Create a new [contentType](../resources/contenttype.md) object.|
+|[Get contentType](../api/contenttype-get.md)|[contentType](../resources/contenttype.md)|Read the properties and relationships of a [contentType](../resources/contenttype.md) object.|
+|[Update contentType](../api/contenttype-update.md)|[contentType](../resources/contenttype.md)|Update the properties of a [contentType](../resources/contenttype.md) object.|
+|[Delete contentType](../api/contenttype-delete.md)|None|Deletes a [contentType](../resources/contenttype.md) object.|
+|[isPublished](../api/contenttype-ispublished.md)|Boolean| Checks if the [contentType](../resources/contenttype.md) is published.|
```
angelgolfer-ms, line 26: After line 25, you can insert two table rows, one for  contenttype-list-columns.md  and the other for contentype-post-columns.md.
The current stub topic generator generates Method tables for a parent resource topic, that would include the "list" and "create" topics for each nav property, if the "list" and "create" operations are supported for that nav property.

angelgolfer-ms (changes requested):
Thanks for the quick response, @swapnil1993 .
Please take a look at my comments, and confirm what's supported (site and list?) for the "create" and "list" operations for contentType.

api-reference/beta/api/contenttype-addcopy.md
swapnil1993: Yes both content types and columns can be created for sites and list. An existing site content type should be copied to the list contentTypes collection to create a content types in a list that is why it is modelled as "addCopy" operation instead of list-post-contenttypes. I will add a file for creating a site content type named "site-post-contenttypes" and also "site-list-contenttypes.md" and "list-list-contenttypes.md" to list content types in both sites and list.
For columns both create an list operations are supported  in both site and list so there also I will need to add the files for each of them.

api-reference/beta/api/site-list-contenttypes.md
```diff
@@ -1,17 +1,17 @@
...
 ms.prod: "sites-and-lists"
 ---
 
-# List contentTypes
+# List contentTypes in a site
 Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
-Get the collection of [content types][contentType] in a [site][] or a [list][]
+Get the collection of [content types][contentType] in a [site][].
```
angelgolfer-ms, line 14: Line 14 and in general:
FYI when we provide a link to a type name (in this case contentType), the link text should match the name of the type. I am changing the link text from "content types" (natural language) to "contentType", and rephrasing the sentence as "Get the collection of contentType resources in a site."

api-reference/beta/resources/contentType.md
```diff
@@ -15,6 +15,25 @@ Namespace: microsoft.graph
 Represents a _content type_ in SharePoint.
 Content types allow you to define a set of columns that must be present on every [**listItem**][listItem] in a [**list**][list].
 
+## Methods
+|Method|Return type|Description|
+|:---|:---|:---|
+|[List contentTypes in a site](../api/site-list-contenttypes.md)|[contentType](../resources/contenttype.md) collection|Get a list of the [contentType](../resources/contenttype.md) objects and their properties in a [site](../resources/site.md).|
+|[List contentTypes in a list](../api/list-list-contenttypes.md)|[contentType](../resources/contenttype.md) collection|Get a list of the [contentType](../resources/contenttype.md) objects and their properties in a [list](../resources/list.md).|
+|[Create a site contentType](../api/site-post-contenttypes.md)|[contentType](../resources/contenttype.md)|Create a new [contentType](../resources/contenttype.md) object in a [site](../resources/site.md).|
```
angelgolfer-ms, line 23: Line 23:
FYI I am changing the first column link text as "Create contentType for a site". "create a site contentType" with 2 nouns one after another may be ambiguous for localization. In general, where possible, we try to avoid having multiple nouns together.

api-reference/beta/api/site-list-columns.md
```diff
@@ -0,0 +1,129 @@
+---
+author: swapnil1993
+title: "List columnDefinitions in a site"
+description: "List columns in a site."
+localization_priority: Normal
+doc_type: apiPageType
+ms.prod: "sites-and-lists"
+---
+
+# List columnDefinitions in a site
```
angelgolfer-ms, line 10: Line 10 and general:
Per the [TOC and topic title](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103176/ToC-and-topic-title) topic in the CDK, the topic title for "list" operation topics is "List {navigation property name}". 
So using this topics as an example, the first part of the topic title should be "List columns..." since columns is the nav property name.
Since we have a few similar topics for listing columns, we differentiate the topic titles by appending "in a site", "in a list", "in a content type". So here, we have "List columns in a site" as topic title.

api-reference/beta/api/contenttype-post-columns.md
```diff
@@ -22,9 +22,9 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type | Permissions (from least to most privileged) |
 |:--------------------|:---------------------------------------------------------|
```
angelgolfer-ms, line 24: Line 15:
I know the description is pre-existing before this PR, but I have a question - should the description be rephrased as "Add a column, represented by a columnDefinition, to a contentType in a site or list"?

angelgolfer-ms: As noted in another comment in this PR, in general, we don't want a few nouns one after another, as this would pose some degree of ambiguity for localization. If we just take the nouns "site or list column", we should say a column in a site or list. But then you can see the problem - it's actually a (existing) content type in a site or list that's getting a new column. Is that correct?

angelgolfer-ms: So I am going ahead to make that change in line 15, but for this one, please make sure to let me know if I interpret it incorrectly.

api-reference/beta/resources/columndefinition.md
```diff
@@ -13,6 +13,19 @@ Namespace: microsoft.graph
...
+|[Create content type columnDefinition](../api/contenttype-post-columns.md)|[columnDefinition](../resources/columndefinition.md)|Create a new [columnDefinition](../resources/columndefinition.md) object in a [content type](../resources/contenttype.md).|
+|[Create site columnDefinition](../api/site-post-columns.md)|[columnDefinition](../resources/columndefinition.md)|Create a new [columnDefinition](../resources/columndefinition.md) object in a [site](../resources/site.md).|
+|[Create list columnDefinition](../api/list-post-columns.md)|[columnDefinition](../resources/columndefinition.md)|Create a new [columnDefinition](../resources/columndefinition.md) object in a [list](../resources/list.md).|
+|[Get columnDefinition](../api/columndefinition-get.md)|[columnDefinition](../resources/columndefinition.md)|Read the properties and relationships of a [columnDefinition](../resources/columndefinition.md) object.|
+|[Update columnDefinition](../api/columndefinition-update.md)|[columnDefinition](../resources/columndefinition.md)|Update the properties of a [columnDefinition](../resources/columndefinition.md) object.|
```
angelgolfer-ms, line 26: My next question is actually for the topic "Update columnDefinition" (columndefinition-update.md):
Line 57 of that topic currently says:
">**Note:** Only required and hidden properties can be updated."

Do we mean only the required property and the hidden property of columnDefinition can be updated?

If so, we must clarify this by bolding "required" and "hidden" in that sentence in line 57. Bolding identifies those terms as actual property names.
Otherwise, I was expecting that to mean a property in the columnDefinition resource being characterized as required, or as hidden.

Please clarify.

swapnil1993: Yes this means that only the required property and the hidden property of columnDefinition can be updated but is applicable to only content type columns. This file was earlier only for content type columns and I missed updating it while adding changes for site and list columns.

angelgolfer-ms: So does it mean for columns in a site or list, any property of columnDefinition, other than the id property, can be updated?

angelgolfer-ms (approved):
@swapnil1993 , FYI I ordered the nodes under "Sites and lists" by first listing "Sites", then "Lists", and then the remaining entities in alphabetical order. This would be consistent with other existing nodes like "Users" and "Groups", under which the main entity comes first, followed by the rest in alphabetical order. I foresee we'll be adding a few more that currently are documented but not in the TOC, so this ordering would work then too.

angelgolfer-ms:
> @swapnil1993 , FYI I ordered the nodes under "Sites and lists" by first listing "Sites", then "Lists", and then the remaining entities in alphabetical order. This would be consistent with other existing nodes like "Users" and "Groups", under which the main entity comes first, followed by the rest in alphabetical order. I foresee we'll be adding a few more that currently are documented but not in the TOC, so this ordering would work then too.

BTW, someone is following up on the TOC warnings for the few Intune topics, so we can ignore those warnings for now.

PR #11959, opened 2021-03-12
Snippets Update

zengin:
The large number of changed files is due to the following fixes in the snippet generation process:
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/493
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/488
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/483
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/478

I did sanity check on the diff locally and haven't observed anything unexpected.

I ran Raptor C# compilation tests, no regressions observed.

PR #11971, opened 2021-03-12
Snippets Update

zengin:
Replaces #11959


zengin:
#11959 was not running apidoctor tests because of the `[skip ci]` in the commit message. I cherry picked the same change to replace that PR and apidoctor is passing as I don't have `[skip ci]` in the commit message now.

PR #11975, opened 2021-03-13
Deduplicate block names

zengin:
## Duplicate block names were causing two issues
1. We generate duplicates of the same snippet in different locations (mismatched snippet output for snippet tabs, e.g. HTTP tab is showing different snippet than C# snippet)
2. If the execution order changes, we swap the snippet output with the other duplicate entry which makes snippet output indeterministic.

## Solution
- Found all the duplicates and appended a counter at the end to make them unique (within each version). Doc writers should feel free to change them into more meaningful names. The only requirement is for them to be unique within the set of each version (v1 vs beta).
- We consider introducing an APIDoctor rule to prevent duplication in future doc writing.

Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/463

## Unblocking 1148 snippets that we were not able to generate before

  | Total Duplicate Entries | Unique Names for Duplicates | Deduped blocks
-- | -- | -- | --
Beta | 325 | 132 | 193
V1 | 158 | 64 | 94
Total | 483 | 196 | 287

**287 (deduped blocks) x 4 (number of languages) = 1148** 

## Regression testing

Ran Raptor C# compilation tests. No new compilation issues in V1 were observed. Beta has 7 new failures, but we have ~70 failures due to known issues (out of total ~2000 tests) and this regression is expected with 193 new test cases.


zengin:
Changing to draft, I will look at the apidoctor failures early next week.

api-reference/beta/api/accessreview-listdecisions.md
```diff
@@ -51,25 +51,25 @@ If successful, this method returns a `200, OK` response code and an array of [ac
 # [HTTP](#tab/http)
 <!-- {
   "blockType": "request",
-  "name": "get_accessReview_decisions"
+  "name": "get_accessReview_decisions_1"
```
MIchaelMainer, line 54: [scope to file]: Should we see a `get_accessReview_decisions_2` if there is `get_accessReview_decisions_1`?

MIchaelMainer: Nevermind, I see the dedup happens across files.

MIchaelMainer (approved):
`sdvdiff main zengin/deduplication-pr`

I observed wide spread dedup. I also observed the addition of missing snippet links due to the dedup of block names (educationuser-post-rubrics.md). I also notice the addition of the msgraph-interactive code block type added. By the time I was at the m*.md files, I started only reviewing change patterns in the change map if they differed from other patterns.

I didn't review the added snippets since those are an artifact of this dedup.

api-reference/beta/api/accessreview-listdecisions.md
zengin: Yes, the scope is all files within a version (v1 or beta)

zengin:
@Lauragra, this is the PR with the changes I proposed last week for deduplication of block names. As long as docs v-team is OK with the change, it should be good to merge.

Please let us know if we can also enable a blocking APIDoctor error to prevent introducing new instances of duplicates. Since the time I forked the branch for this PR (end of last week), there were 4 new instances introduced :)

PR #12004, opened 2021-03-16
Update site-getapplicablecontenttypesforlist.md

JarbasHorst:
I would suggest us displaying the title as we also do for other articles.


api-reference/beta/api/site-getapplicablecontenttypesforlist.md
```diff
@@ -7,7 +7,7 @@ doc_type: apiPageType
 ms.prod: "sites-and-lists"
 ---
 
-# site: getApplicableContentTypesForList
+# Get applicable content types for list
```
Lauragra, line 10: The H1 should match the title, and the existing title is consistent with our guidelines for function/action titles, so I'm going to close this PR. Thanks.

PR #12021, opened 2021-03-18
Typo in Corporate management

GeertVL:


GeertVL:
I will fix this, did not know that it is being used as a reference.

PR #12060, opened 2021-03-22
02 02 2021 nilakhan doc bugs

nilakhan:
Addressing few bugs in "beta" documentation.


jahsu-MSFT:
We probably also need to update the change log

jahsu-MSFT:
Do we also need to update V1.0 to reflect these changes?

PR #12081, opened 2021-03-24
Add a note for updating printer using IPP, and fix permissions in LIST printer shares API

nilakhan:
Add a note for updating printer using IPP, and fix permissions in LIST printer shares API


jahsu-MSFT (approved):

api-reference/beta/api/printdocument-createuploadsession.md
```diff
@@ -82,6 +82,7 @@ The following example shows how to create an upload session that you can use in
 
 <!-- {
   "blockType": "request",
+  "name": "printdocument_createuploadsession"
```
nilakhan, line 85: @Lauragra The code snippets for this request did not get auto-generated. 
Is it because of missing name or we are we missing something else as well?

concepts/upload-data-to-upload-session.md
```diff
@@ -39,6 +39,8 @@ Content-Range: bytes=0-72796/4533322
 Content-Length: 72797
 
 <bytes 0-72796 of the file>
+
```
nilakhan, line 42: @Lauragra we also want to add code samples for this API as well. 
Since, this is a non-graph endpoint, can we create our own samples and add here?

api-reference/beta/api/printdocument-createuploadsession.md
Lauragra: Let me check with @andrueastman .

concepts/upload-data-to-upload-session.md
```diff
@@ -39,6 +39,8 @@ Content-Range: bytes=0-72796/4533322
...
+Here, 0 and 72796 are the start and end indexes of the file segment and 4533322 is the size of document
```
Lauragra, line 43: Did you mean for this to be part of the code example, or move it outside of the ```s? Also make sure the sentence has an ending period.

concepts/upload-data-to-upload-session.md
Lauragra: I think that you can. You want to add tabs for different code languages to the example? Let's also check with @andrueastman .

concepts/upload-data-to-upload-session.md
nilakhan: Fixed. thanks!

esaurabhb-ms (commented):
@nilakhan - should we add a comment in the document that for printers to work with a particular platform, it should meet the requirements of that platform. For e.g. On Windows, it is expected that printer specifies all attributes that are considered mandatory as per MOPRIA specs. Please note MOPRIA specs are available to only the paid members of MOPRIA.

For the time being we can add link to MOPRIA alliance portal. In future when we document our own set of attributes then we can point there and from there we can point to Mopria portal.

Lauragra:
@andrueastman , can you take a look at the comments in this PR please?

concepts/upload-data-to-upload-session.md
andrueastman: @nilakhan @Lauragra You should be able to add your own samples since this is pointing to non-graph endpoint. The snippet generator won't try to generate samples if the endpoint in not a graph endpoint.

Also, since this example does not have a request block annotation, the snippet generator will not pick it up and try to generate snippets off it. 
By request block annotation I mean something like this 
```
<!-- {
  "blockType": "request",
  "name": "example-request"
}-->
```

api-reference/beta/api/printdocument-createuploadsession.md
andrueastman: The missing name is definitely an issue. Once we merge this in it should be generated in the next generation cycle. 
If not please feel free to raise an issue and assign it to me to investigate further.

PR #12085, opened 2021-03-24
Add new fields to pstnCallLogRow

SanderSade:
Adding operator and callDurationSource to callrecords-pstncalllogrow, new pstnCallDurationSource enum to enums


api-reference/beta/api/callrecords-callrecord-getpstncalls.md
```diff
@@ -112,7 +112,9 @@ HTTP/1.1 200 OK
             "destinationName": "United States",
             "conferenceId": null,
             "licenseCapability": "MCOPSTNU",
-            "inventoryType": "Subscriber"
+            "inventoryType": "Subscriber",
+			"operator": "Microsoft",
```
stephenjust, line 116: nit: indentation

api-reference/beta/resources/enums.md
```diff
@@ -2148,3 +2148,9 @@ Possible values for user account types (group membership), per Windows definitio
 |other|
 |unknownFutureValue|
 
+### pstnCallDurationSource values
```
stephenjust, line 2151: This should go in https://github.com/microsoftgraph/microsoft-graph-docs/blob/main/api-reference/beta/resources/enums-callrecords.md

stephenjust (changes requested):

api-reference/beta/api/callrecords-callrecord-getpstncalls.md
SanderSade: Fixed, invisible tabs from JSON copy/paste

api-reference/beta/resources/enums.md
SanderSade: Moved to enum-callrecords

williamlooney (approved):

api-reference/beta/resources/callrecords-pstncalllogrow.md
```diff
@@ -40,6 +40,8 @@ Represents a row of data in the Public Switch Telephone Network (PSTN) call log.
 |conferenceId|String|ID of the audio conference.|
 |licenseCapability|String|The license used for the call.|
 |inventoryType|String|User's phone number type, such as a service of toll-free number.|
+|operator|String|The telecommunications operator which provided PSTN services for this call. This may be Microsoft, or it may be a third-party operator via the [Operator Connect Program](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/introducing-operator-connect-and-more-teams-calling-updates/ba-p/2176398)|
+|callDurationSource|pstnCallDurationSource|The call duration source. When a third-party telecommunications operator is used via the Operator Connect Program, they may provide their own call duration data. In that case, `operator` will be specified. Otherwise `microsoft` will be specified.|
```
angelgolfer-ms, line 44: Line 43:
Please end the description with a period.

angelgolfer-ms, line 44: Line 44:
Please rephrase the first sentence as "The source of the call duration data."
The reason is three or more nouns in a row can become ambiguous and a challenge for localizers.

In general, we avoid passive voice as much as possible. Here, "will be specified" begs the question "by whom".

I suggest rephrasing the other sentences to use active voice:  "If the call uses a third-party telecommunications operator via the Operator Connect Program, the operator may provide their own call duration data. In this case, the property value is `operator`. Otherwise, the value is `microsoft`."

changelog/Microsoft.IC3.DataPlatform.json
```diff
@@ -0,0 +1,34 @@
...
+                    "ChangedApiName": "callDurationSource",
+                    "ChangeType": "Addition",
+                    "Description": "Added the **callDurationSource** property to the [pstnCallLogRow](https://docs.microsoft.com/en-us/graph/api/resources/pstnCallLogRow?view=graph-rest-beta) resource.",
+                    "Target": "pstnCallLogRow"
+                }, {
+                    "Id": "b2951789-8c86-11eb-a57c-31dcdef91872",
+                    "ApiChange": "Enum type",
+                    "ChangedApiName": "pstnCallDurationSource",
+                    "ChangeType": "Addition",
+                    "Description": "Added the [**pstnCallDurationSource**](https://docs.microsoft.com/en-us/graph/api/resources/enums?view=graph-rest-beta#pstncalldurationsource-values) enumeration type, defining the call duration source. When a third-party telecommunications operator is used via the Operator Connect Program, they may provide their own call duration data.",
```
angelgolfer-ms, line 22: Line 22:
FYI in general, if we provide a link to a term (here the term is "pstnCallDurationSource"), then we don't have to bold that term. The link would be displayed the way the publishing site displays links. :-)

api-reference/beta/resources/callrecords-pstncalllogrow.md
angelgolfer-ms, line 44: @SanderSade , thanks for adding the properties. The next comment applies to not only what this PR added, but in general to all resource type topics, including this one.
Can you please reorder the table rows in the Properties section of this topic by alphabetical order?
Reasons:
- Customers would have to scan each row in order to find the property they need. 
- This is not a new guideline. It's one that the generators have always used. Apparently over time, authors and content reviewers haven't actively enforced this guidance in this topic.  Imagine your topic growing to a third of the group resource topic without following this :-)

angelgolfer-ms (changes requested):
Thanks for the PR, @SanderSade ! That's a technically solid PR! My comments are for readability and editorial, please take a look.

api-reference/beta/resources/callrecords-pstncalllogrow.md
SanderSade: Both recommendations applied

SanderSade: Added

SanderSade: Thank you, @angelgolfer-ms!
 
I've sorted the table, but left the id on top - as an unique key, I think it should be always first, so the customer wouldn't have to look for the key when onboarding or reviewing. This is also a pretty much universal behavior for describing SQL tables or any other tabular data - if that is not OK, please let me know.

Also, should the response JSON be sorted to match the table?

changelog/Microsoft.IC3.DataPlatform.json
SanderSade: Un-bolded the term

api-reference/beta/resources/callrecords-pstncalllogrow.md
angelgolfer-ms: Thanks for reordering the Properties table rows, @SanderSade .
For consistency with other resource topics and the future output of the topic generation tool, please place the row for id also in alphabetical order. 

Regarding the "response JSON" - I suppose you're referring to an example response payload? Not the "JSON representation" section, right? 
Well, for the former, I think it's ok to leave the ordering of properties in a response payload in whatever order that a customer would get, say in Graph Explorer. 
For the latter, alphabetical order is desired but not required, because the JSON representation section is technically for API Doctor, customers don't derive a whole lot of value from that section. On the other hand, desired because it would be easier for internal authors like yourself to browse and make sure you guys haven't left out any published properties in that section :-)

angelgolfer-ms (approved):
Thanks, @SanderSade , for incorporating the feedback. I replied to your questions, please take a look. 
I also fixed an issue in the enums.md file that probably was inherited from somebody else's earlier update that generated a build "suggestion". Typically, build suggestions will evolve into build warnings later on, so we proactively fix them.
Since there is only one remaining non-critical issue about ordering the id property (per your question/comment), for efficiency, I am approving this PR to reduce the back-and-forth.
After you incorporate the remaining issue with ordering the id property, please follow this [protocol](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103181/Content-workflow) to get the topic published.

api-reference/beta/resources/callrecords-pstncalllogrow.md
SanderSade: I moved the id row... but really, this is something that most technical users would prefer as the first row, and possibly bolded or with the different background color.

PR #12087, opened 2021-03-24
Adding 2 values to the TAGS property

msharsur:
Adding 2 values to the TAGS property.
'HideApp' value is already documented in the following article - https://docs.microsoft.com/en-us/azure/active-directory/manage-apps/hide-application-from-user-portal which users Graph API in the backend.


jasonjoh:
Please disregard the earlier message from msftbot. Your pull request was incorrectly flagged.

PR #12091, opened 2021-03-24
add links to MGT learn module

elisenyang:


concepts/toc.yml
```diff
@@ -56,6 +56,8 @@ items:
             href: /graph/tutorials/java
           - name: JavaScript
             href: /graph/tutorials/javascript
+          - name: Microsoft Graph Toolkit
```
Lauragra, line 59: A Learn module isn't the same type of tutorial as the other ones that we have here. Are we okay with adding this? Also, if so, we should update the overview topic for the Tutorials as well.

concepts/tutorials.yml
```diff
@@ -75,6 +77,8 @@ landingContent:
             url: /graph/tutorials/spfx
           - text: Teams
             url: /graph/tutorials/teams
+          - text: Microsoft Graph Toolkit
```
Lauragra, line 80: Do we think of the Toolkit as a "Microsoft platform"?

concepts/toc.yml
elisenyang: @Lauragra Is there a place you would recommend adding it instead? Is there a section just for learn content?

jasonjoh: No this should not show up in the TOC. We have added links to Graph learning modules on the Tutorials overview page itself, but not as part of our TOC.

concepts/tutorials.yml
jasonjoh: I suggested it could go into that card. Calling it a platform might be a stretch, but the only other place it would make sense is under single-page apps, but the module doesn't really build a full app, just a very basic static HTML page. Maybe we just add another card for "Microsoft Graph Learning paths" or something similar? Or move all Learn modules off this page altogether and maybe add another landing page under Explore dedicated to Learn modules?

Lauragra: For now I think adding another card where we can link to the Learn content would work okay, as long as we're okay with having a mix of Learning paths and modules. We can list the two Microsoft Graph learning paths, and this Toolkit one that's a module.

jasonjoh: In that case repurpose the "Scenarios" card. It's just Learn modules :D

elisenyang: @Lauragra @jasonjoh is there any action needed from me to get this merged?

concepts/toc.yml
Lauragra: @elisenyang , please remove the TOC entry.

concepts/tutorials.yml
Lauragra: @elisenyang , please move the link to the Toolkit Learn module from the platforms to the Scenarios card.

elisenyang: @Lauragra with the changes that Jason has made, are we good to go?

Lauragra: Sure, I went ahead and set Ready to merge so it will be in today's publishing queue.

Issue #913, opened 2017-02-27
Beta nameditem_add not working as expected


ValerioSevilla:
Issue:

nameditem_add method in beta endpoint not working as expected

Article:

Hi,

I have been trying to add a named item using [nameditem_add method in beta](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/nameditem_add.md). The most I can get now is this error:

```javascript
{
  "error": {
    "code": "ItemNotFound",
    "message": "The requested resource doesn't exist.",
    "innerError": {
      ...
    }
  }
}
```

This is the request URL I have:

`https://graph.microsoft.com/beta/me/drive/items/<itemId>/workbook/names/add`

(replacing `<itemId>` with an actual item Id).

With this request body:

```javascript
{
	"name": "APITestName1",
	"reference": "=B3+A3",
	"comment": "Awesome named range"
}
```

I'm opening this issue here as I think it may be a documentation issue rather than a code issue. It looks like this method is not being exposed at the moment in the Beta API.

ValerioSevilla:
Update:

After multiple tests, we have found out that:

- There is no explicit way of setting the type of the named item through this method.
- In general, it is created with type String.
- The only way to create a correct named item with type Range is by setting a reference like this: **=Sheet1!$A$1:$B$2** (including the sheet name and the ‘$’ characters).
- A Range item can also be created with a reference that does not include the ‘$’ character: **=Sheet!A1:B2**. But when including the named range in a formula from other cell, the formula is not correctly calculated. For example, if all cells in the range contain numbers, and we create this formula in other cell: **=SUM(<namedItem>)**, it always returns 0. However, if the named range includes the ‘$’ signs, it works fine.

- There is a Delete Named Item method in Beta version that is not documented at all:
DELETE https://graph.microsoft.com/beta/me/drive/items/<id>/workbook/names(<name>)

isvargasmsft:
This one is fixed. My deepest apologies for the delay.

PR #12173, opened 2021-03-31
New doc for namedItem delete operation in beta

isvargasmsft:
This is the first draft

### DIFF
--- a/api-reference/beta/api/nameditem-add.md
+++ b/api-reference/beta/api/nameditem-add.md
@@ -1,6 +1,6 @@
 ---
 title: "workbookNamedItem: add"
-description: "Adds a new namedItem to the collection of the given scope"
+description: "Adds a new namedItem to the collection of a given scope"
 localization_priority: Normal
 doc_type: apiPageType
 ms.prod: "excel"
@@ -13,7 +13,7 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Adds a new namedItem to the collection of the given scope.
+Adds a new namedItem to the collection of a given scope.
 
 ## Permissions
 One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
--- a/api-reference/beta/api/nameditem-add.md
+++ b/api-reference/beta/api/nameditem-add.md
@@ -1,9 +1,9 @@
 ---
-title: "Add Named Item"
+title: "workbookNamedItem: add"
 description: "Adds a new name to the collection of the given scope using the user's locale for the formula."
 localization_priority: Normal
 doc_type: apiPageType
-ms.prod: ""
+ms.prod: "excel"
 author: "ruoyingl"
 ---
 
@@ -20,9 +20,9 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type      | Permissions (from least to most privileged)              |
 |:--------------------|:---------------------------------------------------------|
-|Delegated (work or school account) | Files.ReadWrite, Sites.Read.All    |
-|Delegated (personal Microsoft account) | Files.ReadWrite    |
-|Application | Sites.Read.All |
+|Delegated (work or school account) | Files.ReadWrite.Selected, Files.ReadWrite.AppFolder, Files.ReadWrite, Files.ReadWrite.All, Sites.ReadWrite.All, Sites.Manage.All, Sites.FullControl.All    |
+|Delegated (personal Microsoft account) | Files.ReadWrite, Files.ReadWrite.All    |
+|Application | Files.ReadWrite.All, Sites.ReadWrite.All, Sites.Manage.All, Sites.FullControl.All   |
 
 ## HTTP request
 <!-- { "blockType": "ignored" } -->
@@ -37,7 +37,7 @@ POST /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/names/add
 | Name       | Description|
 |:---------------|:----------|
 | Authorization  | Bearer {token}. Required. |
-| Workbook-Session-Id  | Workbook session Id that determines if changes are persisted or not. Optional.|
+| Workbook-Session-ID  | Workbook session ID that determines if changes are persisted or not. Optional.|
 
 ## Request body
 In the request body, provide a JSON object with the following parameters.
--- a/api-reference/beta/api/nameditem-add.md
+++ b/api-reference/beta/api/nameditem-add.md
@@ -1,6 +1,6 @@
 ---
 title: "workbookNamedItem: add"
-description: "Adds a new name to the collection of the given scope using the user's locale for the formula."
+description: "Adds a new namedItem to the collection of the given scope using the user's locale for the formula."
 localization_priority: Normal
 doc_type: apiPageType
 ms.prod: "excel"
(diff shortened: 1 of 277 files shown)


jasonjoh (changes requested):

jasonjoh:
Couple of additional suggestions:

- Delete is also missing from v1. Check with the owner but I would assume it should be doc'ed there too :D.
- Don't forget to add this to [toc.yml](https://github.com/microsoftgraph/microsoft-graph-docs/blob/main/api-reference/beta/toc.yml)
- Edit the body of your pull request to add a line `Fixes #913`. This will auto-close the GitHub issue when the PR is merged.

api-reference/beta/api/nameditem-delete.md
```diff
@@ -0,0 +1,66 @@
+---
+title: "Delete namedItem"
```
chiz-ms, line 2: It seems for add the title is "Add Named Item". Shall we align?

api-reference/beta/api/nameditem-delete.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Delete namedItem
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Delete the specified [namedItem](../resources/workbooknameditem.md) object.
+
+## Permisions
```
chiz-ms, line 18: Seems a typo.

api-reference/beta/api/nameditem-delete.md
```diff
@@ -0,0 +1,67 @@
...
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Delete the specified [namedItem](../resources/nameditem.md) object.
+
+## Permisions
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | Files.ReadWrite, Sites.Read.All    |
```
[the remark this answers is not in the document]
chiz-ms: This is interesting. We didn't notice this before. We need to double check this (I agree it doesn't look correct).

api-reference/beta/api/nameditem-delete.md
```diff
@@ -0,0 +1,66 @@
...
+DELETE /me/drive/items/{id}/workbook/names/{name}
+DELETE /me/drive/root:/{item-path}:/workbook/names/{name}
+DELETE /me/drive/items/{id}/workbook/worksheets/{id|name}/names/{name}
+DELETE /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/names/{name}
+```
+
+## Request headers
+| Name       | Description|
+|:---------------|:----------|
+| Authorization  | Bearer {token}. Required. |
```
chiz-ms, line 39: Please add Workbook-Session-Id similar to the add request.

api-reference/beta/api/nameditem-delete.md
angelgolfer-ms: Good catch, chiz-ms!
Because the "create" method is actually an "add" action, not a usual POST operation, please change "Add Named Item" to "workbookNamedItem: add", following the [topic title guidance](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103176/ToC-and-topic-title).
Please make sure the topic title casing follows the camel case of the binding parameter (i.e., workbookNamedItem) and the method "add".

"Delete namedItem" follows the topic title guidance.

api-reference/beta/api/nameditem-delete.md
```diff
@@ -0,0 +1,66 @@
+---
+title: "Delete namedItem"
+description: "Delete the specified named item in a workbook."
+author: "isvargasmsft"
+localization_priority: Normal
+ms.prod: ""
```
angelgolfer-ms, line 6: Please always address any build warnings, suggestions, and of course build errors, if there are any.
In this case, there are 2 warnings about the ms.prod attribute (missing value). Please fill that in.

api-reference/beta/api/nameditem-delete.md
angelgolfer-ms: @isvargasmsft , when you add the Workbook-Session-Id header and its description, please make sure you use "ID" instead of "Id", in "Workbook session Id that determines if changes are persisted or not. Optional." "ID" is the proper acronym for identifier, not "Id".

It would be good if you can fix the same in the nameditem-add.md topic as well.

For

angelgolfer-ms (changes requested):
@isvargasmsft , good job particularly this is your first PR!
Just a few comments, please take a look!

api-reference/beta/api/nameditem-delete.md
chiz-ms: If that's the case probably we want to also review others? It seems current titles have various formats.

api-reference/v1.0/api/nameditem-add.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: "Add Named Item"
+title: "namedItem: add"
```
jasonjoh, line 2: In the beta topic it's workbookNamedItem. Which one is right?

jasonjoh (changes requested):
- Make sure the name of the resource is consistent. (workbookNamedItem vs. namedItem)
- Be sure to add the new delete topics to their respective TOC.yml files.

api-reference/v1.0/api/nameditem-add.md
isvargasmsft: @chiz-ms could you please clarify this?

chiz-ms: v1.0 should be the same as beta, but I'm not sure whether you should write `workbookNamedItem` or `namedItem`. It looks more about documentation convention. The type is currently described in https://github.com/microsoftgraph/microsoft-graph-docs/blob/f634fb80d5f87345782c3dc42a02a98fb2b7d0ed/api-reference/v1.0/resources/nameditem.md, and you can see even in that page it is sometimes called "`NamedItem`" while in some other cases "`WorkbookNamedItem`". Whichever we call it, I would prefer names aligned across the whole Graph documentation unless there's a reason to make them different :)

api-reference/beta/api/nameditem-add.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: "Add Named Item"
```
angelgolfer-ms, line 2: Please also make the update in line 10.
Here line 2 is just the metadata. It doesn't show in the published topic.

api-reference/v1.0/api/nameditem-add.md
isvargasmsft: Setting this to workbookNamedItem

angelgolfer-ms (approved):
Thanks for the prompt updates!

isvargasmsft:
@chiz-ms may I please have your approval to proceed further? Is there any additional changes you'd like for these particular articles?

api-reference/v1.0/api/nameditem-delete.md
```diff
@@ -0,0 +1,65 @@
...
+localization_priority: Normal
+ms.prod: "excel"
+doc_type: apiPageType
+---
+
+# Delete workbookNamedItem
+
+Namespace: microsoft.graph
+
+Delete the specified [namedItem](../resources/nameditem.md) object.
```
chiz-ms, line 14: Not sure whether we should use `namedItem` or `workbookNamedItem` here. And I wonder whether we should align such names across the whole Graph documentation.

chiz-ms (approved):

PR #12198, opened 2021-04-04
Abheek Das user purpose

abheek-das:
Doc update to notify users that userPurpose API would be temporarily un-available.


jasonjoh:
Please disregard the earlier message from msftbot. Your pull request was incorrectly flagged.

PR #12203, opened 2021-04-05
Generate stubs for updating noncustodial source in source collection as a list of NCDS.

roankari-msft:


Lauragra:
@roankari-msft , did you use the Upload files option in the GitHub UI to add these files? Please add file through a git client rather than uploading directly. Your PR is currently failing the CRLF check, which happens when contributors upload files without going through git.

PR #12231, opened 2021-04-08
updated audioConferencing

frankpeng7:


api-reference/beta/resources/audioconferencing.md
```diff
@@ -17,12 +17,19 @@ Represents phone access information for an [onlineMeeting](onlinemeeting.md).
...
+| dialinUrl       | String            | A URL to the externally-accessible web page that contains dial-in information. |
+| conferenceId    | String            | The conference id of the online meeting.                                       |
+| tollFreeNumber  | String            | The toll-free number that connects to the Audio Conference Provider.           |
+| tollFreeNumbers | String collection | List of toll-free numbers that are displayed in the meeting invite.            |
+| tollNumber      | String            | The toll number that connects to the Audio Conference Provider.                |
+| tollNumbers     | String collection | List of toll numbers that are displayed in the meeting invite.                 |
+
+> [!NOTE]
+>
+>- **tollFreeNumber** and **tollNumber** will be deprecated. Please use **tollFreeNumbers** and **tollNumbers** instead.
```
Lauragra, line 31: See the guidance for deprecating content: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103183/Deprecate-docs. Next to the property names in the Property column, put (deprecated).

frankpeng7: Added `(deprecated)` and changed "will be" to "are". 

Regarding the second sentence tho, Darrel mentioned this during Graph API review that once we declare deprecation, we have to 1) wait for at least 2 years 2) make sure there's no traffic/usage after that and then the property can be removed - so it's really hard for us to give a specific date when this will be removed.

Lauragra (approved):

PR #12261, opened 2021-04-12
Update onpremisesextensionattributes.md

weyto:
Adding attribute authority clarification in a cloud-only scenario.


weyto:
The change should be added as well on the beta version. Forgot to add them on that.

PR #12275, opened 2021-04-12
Send email with MIME content

isvargasmsft:
Starting with draft


api-reference/beta/api/message-reply.md
```diff
@@ -55,13 +55,14 @@ In the request body, provide a JSON object with the following parameters.
 | Parameter	   | Type	|Description|
 |:---------------|:--------|:----------|
 |comment|String|A comment to include. Can be an empty string.|
-|message|[message](../resources/message.md)|Any writeable properties to update in the reply message.|
+|message <br/> mimeContent|[message](../resources/message.md) <br/> Edm.String | Any writeable properties to update in the reply message. <br/> Base64Binary-encoded MIME content of the message being sent.|
```
jasonjoh, line 58: mimeContent should be it's own row in the table. Don't use <br/> here, instead add an additional line with cells broken up by pipes (`|`). I would add a note to the description of both message and mimeContent that only one should be present (if using message, don't use mimeContent, and vice-versa)

concepts/outlook-send-mime-message.md
```diff
@@ -0,0 +1,309 @@
...
+
+
+## Properties
+For sending in MIME content for our message-related actions, we are adding an optional parameter of type `Edm.String` called `mimeContent`. This `mimeContent` string contains the Base64Binary-enconded format of the MIME content text.
+
+## Actions
+| Action | Parameter       | Type           | Description     | Required  |
+| -------------- | -------------- | -------------- | --------------- | --------- |
+| sendMail | `message` | `graph.Message` | Message object of the message being sent | No |
+| sendMail | `mimeContent` | `Edm.String` | Base64Binary-encoded MIME content of the message being sent | No |
```
jasonjoh, line 59: All of these also need to be updated right? Each action has it's own topic.

jasonjoh (changes requested):

api-reference/beta/api/message-createforward.md
```diff
@@ -40,25 +40,30 @@ POST /me/mailFolders/{id}/messages/{id}/createForward
 POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/createForward
 ```
 ## Request headers
-| Name       | Type | Description|
-|:---------------|:--------|:----------|
-| Authorization  | string  | Bearer {token}. Required. |
-| Content-Type | string  | Nature of the data in the body of an entity. Required. |
+| Name       | Type | Description| Required |
```
angelgolfer-ms, line 43: Lines 43-46:
FYI in general for API reference topics, there are fixed templates defining H2 section titles, and corresponding table column titles. We can't add a column without affecting generator tool and API Doctor.

api-reference/beta/api/message-createforward.md
```diff
@@ -40,25 +40,30 @@ POST /me/mailFolders/{id}/messages/{id}/createForward
 POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/createForward
```
angelgolfer-ms, line 40: This comment is for the description section for the method, lines 16-23. (Please also apply this feedback to other method topics in this PR.)
In general, it's a good idea to clearly state in the top description what the method supports. That sentence in line 16 as is only addresses the original approach of specifying a body in JSON, which supports including a comment. That doesn't apply to the MIME body approach.
Suggestion:
"Create a draft message (add link to message.md file on first reference) for forwarding the specified message. You can specify the draft message body in JSON or MIME format.

When specifying the body in JSON format, you can include a comment, specify recipients, and update any message properties all in one **createForward** call. You can then [send](../api/message-send.md) the draft message.

**Note**
- You can specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying neither will return an HTTP 400 Bad Request error.

When specifying the body in MIME format, include only the MIME content in the request body. You must specify the recipients, and any updates to **message** properties in a subsequent [send](../api/message-send.md) operation to forward the draft message."

api-reference/beta/api/message-createforward.md
```diff
@@ -40,25 +40,30 @@ POST /me/mailFolders/{id}/messages/{id}/createForward
...
-| Authorization  | string  | Bearer {token}. Required. |
-| Content-Type | string  | Nature of the data in the body of an entity. Required. |
+| Name       | Type | Description| Required |
+|:---------------|:--------|:----------|:--------|
+| Authorization  | string  | Bearer {token}.| Yes |
+| Content-Type | string  | Nature of the data in the body of an entity. <br/> Use text/plain for MIME content and application/json for a json object| Yes |
 
 ## Request body
-In the request body, provide a JSON object with the following parameters.
+To send a json object provide the following parameters:
```
angelgolfer-ms, line 49: Line 49:
Please clarify this way:
When sending the draft body in JSON format, specify the following parameters.

Please note the appropriate upper-case for "JSON", and the lead-in sentence to a table always ends in a period, not a colon, per Microsoft Writing Style.
Specify the following parameters when

api-reference/beta/api/message-createforward.md
isvargasmsft: @angelgolfer-ms should removing the additional column be enough? I can move the Required piece to the second column and remove that one.

api-reference/beta/api/message-createforward.md
```diff
@@ -40,25 +40,30 @@ POST /me/mailFolders/{id}/messages/{id}/createForward
...
 
 | Parameter	   | Type	|Description|
 |:---------------|:--------|:----------|
 |comment|String|A comment to include. Can be an empty string.|
 |toRecipients|[recipient](../resources/recipient.md) collection|The list of recipients.|
 |message|[message](../resources/message.md)|Any writeable properties to update in the reply message.|
 
+> [!IMPORTANT]
+> To send MIME content no parameters are required, just paste the MIME string in the body of the request.
+
```
angelgolfer-ms, line 59: Please don't note this as "important", but instead present the MIME approach as just a peer alternative to the JSON approach.
Also, we want to draw a parallel to how we list the differences between the two approaches in the preceding section, to aid scanning.
So I suggest continuing to start with the phrase "When specifying a MIME body format,..."
Something to the effect of:
"When specifying a MIME body format, include only the MIME content in the request body."

api-reference/beta/api/message-createforward.md
```diff
@@ -40,25 +40,30 @@ POST /me/mailFolders/{id}/messages/{id}/createForward
...
 ## Response
 
 If successful, this method returns `201 Created` response code and [message](../resources/message.md) object in the response body.
 
-## Example
+## Examples
+
+### Example 1: Create a draft forward message
```
angelgolfer-ms, line 66: Line 66:
We need to better clarify what this example does, for example:
"Example 1: Create a draft message in JSON format for forwarding the specified message"

BTW, the original wording "draft forward message" has 3 nouns in a row, that's generally to avoid to reduce confusion for loc. I have tried to break up the noun phrase to make it less ambiguous.

api-reference/beta/api/message-createforward.md
```diff
@@ -146,6 +151,27 @@ Content-length: 272
 }
 ```
 
+
+### Example 2: Create a draft forward message with MIME content
+
```
angelgolfer-ms, line 156: Line 155:
For the same as reason as for line 66: Please break up the 3-noun phrase "draft forward message".

angelgolfer-ms (changes requested):
Hi @isvargasmsft , thank you for the quick turnaround!
Even though this is supposed to be in tech review stage and content review comes only when tech review feedback is incorporated, I wanted to give some general comments so that you can apply them to the topics currently in this PR and the ones you work on for v1.
Basically, please view each method topic more holistically, make sure at the start, it's clear that there are two approaches supported, and every subsequent section clearly states what the customer should do for each approach.

api-reference/beta/api/message-createforward.md
angelgolfer-ms: Yes please.

api-reference/beta/api/message-createreply.md
```diff
@@ -156,6 +156,92 @@ Content-type: application/json
...
+##### Request
+
+<!-- {
+  "blockType": "response",
+  "@odata.type": "microsoft.graph.message",
+  "truncated": true
+} -->
+
+```http
+POST https://graph.microsoft.com/beta/me/messages/AAMkADA1MTAAAH5JaLAAA=/createreply
```
msft-karlcoelho, line 169: Let's keep the camel case consistency. This could be `createReply` like the JSON example.

api-reference/beta/api/message-createreplyall.md
```diff
@@ -151,6 +148,88 @@ Content-type: application/json
...
 ```
 
+### Example 2: Create a draft using MIME format to reply-all to an existing message
+<!-- {
+  "blockType": "request",
+  "name": "message_createreplyall_mime_beta"
+}-->
+
+```http
+POST https://graph.microsoft.com/beta/me/messages/AAMkADA1MTAAAH5JaLAAA=/createreplyall
```
msft-karlcoelho, line 158: Same for `createReply`. Consistency on camel casing

api-reference/beta/api/message-send.md
```diff
@@ -33,22 +34,26 @@ One of the following permissions is required to call this API. To learn more, in
 ```http
 POST /me/messages/{id}/send
 POST /users/{id | userPrincipalName}/messages/{id}/send
+POST https://graph.microsoft.com/v1.0/me/sendMail
```
msft-karlcoelho, line 37: `sendMail` is different from `/send` the implementation doesn't make any changes to this endpoint. We can leave it alone.

msft-karlcoelho (changes requested):

api-reference/beta/api/message-send.md
isvargasmsft: Done!

api-reference/beta/api/message-createreply.md
isvargasmsft: Done

api-reference/beta/api/message-createreplyall.md
isvargasmsft: Done

api-reference/v1.0/api/message-createreply.md
```diff
@@ -31,18 +37,20 @@ POST /me/mailFolders/{id}/messages/{id}/createReply
...
-| Name       | Type | Description|
+| Name       | Type | Description| 
 |:---------------|:--------|:----------|
-| Authorization  | string  | Bearer {token}. Required. |
+| Authorization  | string  | Bearer {token}. Required|
+| Content-Type | string  | Nature of the data in the body of an entity. <br/> Use text/plain for MIME content and application/json for a JSON object|
 
 ## Request body
-Do not supply a request body for this method.
+This method does not require a request body. However, for creating a reply draft with MIME content, the MIME string should be provided in the body of the request. No parameters are required, just paste the MIME string in the body of the request.
```
msft-karlcoelho, line 46: Should mention that the MIME string needs to be Base64-encoded. Not just the MIME plaintext.

api-reference/v1.0/api/message-createreplyall.md
```diff
@@ -11,7 +11,9 @@ doc_type: apiPageType
 
 Namespace: microsoft.graph
 
-Create a draft to reply to the sender and all the recipients of the specified [message](../resources/message.md). You can then [update](../api/message-update.md) the draft to add reply content to the **body** or change other message properties, or, simply [send](../api/message-send.md) the draft.
+Create a draft to reply to the sender and all the recipients of the specified [message](../resources/message.md) in either JSON or MIME format. You can then [update](../api/message-update.md) the draft to add reply content to the **body** or change other message properties, or, simply [send](../api/message-send.md) the draft.
+
+When specifying the body in MIME format, include only the MIME content in the request body.
```
msft-karlcoelho, line 16: Again, should we be explicit about the MIME string being in Base64 encoding?

api-reference/v1.0/api/message-createreplyall.md
```diff
@@ -31,18 +33,20 @@ POST /me/mailFolders/{id}/messages/{id}/createReplyAll
...
-| Name       | Type | Description|
+| Name       | Type | Description| 
 |:---------------|:--------|:----------|
-| Authorization  | string  | Bearer {token}. Required. |
+| Authorization  | string  | Bearer {token}. Required |
+| Content-Type | string  | Nature of the data in the body of an entity. <br/> Use text/plain for MIME content and application/json for a JSON object|
 
 ## Request body
-Do not supply a request body for this method.
+This method does not require a request body. However, for creating a replyAll draft with MIME content, the content should be provided in the body of the request. No parameters are required, just paste the MIME string in the body of the request.
```
msft-karlcoelho, line 42: Similar comment, should be explicit about Base64-encoding the MIME string.

msft-karlcoelho (changes requested):

api-reference/v1.0/api/message-createreply.md
isvargasmsft: Solved

msft-karlcoelho (approved):
👍

api-reference/beta/api/message-createforward.md
```diff
@@ -13,14 +13,18 @@ Namespace: microsoft.graph
...
+When using JSON format: 
+- You can include a comment, specify recipients and update any message properties all in one **createForward** call. 
 - You can specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
-- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying 
-neither will return an HTTP 400 Bad Request error.
+- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying neither will return an HTTP 400 Bad Request error.
+
+When using MIME format:
+- Microsoft Graph does not suppport editing MIME properties individually, the complete MIME content must be provided in a base64-encoded string.
+- S/MIME properties must be previously included in the base64-encoded string.
```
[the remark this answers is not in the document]
isvargasmsft: Fixed.

api-reference/beta/api/message-createreplyall.md
```diff
@@ -13,17 +13,19 @@ Namespace: microsoft.graph
...
-([RFC 2822](https://www.rfc-editor.org/info/rfc2822)), you should send the reply to the recipients in the  
-**replyTo** and **toRecipients** properties, and not the recipients in the **from** and **toRecipients** properties. 
+When using JSON format:
+- Include a comment or update any message properties, all in one **createReplyAll** call.
+- Subsequently [update](../api/message-update.md) the draft message.
+- Specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
+- If the **replyTo** property is specified in the original message, per Internet Message Format ([RFC 2822](https://www.rfc-editor.org/info/rfc2822)), you should send the reply to the recipients in the **replyTo** and **toRecipients** properties, and not the recipients in the **from** and **toRecipients** properties. 
 
+When using MIME format:
+- Microsoft Graph does not suppport editing MIME properties individually, the complete MIME content must be provided in a base64-encoded string.
```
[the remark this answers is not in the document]
isvargasmsft: Thank you. Solved.

jasonjoh (approved):

api-reference/beta/api/message-createforward.md
```diff
@@ -13,14 +13,18 @@ Namespace: microsoft.graph
...
+When using JSON format: 
+- You can include a comment, specify recipients and update any message properties all in one **createForward** call. 
 - You can specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
-- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying 
-neither will return an HTTP 400 Bad Request error.
+- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying neither will return an HTTP 400 Bad Request error.
+
+When using MIME format:
+- Microsoft Graph does not support editing MIME properties individually, the complete MIME content must be provided in a base64-encoded string.
+- S/MIME properties can be included in the base64-encoded string.
```
angelgolfer-ms, line 27: The section for MIME:
In my earlier [comment](https://github.com/microsoftgraph/microsoft-graph-docs/pull/12275#discussion_r617074997), I suggested the following sentence:
"When specifying the body in MIME format, include only the MIME content in the request body. You must specify the recipients, and any updates to message properties in a subsequent send operation to forward the draft message."

While lines 26 and 27 sort of addressed the MIME content in the request body, there is the second phrase about specifying the recipients, that you should contrast with the JSON case. Otherwise that's a hole in the new scenario.

angelgolfer-ms, line 27: Lines 26 and 27:
The next comment is about parallel sentence structure. You can see examples [here](https://writing-center.phsc.edu/grammar/sentence-structure/problems-sentences/parallelism) or search for "parallel writing" on the Internet.
The group of sentences for the JSON format and for the MIME format should use a parallel grammatical sentence structure, i.e., tell customers what they should do for each case.
But instead, line 26 starts with "Microsoft Graph" as the subject and uses active voice, followed by a phrase in passive voice (second person implied).
And then line 27 is a sentence in passive voice.

angelgolfer-ms: So just to illustrate my point, I will provide an example of how to address my last two comments, for the section starting with line 25 (MIME case):

"When using MIME format:
- You must specify the complete MIME content in the request body as a base64-encoded string. 
- You may include S/MIME properties as part of the base64-encoded string. Microsoft Graph does not support editing these properties individually.
- You must specify the recipients, and any updates to message properties in a subsequent send operation to forward the draft message.

api-reference/beta/api/message-createforward.md
```diff
@@ -42,23 +46,27 @@ POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/createForwar
 ## Request headers
 | Name       | Type | Description|
 |:---------------|:--------|:----------|
-| Authorization  | string  | Bearer {token}. Required. |
-| Content-Type | string  | Nature of the data in the body of an entity. Required. |
+| Authorization  | string  | Bearer {token}. Required|
+| Content-Type | string  | Nature of the data in the body of an entity. Required <br/> Use text/plain for MIME content and application/json for a JSON object|
```
angelgolfer-ms, line 51: Line 50:
The preceding part of your topic describes the JSON case first, then the MIME case.
So please reverse the order of the two in the phrase you added.
Also, because there is normal text mixed with "text/plain" and "application/json" in the same table cell content, you should style the two by enclosing them in tick marks, like `text/plain` and `application/json`.
In general, any programmatic strings that are not part of the Graph API should be styled this way, especially when the strings are embedded with normal text. This would promote customer reading/scanning.

api-reference/beta/api/message-createforward.md
```diff
@@ -42,23 +46,27 @@ POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/createForwar
...
+To send a JSON object provide the following parameters.
 
 | Parameter	   | Type	|Description|
 |:---------------|:--------|:----------|
 |comment|String|A comment to include. Can be an empty string.|
 |toRecipients|[recipient](../resources/recipient.md) collection|The list of recipients.|
 |message|[message](../resources/message.md)|Any writeable properties to update in the reply message.|
 
+When specifying the body in MIME format no parameters are required, include only the MIME content as **a Base64-encoded string** in the request body.
+
```
angelgolfer-ms, line 62: The next comment applies to the "Request body" section:
Please apply parallel writing again to line 53 (the JSON case) and line 61 (MIME case).
I suggested that in a previous [comment](https://github.com/microsoftgraph/microsoft-graph-docs/pull/12275#discussion_r617075645). If you prefer a shorter version, you can start line 53 like this:
"When specifying a draft message in JSON format, include the following parameters."

Then in line 61, you can do this please:
"When specifying a draft message in MIME format, include only the MIME content as **a base64-encoded string** in the request body. Do not include any parameters."
Please note the lower case 'b' for Base64.

FYI there are 2 issues with line 61. First, as is, that line is a run-on sentence. There are two ideas ("no parameters are required" and "include only the MIME content..."), you don't connect them with just a comma. Second, remember, use second person singular or imperative as much as you can. But here, again, you introduced passive voice and broke the parallelism.

api-reference/beta/api/message-createforward.md
```diff
@@ -146,6 +154,106 @@ Content-length: 272
...
+} -->
+
+```http
+HTTP/1.1 201 Created
+Content-type: application/json
+
+{
+    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#messages/$entity",
+    "@odata.etag": "W/\"CQAAABYAAADt09JRVj4bRqYt7q6MVBpAAAAiQc5k\"",
+    "id": "AAMkAGNhOWVmNmEyLTgwNjItNDI0NC05ZGI5LWZhYjc4MzcxZWE3YQBGAAAAAADW_NNy5NNRT51Qs4RFfm_PBwDt09JRVj4bRqYt7q6MVBpAAAAAAAEPAADt09JRVj4bRqYt7q6MVBpAAAAiTjnQAAA=",
```
angelgolfer-ms, line 188: Line 188:
You forgot to match this id value with the truncated id segment you introduced in the example request in line 167.

In general, whenever you truncate an ID property value, you can do a search and replace to make sure you truncate the same value consistently.

api-reference/beta/api/message-createforward.md
```diff
@@ -146,6 +154,106 @@ Content-length: 272
...
+        "emailAddress": {
+            "name": "Alex Wilber",
+            "address": "AlexW@M365x765140.OnMicrosoft.com"
+        }
+    },
+    "toRecipients": [
+        {
+            "emailAddress": {
+                "name": "Megan Bowen",
+                "address": "MeganB@M365x765140.OnMicrosoft.com"
```
angelgolfer-ms, line 229: Line 229:
Please replace the actual tenant domain by something like "contoso". It's important that as authors and content reviewers, we make sure our docs use only CELA-approved data and do not inadvertently expose real data.
Please look up the related CDK topic: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103160/Fictitious-names

api-reference/beta/api/message-createforward.md
```diff
@@ -146,6 +154,106 @@ Content-length: 272
...
+    "bccRecipients": [],
+    "replyTo": [],
+    "flag": {
+        "flagStatus": "notFlagged"
+    }
+}
+
+```
+
+If the MIME content in the request body is malformed, this method returns the following error message.
```
angelgolfer-ms, line 243: Line 244:
Include this line for API Doctor:
```
```

api-reference/beta/api/message-createforward.md
```diff
@@ -146,6 +154,106 @@ Content-length: 272
...
+
+```http
+HTTP/1.1 400 Bad Request
```
angelgolfer-ms, line 246: Also, I thought we had talked about this - please add this error case to the "Response" section.
This being the very last section in an API topic, and without any H2 section such that it wouldn't appear in the right-side TOC, customers will miss this info.

You can add this as a separate paragraph in the "Response" section:

If successful, this method returns `201 Created` response code and [message](../resources/message.md) object in the response body.
If the request body includes malformed MIME content, this method returns `400 Bad request` and the following error message: 
```
Invalid base64 string for MIME content.
```

api-reference/beta/api/message-createforward.md
```diff
@@ -13,14 +13,18 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Create a draft forward message to include a comment or update any message properties  
-all in one **createForward** call. You can then [send](../api/message-send.md) the draft message.
+Create a draft to forward an existing [message](/graph/api-reference/beta/resources/message.md). Specify the request body in either JSON or MIME format and [send](../api/message-send.md) the draft message when ready.
 
-**Note**
+Alternatively, [forward a message](../api/message-forward.md) in a single operation.
```
angelgolfer-ms, line 18: Line 18:
I'd suggest moving this line all the way as the last sentence in this section, before line 29, the next H2 section "Permissions".
The reason is lines 20-27 still apply to the createForward method, so it's better to keep them together.
Besides, some customers may get confused and mistake line 20 applies to the method mentioned in the preceding line, line 18.

api-reference/beta/api/message-createreplyall.md
```diff
@@ -13,17 +13,19 @@ Namespace: microsoft.graph
...
-**Note**
+When ready, [send](../api/message-send.md) the draft message.
 
-- You can specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
-- If the **replyTo** property is specified in the original message, per Internet Message Format 
-([RFC 2822](https://www.rfc-editor.org/info/rfc2822)), you should send the reply to the recipients in the  
-**replyTo** and **toRecipients** properties, and not the recipients in the **from** and **toRecipients** properties. 
+When using JSON format:
+- Include a comment or update any message properties, all in one **createReplyAll** call.
+- Subsequently [update](../api/message-update.md) the draft message.
```
angelgolfer-ms, line 22: Line 22:
Is that true?
Per line 21, the customer can already set property values of the draft message in the request body. Why would we suggest the customer to do a separate PATCH operation to that draft message?

angelgolfer-ms (changes requested):
@isvargasmsft , thanks for incorporating the last set of feedback for the message-createforward.md topic.
I have reviewed the same beta version of the topic and added a few more comments. A few of the comments relate to writing parallel sentences, a basic writing technique.
Please take a look and apply the comments to the other topics as well. I eyeballed a couple more topics and noticed they all follow a similar pattern.
I'll take a look at the rest of the topics in a day or so.

api-reference/beta/api/message-createforward.md
isvargasmsft: Fixed!

api-reference/beta/api/message-createreplyall.md
```diff
@@ -13,20 +13,22 @@ Namespace: microsoft.graph
...
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Create a draft of a reply-all message to include a comment or update any message properties, 
-all in one **createReplyAll** call. You can then [update](../api/message-update.md) or 
-[send](../api/message-send.md) the draft.
+Create a draft to reply to all recipients of a [message](../resources/message.md) in either JSON or MIME format.
 
-**Note**
+When using JSON format:
+- Include a comment or update any message properties, all in one **createReplyAll** call.
```
angelgolfer-ms, line 19: Line 19:
@isvargasmsft , I took a closer look at the current state of this bullet list, just want to make sure we got on the same page with this first bullet item. As is:
"Include a comment or update any message properties, all in one **createReplyAll** call."
this sentence is in imperative voice, implying the customer should do this.
That would be ambiguous with the next bullet point in line 20, which says the customer can choose to EITHER specify a comment, OR, not specify the comment parameter and specify the body property of the message parameter.

Somehow we still need to convey that line 19 is not imperative, and is something the customer can choose to do.

api-reference/v1.0/api/message-createreplyall.md
```diff
@@ -11,10 +11,21 @@ doc_type: apiPageType
 
 Namespace: microsoft.graph
 
-Create a draft to reply to the sender and all the recipients of the specified [message](../resources/message.md). You can then [update](../api/message-update.md) the draft to add reply content to the **body** or change other message properties, or, simply [send](../api/message-send.md) the draft.
+Create a draft to reply to the sender and all the recipients of the specified [message](../resources/message.md) in either JSON or MIME format.
```
angelgolfer-ms, line 14: @isvargasmsft , can you please confirm, is it true that the v1 and beta version of createReplyAll do not use the same syntax? That is, the v1 version doesn't take any parameters (and hence we explicitly state in the original line 14 that the customer needs to do a PATCH operation to specify the reply content to the body and so forth)?
Whereas the beta version takes the comment and message parameters?

api-reference/v1.0/api/message-createreplyall.md
```diff
@@ -11,10 +11,21 @@ doc_type: apiPageType
 
 Namespace: microsoft.graph
 
-Create a draft to reply to the sender and all the recipients of the specified [message](../resources/message.md). You can then [update](../api/message-update.md) the draft to add reply content to the **body** or change other message properties, or, simply [send](../api/message-send.md) the draft.
+Create a draft to reply to the sender and all the recipients of the specified [message](../resources/message.md) in either JSON or MIME format. 
+
+When using JSON format:
+- Optionally, [update](../api/message-update.md) the draft later to add reply content to the **body** or change other message properties.
```
angelgolfer-ms, line 17: Line 17:
If you confirm my last comment as positive - that this method doesn't take any comment or message parameter as in the beta version - then I'd suggest in line 17, we don't specify "Optionally", even though technically our API may not require it. The typical reply-all use case is there's got to be a message body, otherwise why do a reply-all?

Also, since this is just a single bullet point (as opposed to a few bullet points in some other method topics), we shouldn't have a bullet list (per our writing style guide). Perhaps in this case, we don't separate the steps subsequent to calling createReplyAll into line 17, then skip lines 19-21, and resume in line 23?

Instead, for line 16-17, we can just say:
"When specifying the JSON format, you can then update the draft to add reply content to the body or change other message properties, or, simply send the draft."

I think that's clearer than saying "optionally", because in that context, it's not clear what that thing is an option for, as opposed to what.

(Then of course make line 23 a bullet point for the MIME format case.)

concepts/outlook-send-mime-message.md
```diff
@@ -0,0 +1,53 @@
...
+ms.prod: "outlook"
+---
+
+# Send emails with MIME content
+
+## Overview
+Email clients have the ability to send emails through Exchange in a MIME message format and communicate across multiple email platforms.
+
+Supported MIME features:
+- Send emails in MIME format: message body contents, attachments and sender/recipient information (CC and BCC).
```
angelgolfer-ms, line 15: Line 15:
Hm, you refer to "attachments".
The attachment feature is not a simple story. There are 3 vectors to consider:
(a) the types of attachments (file attachment, reference attachment, and Outlook item attachment which in turn can be a contact, event, or message). 
(b) What can you attach to - you can add an attachment to an event, message, or group post.
(c) How do you add the attachment - in general, you can attach to an existing Outlook item, attach to a draft of a message, or add the attachment when creating and sending a message (sendMail method).

So you need to be precise about this - is the supported scenario about attaching a message represented in MIME format, to a draft message? How about attaching to an event or group post? We need to make that clear.

If indeed we support attaching messages, events, or contacts in MIME format to a draft message, then we also need a separate section to illustrate that in the concept topic.

angelgolfer-ms, line 15: Also line 15:
What is the MIME support about "sender/recipient information (CC and BCC)"?
Are we referring to the "cc" and "bcc" Internet message headers (which are Internet message headers per RFC 2076), or are we referring to some other MIME headers that serve the cc and bcc purpose?
Please make sure if we're referring to the Internet message headers, then that's not MIME and shouldn't be included here.

api-reference/beta/api/message-createreply.md
```diff
@@ -13,15 +13,20 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Create a draft of a reply message to include a comment or update any message properties 
-all in one **createReply** call. You can then [update](../api/message-update.md) or 
-[send](../api/message-send.md) the draft.
+Create a draft to reply to the sender of a [message](../resources/message.md) in either JSON or MIME format.
 
-**Note**
+When using JSON format, you can:
```
angelgolfer-ms, line 18: Line 18:
FYI I'm moving the "you can" to line 20, because keeping it in line 18 wouldn't connect with the second bullet point in line 20.
BTW, even though we're doing it in line 18 in message-createforward.md, just so you know in the [style guide](https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=36412): 
"Global tip
If your content will be localized, avoid lists where an introductory fragment is completed by the list items below it. This can be difficult to translate."

api-reference/beta/api/message-createforward.md
```diff
@@ -13,17 +13,23 @@ Namespace: microsoft.graph
...
+- Specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying neither will return an HTTP 400 Bad Request error.
+- [Update](../api/message-update.md) the draft later to add content to the **body** or change other message properties.
 
-- You can specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
-- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying 
-neither will return an HTTP 400 Bad Request error.
+When using MIME format:
+- Provide the applicable [Internet Message Headers](https://tools.ietf.org/html/rfc2076) and the [MIME content](https://tools.ietf.org/html/rfc2045), all encoded in **base64** format in the request body.
+- Add any attachments and S/MIME properties into the MIME content.
+
```
angelgolfer-ms, line 26: Line 25:
FYI I'm making this change to 16 files in your PR:
"... add {something} to..."

api-reference/beta/api/message-createforward.md
```diff
@@ -42,23 +48,29 @@ POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/createForwar
 ## Request headers
 | Name       | Type | Description|
 |:---------------|:--------|:----------|
-| Authorization  | string  | Bearer {token}. Required. |
-| Content-Type | string  | Nature of the data in the body of an entity. Required. |
+| Authorization  | string  | Bearer {token}. Required|
+| Content-Type | string  | Nature of the data in the body of an entity. Required <br/> Use `application/json` for a JSON object and `text/plain` for MIME content|
```
angelgolfer-ms, line 53: Hm, FYI every description should end with a period, particularly because each table cell already has a phrase that ends with a period.
https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28751

api-reference/beta/api/message-createforward.md
```diff
@@ -42,23 +48,29 @@ POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/createForwar
...
 |message|[message](../resources/message.md)|Any writeable properties to update in the reply message.|
 
+When specifying the body in MIME format, provide the MIME content with the applicable Internet Message Headers ("To", "CC", "BCC", "Subject"), all encoded in **base64** format in the request body.
+
 ## Response
 
 If successful, this method returns `201 Created` response code and [message](../resources/message.md) object in the response body.
 
-## Example
+If the request body includes malformed MIME content, this method returns `400 Bad request` and the following error message: `Invalid base64 string for MIME content`.
```
angelgolfer-ms, line 69: Line 69:
FYI I enclosed the error message with double quotes rather than the tick marks.
I chose to do that mainly because of the styling effect (gray background) being applied to the longer length of the sentence. In general, we want to use special styling (like gray background, bolding, etc.) that calls for attention sparingly. The other reason is, in the same context, the phrases `201 Created` and `400 Bad request` are HTTP 1.0+ standards, which might make sense to differentiate them from error messages that are specific to Microsoft.
In general, the current means of styling are limited and doesn't cover all our needs.

api-reference/beta/api/message-createforward.md
```diff
@@ -13,17 +13,23 @@ Namespace: microsoft.graph
...
+When using JSON format, you can: 
+- Specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
+- Specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying neither will return an HTTP 400 Bad Request error.
+- [Update](../api/message-update.md) the draft later to add content to the **body** or change other message properties.
 
-- You can specify either a comment or the **body** property of the `message` parameter. Specifying both will return an HTTP 400 Bad Request error.
-- You must specify either the `toRecipients` parameter or the **toRecipients** property of the `message` parameter. Specifying both or specifying 
-neither will return an HTTP 400 Bad Request error.
+When using MIME format:
+- Provide the applicable [Internet Message Headers](https://tools.ietf.org/html/rfc2076) and the [MIME content](https://tools.ietf.org/html/rfc2045), all encoded in **base64** format in the request body.
```
angelgolfer-ms, line 24: Line 24:
FYI "Internet Message Header" should be "Internet message header". I am going to do a search and replace.

api-reference/beta/api/user-post-messages.md
```diff
@@ -13,15 +13,24 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Use this API to create a draft of a new message. Drafts can be created in any folder and optionally updated before sending. To save to the Drafts folder, use the /messages shortcut.
+Create a draft of a new message in either JSON or MIME format.
 
-While creating the draft in the same **POST** call, you can:
+When using JSON format you can:
+- Include an [attachment](../resources/attachment.md) and use a [mention](../resources/mention.md) to call out another user in the new message.
+- [Update](../api/message-update.md) the draft later to add content to the **body** or change other message properties.
```
angelgolfer-ms, line 20: Line 19:
I broke these 2 phrases into 2 bullet points. They are unrelated.

api-reference/beta/api/user-post-messages.md
```diff
@@ -13,15 +13,24 @@ Namespace: microsoft.graph
...
 
-- Include an [attachment](../resources/attachment.md) 
-- Use a [mention](../resources/mention.md) to call out another user in the new message
+When using MIME format:
+- Provide the applicable [Internet Message Headers](https://tools.ietf.org/html/rfc2076) and the [MIME content](https://tools.ietf.org/html/rfc2045), all encoded in **base64** format in the request body.
+- Add any attachments and S/MIME properties into the MIME content.
+
+You can save the draft in any folder. To save it in the Drafts folder, use the /messages shortcut.
```
angelgolfer-ms, line 26: Line 26:
I know this sentence has been there before this PR: "To save it in the Drafts folder, use the /messages shortcut."
FYI I am going to rephrase it. The way it is makes it sound like "using the /messages shortcut" is an option. AFAIK it's not an option, the CSDL defines the user entity has a messages nav property ("shortcut"?) ; what else can the query URL be otherwise.

angelgolfer-ms, line 26: Line 26:
I know this sentence has been there before this PR: "To save it in the Drafts folder, use the /messages shortcut."
FYI I am going to rephrase it. The way it is makes it sound like "using the /messages shortcut" is an option. AFAIK it's not an option, the CSDL defines the user entity has a messages nav property ("shortcut"?) ; what else can the query URL be otherwise.

api-reference/beta/api/user-post-messages.md
```diff
@@ -38,26 +47,30 @@ POST /me/mailFolders/{id}/messages
...
 
 If successful, this method returns a `201 Created` response code and a [message](../resources/message.md) object in the response body.
 
-## Example
+If the request body includes malformed MIME content, this method returns `400 Bad request` and the following error message: `Invalid base64 string for MIME content`.
+
+## Examples
+### Example 1: Create a message draft in JSON format
 ##### Request 1
 Here is an example of the request to create a draft of a new message.
```
angelgolfer-ms, line 75: Examples section - please note the structure we should use.
Level 2 heading: "Examples"
Level 3 heading: "Example x: {short description}"
Level 4 heading: "Request", "Response". With these headings, we don't keep the count, because the level 3 heading already keeps the count.

api-reference/v1.0/resources/mail-api-overview.md
```diff
@@ -67,6 +67,7 @@ The Microsoft Graph API also provides methods and actions that support common us
 | Add custom app data as Internet message headers of a message | [message](../resources/message.md) | Add custom data to the **internetMessageHeaders** property of the message. |
 | Add custom app data to a message by using extensions | [openTypeExtension](../resources/opentypeextension.md), <br>[schemaExtension](../resources/schemaextension.md) | [Add custom data to resources using extensions](/graph/extensibility-overview) |
 | Access custom data for under-exposed Outlook MAPI properties | [singleValueLegacyExtendedProperty](../resources/singlevaluelegacyextendedproperty.md), <br> [multiValueLegacyExtendedProperty](../resources/multivaluelegacyextendedproperty.md) | [Outlook extended properties overview](../resources/extended-properties-overview.md) |
+| Send emails with MIME content | [message](../resources/message.md) | [Send MIME content](/graph/outlook-send-mime-message) |
```
angelgolfer-ms, line 70: Line 70:
I'm moving this to the section for user-centric actions.

concepts/outlook-mail-concept-overview.md
```diff
@@ -30,29 +30,31 @@ Customers like how Outlook helps them stay organized. Microsoft Graph brings the
...
+
+- Email clients can send and receive messages in MIME formats. Even though Outlook does not save messages in MIME format, there are two ways you can [get an Outlook message body in MIME format](outlook-get-mime-message.md). In addition, email clients have the ability to [send emails through Exchange in a MIME message format](outlook-send-mime-message.md), attach S/MIME digital signatures to emails as well as S/MIME encrypted email contents and update message drafts in Exchange with MIME content.
 
 ### Write smarter apps that leverage intelligence
 
 Use Microsoft Graph to suggest contextual data to your app users:
 
-- Integrate with [Focused Inbox](/graph/api/resources/manage-focused-inbox?view=graph-rest-1.0) and [@-mentions (preview)](/graph/api/message-get?view=graph-rest-beta#request-2) and let your app users read and respond to what's relevant to them first.
+- Integrate with [Focused Inbox](/graph/api/resources/manage-focused-inbox?view=graph-rest-1.0&preserve-view=true) and [@-mentions (preview)](/graph/api/message-get?view=graph-rest-beta&preserve-view=true#request-2) and let your app users read and respond to what's relevant to them first.
```
angelgolfer-ms, line 46: Thanks for fixing these links. FYI including the preserve-view query parameter to versioned links came about only last year. Along with that change was that links to the reference that do not include the ?view query parameter default to v1.0. So we can actually remove the ?view query parameter altogether, and there's no view to preserve either (i.e., don't need to include the preserve-view query parameter).

concepts/outlook-send-mime-message.md
```diff
@@ -0,0 +1,191 @@
...
+- S/MIME encrypted email contents.
+
+> [!IMPORTANT]
+> * S/MIME provides two key features: digital signatures for verification of sender and email contents, and message encryption for preventing third parties from viewing email contents.
+> * Add S/MIME digital signatures and encrypted contents to emails only in their MIME message format.
+
+Visit the documentation about [getting MIME content of a message](../concepts/outlook-get-mime-message.md) for more information about MIME.
+
+## Details
+There are two supported formarts for creating a [message](/graph/api-reference/v1.0/resources/message.md): (1) JSON metadata (2) MIME content. Specify the selected format in the request header.
```
angelgolfer-ms, line 28: FYI, in general, we reserve numbered lists to procedures where the order of the list items matters.
When the order doesn't matter, we use bullet lists.

concepts/outlook-send-mime-message.md
```diff
@@ -0,0 +1,191 @@
...
+AAATAAAAAAAAAAAAAAAAAFy6CgBjdXN0b21YbWwvaXRlbTIueG1sUEsFBgAAAAAhACEAwwgAAHHB
+CgAAAA==
+
+--_004_MN2PR13MB3152FA5D970E0F158DF41456D5579MN2PR13MB3152namp_--
+
+```
+
+## Use cases
+| Use cases | Resources |
+| ------| ----- |
```
angelgolfer-ms, line 175: Looks like the links in the right column of this table are not specified properly. Site-relative links don't take the file extension (.md).

concepts/outlook-send-mime-message.md
```diff
@@ -0,0 +1,191 @@
...
+title: "Sending emails with MIME content"
+description: "Email clients have the ability to send emails through Exchange in a MIME message format."
+author: "isvargasmsft"
+localization_priority: Priority
+ms.prod: "outlook"
+---
+
+# Send emails with MIME content
+
+## Overview
```
angelgolfer-ms, line 11: Lines 11 and 27:
In general, it's a good practice to use section headers for education and SEO purpose, so titling sections within a topic (which is typically shorter than an article) with "Overview" and "Details" are too generic and waste of opportunity.
In a conceptual topic that's usually 1-2 pages, the first part of the topic should summarize the purpose of the topic, so  that being an "overview" is already implicit. We don't usually start a concept topic with a header other than the topic title.

angelgolfer-ms (changes requested):
@isvargasmsft , I made some fixes directly in your PR so that we can publish on 5/24 just before Build. Please take a look at my [commit](https://github.com/microsoftgraph/microsoft-graph-docs/commit/1e3c302fbca82c381fff061070fc40d8a81084e9) and related comments.
There is one thing you should follow up on - the changelog.

PR #12288, opened 2021-04-13
Add missing properties & complex type

FaithOmbongi:


Lauragra:
@FaithOmbongi , we don't need a changelog entry for the complex type in v1.0?

FaithOmbongi:
> @FaithOmbongi , we don't need a changelog entry for the complex type in v1.0?

@Lauragra , these are not new/promoted properties on the resource. Just pieces of undocumented info we're resolving.

PR #12294, opened 2021-04-14
Update tablerow-delete.md

grangeryy:
update more information

### DIFF
--- a/api-reference/beta/api/table-post-rows.md
+++ b/api-reference/beta/api/table-post-rows.md
@@ -35,10 +35,6 @@ POST /me/drive/items/{id}/workbook/tables/{id|name}/rows/add
 POST /me/drive/root:/{item-path}:/workbook/tables/{id|name}/rows/add
 POST /me/drive/items/{id}/workbook/worksheets/{id|name}/tables/{id|name}/rows/add
 POST /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/tables/{id|name}/rows/add
-POST /me/drive/items/{id}/workbook/tables/{id|name}/rows
-POST /me/drive/root:/{item-path}:/workbook/tables/{id|name}/rows
-POST /me/drive/items/{id}/workbook/worksheets/{id|name}/tables/{id|name}/rows
-POST /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/tables/{id|name}/rows
 
 ```
 ## Request headers
--- a/api-reference/beta/api/table-post-rows.md
+++ b/api-reference/beta/api/table-post-rows.md
@@ -35,6 +35,10 @@ POST /me/drive/items/{id}/workbook/tables/{id|name}/rows/add
 POST /me/drive/root:/{item-path}:/workbook/tables/{id|name}/rows/add
 POST /me/drive/items/{id}/workbook/worksheets/{id|name}/tables/{id|name}/rows/add
 POST /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/tables/{id|name}/rows/add
+POST /me/drive/items/{id}/workbook/tables/{id|name}/rows
+POST /me/drive/root:/{item-path}:/workbook/tables/{id|name}/rows
+POST /me/drive/items/{id}/workbook/worksheets/{id|name}/tables/{id|name}/rows
+POST /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/tables/{id|name}/rows
 
 ```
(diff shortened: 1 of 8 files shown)


api-reference/beta/api/table-post-rows.md
```diff
@@ -35,6 +35,10 @@ POST /me/drive/items/{id}/workbook/tables/{id|name}/rows/add
 POST /me/drive/root:/{item-path}:/workbook/tables/{id|name}/rows/add
```
Lauragra, line 35: It looks like we need an action topic with file name workbooktablerow-add and topic title workbookTableRow: add for this syntax. I can't see in the metadata what the resource/nav property is that you're documenting here. Can you point me to it?

api-reference/beta/api/tablerow-delete.md
```diff
@@ -30,6 +30,10 @@ DELETE /me/drive/items/{id}/workbook/tables/{id|name}/rows/{index}
 DELETE /me/drive/root:/{item-path}:/workbook/tables/{id|name}/rows/{index}
```
Lauragra, line 30: I don't see a delete action defined with a binding parameter of tableRow. The only delete action is for workbookRange. Can you point me to the metadata for this operation?

api-reference/beta/api/table-post-rows.md
grangeryy: Sorry Laura, not sure I get you point on this. I only modify the existing doc [HERE]: (https://docs.microsoft.com/en-us/graph/api/table-post-rows?view=graph-rest-1.0&tabs=http)

api-reference/beta/api/tablerow-delete.md
grangeryy: Is this one the correct? (https://docs.microsoft.com/en-us/graph/api/resources/tablerow?view=graph-rest-1.0)
I saw:
image

api-reference/beta/api/table-post-rows.md
grangeryy: image

Lauragra: I'm pointing out an issue with the documentation that your update has uncovered. We actually should have a separate method topic for the add action. It looks like the file name here should be workbooktable-post-rows. Do you know why the resource names in the file names and topic titles don't include the "workbook" preface? Note that in /beta, we have a resource topic "workbooktablerow", but in v1.0 we only have a resource topic "tablerow". In both beta and v1.0, we should use "workbooktablerow" for the file name and "workbookTableRow resource type" for the title, and it looks like that resource and its methods should be surfaced in the TOC?

grangeryy: @Lauragra I agree that these resources and methods should be surfaced in TOC. "table-post-rows.md" is the one for add action. I searched in the system and find this existing documentation, but it's not in TOC. I also checked with engineer and this one is used by customers to add table row/s in workbook. Then I directly modified these contents in this PR. I don't know why the names are different in V1.0 and beta. Guess it's a problem carried over from the past. Should I directly modify these existing file names and titles or what's the best practice here?

Lauragra: I think that we rename this file to workbooktable-post-rows and keep the rows syntax in this topic, and then we should create a new method topic, workbooktablerow-add, for the add method.

grangeryy: Guess this time I got the point. Sorry for bringing the confusion. For the scenario wise, "create table row" and "add table row"  is doing exactly the same thing from user perspective. Is it better to modify the title from "Create Row" -> "Add Row" directly and no change on the existing files??

Lauragra: I guess that ultimately I'm not clear about why we need two methods to do the same task to add a table row. However, the action is defined in the metadata and should have a standalone topic "workbookTableRow: add".

api-reference/beta/api/tablerow-delete.md
Lauragra: That is the resource topic. These POST operations to delete rows look like an action but I don't see the action in the metadata. Can you point me to the CSDL for the action syntax you're adding? For the existing topic, the title should be Delete workbookTableRow. The new syntax that you're adding should be in a separate action topic - assuming that the action exists (again, I'm not seeing it in the metadata), with a title "workbookTableRow: delete".

grangeryy:
@Lauragra, for this PR, I only keep the change on TOC.

PR #12436, opened 2021-04-23
Update teamworktag-update.md

anniecolonna:
Updated permission support to add application and remove delegated. We will add delegated back in once it's unblocked and out.


anniecolonna:
Getting an error, trying again

PR #12369, opened 2021-04-19
Update Assignments GET APIs to include AppOnly support

zoexi:
Updating the EduAssignments docs to specify that we now support Applications Permissions for certain Assignments GET APIs (READ operations).

PR #12507, opened 2021-04-27
Update to specify Assignments application permissions is in private preview

zoexi:
Follow-up to: https://github.com/microsoftgraph/microsoft-graph-docs/pull/12369


mlafleur:
Is there a link/address a developer should use to gain access to the private preview?

Lauragra (approved):

zoexi:
> Is there a link/address a developer should use to gain access to the private preview?

This will be available to the general public in a few days! And based on telemetry, we have also reached out to a few customers to private preview this functionality.

PR #12570, opened 2021-05-03
Remove resourse type from response metadata if response has no body to validate

millicentachieng:
Currently, validation of resource type specified in response metadata only happens if the response has a body. We have seen cases where the resource type does not exist yet it's not being validated by API Doctor due to the reason above.

A new API Doctor rule has been added to ensure that a resource type should only be specified in the response metadata if there's a response body. That way, the validity of the resource type is also checked.

This PR is to bring the docs up to date with this new rule.


millicentachieng:
We have three errors in both v1.0 and beta that will be fixed with the next API Doctor update.

Lauragra (approved):

PR #12677, opened 2021-05-10
Adding disabledByMicrosoftStatus property for App and SP resources

tilarso:
Adding disabledByMicrosoftStatus to App and SP resources to reflect when an application has been disabled by Microsoft due to a violation of our ToU / found to be malicious in nature.


tilarso:
Minor changes required to application and sp enum description.

PR #12703, opened 2021-05-12
Hidden Mail Folders - GA Release - Doc Updates

abheek-das:
Promoting new mailFolder property: "isHidden" & query parameter - includeHiddenFolders from Graph Beta to GA.


api-reference/v1.0/api/user-post-mailfolders.md
```diff
@@ -12,6 +12,9 @@ doc_type: apiPageType
 Namespace: microsoft.graph
 
 Use this API to create a new mail folder in the root folder of the user's mailbox.
```
angelgolfer-ms, line 14: @abheek-das , this comment isn't specific to this file but applicable for this PR:
In Graph Explorer, the operation documented in mailfolder-post-childfolders.md also supports setting isHidden, in beta and v1. 

So looks like we overlooked updating these 2 files in beta before, and we need to do that for both beta and v1 now:
mailfolder-post-childfolders.md 
mailfolder-list-childfolders.md

Do you mind following up on that please?

angelgolfer-ms: Also, please add these 2 operations to the Methods table in mailfolder.md. 
To the same Methods table, please add the "list mailFolder" operation (user-list-mailfolders.md) as well.
Add these 3 operations to the Methods table, in v1 and in beta, please.

angelgolfer-ms (changes requested):
Hi @abheek-das , thanks for the PR. Please take a look at my comments about updating the 2 topics mailfolder-post-childfolders.md and mailfolder-list-childfolders.md, and adding the 2 operations plus the user-list-mailfolders.md operation to the Methods table.

FYI I reorganized the order of the children topics under the "Mail folder" node in the v1 and beta TOC.

Thanks!

abheek-das:
Hi @angelgolfer-ms:
Thanks for your review, I've addressed the comments that you left previously. 
One question, shouldn't the title of pages mailfolder-post-childfolders.md be "Create ChildFolder" instead of "Create MailFolder". We already have user-post-mailfolders.md which tells us how to "Create MailFolder".

api-reference/v1.0/api/user-post-mailfolders.md
abheek-das: Indeed looks like mailfolder-post-childfolders.md & mailfolder-list-childfolders.md got missed. I've added them now for both Beta and v1.0.

Updated the Methods table in mailfolder.md as well. Ensured that following pages are listed as one of the operations:
- mailfolder-get.md
- user-list-mailfolders.md
- user-post-mailfolders.md
- mailfolder-list-childfolders.md
- mailfolder-post-childfolders.md

angelgolfer-ms:
> Hi @angelgolfer-ms:
> Thanks for your review, I've addressed the comments that you left previously.
> One question, shouldn't the title of pages mailfolder-post-childfolders.md (both beta/v1.0) be "Create ChildFolder" instead of "Create MailFolder". We already have user-post-mailfolders.md which tells us how to "Create MailFolder".

If we look at only the topic titling rules used by the generator, "create" operation topics follow the "Create {resource name}" rule. So "Create mailFolder" would be what's generated in this case. (Also documented [here](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103176/ToC-and-topic-title).)

On the other hand, I understand from the perspectives of SEO and docs.microsoft.com-generated warning, having two or more topics in the same version ("doc set") of the docs with the same topic title is not ideal. Since the warning will turn into a build error at some point, we should mitigate it by retitling. I'd suggest "Create child folder" for both the TOC and topic title.

api-reference/v1.0/api/mailfolder-post-childfolders.md
```diff
@@ -43,7 +45,7 @@ Specify the parent folder in the query URL as a folder ID, or a well-known folde
 
 ## Request body
 
-In the request body, provide a JSON object with the following parameters. **displayName** is the only writable property for a
+In the request body, provide a JSON object with the following parameters. **displayName** and **isHidden** are the only writable property for a
 [mailFolder](../resources/mailfolder.md) object.
```
angelgolfer-ms, line 49: Line 54:
I think we should add a table row for the isHidden body parameter, right?
In the interest of time (got a few things to do before Build), I'm going to add that table row, in both this topic and the beta version of the topic.

PR #12745, opened 2021-05-17
UPN support in urls

anandab-msft:
users will be able to use UPN as well in URL in addition to user id

### DIFF
--- a/changelog/Microsoft.Teams.Core.json
+++ b/changelog/Microsoft.Teams.Core.json
@@ -1,6 +1,56 @@
 {
 
   "changelog": [
+        {
+      "ChangeList": [
+        {
+          "Id": "185e5343-c8a2-4716-995f-1427973c31ca",
+          "ApiChange": "Parameter",
+          "ChangedApiName": "UserPrincipalName",
+          "ChangeType": "Addition",
+          "Description": "UserPrincipalName can be used in the request URL for [looking up chats](https://docs.microsoft.com/en-us/graph/api/chat-list?view=graph-rest-beta) of a user.",
+          "Target": "chat"
+        },
+        {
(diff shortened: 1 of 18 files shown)


changelog/Microsoft.Teams.Core.json
```diff
@@ -1,6 +1,56 @@
 {
 
   "changelog": [
+        {
+      "ChangeList": [
+        {
+          "Id": "185e5343-c8a2-4716-995f-1427973c31ca",
+          "ApiChange": "Parameter",
```
shantanu1singh, line 8: So, apparently, all the IDs here need to be the same. I was told so on my email provisioning PR. The doc generator tool currently creates unique IDs so I was under the assumption that they are supposed to be unique.

shantanu1singh (approved):
Just need to fix ChangeList IDs. LGTM otherwise.

changelog/Microsoft.Teams.Core.json
```diff
@@ -1,6 +1,56 @@
 {
 
   "changelog": [
+        {
+      "ChangeList": [
+        {
+          "Id": "185e5343-c8a2-4716-995f-1427973c31ca",
```
Lauragra, line 7: Can we combine these into a single ChangeList item that says "Added support for user principal names in requests to the following methods: [method links]." That will make the entry more concise and reduce repetition.

anandab-msft: the target of each changelist item is different, so i have created multiple items.

anandab-msft: @Lauragra - since the target of each changelist item is different, so i have created multiple items. is there a way to merge all of them?

Lauragra (approved):

PR #12755, opened 2021-05-17
Update the Universal Print pricing header

braedenp-msft:
We've decided not to charge for API usage. Updating the docs pages to reflect this decision.

The string has been fully reviewed by CELA and Marketing.

### DIFF
--- a/api-reference/includes/cloudprinting-pricing-disclaimer.md
+++ b/api-reference/includes/cloudprinting-pricing-disclaimer.md
@@ -1,10 +1,10 @@
 ---
 author: nilakhan
 ms.topic: include
-ms.date: 03/02/2021
+ms.date: 05/17/2021
 ---
 
 <!-- markdownlint-disable MD041-->
 
 > [!IMPORTANT]
-> Microsoft is temporarily offering usage of the cloud printing APIs to manage Universal Print at no charge. Microsoft expects to charge for the use of some or all of these APIs in the future. Microsoft will provide advanced notice of pricing changes.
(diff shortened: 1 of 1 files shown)


braedenp-msft:
@Lauragra this is ready for review and publishing 😊

api-reference/includes/cloudprinting-pricing-disclaimer.md
```diff
@@ -1,10 +1,10 @@
...
-ms.date: 03/02/2021
+ms.date: 05/17/2021
 ---
 
 <!-- markdownlint-disable MD041-->
 
 > [!IMPORTANT]
-> Microsoft is temporarily offering usage of the cloud printing APIs to manage Universal Print at no charge. Microsoft expects to charge for the use of some or all of these APIs in the future. Microsoft will provide advanced notice of pricing changes.
\ No newline at end of file
+> Microsoft is offering usage of the cloud printing APIs to manage Universal Print at no charge and there are no foreseeable plans to charge for the use of these APIs.
```
Lauragra, line 10: Saying that there are no "foreseeable plans" is ambiguous because it implies that this is not foreseen but it can still happen. Is that what you mean to imply? Otherwise, I would suggest revising to say "and we do not plan to charge for use of these APIs."

braedenp-msft: Yep, that's what we mean to imply. The legal team wants to leave some room for flexibility.

Lauragra (approved):

PR #12832, opened 2021-05-21
Document advanced query capabilities on Azure AD objects

FaithOmbongi:


api-reference/beta/resources/user.md
```diff
@@ -155,19 +155,19 @@ This resource supports:
...
-| creationType | String | Indicates whether the user account was created as a regular school or work account (`null`), an external account (`Invitation`), a local account for an Azure Active Directory B2C tenant (`LocalAccount`) or self-service sign-up using email verification (`EmailVerified`). <br><br>Returned only on `$select`. Read-only. |
+| aboutMe | String | A freeform text entry field for the user to describe themselves. <br>Returned only on `$select` on the Get user API (`GET /users/{ID}` or `/me`). |
+| accountEnabled | Boolean | `true` if the account is enabled; otherwise, `false`. This property is required when a user is created. <br><br>Supports `$filter` (`eq`,`in`) and `ne` operator only in eventual consistency. |
```
jasonjoh, line 159: What is "eventual consistency"? It's used repeatedly but I don't see anywhere in our docs where it's actually explained. The only thing I see is this in some docs:

```
| ConsistencyLevel | eventual. This header and `$count` are required when using the `$search`, `$filter`, `$orderby`, or OData cast query parameters. It uses an index that might not be up-to-date with recent changes to the object. |
```

Which I'm sure is what it refers to, but even then it seems very "hand-wavy".

FaithOmbongi: I've introduced it in the query parameters doc. We're trying to find a good way to 
1. Introduce why the **ConsistencyLevel:eventual** header and **$count=true** query string are sometimes required.
2. Document properties & parameters that work only when 1 is fulfilled.

Still a WIP and your guidance is highly welcome.

jasonjoh: Awesome. I don't have a good understanding of it myself so I'm not sure what to recommend. If it only applies to those query params, then it probably is OK to put in the query params doc. I'd put it in its own section with a heading so it could be deep-linked. Then you could convert those instances of `eventual consistency` to a link.

jasonjoh: Oh and include an example maybe that illustrated the impact of using that header. It sounds like you might be searching against old data?

Licantrop0: I don't think specifying "on the Get user API (`GET /users/{ID}` or `/me`)" is necessary.

api-reference/beta/resources/user.md
```diff
@@ -155,19 +155,19 @@ This resource supports:
...
-| createdDateTime | DateTimeOffset | The date and time the user was created. The value cannot be modified and is automatically populated when the entity is created. The DateTimeOffset type represents date and time information using ISO 8601 format and is always in UTC time. Property is nullable. A null value indicates that an accurate creation time couldn't be determined for the user. <br><br>Returned only on `$select`. Read-only. Supports `$filter` with the `eq`, `ne`, `le`, and `ge` operators. |
-| creationType | String | Indicates whether the user account was created as a regular school or work account (`null`), an external account (`Invitation`), a local account for an Azure Active Directory B2C tenant (`LocalAccount`) or self-service sign-up using email verification (`EmailVerified`). <br><br>Returned only on `$select`. Read-only. |
+| aboutMe | String | A freeform text entry field for the user to describe themselves. <br>Returned only on `$select` on the Get user API (`GET /users/{ID}` or `/me`). |
```
Licantrop0, line 158: I don't think specifying "on the Get user API (`GET /users/{ID}` or `/me`)" is necessary.

concepts/known-issues.md
```diff
@@ -391,15 +391,22 @@ Requesting objects using [Get directory objects from a list of IDs](/graph/api/d
...
   * Cross-workload filtering is not supported.
 * `$search`:
   * Full-text search is only available for a subset of entities such as messages.
   * Cross-workload searching is not supported.
+  * Searching is not supported on Azure AD B2C tenants.
+* `$count`:
+  * Not supported on Azure AD B2C tenants.
+  * If you're using OData Count (adding `$count=true` in the query string), the `@odata.count` property will be present only in the first page of the paged data.
+* In paging, any additional request headers such as the **ConsistencyLevel** header, are not included by default in subsequent page requests. If those headers need to be sent on subsequent requests, you must set them explicitly.
+* Query parameters specified in a request might fail silently. This can be true for unsupported query parameters as well as for unsupported combinations of query parameters.
```
Licantrop0, line 409: This is only true for AAD workload.

concepts/search-query-parameter.md
```diff
@@ -0,0 +1,169 @@
+---
```
Licantrop0, line 1: if we are going to make $search a separate page, maybe we should do it for all the other parameters for consistency?

Licantrop0 (changes requested):
this change requires a little more discussion. Let's sync.

concepts/search-query-parameter.md
FaithOmbongi: As agreed, only the $search and advanced query capabilities would be separate pages due to their many differences across resources.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,211 @@
...
+- Properties that are not listed here do not support `$filter`.
+
+> [!NOTE]
+>
+> 1. Properties with the same name across directory resources support the same `$filter` operators. For example, the **createdDateTime** property is available in **application**, **group**, **organization**, and **user** resources. It supports the `eq`, `ge`, and `le` operators by default and the `in` and `startsWith` operators only in advanced queries.
+> 2. Queries that are supported by default will also work in advanced queries.
+> 3. `NOT` and `ne` operators are only supported in advanced queries.
+>>
+
+| Property                      |  eq   |  ne   |  ge   |  le   |  in   | startsWith | endsWith | null value |
```
Licantrop0, line 82: @FaithOmbongi null value indexing is only supported with Advanced Queries.
for example, [this query](https://developer.microsoft.com/en-us/graph/graph-explorer?request=users%3F%24count%3Dtrue%26%24filter%3Dcountry%20eq%20null%26%24select%3Did%2CdisplayName%2Ccountry&method=GET&version=beta&GraphUrl=https://graph.microsoft.com) doesn't work without mezzo parameters.

We also need to find a way to indicate whenever the "null default value" is implemented (e.g. for accountEnabled all nulls fall back to 'false' value). Null indexing and Null default value are mutually exclusive and they can be represented in the same column.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,211 @@
...
+| :---------------------------- | :---: | :---: | :---: | :---: | :---: | :--------: | :------: | :--------: |
+| accountEnabled                |   ⚪   |   ⚫   |   ⚫   |   ⚫   |   ⚪   |            |          |            |
```
Licantrop0, line 84: using only colored dots is not a best practice for accessibility, can we find symbols that have different shapes? (also considering the 3rd one regarding null indexes / null default values.

Licantrop0 (changes requested):

concepts/aad-advanced-queries.md
FaithOmbongi: Let me revisit the accessibility guidelines. I see [Azure DB docs use ticks](https://azure.microsoft.com/en-us/product-categories/databases/).

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,231 @@
...
+- Properties that are not listed here do not support `$filter`.
+
+> [!NOTE]
+>
+> 1. Properties with the same name across directory resources support the same `$filter` operators. For example, the **createdDateTime** property is available in **application**, **group**, **organization**, and **user** resources. It supports the `eq`, `ge`, and `le` operators by default and the `in` and `ne` operators only in advanced queries.
+> 2. Queries that are supported by default will also work in advanced queries.
+> 3. `NOT` and `ne` operators are supported only in advanced queries.
+>>
+
+| Property                      | eq | ne | NOT | ge | le | in | startsWith | endsWith | null values |
```
Licantrop0, line 83: ne and NOT are **always** supported with advanced queries, probably not useful to call out in the table since we told in the line above?

Licantrop0, line 83: we need to find a way to clarify when `$filter=property eq null` is supported
and when `$filter=property eq false` actually means false or null.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,231 @@
...
+| proxyAddresses                | ⚪  | ⚫  |     | ⚪  | ⚪  |    | ⚪          |          | ⚫           |
+| showInAddressList             | ⚫  | ⚫  |     | ⚫  | ⚫  | ⚫  |            |          |             |
+| signInActivity                | ⚪  | ⚫  |     | ⚪  | ⚪  |    |            |          |             |
+| state                         | ⚪  | ⚫  |     | ⚪  | ⚪  | ⚪  | ⚪          |          | ⚫           |
+| streetAddress                 | ⚫  | ⚫  |     | ⚫  | ⚫  | ⚫  | ⚫          |          | ⚫           |
+| surname                       | ⚪  | ⚫  |     | ⚪  | ⚪  | ⚪  | ⚪          |          | ⚫           |
+| usageLocation                 | ⚪  | ⚫  |     | ⚪  | ⚪  | ⚪  | ⚪          |          |             |
+| userPrincipalName             | ⚪  | ⚫  |     |    | ⚪  | ⚪  | ⚪          | ⚫        |             |
+| userType                      | ⚪  |    |     |    |    | ⚪  |            | ⚫        | ⚫           |
+
```
Licantrop0, line 134: we should add a similar table for the orderBy support

Licantrop0 (changes requested):

concepts/aad-advanced-queries.md
FaithOmbongi: I think we can add that to the individual property descriptions.

concepts/aad-advanced-queries.md
FaithOmbongi: We can discuss this further.

FaithOmbongi: Anything to simplify the table and group shared scenarios would be good.

1. But, do all properties support NOT and ne?
2. Or, can we say a common rule is that all properties that support eq also support NOT and ne? With a note that ne cannot be used inside any/all clauses (hit this error when I tried)

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,184 @@
...
+title: "Advanced query capabilities on Azure AD directory objects"
+description: "Azure AD directory objects support advanced queries based on the OData query parameters."
+author: "FaithOmbongi"
+localization_priority: Priority
+ms.custom: graphiamtop20, scenarios:getting-started
+---
+
+# Advanced query capabilities on Azure AD directory objects
+
+As Azure AD continues to deliver more capabilities and improvements in stability, availability, and performance, Microsoft Graph also continues to evolve and scale to efficiently access the data. One way is through Microsoft Graph's increasing support for advanced query capabilities on various Azure AD objects and their properties. For example, the addition of **Not** (`NOT`), **Not equals** (`ne`), and **Ends with** (`endsWith`) operators on the `$filter` query parameter in October 2020.
```
davidmu1, line 11: "in October 2020" isn't needed.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,184 @@
...
+| Use of `$filter` with `NOT` operator                                                                                                                           | [GET](https://developer.microsoft.com/graph/graph-explorer?request=users%3F%24filter%3DNOT%20startsWith(displayName%2C%20'Luca')%26%24count%3Dtrue&method=GET&version=beta&GraphUrl=https://graph.microsoft.com&headers=W3sibmFtZSI6IkNvbnNpc3RlbmN5TGV2ZWwiLCJ2YWx1ZSI6ImV2ZW50dWFsIn1d) `../users?$filter=NOT startsWith(displayName, 'Luca')&$count=true`                                                                         |
+
+> [!NOTE]
+> These advanced query capabilities are not available in Azure AD B2C tenants.
+
+## Support for filter on properties of Azure AD directory objects
+
+Properties of directory objects behave differently in their support for query parameters. The following table summarizes support for `$filter` operators by properties on the [users](/graph/api/resources/user) directory object. Properties for other directory objects will be added soon.
```
davidmu1, line 66: We shouldn't include text that implies promises of future functionality: "Properties for other directory objects will be added soon."

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,184 @@
...
+
+- ![Requires Advanced Query Parameters.](/graph/images/advanced-query-parameters/AQP.png) The property supports `$filter` with the operator by default.
```
davidmu1, line 68: Since advanced query parameters isn't a brand name, I don't think the term should be capitalized.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,184 @@
...
+- ![Works by default. Does not require Advanced Query Parameters.](/graph/images/advanced-query-parameters/default.png) The specific `$filter` operator requires *Advanced Query Parameters*:
+  - `ConsistencyLevel=eventual` header
+  - `$count=true` query string
+- Blank cells indicate that the property does not support the use of `$filter` with the operator.
+- The null value column indicates that the property is filterable on `null` values.
+- Properties that are not listed here do not support `$filter`.
+
+> [!NOTE]
```
davidmu1, line 76: I don't think the note format is necessary. And the bullet items are not steps to complete so shouldn't be numbered.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,184 @@
...
+> 3. `NOT` and `ne` operators are supported only in advanced queries.
+>>
+
+
+[!INCLUDE [filter-directory-objects](../includes/filter-directory-objects.md)]
+
+
+## Error handling for advanced queries on directory objects
+
+Counting Directory Objects is only supported using the advanced queries parameters.
```
davidmu1, line 90:
```suggestion
Counting directory objects is only supported using the advanced queries parameters.  
```

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,184 @@
...
+>>
+
+
+[!INCLUDE [filter-directory-objects](../includes/filter-directory-objects.md)]
+
+
+## Error handling for advanced queries on directory objects
+
+Counting Directory Objects is only supported using the advanced queries parameters.  
+If the `ConsistencyLevel=eventual` header is not specified, the request will return an error in case raw count is used (`/$count` query segment) or silently ignoring the count parameter in case inline count is used (`$count=true` query parameter).
```
davidmu1, line 91:
```suggestion
If the `ConsistencyLevel=eventual` header is not specified, the request returns an error when the $count segment is used or silently ignores the $count query parameter (?$count=true) if it's used.
```

concepts/query-parameters.md
```diff
@@ -79,7 +80,12 @@ GET https://graph.microsoft.com/v1.0/me/messages?$filter=subject eq 'let''s meet
 
 ## count parameter
 
-Use the `$count` query parameter to include a count of the total number of items in a collection alongside the page of data values returned from Microsoft Graph. 
+Use the `$count` query parameter to include a count of the total number of items in a collection alongside the page of data values returned from Microsoft Graph.
+
+> [!NOTE]
+> `$count` can also be used as a [URL segment](#other-odata-url-capabilities) to retrieve the integer total of the collection. On resources that derive from [directoryObject](/graph/api/resources/directoryobject), is it only supported in an [advanced query](/graph/aad-advanced-queries). See [Advanced query capabilities in Azure AD directory objects](/graph/aad-advanced-queries).
+>
+> Use of `$count` is currently not supported in Azure AD B2C tenants.
```
davidmu1, line 88:
```suggestion
> Use of `$count` is not supported in Azure AD B2C tenants.
```

concepts/query-parameters.md
```diff
@@ -422,12 +297,16 @@ GET  https://graph.microsoft.com/v1.0/me/events?$orderby=createdDateTime&$skip=2
...
+>
+> The **ConsistencyLevel** header required for advanced queries against directory objects is not included by default in subsequent page requests. It must be set explicitly in subsequent pages.
 
 ## skipToken parameter
 
 Some requests return multiple pages of data, either due to server-side paging or due to the use of the [`$top`](#top-parameter) parameter to limit the page size of the response. Many Microsoft Graph APIs use the `skipToken` query parameter to reference subsequent pages of the result.  
 The `$skiptoken` parameter contains an opaque token that references the next page of results and is returned in the URL provided in the `@odata.nextLink` property in the response. To learn more, see [Paging](./paging.md).
-> **Note:** if you're using OData Count (adding `$count=true` in the querystring), the `@odata.count` property will be present only in the first page.
+> **Note:** If you're using OData Count (adding `$count=true` in the query string) for queries against directory objects, the `@odata.count` property will be present only in the first page.
```
davidmu1, line 307:
```suggestion
> **Note:** If you're using OData Count (adding `$count=true` in the query string) for queries against directory objects, the `@odata.count` property is present only in the first page.
```

concepts/aad-advanced-queries.md
FaithOmbongi: Well noted. The patterns have also grown from the origin 2.

concepts/aad-advanced-queries.md
FaithOmbongi: Added change in local branch.

concepts/aad-advanced-queries.md
FaithOmbongi: Was thinking of this in the context of "improvements, evolution, and scalability" pointed out in the 1st sentence.

api-reference/beta/resources/user.md
FaithOmbongi: Moving the explanation of consistency levels in Azure AD to another PR/topic. This PR is now only for advanced query capabilities on directory objects.

concepts/aad-advanced-queries.md
```diff
@@ -0,0 +1,205 @@
...
+
+The Microsoft Graph query engine uses an index store to fulfill query requests. To add support for additional query capabilities on some properties, these properties are now indexed in a separate server. This separate indexing allows Azure AD to increase support and improve the performance of the query requests. However, these advanced query capabilities are not available by default but, the requestor must also set the **ConsistencyLevel** header with the value `eventual` *and*, with the exception of `$search`, use the `$count` query parameter (either as a [URL segment](/graph/query-parameters#other-odata-url-capabilities) or `$count=true` query string). The **ConsistencyLevel** header and `$count` are referred to as *advanced query parameters*.
+
+For example, if you wish to retrieve only inactive user accounts, you can run either of these queries that use the `$filter` query parameter.
+
+1. Use the `$filter` query parameter with the `eq` operator. This request will work by default, that is, the request does not require the advanced query parameters.
```
davidmu1, line 17: These should not be numbered. Numbering implies a sequential order of operation.

davidmu1 (approved):

PR #12844, opened 2021-05-24
Fixed user flow language page delete example

jkdouglas:
Changed b2cUserFlows to b2xUserFlows


jkdouglas:
Just a simple change in the example where I had a typo referencing b2cUserFlows instead of b2xUserFlows, which are what is in v1.

PR #12943, opened 2021-06-01
Remove documentation and pointer for assignment resource patch, now t…

sharaaU7:
Now that PATCH assignment resource action support has been removed (405) for graph callers (with effect from 5/17), there is need to cleanup related documentation.


Lauragra (approved):

ssriniva-msft (approved):
Is there a changelog update needed for this?

PR #12978, opened 2021-06-02
Update chatmessage.md

JarbasHorst:
Replaced non-English word


Lauragra:
@JarbasHorst , good catch, also note that we avoid "etc." for globalization - https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/etc

PR #13007, opened 2021-06-04
transitive count API

davidmu1:


api-reference/beta/api/user-get-transitivereports.md
```diff
@@ -0,0 +1,97 @@
...
+| Application | User.Read, User.Read.All, Directory.Read.All |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+GET /users/{id}/transitiveReports
+```
+## Optional query parameters
+
+This method requires the `$count` query parameter. For more information, see [OData Query Parameters](/graph/query-parameters).
```
Licantrop0, line 36: as the comment above, it requires the $count segment, not the query parameter `$count=true`, as listing the reports is not supported, only counting them.
That's why in the sample above we need to add /$count otherwise it will not work.

Licantrop0 (changes requested):

api-reference/beta/api/orgcontact-get-transitivereports.md
```diff
@@ -0,0 +1,99 @@
...
+GET https://graph.microsoft.com/beta/contacts/45b7d2e7-b882-4a80-ba97-10b7a63b8fa4/transitiveReports/$count
+```
+
+#### Response
+
+The following is an example of the response.
+<!-- {
+  "blockType": "response",
+  "truncated": true,
+  "@odata.type": "microsoft.graph.transitivereports",
```
millicentachieng, line 77: Remove this line since response has no body to validate against the odata type

api-reference/beta/api/orgcontact-get-transitivereports.md
```diff
@@ -0,0 +1,99 @@
...
+  "isCollection": true
+} -->
+```http
+HTTP/1.1 200 OK
+Content-type: text/plain
+```
+
+5
```
millicentachieng, line 85: Is the number 5 expected here?

api-reference/beta/api/orgcontact-get-transitivereports.md
davidmu1: Removed.

api-reference/beta/api/orgcontact-get-transitivereports.md
davidmu1: Yes.

api-reference/beta/api/orgcontact-get-transitivereports.md
```diff
@@ -0,0 +1,96 @@
...
+## HTTP request
+
+<!-- { "blockType": "ignored" } -->
+```http
+GET /contacts/{id}/transitiveReports/$count
+```
+
+## Optional query parameters
+
+This method does not support the use of query parameters, but does require the `$count` query segment.
```
Licantrop0, line 38: it does support $filter=accountEnabled eq 'value'
Only accountEnabled property is supported on filter clause and the two supported values are 'true' or 'false'
See https://microsoftgraph.visualstudio.com/_git/onboarding/pullrequest/4612?_a=files&path=%2Freviews%2F4379-add-gettransitivereportscount-api-for-a-given-user%2Fapi.md

api-reference/beta/api/user-get-transitivereports.md
```diff
@@ -0,0 +1,94 @@
...
+| Application | User.Read, User.Read.All, Directory.Read.All |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+GET /users/{id}/transitiveReports/$count
+```
+## Optional query parameters
+
+This method does not support the use of query parameters, but does require the `$count` query segment.
```
Licantrop0, line 36: see my comment above

Licantrop0 (changes requested):

api-reference/beta/api/orgcontact-get-transitivereports.md
millicentachieng: What's the number representing? It looks misplaced in the docs.

image

davidmu1: It's representing the response. It doesn't work to put it in the code block.

Licantrop0 (approved):

api-reference/beta/api/orgcontact-get-transitivereports.md
```diff
@@ -0,0 +1,96 @@
...
+
+Do not supply a request body for this method.
+
+## Response
+
+If successful, this method returns a `200 OK` response code and a count of transitive reports for an organizational contact in the response body.
+
+## Examples
+
+### Get the count of transitive reports for an organizational contact
```
Lauragra, line 57: There's only one example so we don't need this H3, just make the Request and Response headings H3s.

api-reference/beta/api/user-get-transitivereports.md
```diff
@@ -0,0 +1,94 @@
+---
+title: "user: Get transitiveReports"
```
Lauragra, line 2: The title and H1 should be "Get transitiveReports for a user".

api-reference/beta/api/user-get-transitivereports.md
```diff
@@ -0,0 +1,94 @@
...
+
+Do not supply a request body for this method.
+
+## Response
+
+If successful, this method returns a `200 OK` response code and a count of transitive reports for the user in the response body.
+
+## Examples
+
+### Get the count of transitive reports for a user
```
Lauragra, line 55: Remove and elevate Request and Response to H3s.

Lauragra (approved):

PR #13017, opened 2021-06-07
Update worksheetprotection-unprotect.md

grangeryy:
Correct the wrong documentation on worksheetprotection. Excel Graph doesn't support password yet.

### DIFF
--- a/api-reference/beta/api/worksheetprotection-unprotect.md
+++ b/api-reference/beta/api/worksheetprotection-unprotect.md
@@ -37,11 +37,7 @@ POST /me/drive/root:/{item-path}:/workbook/worksheets/{id|name}/protection/unpro
 | Workbook-Session-Id  | Workbook session Id that determines if changes are persisted or not. Optional.|
 
 ## Request body
-In the request body, provide a JSON object with the following parameters.
-
-| Parameter	   | Type	|Description|
-|:---------------|:--------|:----------|
-|password|string|Optional. sheet protection password.|
+Do not supply a request body for this method.
 
 ## Response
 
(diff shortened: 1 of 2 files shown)


api-reference/beta/api/worksheetprotection-unprotect.md
```diff
@@ -60,10 +56,8 @@ Here is an example of the request.
 ```http
 POST https://graph.microsoft.com/beta/me/drive/items/{id}/workbook/worksheets/{id|name}/protection/unprotect
 Content-type: application/json
-Content-length: 34
-
+workbook-session-id: {session-id}
 {
-  "password": "password-value"
```
Lauragra, line 66: Do you still want to have an empty JSON body here? Also do we need the Content-type header? Note that it isn't listed in the request headers section.

api-reference/v1.0/api/worksheetprotection-unprotect.md
```diff
@@ -58,10 +54,8 @@ Here is an example of the request.
 ```http
 POST https://graph.microsoft.com/v1.0/me/drive/items/{id}/workbook/worksheets/{id|name}/protection/unprotect
 Content-type: application/json
-Content-length: 34
-
+workbook-session-id: {session-id}
 {
-  "password": "password-value"
```
Lauragra, line 64: The same questions apply to this topic re: the JSON body and content-type header.

api-reference/beta/api/worksheetprotection-unprotect.md
grangeryy: Thanks for pointing out. I'll remove the empty body. Also, no need for content-type header.

grangeryy: sessionid header is optional.

PR #13035, opened 2021-06-08
Remove SharePoint/ODB limitation for drive item delta on subfolders.

ocreeva:
SharePoint/ODB delta requests now support queries against a subfolder. This change removes the documented limitation stating that SharePoint/ODB delta requests must target the root folder.


chackman (approved):

JeremyKelley (approved):
Looks good

PR #13039, opened 2021-06-08
Update Drive and Sites Search App Permission Not Supported

BarryShehadeh:


chackman (approved):

api-reference/beta/api/site-search.md
```diff
@@ -82,7 +84,10 @@ Content-type: application/json
   ]
 }
 ```
->**Note:** The only property that works for sorting is **createdDateTime**. The search filter is a free text search that uses multiple properties when retrieving the search results.
+
+## Remarks
```
Lauragra, line 88: Let's avoid introducing this non-standard H2 into the topics, although it does exist in other topics. Instead, please add this information to the summary section, in a new paragraph after the introductory statement.

api-reference/v1.0/api/site-search.md
```diff
@@ -82,7 +84,10 @@ Content-type: application/json
   ]
 }
 ```
->**Note:** The only property that works for sorting is **createdDateTime**. The search filter is a free text search that uses multiple properties when retrieving the search results.
+
+## Remarks
```
Lauragra, line 88: Please do the same in this topic - remove and add to the summary section.

api-reference/beta/api/site-search.md
```diff
@@ -27,6 +27,8 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account) | Not supported.
 |Application                            | Sites.Read.All, Sites.ReadWrite.All
 
+>**Note:** This method does not support the Sites.Selected application permission.
+
 ## HTTP request
```
Lauragra, line 32: Global: Please be sure to set the HTTP request section to ignore so that we don't generate SDK snippets for that syntax. We only want the snippets in the examples section. This topic seems to be missing an example. The HTTP request section is not intended to be a complete example, just to show the syntax of the call. So you'll want to add an example request and response.

api-reference/v1.0/api/site-search.md
```diff
@@ -25,6 +25,8 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account) | Not supported.
 |Application                            | Sites.Read.All, Sites.ReadWrite.All
 
+>**Note:** This method does not support the Sites.Selected application permission.
+
```
Lauragra, line 29: Same comment here regarding the HTTP request section - please ignore and add an Examples section with a request and response example.

api-reference/beta/api/site-search.md
BarryShehadeh: Yeah, it's an old style one. Updated.

PR #13130, opened 2021-06-15
Update passwordprofile.md

bpcastilho:
JSON representation missing:
"passwordProfile": { ... }

This is misleading the customers on what to pass in the body request.

### DIFF
--- a/api-reference/v1.0/resources/passwordprofile.md
+++ b/api-reference/v1.0/resources/passwordprofile.md
@@ -35,9 +35,11 @@ Here is a JSON representation of the resource
 
 ```json
 {
-  "forceChangePasswordNextSignIn": true,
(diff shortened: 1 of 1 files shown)


FaithOmbongi:
Hi @bpcastilho ,

Thank you for raising this PR. 

However, the JSON representation indicates only the properties and OData type of the subject resource. **passwordProfile** is not a property of the [passwordProfile ](https://docs.microsoft.com/en-us/graph/api/resources/passwordprofile?view=graph-rest-1.0) complex resource, rather a property of the [User ](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties)resource.

FaithOmbongi:
Hi @bpcastilho: I believe we have now resolved this issue by adding an example to the Update article. Closing this PR.

PR #13140, opened 2021-06-15
Create link

k-tsoi:
update "item" to "driveItem"


angelgolfer-ms (changes requested):
Excellent, @k-tsoi ! If you don't mind, can you please bold the instances of driveItem in that topic? FYI this is from the Microsoft conventional docs style guide - on first reference of an API element (e.g., resource, method, property), we should provide a link (resource or method), and bold subsequent references. Styling API elements differently from normal text makes the content more scannable, making it easier to identify API elements from natural language in the content. An exception in place in the Graph docs is in the Properties or Methods table, where possibly many terms would be styled bold, which can have the adverse glaring effect. I looked at the preview of this topic on the staging site. With relatively few terms that need to be bolded in your tables, bolding is the better approach.

angelgolfer-ms (approved):
Thanks, @k-tsoi , looks good!

PR #12958, opened 2021-06-02
add all ipp values of PrinterProcessingStateDetails

tomsato-ms:
Adding all 825 printer-state-reasons values to Graph
https://www.iana.org/assignments/ipp-registrations/ipp-registrations.xhtml

PR #13174, opened 2021-06-17
2021 06 17 tomsato printer state reasons

tomsato-ms:
Adding all 825 printer-state-reasons values to Graph
https://www.iana.org/assignments/ipp-registrations/ipp-registrations.xhtml

Copying PR from https://github.com/microsoftgraph/microsoft-graph-docs/pull/12958 against main branch


api-reference/beta/resources/printerstatus.md
```diff
@@ -70,7 +70,798 @@ Represents the processing status of the printer, including any errors.
 |developerLow|33|The device is low on developer.|
 |developerEmpty|34|The device is out of developer.|
 |interpreterResourceUnavailable|35|An interpreter resource is unavailable (e.g. font, form).|
-|unknownFutureValue|36|Evolvable enumeration sentinel value. Do not use.|
+|alertRemovalOfBinaryChangeEntry|36|This is a standard IPP Printer Attribute value described in PWG5100.9.|
```
Lauragra, line 73: This is such a long list of state values that all have identical descriptions. If there isn't a resource that we can link to for the user to get the list of state values, I would recommend that we apply Option 3 for documenting enums, although we do that only rarely: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103158/Enumerations?anchor=third-option%3A-in-a-separate-dedicated-enum-topic. That avoids overburdening this topic with the list of values.

tomsato-ms: Sounds good. I've updated the Enum to be in its own file, and added a link from the original file.

Lauragra (approved):

api-reference/beta/resources/printerprocessingstatedetail.md
```diff
@@ -0,0 +1,847 @@
...
+description: Represents the processing status detail of the printer.
+author: tomsato-ms
+ms.localizationpriority: medium
+ms.prod: universal-print
+doc_type: enumPageType
+---
+
+# printerProcessingStateDetail enum type
+
+Namespace: microsoft.graph
```
nilakhan, line 12: @tomsato-ms please mention that to use the enum values beyond unknownFutureValue, client needs to pass this header : "Prefer" : "include-unknown-enum-members" 

For v1.0 as well.

nilakhan (approved):

api-reference/beta/resources/printerprocessingstatedetail.md
Tomothehiro: Yup. Added these lines to both files
Please include the `Prefer: include-unknown-enum-members` header to explicitly request for enum values beyond 'unknownFutureValue'.
https://docs.microsoft.com/en-us/graph/best-practices-concept#handling-future-members-in-evolvable-enumerations

nilakhan (approved):

PR #13193, opened 2021-06-18
Adding a new element to the Learning paths section in the hub page

lramosvea:
Adding the Build your first custom connector quickstart to the Learning paths section in the Graph hub page

### DIFF
--- a/concepts/index.yml
+++ b/concepts/index.yml
@@ -112,6 +112,8 @@ additionalContent:
               url: https://docs.microsoft.com/learn/paths/m365-msgraph-scenarios
             - text: Develop apps with Microsoft Graph Toolkit
               url: https://docs.microsoft.com/learn/paths/m365-msgraph-toolkit
+            - text: Build your first custom connector
+              url: https://docs.microsoft.com/en-us/graph/connecting-external-content-build-quickstart              
         # Card
         - title: How-to
           links:
--- a/concepts/index.yml
+++ b/concepts/index.yml
@@ -115,15 +115,15 @@ additionalContent:
         # Card
         - title: How-to
           links:
-            - text: Connectors':' Build your first custom connector
+            - text: Connectors - Build your first custom connector
               url: https://docs.microsoft.com/en-us/graph/connecting-external-content-build-quickstart  
-            - text: Microsoft Search":" Search content in OneDrive and Sharepoint
+            - text: Microsoft Search - Search content in OneDrive and Sharepoint
               url: https://docs.microsoft.com/graph/search-concept-files
-            - text: Identity':' Manage access to resources
+            - text: Identity - Manage access to resources
               url: https://docs.microsoft.com/en-us/graph/tutorial-access-package-api
-            - text: Outloo':' Attach large files to a message or event
+            - text: Outlook - Attach large files to a message or event
               url: https://docs.microsoft.com/graph/outlook-large-attachments
-            - text: Outlook':' Get MIME content of a message
+            - text: Outlook - Get MIME content of a message
               url: https://docs.microsoft.com/graph/outlook-get-mime-message
         # Card
         - title: Resources and community
--- a/concepts/index.yml
+++ b/concepts/index.yml
@@ -115,20 +115,16 @@ additionalContent:
         # Card
         - title: How-to
           links:
-            - text: Build your first custom connector
+            - text: Connectors: Build your first custom connector
               url: https://docs.microsoft.com/en-us/graph/connecting-external-content-build-quickstart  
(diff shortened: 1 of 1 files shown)


lramosvea:
Hi @CelesteDG, could you please review these changes? I only added a new element to the Learning paths section. Please set to Ready to merge if you approve these changes. Thanks!

concepts/index.yml
```diff
@@ -112,6 +112,8 @@ additionalContent:
               url: https://docs.microsoft.com/learn/paths/m365-msgraph-scenarios
             - text: Develop apps with Microsoft Graph Toolkit
               url: https://docs.microsoft.com/learn/paths/m365-msgraph-toolkit
+            - text: Build your first custom connector
```
Lauragra, line 115: Actually this one is a "tutorial", not technically a learning path. Are we okay with mixing content types? Maybe we rename this tile "Tutorials and learning paths"? But that might then lead us to linking to all the tutorials here.

lramosvea: I was discussing this with @angelgolfer-ms, and even though this quick start has a different format or structure, we were thinking it still can be considered as a learning path. We are not planning to link to all the tutorials but we wanted to link to this quick start to promote it since we received feedback saying that users are not aware of its existence.

Lauragra: I think that "Learning path" has a specific meaning and that this can be misleading/confusing. If we want to add the link, are we okay with renaming the section "Learning paths and tutorials"?

lramosvea: @Lauragra do you think we may need to have a separate box just for this type of quickstarts or tutorials? I see this other quickstart: https://docs.microsoft.com/en-us/graph/connecting-external-content-build-quickstart?view=graph-rest-1.0, although it is In Preview content.

concepts/index.yml
```diff
@@ -111,12 +111,12 @@ additionalContent:
...
             - text: Develop apps with Microsoft Graph Toolkit
-              url: https://docs.microsoft.com/learn/paths/m365-msgraph-toolkit
-            - text: Build your first custom connector
-              url: https://docs.microsoft.com/en-us/graph/connecting-external-content-build-quickstart              
+              url: https://docs.microsoft.com/learn/paths/m365-msgraph-toolkit             
         # Card
         - title: How-to
           links:
+            - text: Build your first custom connector
+              url: https://docs.microsoft.com/en-us/graph/connecting-external-content-build-quickstart
```
angelgolfer-ms, line 119: @lramosvea , shall we try giving the list of items under "How-to" more context? Right now, because each of those items is a phrase describing a task or objective, customers cannot scan to get any idea what each item is, they have to read the entire item to decide if it's interesting or relevant to them.
I suggest formatting those list items this way:

Connectors: Build your first custom connector
Identity: Manage access to resources
Microsoft Search: Search content in OneDrive and SharePoint
Outlook: Attach large files to a message or event
Outlook: Get MIME content of a message

The prefix (product or service names) for each item can serve as a cue to make scanning easier. Also, these items themselves don't make a lot of sense for newer customers unfamiliar with each of the spaces. The prefix would hopefully help cue them in, or out, and skip to the next one.

And after we update the Tutorial landing page to include the (informal) tutorials in Identity, and how-to topics in the conceptual section under a few workloads, then we can add a trailing item "More tutorials..." pointing to the Tutorials landing page.

lramosvea: @angelgolfer-ms took me a while to figure out how to be able to use colons in the hub page but changes have been applied now.

angelgolfer-ms: Thanks for the persistence. IMHO the updated list in the How-to card now looks a bit easier to make sense of. We can't change the fact that it's at the bottom of the hub page,... over time, perhaps we'd become more informed to move things around.

concepts/index.yml
```diff
@@ -111,22 +111,20 @@ additionalContent:
...
-            - text: Attach large files to a message or event
+            - text: "Identity: Manage access to resources"
+              url: https://docs.microsoft.com/en-us/graph/tutorial-access-package-api
+            - text: "Outlook: Attach large files to a message or event"
               url: https://docs.microsoft.com/graph/outlook-large-attachments
-            - text: Get MIME content of a message
+            - text: "Outlook: Get MIME content of a message"
               url: https://docs.microsoft.com/graph/outlook-get-mime-message
-            - text: Schedule a Windows update deployment
-              url: https://docs.microsoft.com/graph/windowsupdates-schedule-deployment
```
Lauragra, line 129: Were we going to add a "More" link here that links to the Tutorials landing page?

Lauragra (approved):

PR #13204, opened 2021-06-21
Added Limitations description in event-delta CalendarView and None, N…

Swathi-Birlangi:
…otResponded attribute description in ResponseStatus


Swathi-Birlangi:
@angelgolfer-ms please help with your review comments

api-reference/v1.0/api/event-delta.md
```diff
@@ -57,7 +63,7 @@ includes the encoded, desired parameters.
 | $skiptoken | string | A [state token](/graph/delta-query-overview) returned in the `nextLink` URL of the previous **delta** function call, indicating there are further changes to be tracked in the same calendar view. |
 
 When you do a delta query on a calendar view, expect to get all the properties you'd normally get from 
-a `GET /calendarview` request. `$select` is not supported in this case. 
+a `GET /calendarview` request. `$select`, `$orderby`,`$filter`,`$search`,`$expand`is not supported in this case.
```
angelgolfer-ms, line 66: FYI in the original line 65, the main purpose is to say a delta function call on a calendarView returns the same properties as a GET /calendarView request, and that you can't use a $select to just get a subset of those properties.

So if we want to add a few more OData query parameters that are not supported, we should add that as a separate item so as not to convolute the original idea in line 65.

angelgolfer-ms: Also want to call out to you that now there is more info we want to bring up about OData query parameters, please refer to this section in the CDK about how to format that info: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103170/Parameters?anchor=documenting-odata-system-query-options
Specifically you can see I am going to add a separate H3 custom section, under the existing H2 Query parameters section, to separate the request parameters from the OData parameters for greater clarify. 
Note - all custom sections should only be H3s under the corresponding H2 sections. H2 sections are reserved for the formal reference topic template.

api-reference/v1.0/api/event-delta.md
```diff
@@ -21,6 +21,12 @@ you can query for incremental changes in that calender view. This allows you to
 a local store of a user's events in the primary calendar, without having to fetch all the events of that calendar 
 from the server every time.
 
+## Limitations
```
angelgolfer-ms, line 24: Line 24: 
Note - all custom sections should only be H3s under the corresponding H2 sections. H2 sections are reserved for the formal reference topic template.

api-reference/v1.0/resources/responsestatus.md
```diff
@@ -17,7 +17,7 @@ The response status of a meeting request.
 
 | Property | Type           | Description |
 |:---------|:---------------|:------------|
-| response | responseType   | The response type. The possible values are: `None`, `Organizer`, `TentativelyAccepted`, `Accepted`, `Declined`, `NotResponded`.
+| response | responseType   | The response type. The possible values are: `None`, `Organizer`, `TentativelyAccepted`, `Accepted`, `Declined`, `NotResponded`. <br><br>To differentiate between `None` and `NotResponded`, use an attendee, Alex, as an example. Alex hasn't responded to a meeting request (which programmatically is an [event](event.md)). An operation that returns the properties of the event, such as a [GET event](../api/event-get.md) operation, called on behalf of Alex returns `NotResponded` for Alex in the **attendees** and **responseStatus** properties. For the same event, a GET /event operation called on behalf of any other attendee or the organizer would return `None` in the **attendees** property for Alex.
```
angelgolfer-ms, line 20: @Harini84 , I rewrote the explanation trying to make it a bit more understandable. Can you please confirm if I got it correct?

angelgolfer-ms: @Swathi-Birlangi , once @Harini84 confirms, you will please follow up with updating the beta version of responsestatus.md as well.

Harini84: @angelgolfer-ms, yes this explanation is correct. 

+ @Swathi-Birlangi Swathi-Birlangi

api-reference/beta/api/event-delta.md
```diff
@@ -13,7 +13,7 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Get a set of [event](../resources/event.md) resources that have been added, deleted, or updated in one or more calendars. 
+Get a set of events that have been added, deleted, or updated in a **calendarView** (a range of events defined by start and end dates) of the user's primary calendar.
```
angelgolfer-ms, line 17: @Swathi-Birlangi ,
Line 16, why did you decide to remove the link to the event resource?
There's a discrepancy on the live site between the beta topic (that has the link), and the v1 topic (that doesn't have the link).
FYI in general, when we refer to an API element (e.g., resource, method) that has its own topic, we should provide the link on first reference in the article. In this case in line 16, "event" is the first reference of that resource in the topic, so there should be a link.
To save time, I am going to fix this in both the v1 and beta event-delta.md topics.

angelgolfer-ms: Also in line 16:
FYI the original description is specific to beta and doesn't apply to v1. Looks like you copied the v1 description over to beta. By doing that, you actually narrowed the functionality in beta back down to only calendarView. 
In beta, delta works for calendarView and for events (possibly unbounded by any date range) in any user calendar(s).

angelgolfer-ms: To save time, I am fixing the issues in both v1 and beta topics.
Perhaps in the future, you can focus on what exactly needs to be changed for the corresponding user story?

api-reference/v1.0/api/event-delta.md
```diff
@@ -13,6 +13,8 @@ Namespace: microsoft.graph
 
 Get a set of events that have been added, deleted, or updated in a **calendarView** (a range of events defined by start and end dates) of the user's primary calendar.
 
+You can get specific types of these incremental changes in the events in all the calendars of a mailbox or in a specific calendar, or in an event collection of a **calendarView** (range of events defined by start and end dates) of a calendar. The calendar can be the default calendar or some other specified calendar of the user's. In the case of getting incremental changes on **calendarView**, the calendar can be a group calendar as well.
```
angelgolfer-ms, line 16: Line 16 (in the v1 topic):
This change is also inappropriate. Looks like you copied the paragraph from the beta event-delta.md topic?
This change widened the functionality of the delta function in v1, to not only support calendarView but also events in any user calendar.
This is not in the scope of the corresponding user story.

In the interest of time, I am fixing this on your behalf.

PR #13269, opened 2021-06-24
Surface the changePassword action on /me

### LINKED ISSUE #7376: Change Password is not documented
If we refer to Graph SDK it has changePassword Method which makes a post-call. graph.microsoft.com/v1.0/me/changePassword.

C#-&gt;await graphClient.Me.ChangePassword("currentPassword", "newPassword").Request().PostAsync();

I am able to find the document here ? 

---
#### Document Details

* ID: e8446021-6786-c54c-066f-d99ba4644312
* Version Independent ID: 34e2e6e4-578a-9309-ebcf-94b640dc4e4c
* Content: [user resource type - Microsoft Graph beta](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-beta)
* Content Source: [api-reference/beta/resources/user.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/resources/user.md)
* Product: **microsoft-identity-platform**
* GitHub Login: @dkershaw10

FaithOmbongi:
Fixes #7376


FaithOmbongi:
@dkershaw10 I've tested only with Directory.AccessAsUser.All permission and with Global Administrator role.

api-reference/beta/api/user-changepassword.md
```diff
@@ -0,0 +1,94 @@
...
+
+Update the password of the currently signed-in user.
+
+## Permissions
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | User.ReadWrite.All, Directory.Read.All, User.ReadWrite.All, Directory.ReadWrite.All, Directory.AccessAsUser.All    |
```
dkershaw10, line 24: You have User.ReadWrite.All twice
Directory.Read.All clearly won't work.  I have a suspicion that Directory.ReadWrite.All won't either.

Be good if User.ReadWrite.All works here, but it's probably only Directory.AccessAsUser.All.

Regular users **must** be able to use this API.  No special privileged role is required.

api-reference/beta/api/user-changepassword.md
```diff
@@ -0,0 +1,94 @@
...
+doc_type: apiPageType
+---
+
+# user: changePassword
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Update the password of the currently signed-in user.
```
dkershaw10, line 16: Maybe we can be more specific.

Allows the signed-in user to change their own password.  Does not allow resetting passwords of other users.  For password reset see xxx.

We can decide if we want to link to the password reset example you are creating.

api-reference/beta/api/user-changepassword.md
```diff
@@ -0,0 +1,94 @@
...
+|Permission type      | Permissions (from least to most privileged)              |
+|:--------------------|:---------------------------------------------------------|
+|Delegated (work or school account) | User.ReadWrite.All, Directory.Read.All, User.ReadWrite.All, Directory.ReadWrite.All, Directory.AccessAsUser.All    |
+|Delegated (personal Microsoft account) | Not supported.    |
+|Application | Not supported. |
+
+## HTTP request
+<!-- { "blockType": "ignored" } -->
+```http
+POST /me/changePassword
```
dkershaw10, line 31: In theory POST /users/{id}/changePassword (where {id} = "the signed-in user's ID") would also work.  Not sure that we should show it though.

api-reference/beta/api/user-changepassword.md
```diff
@@ -0,0 +1,94 @@
+---
+title: "user: changePassword"
```
dkershaw10, line 2: I believe this API is in v1.0 too.  Will you update those files too?

api-reference/beta/api/user-changepassword.md
FaithOmbongi: Maybe not? We mostly use /me to indicate requests specific to the signed-in user. Removes any ambiguity.

api-reference/beta/api/user-changepassword.md
FaithOmbongi: > I have a suspicion that Directory.ReadWrite.All won't either. Be good if User.ReadWrite.All works here, but it's probably only Directory.AccessAsUser.All.

You're right. Got a 403 Forbidden with Directory.ReadWrite.All and User.ReadWrite.All permissions. Only Directory.AccessAsUser.All works.

changelog/Microsoft.DirectoryServices.json
```diff
@@ -1,5 +1,41 @@
 {
   "changelog": [
+    {
+      "ChangeList": [
```
dkershaw10, line 4: QQ:  This is capturing, in the changelog, when we added the documentation.
Should we come clean and actually provide the date when the API actually went live in beta and v1.0?

@FaithOmbongi Let me know and I can dig up the dates.

api-reference/beta/api/user-changepassword.md
```diff
@@ -0,0 +1,94 @@
...
+doc_type: apiPageType
+---
+
+# user: changePassword
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Update the password of the currently signed-in user. This API does not allow resetting passwords of other users. To reset the password of other users, update the **passwordProfile** property of the [user](../resources/user.md) resource. See an example to [Update the passwordProfile of a user](../api/user-update.md#example-3-update-the-passwordprofile-of-a-user).
```
dkershaw10, line 16: I realize that you are keeping to the doc style etc, but for me, it doesn't quite hammer home what this API does.
"Update the password of the currently signed-in user" still sounds a bit like an admin action.  I want to make it clear that that this API lets any user change their own password.

changelog/Microsoft.DirectoryServices.json
FaithOmbongi: @dkershaw10 , no need to. The changelog shows the date we publish the documentation.

api-reference/beta/api/user-changepassword.md
FaithOmbongi: Noted. Let me tweak it and share for review.

api-reference/v1.0/toc.yml
```diff
@@ -217,6 +217,8 @@ items:
           href: api/user-update.md
         - name: Delete user
           href: api/user-delete.md
+        - name: Change password
```
dkershaw10, line 220: I don't think this is incorrect, but it brings up an inconsistency on what we show under the Users' User node.
This is the ONLY method (other than delta and regular HTTP verbs) that we show under this node.

We put a bunch of other actions under the "Directory objects" node (which is under Users).  That *might* make sense if the actions there applied ALSO to other directory objects.  But a number of those actions ONLY apply to users.

Then when I compare to the Group node (under Groups) all the methods/actions are listed under Group.  There is no "directory objects" node under Groups, like there is for Users.

This seems pretty inconsistent.

And when I look at Service principal it has a Membership node which has 4 of the same actions :)

dkershaw10 (approved):
Left a comment about the TOC that you could fix now or later.  The rest looks good to me.

api-reference/v1.0/toc.yml
FaithOmbongi: Let me look at these findings more keenly then get back to you on this. I'll also check in with Laura & David with both findings.

> Then when I compare to the Group node (under Groups) all the methods/actions are listed under Group. There is no "directory objects" node under Groups, like there is for Users.

We've applied this concept for (i believe) all APIs in the Identity and access node.

api-reference/beta/api/user-changepassword.md
```diff
@@ -0,0 +1,97 @@
...
+doc_type: apiPageType
+---
+
+# user: changePassword
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Use this API to update your own password. Any user can update their own password without belonging to any administrator role.
```
Lauragra, line 16: We generally refer to the developer as "you", so would it be appropriate to revise this to say "Update a user's password", or "Enable the user to update their password"? Also, we generally begin the summary description of an API topic with an imperative verb.

changelog/Microsoft.DirectoryServices.json
```diff
@@ -1,5 +1,41 @@
...
+      { 
+        "Id": "e114cd2d-6d20-498d-a61e-623baee7e0f1",
+        "ApiChange": "Resource",
```
Lauragra, line 7: The ApiChange value here should be Method.

api-reference/beta/api/user-changepassword.md
FaithOmbongi: Changing to "_Update your own password_".

changelog/Microsoft.DirectoryServices.json
FaithOmbongi: Thanks for catching that.

api-reference/beta/api/user-changepassword.md
Lauragra: I'm still unclear about the "your own". Can we change that to one of the suggestions, like "Enable the user to update their password"? By "your own", we don't mean the developer of the app. Let's preserve "you" for referring to the developer target audience.

FaithOmbongi: Updated.

Lauragra (approved):

PR #13271, opened 2021-06-24
Add example to update passwordProfile of a user

FaithOmbongi:
Related to #13130 .
Adding an example to guide customers on how to reset a user password using the passwordProfile proeprty.

### DIFF
--- a/api-reference/beta/api/user-update.md
+++ b/api-reference/beta/api/user-update.md
@@ -205,11 +205,9 @@ HTTP/1.1 204 No Content
 
 ### Example 3: Update the passwordProfile of a user to reset their password
 
-The following is an example of a request that resets the password of another user.
+The following example shows a request that resets the password of another user.
 
 #### Request
-
-The following example shows a request.
 
 <!-- {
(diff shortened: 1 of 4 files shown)


bpcastilho (approved):
It's great to have a sample.

api-reference/beta/api/user-update.md
```diff
@@ -203,6 +203,37 @@ The following example shows the response.
 HTTP/1.1 204 No Content
 ```
 
+### Example 3: Update the passwordProfile of a user
```
dkershaw10, line 206: While this title is strictly correct, it's not scenario based.  Can we change this or add a sentence before the request to something like "Reset the password of another user"

dkershaw10 (approved):
Feel free to make a change or ignore my comment.  Either way, I'm happy with this improvement.

api-reference/beta/api/user-update.md
FaithOmbongi: Totally acceptable & done.

PR #13286, opened 2021-06-25
Make one correction for our doc: permissions does not support MSA acc…

payiAzure:
Make a correction in our documentation.
Permission type in the following is not supported in our API.
      Delegated (personal Microsoft account) | Not supported.


angelgolfer-ms (approved):
Makes sense, thanks for the updates.

PR #13299, opened 2021-06-27
Simplier examples

AntoineJo:
The provided example were too complex, it makes people think they need to have a managementScript item and running state to use it, while a managedDevice object is enough


AntoineJo:
up, @Lauragra do you know who can approve or deny these updates ?

Lauragra:
> up, @Lauragra do you know who can approve or deny these updates ?

Intune has a different doc generation process so before we can accept PRs we need to update the source that generates the content. I'll follow up internally with Erik, the new owner.

PR #13325, opened 2021-06-29
[Doc bug][API changes made]

v-abirade:


Lauragra (changes requested):
Please update the examples in this file to match the updated syntax in the HTTP request section, and please make the same changes to the comparable file in the /beta folder.

PR #13351, opened 2021-07-01
Update intune-deviceconfig-emailsyncschedule.md

JarbasHorst:
Added beta disclaimer


Lauragra:
Intune has a different doc generation process. They should be using the include file; I'll follow up offline about getting that updated globally.

PR #13352, opened 2021-07-01
Windows updates: Add known issue on accessing and updating deployment audiences

Alice-at-Microsoft:
APIs for **deploymentAudience** resources do not work if the parent **deployment** resource was created via Intune, rather than directly via Graph APIs.


isvargasmsft (approved):
Thank you very much for your work on this PR. 

@Lauragra, this one is ready for final review.

PR #13364, opened 2021-07-03
AppPublisher: Removing hidden APIs

kanchm:
Removing hidden APIs that are documented.


kanchm:
Removing the PR. Schema is being fixed to match docs.

PR #13367, opened 2021-07-03
IdentityProtectionServices: IdentityProtection singleton is hidden in v1.0

kanchm:
IdentityProtectionServices: IdentityProtection singleton is hidden in v1.0

- Removing docs that describe the hidden element.


Lauragra:
@kanchm , can you please also remove the Changelog entry associated with adding these APIs to v1.0, so that we don't have broken links on the site? Thanks.

PR #11591, opened 2021-02-09
Update auth-limit-mailbox-access.md

michelderooij:
Important to mention that for Exchange Web Services, full_access_as_app API permission also honors the Application Access Policy.

PR #13395, opened 2021-07-07
Incorporate external and internal feedback to add EWS support for application access policy.

angelgolfer-ms:


angelgolfer-ms:
Made content clearer as initiated by PR #11591 .

PR #13414, opened 2021-07-08
Update with activated roles details.

MatthewJDavis:
Add details about only returning activated roles, same as in [v1 documentation](https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-docs/main/api-reference/v1.0/api/directoryrole-list.md) and still relevant to the beta endpoint.


Lauragra:
Looks good - thanks!

PR #13423, opened 2021-07-08
FIX: example of adding multiple members to a group in one request (for C#)

mbratukha:
an array has to be set as value for the additional parameter `members@odata.bind`


FaithOmbongi:
Thank you @mbratukha for raising this PR. However, the code snippets are auto-generated from the SDK and I don't believe you can update them via a PR. @andrueastman , kindly advise.

mbratukha:
@FaithOmbongi, thanks for checking my PR.

@FaithOmbongi, @andrueastman Would it be possible to make the auto-generated example look correctly?

andrueastman:
We will need to fix this in the snippet generator. I have created this issue on the generator to track this. https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/631
 
We can close this PR so that it may be resolved from the generator as these changes will be overwritten if it is merged as is.

PR #13506, opened 2021-07-14
add guidance on retrieving access package resource scopes

markwahl-msft:
Feedback on recent issue was that the ability to expand the resource scope relationship on an access package resource was not obvious.  Adding more mentions to this.

### DIFF
--- a/api-reference/beta/resources/accesspackageresourcescope.md
+++ b/api-reference/beta/resources/accesspackageresourcescope.md
@@ -15,7 +15,9 @@ Namespace: microsoft.graph
 
 In [Azure AD entitlement management](entitlementmanagement-root.md), an access package resource scope is a reference to a scope within a resource, for those resources that have multiple scopes.
 
-You can determine the access package resource scope, for a resource that has already been added to an access package, by using [list accessPackageResourceRoleScopes](../api/accesspackage-list-accesspackageresourcerolescopes.md) to return a collection of [accessPackageResourceRoleScope](accesspackageresourcerolescope.md) objects.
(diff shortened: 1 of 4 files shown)


api-reference/beta/resources/accesspackageresourcescope.md
```diff
@@ -15,7 +15,9 @@ Namespace: microsoft.graph
...
 
-You can determine the access package resource scope, for a resource that has already been added to an access package, by using [list accessPackageResourceRoleScopes](../api/accesspackage-list-accesspackageresourcerolescopes.md) to return a collection of [accessPackageResourceRoleScope](accesspackageresourcerolescope.md) objects.
+You can determine the access package resource scope, for a resource which has roles already added to an access package, by using [list accessPackageResourceRoleScopes](../api/accesspackage-list-accesspackageresourcerolescopes.md) to return a collection of [accessPackageResourceRoleScope](accesspackageresourcerolescope.md) objects.
+
+If the resource is in an access package catalog but has not yet had its roles added to an access package, you can determine the access package resource scope by using [list accessPackageResources](../api/accesspackagecatalog-list-accesspackageresources.md) and including `$expand=accessPackageResourceScopes,accessPackageResourceEnvironment` in the query.
```
FaithOmbongi, line 20: Proposing you add this as an example in the api topic.

FaithOmbongi (approved):
- Left a comment. 
- ELM APIs also seem to support many OData query parameters but the resource pages don't indicate this. I would recommend indicating these supported query parameters in both the property descriptions & relationships. Please see the [User](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0) resource topic for reference.

api-reference/beta/resources/accesspackageresourcescope.md
markwahl-msft: Added filter and expand to the API example.

PR #13527, opened 2021-07-15
Update sdks-overview.md

maparaschivei:
Add reference to PowerShell SDK


maparaschivei:
I matched the original source file link syntax for consistency, however,  I can remove the locale code and use a relative link if preferred (as per warnings above).

maparaschivei:
Thank you for fixing the link paths @Lauragra !

PR #13558, opened 2021-07-19
Update toc.yml

JarbasHorst:
Added Resource-specific permission grants as a toc entry.
In beta, the resource file is available in our navigation. However, the v1.0 resource file isn't listed in the toc.


Lauragra:
Hi @JarbasHorst , no hurry at all but when you get a chance, can you also expose the method topics in the TOC under the resource topic? In beta, we need to expose the three methods and in v1.0 there's just one method.

JarbasHorst:
Added changes to the BETA toc via PR #13641.

PR #13581, opened 2021-07-20
Update conversationthread-get.md

ArunMSFT:
Update the API response with right default fields


Lauragra:
@ArunMSFT , please update the beta version of this topic as well. Thanks.

PR #13605, opened 2021-07-21
Added new string for new detection

vinayakdshenoy:
Added a string for "anomalous token" sign-in risk detection


vinayakdshenoy:
Barbara is doing a pull right now.

PR #13687, opened 2021-07-29
Intune 2107 beta

davidmu1:


Lauragra:
@davidmu1 , this PR also includes the file with the merge conflict notations. Also I'm seeing some files with the description "Not yet documented", is that something that we can take care of before merging?

Issue #13704, opened 2021-07-31
generate new page to get a single outcome


cristobal-buenrostro:
Create new page to document GetOutcome endpoint.

GET /education/classes/{​​classId}​​/assignments/{​​assignmentId}​​/submissions/{​​submissionId}​​/outcomes/{​​outcomeId}​​

GET Outcomes, If an invalid Outcome Id is specified, a 404 error is returned
(e.g. { "error": { "code": "20241", "message": "Entity not found. Outcome id: 05d0f76c-1dfa-4442-926c-1b094828b505" }}

Currently only exists GetOutcomes.
List outcomes - Microsoft Graph v1.0 | Microsoft Docs
The doc page must be created in beta and v1.0 paths

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8f3d0cd6-a23a-77bb-178e-7e27e1f90901
* Version Independent ID: 0f798da9-c9a5-c18b-88e9-931315ee7a77
* Content: [List outcomes - Microsoft Graph v1.0](https://docs.microsoft.com/en-us/graph/api/educationsubmission-list-outcomes?view=graph-rest-1.0&tabs=http)
* Content Source: [api-reference/v1.0/api/educationsubmission-list-outcomes.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/main/api-reference/v1.0/api/educationsubmission-list-outcomes.md)
* Product: **education**
* Technology: **microsoft-graph**
* GitHub Login: @sharad-sharma-msft
* Microsoft Alias: **MSGraphDocsVteam**

PR #13706, opened 2021-07-31
july21 train - adding new getOutcome page

cristobal-buenrostro:
Issue created for this PR
#13704

### DIFF
--- a/api-reference/v1.0/api/educationoutcome-get.md
+++ b/api-reference/v1.0/api/educationoutcome-get.md
@@ -1,0 +1,117 @@
+---
+title: "Get educationOutcome"
+description: "Retrieve the properties of a specific outcome."
+author: "cristobal-buenrostro"
+localization_priority: Normal
+ms.prod: "education"
+doc_type: apiPageType
+---
+
+# Get educationOutcome
+
+Namespace: microsoft.graph
+
+Retrieves an [educationOutcome](../resources/educationoutcome.md) object.
+
+## Permissions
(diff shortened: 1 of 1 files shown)


api-reference/beta/toc.yml
```diff
@@ -10025,6 +10025,8 @@ items:
           href: api/educationuser-delta.md
         - name: Create rubric
           href: api/educationuser-post-rubrics.md
+        - name: List assignments
```
Lauragra, line 10028: The beta TOC doesn't appear to match the v1.0 TOC. Can you please double check? We should list the CRUD operations first followed by other methods. So I would example to see List, Create, Get, Update, Delete in this TOC.

cristobal-buenrostro: DONE, order has been changed according to v1.0, I see v1 has the operations listed as you mentioned.
I noticed that 4 pages are not available in beta; @sharad-sharma-msft do you remember if there is any special reason about that? or can I include them in beta too? Let me know please.
        - name: List users
          href: api/educationuser-list.md
        - name: Create user
          href: api/educationuser-post.md
        - name: Get directory user
          href: api/educationuser-get-user.md
        - name: Get changes to users
          href: api/educationuser-delta.md
        - name: List taught classes
          href: api/educationuser-list-taughtclasses.md

cristobal-buenrostro: I got clarification on those pages, does not belong to assignments.

PR #13709, opened 2021-08-02
revert #13634

FaithOmbongi:
Reverting to allow the product team more time to investigate whether this permission is truly required or it's a product bug.


CelesteDG:
Updated branch to sync with latest changes in main. Merging to main now.

PR #13756, opened 2021-08-05
Updating Connection API conceptual topic for MS Graph connectors

mecampos:


mecampos:
09/13/2021 Checked this PR, still pending comments from reviewers.

PR #13759, opened 2021-08-05
[Teams] Add toc entries under resource-specific permission grants

jessiechan27:


AkJo (approved):

api-reference/beta/toc.yml
```diff
@@ -14250,6 +14250,13 @@ items:
           href: api/onenote-post-notebooks.md
       - name: Resource-specific permission grants
         href: resources/resourcespecificpermissiongrant.md
+        items:
+        - name: List permission grants of team
```
Lauragra, line 14254: For the TOC entries, can we streamline a bit, i.e. "List for team", and so on? Also, I see that we surface the API under the Team resource in the TOC as well, should we do the same for the group and chat resources?

jessiechan27: Sure, I can shorten them

For your second question, it's already under the chat resource. As for the group resource, we want people to use the team permission grants API instead, so it's probably best to leave it out of the toc

Lauragra (approved):

PR #13779, opened 2021-08-09
updates to Advanced Queries:

Licantrop0:
- specifying limitation of $expand
- updated property table, splitted by object
- removed ne/not/in column as they are always supported
- updated property format to help developers query complex properties
- removed onPremisesUserPrincipalName from the list (bug in RDS)
- updated iconography
- fixed markdown editor suggestions


concepts/aad-advanced-queries.md
```diff
@@ -60,31 +63,36 @@ The following table lists query scenarios on directory objects that are supporte
...
 
-+ Unless otherwise indicated, properties with the same name across directory resources support the same `$filter` operators. For example, the **createdDateTime** property is available in **application**, **group**, **organization**, and **user** resources. It supports the `eq`, `ge`, and `le` operators by default and the `in`, `ne`, and `NOT` operators only in advanced queries.
-+ The `endsWith` operator is supported only on **mail** and **userPrincipalName** properties.
-+ Queries that are supported by default will also work in advanced queries.
-+ The `NOT` and `ne` negation operators are supported only in advanced queries. 
-  + All properties that support the `eq` operator also support the `ne` or `NOT` operators.
-  + The `ne` operator negates where the `eq` operator would otherwise evaluate to `true`. For queries that use the `any` lambda operator, use the `NOT` operator. See [Filter using lambda operators](/graph/query-parameters#filter-using-lambda-operators).
++ Queries that are supported by default will also work in advanced queries, but the response will be eventually consistent.
++ The `in` operator is supported by default whenever `eq` operator is supported.
```
FaithOmbongi, line 76: Previous tests show this is not always true. For example, approximateLastSignInDateTime, classification, deviceId, customerId, disabledByMicrosoftStatus etc. Could you confirm?
```suggestion
+ The `in` operator is supported by default whenever `eq` operator is supported by default.
```

FaithOmbongi (commented):
Hi Luca,

Comments about using the **Not supported** icon.
- If we have this icon, why have blank cells at all? I think for the customer, _Not valid_ and _not supported_ translate to the same thing - "it won't work, so don't use it".
- Doesn't what's not supported stand out too much, even more than what is?

concepts/aad-advanced-queries.md
Licantrop0: I tried few of them and it looks like they are all working:

`https://graph.microsoft.com/beta/applications?$filter=disabledByMicrosoftStatus in ('DisabledDueToViolationOfServicesAgreement', 'NotDisabled')`

`https://graph.microsoft.com/beta/devices?$filter=approximateLastSignInDateTime in ('2021-01-02T12:00:00Z', '2021-02-02T12:00:00Z')`

`https://graph.microsoft.com/beta/groups?$filter=classification in ('Highly Confidential', 'Confidential')`

`https://graph.microsoft.com/beta/devices?$filter=deviceId in ('aa3674c6-2119-4222-aedc-aad2fc720ac9', 'aa3674c6-2119-4222-aedc-aad2fc720ac3')`

if you have a specific sample, it might be a bug.

Licantrop0:
> * If we have this icon, why have blank cells at all? I think for the customer, _Not valid_ and _not supported_ translate to the same thing - "it won't work, so don't use it".

@FaithOmbongi there is a significant difference between "what **could be supported** with advanced queries in the future" and what doesn't make sense at all as a query. We offer a link to provide this kind of feedback here:
image

> * Doesn't what's not supported stand out too much, even more than what is?
We should not be worried about highlighting the shortcoming of our APIs.

FaithOmbongi (approved):

PR #13794, opened 2021-08-10
Danielwood95resilience defaults

danielwood95:
Added resilience defaults session control for Conditional Access


FaithOmbongi (approved):

FaithOmbongi:
The PR's hitting an OPS build error. Should try to close and reopen later to resolve it.

FaithOmbongi:
Hi @danielwood95 , I see this change is now in the public schema. I'm adding the **Ready to merge** label so that the docs may also be updated. Feel free to remove the label if we still need to hold back the PR.

PR #13854, opened 2021-08-17
Remove public preview note for Graph connectors

BrianTJackett:
Graph connectors have exited public preview.  Removing note regarding this.


BrianTJackett:
@nmoreau If I'm not mistaken, Graph connectors are also searchable on the /v1.0 version of Search APIs.  Should we update sample queries on this page to reflect?  I haven't fully tested but quick check returned results when specifying "entitytypes" of externalItem as well as a queryString.

Lauragra (approved):

PR #13938, opened 2021-08-23
Create excel.md

grangeryy:
correct the link

### DIFF
--- a/api-reference/v1.0/resources/excel.md
+++ b/api-reference/v1.0/resources/excel.md
@@ -25,7 +25,7 @@ The Excel REST API supports only Office Open XML file formatted workbooks. The `
 
 ## Authorization and scopes
 
-You can use the [Azure AD v.2 endpoint](../index.md) to authenticate Excel APIs. All APIs require the `Authorization: Bearer {access-token}` HTTP header.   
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/resources/excel.md
```diff
@@ -25,7 +25,7 @@ The Excel REST API supports only Office Open XML file formatted workbooks. The `
 
 ## Authorization and scopes
 
-You can use the [Azure AD v.2 endpoint](../index.md) to authenticate Excel APIs. All APIs require the `Authorization: Bearer {access-token}` HTTP header.   
+You can use the [Azure AD v.2 endpoint](../permissions-reference.md) to authenticate Excel APIs. All APIs require the `Authorization: Bearer {access-token}` HTTP header.
```
Lauragra, line 28: This is a link to the permissions reference, is that what you meant to link to? Or this topic: https://docs.microsoft.com/en-us/graph/auth-register-app-v2?

Lauragra: Also I think the link will work if you use the following: /graph/auth-register-app-v2

grangeryy: The link you provided is what we need. Thanks! I've updated the PR

Lauragra (approved):

PR #13947, opened 2021-08-23
Adding warning about use of multiple Graph API for WUfB settings

Brenduns:
This note helps surface issues between Intune Graph API and WUfB-DS graph API which both manage the same settings but which are not necessarily aware of the settings edit sources.  This is a compliment to the warning found here: https://docs.microsoft.com/windows/deployment/update/deployment-service-overview#general

### DIFF
--- a/api-reference/beta/resources/windowsupdates-updates.md
+++ b/api-reference/beta/resources/windowsupdates-updates.md
@@ -44,3 +44,13 @@ The following is a JSON representation of the resource.
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Brenduns:
@Alice-at-Microsoft   Hi Alice, this is a PR to address a graph-api issue we discussed via email back in July. Looks like I failed to follow up at that time, or to get this assigned to you for review.  Can you take a look to see if we can get this added to docs?  Thanks!

PR #14020, opened 2021-08-27
Add changelog for app role assignment paging in beta

psignoret:


changelog/Microsoft.DirectoryServices.json
```diff
@@ -3,34 +3,34 @@
...
-          "Description": "Added support for paging for the **appRoleAssignments** navigation property on the [group](https://docs.microsoft.com/en-us/graph/api/resources/group.md) resource type. The **appRoleAssignments** navigation property returns app roles assigned directly to the group. Support for paging will be added to Microsoft Graph v1.0 starting in September 2021.",
+          "Description": "Added support for paging for the **appRoleAssignments** navigation property on the [group](https://docs.microsoft.com/en-us/graph/api/resources/group?view=graph-rest-beta) resource type. The **appRoleAssignments** navigation property returns app roles assigned directly to the group. Support for paging will be added to Microsoft Graph v1.0 starting in September 2021.",
           "Target": "group"
         }
       ],
       "Id": "479558a4-5491-4463-bf2f-4474590767db",
       "Cloud": "prd",
       "Version": "beta",
-      "CreatedDateTime": "2021-08-01T00:00:00Z",
+      "CreatedDateTime": "2021-08-27T00:00:00Z",
```
psignoret, line 33: @Lauragra I had back-dated it to earlier in August intentionally, to align with when the change was actually made. Is it more important to align with when the change happened, or when the update is submitted?

FaithOmbongi: Hi @psignoret , confirming that the date should/will be the publishing date.

FaithOmbongi (commented):
Hi @psignoret , regarding this statement in both the changelog and the What's new article:

> This change will be brought to v1.0 starting in September 2021

We try to avoid messaging of future intentions (we use the [external roadmap](https://www.microsoft.com/en-us/microsoft-365/roadmap?filters=Microsoft%20Graph) for that). I recommend that you remove the statement and instead raise a changelog for v1.0 when the feature is live in v1.0. 

cc @Lauragra

changelog/Microsoft.DirectoryServices.json
Lauragra: Yep, we generally use the date that we publish here; that way the entry shows up at the top of the Changelog so folks can easily identify what's new.

psignoret:
@FaithOmbongi Thanks, I've done as you suggested.

PR #14220, opened 2021-09-09
Fixed type DeviceManagementApps.ReadW.All > DeviceManagementApps.Read…

merill:
….All

Name of permission was incorrect.


jaiprakashmb:
This has been fixed. 
https://docs.microsoft.com/en-us/graph/api/intune-shared-deviceappmanagement-get?view=graph-rest-beta

PR #14223, opened 2021-09-10
Add example to transfer from group call

jackry:


changelog/Microsoft.Skype.Calling.json
```diff
@@ -1,5 +1,93 @@
...
+      ],
+      "Id": "db13b114-ad8b-42c0-95fd-d03d3b102d09",
+      "Cloud": "Prod",
+      "Version": "beta",
+      "CreatedDateTime": "2021-09-28T21:32:00.9962748Z",
+      "WorkloadArea": "Cloud communications",
+      "SubArea": "Call"
+    },
+    {
+      "ChangeList": [
```
Lauragra, line 22: I don't understand the two ChangeList entries here that mention deleting and adding a transfer method, for both beta and v1.0. I assume those were added in error - please delete them before we merge/publish. Thanks.

Lauragra:
@jackry , can you please take a look at the API Doctor errors on this PR? Thanks.

PR #14265, opened 2021-09-14
2021 09 14 code matchv1 cleanup

mjsantani:
Cleaned up some incorrectly added properties from V1. Graph team has provided the feature team an exception here to not go through standard deprecation.


andres-canello (approved):
Looks good

Lauragra (approved):

PR #14333, opened 2021-09-21
Update intune-devices-manageddevice-locatedevice.md

JarbasHorst:
Added correct reference to beta disclaimer.
Fixed punctuation.


Lauragra:
@Erikre , can you please follow up to get this fixed globally in the next drop? @JarbasHorst , FYI, the Intune docs are autogenerated so we can't accept PRs against them, the changes will get overwritten. So we can hopefully get this issue fixed at the source.

PR #14375, opened 2021-09-24
Updating validation token requirements and throttling guidelines

Jumaodhiss:
Updating Throttling guidelines in Webhooks and ValidationTokens requirements for delivering change notifications through Azure Event Hubs


concepts/webhooks.md
```diff
@@ -265,8 +265,6 @@ Your process should process every change notification it receives. The following
 
 1. Send a `202 - Accepted` status code in your response to Microsoft Graph. If Microsoft Graph doesn't receive a 2xx class code, it will try to publishing the change notification a number of times, for a period of about 4 hours; after that, the change notification will be dropped and won't be delivered.
```
robward-ms, line 266: Can we say "If your response time is greater than 3 seconds, consider sending..."

robward-ms, line 266: And here let's add "within 3 seconds, your subscription may be subject to throttling and Microsoft Graph will try...

concepts/webhooks.md
```diff
@@ -275,6 +273,20 @@ Your process should process every change notification it receives. The following
 
 Repeat for other change notifications in the request.
```
robward-ms, line 274: Just before this, can we add "4.) Send a '200 - Success' status code in your response to Microsoft Graph to acknowledge completion of the notification or a non 5xx status code to indicate an error. Microsoft Graph will try to publishing the change notification a number of times, for a period of about 4 hours; after that, the change notification will be dropped and won't be delivered."

If the client can process the message inside of 3 seconds, then sending a 200 at the end is better than sending a 202 in step 1 since a failure in the client, and no 200 response,

priya-hs (approved):

robward-ms (changes requested):

concepts/webhooks.md
```diff
@@ -265,8 +265,6 @@ Your process should process every change notification it receives. The following
```
robward-ms, line 265: Can we change the above to say:

"Your process should process every change notification it receives and send a 2xx class code. If Microsoft Graph doesn't receive a 2xx class code within 3 seconds, it will try to publishing the change notification a number of times, for a period of about 4 hours; after that, the change notification will be dropped and won't be delivered. If your process consistently does not respond within 3 seconds, your notification may be subject to throttling.

If your processing is expected to take more than 3 seconds, you should persist the notification, return a `202 - Accepted` status code in your response to Microsoft Graph, then process the notifications. If the notification is not persisted, return a 5xx class code to indicate an error so the notification will be retried.

If your processing is expected to take less than 3 seconds, you should process the notifications and return a `200 - Accepted` status code in your response to Microsoft Graph, then process the notifications. If the notification is not processes correctly, return a 5xx class code to indicate an error so the notification will be retried.

The following are the minimum tasks that your app must perform to process a change notification:"

concepts/webhooks.md
robward-ms: Actually the "If Microsoft Graph doesn't receive a 2xx class code, it will try to publishing the change notification a number of times, for a period of about 4 hours; after that, the change notification will be dropped..." is now covered above to just cut this portion of the sentence,

robward-ms (changes requested):

concepts/webhooks.md
```diff
@@ -263,9 +263,11 @@ When many changes occur, Microsoft Graph may send multiple notifications that co
...
+> If your processing is expected to take more than 3 seconds, you should persist the notification, return a 202 - Accepted status code in your response to Microsoft Graph, then process the notifications. If the notification is not persisted, return a 5xx class code to indicate an error so the notification will be retried.
+
+> If your processing is expected to take less than 3 seconds, you should process the notifications and return a 200 - Accepted status code in your response to Microsoft Graph. If the notification is not processes correctly, return a 5xx class code to indicate an error so the notification will be retried.
 
 1. Validate the `clientState` property. It must match the value originally submitted with the subscription creation request.
```
robward-ms, line 272: 2. instead of 1.

robward-ms (approved):

PR #14388, opened 2021-09-27
Corrected scope for client credentials provider

### LINKED ISSUE #14323: The examples listed here do not work
In my scenario, I need to use the [Client credentials provider](https://docs.microsoft.com/en-us/graph/sdks/choose-authentication-providers?tabs=CS#client-credentials-provider), but the configuration is different from the one listed.

Using the `var scopes = new[] { "User.Read.All" };` will cause an error:
> AADSTS1002012: The provided value for scope User.Read.All is not valid. Client credential flows must have a scope value with /.default suffixed to the resource identifier (application ID URI.

This error message suggest that we need to use `https://graph.microsoft.com/.default` but acquiring a token with this scope will cause unauthorized errors when trying to use the `GraphServiceClient`.

How can I acquire a token with the `User.Read.All` scope? The example here is wrong.

---
#### Document Details

* ID: 4537f77e-a8e8-e3f1-9cc4-2fd930db7ada
* Version Independent ID: 46396fa2-f7e2-322e-5e76-f9e2d17d1b48
* Content: [Choose a Microsoft Graph authentication provider - Microsoft Graph](https://docs.microsoft.com/en-us/graph/sdks/choose-authentication-providers?tabs=CS)
* Content Source: [concepts/sdks/choose-authentication-providers.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/main/concepts/sdks/choose-authentication-providers.md)
* GitHub Login: @MIchaelMainer

jasonjoh:
Fixes #14323

### DIFF
--- a/concepts/sdks/choose-authentication-providers.md
+++ b/concepts/sdks/choose-authentication-providers.md
@@ -168,7 +168,11 @@ The client credential flow enables service applications to run without user inte
 ### Using a client secret
 
 ```csharp
-var scopes = new[] { "User.Read.All" };
+// The client credentials flow requires that you request the
+// /.default scope, and preconfigure your permissions on the
+// app registration in Azure. An administrator must grant consent
+// to those permissions beforehand.
+var scopes = new[] { "https://graph.microsoft.com/.default" };
 
 // Multi-tenant apps can use "common",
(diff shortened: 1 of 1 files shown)


JohannesPertl:
Can you add the same scope clarification to the Java example?

PR #14428, opened 2021-09-29
add download URL annotation to driveItemVersion

zachzi:


api-reference/v1.0/resources/driveItemVersion.md
```diff
@@ -63,11 +66,20 @@ In the previous table, the examples use `/drive`, but there are many valid reque
 | **size**                 | Int64                                                | Indicates the size of the content stream for this version of the item.  |
 | **content**              | Stream                                               | The content stream for this version of the item.                        |
 
+## Instance Attributes
```
kevklam, line 69: should this be "Instance Annotations"?

zachzi: maybe, but I copied this from driveitem, so probably want to stay consistent?

k-tsoi: Other resources in the Graph docs also say "instance attributes" so this is likely correct

k-tsoi (approved):

PR #14435, opened 2021-09-29
Link for sharedDriveItem

learafa:
The sharedDriveItem resource link is added under Permissions.

### DIFF
--- a/api-reference/beta/toc.yml
+++ b/api-reference/beta/toc.yml
@@ -9911,6 +9911,8 @@ items:
       - name: Permissions
         href: resources/permission.md
         items:
+        - name: "Shared driveItem"
+          href: resources/shareddriveitem.md
         - name: Create sharing link
           href: api/driveitem-createlink.md
         - name: Use sharing links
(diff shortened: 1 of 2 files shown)


api-reference/beta/toc.yml
```diff
@@ -9911,6 +9911,8 @@ items:
       - name: Permissions
         href: resources/permission.md
         items:
+        - name: "Shared driveItem"
```
Lauragra, line 9914: Why are you adding a link to this resource under the Permissions resource? The child topics under Permissions should be methods that are related to the Permissions resource. Regarding the TOC entry itself, we use friendly names in the TOC so it would be "Shared drive item". Although it seems to work, I recommend removing the quotes for consistency in the TOC file.

There is an issue in the sharedDriveItem resource file itself - the link in the Methods table is named Get shared item. It targets a method topic named "Accessing shared DriveItems" with a TOC entry of "Access shared items". Can you please make all three of those consistent? And should "Shared drive item" be a peer of Permissions with "Get shared item" a child of it? One final note - resource names in Microsoft Graph are lower camel case, so DriveItem should be driveItem in all occurrences. Thanks.

PR #14436, opened 2021-09-29
Minor docs improvements

BarryShehadeh:


k-tsoi (approved):

api-reference/beta/resources/driverecipient.md
```diff
@@ -1,19 +1,31 @@
...
+# driveRecipient resource
 
 Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-The **DriveRecipient** resource represents a person, group, or other recipient to share with using the [invite](../api/driveitem-invite.md) action.
+Represents a person, group, or other recipient to share with using the [invite](../api/driveitem-invite.md) action.
+
+When using [invite](../api/driveitem-invite.md) to add permissions, the **driveRecipient** can specify **email**, **alias**, or **objectId**.
```
Lauragra, line 18: Actually are you sure that this is what you meant to say? It's not the recipient who would specify the alias, it would be the sender of the invitation, right? Please clarify in all instances. Please also note that in Microsoft Graph, resource names are lower camel case, so it should be driveRecipient. When referring to resources in text, the resource name should be bold.

api-reference/beta/resources/driverecipient.md
```diff
@@ -1,19 +1,31 @@
...
 ---
-# DriveRecipient resource
+# driveRecipient resource
 
 Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-The **DriveRecipient** resource represents a person, group, or other recipient to share with using the [invite](../api/driveitem-invite.md) action.
+Represents a person, group, or other recipient to share with using the [invite](../api/driveitem-invite.md) action.
```
Lauragra, line 16: Would it be more clear to update this description to say "Represents a person, group, or other recipient to share a drive item with using the invite action."?

BarryShehadeh: Yes, updated. @k-tsoi (the owner) let me know if you don't agree.

api-reference/beta/resources/driverecipient.md
BarryShehadeh: Ma'am I was only adding "multiple values are not accepted." to this 2017 document. Lol
Absolutely Laura. I updated the V1.0 version as well.

I think you got confused between the "Recipient", and the "driveRecipient" resource. Which is fair.
I updated it. Does it make more sense now?

Lauragra (approved):

PR #14473, opened 2021-10-04
Update SPO example in EM

markwahl-msft:
In entitlement management, it is necessary when creating a resource role for an access package, to ensure that the isRootScope is included if that was part of the scope definition of the resource.  Developers were omitting it as the example shown here did not have that value, since this example wasn't a root scope.


FaithOmbongi (approved):
Left a minor suggestion; otherwise, approved.

PR #14667, opened 2021-10-19
peopleInsights (Working With Changes)

PeterNjorogeMS:
New PR for Working with changes.


angelgolfer-ms:
> @angelgolfer-ms Could you, please, tell me why we changes method from GET insightSettings to List?

Hi @elmakhmu ,
Are you referring to, since we define plural "insightsSettings" and we don't define "insightSetting", why do the TOC and topic title use "list" and not "get" for the GET operation?
We use "list" in the TOC and topic title in 2 cases:
(A) To differentiate between getting a collection and getting a single instance, if both are supported.
As you probably know, the generator generates API stub topics based on the metadata. For the generator to generate CRUD operation topics in a single pass, it generates a topic for getting a collection (typically when parsing the CSDL  definition for a nav property) and uses a "list" in the topic title, and another topic for getting a single instance (by default, because that's not even indicated in the metadata) and uses a "get" in the topic title. The author simply eliminates what they don't support. 

(B) In cases where the supported API doesn't support both getting a collection and getting a single instance, then whether to use "list" or "get" is a natural language consideration targeting to help educate customers. In general, in the TOC we try to reduce confusion, because customers tend to ditch using the TOC when TOC terms are not intuitive and make them guess. In the case of insightsSettings - "setting" is countable, so when we have the term "insightsSettings" in plural, using "list" instead of "get" is actually consistent with how we use "list" vs "get" - list supports returning multiple objects, get returns a single object.
Sure you may counter by saying in the metadata, we only define "insightsSettings" and one instance is still "insightsSettings". That's when we apply common natural language principles.

elmakhmu:
@angelgolfer-ms many thanks for the thorough explanation! it became clearer

angelgolfer-ms (commented):
@PeterNjorogeMS , of the list I specified in my earlier [review submission](https://github.com/microsoftgraph/microsoft-graph-docs/pull/14667#pullrequestreview-789077920), FYI:
- Point#1 fixing the permissions for user-list-people.md - I fixed beta to make it the same as v1.
- Point#3 adding changelist records for this PR - I manually created the changelist records in Microsoft.OfficeGraph.json. Please take a look. If you had problem generating it, I encourage you to contact the owner team on Teams (as we talked about last week) to sort that out still. It might be some problem that can need to be fixed regardless, or it might be some misconception of how the generation works, and sorting that out will help you and other partners down the road. Please use a different branch for that purpose, so that it won't affect this the progress of this PR.
- In addition, I reassigned the new concept topic you created in this PR, and the existing user-list-people.md topics, to use the author value "anthona". In general, there should be just one value that each team specifies for author for all their topics, to strike a balance for maintenance reasons. 

Of that list, I left point#2 alone. I can have you demo what you mean. IMHO the way the related content is worded right now is not quite clear and can be misleading.

There are now 2 things that are related to docs that I can think of:
- Point#2, per my earlier [review submission](https://github.com/microsoftgraph/microsoft-graph-docs/pull/14667#pullrequestreview-789077920).
- New point#4: Have LauraGra set up redirects, because this PR renamed some existing files. For this time, I have identified them for you, so that you can do the similar in the future, if you do any file renaming in the docs.

As soon as your team publishes the API and CSDL, you can follow the [workflow](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103181/Content-workflow) and set the label "ready to merge", so that Lauragra can publish the PR.

PeterNjorogeMS:
@angelgolfer-ms 

I added the changelog to a new file Microsoft.People.json and added it's reference to Validation file. Luckily, someone showed us that the Changelog is automatically generated when a schema PR is created which made it easier. I was blocked on that for a while.

Regarding number two, this is from a security review, I had to make an edit to the insightsSettings file paragraph 2  to mention that when disabled the controls, it's not just ensuring that there's no visibility also that we don't calculate/generate the working-with list or lists for itemInsights. cc @elmakhmu 

I will work with @Lauragra on the links.

PR #14672, opened 2021-10-20
Convert Exchange and Microsoft Teams usage reports APIs to weak type

kszb:


changelog/Microsoft.O365Reporting.json
```diff
@@ -1,165 +1,495 @@
...
+          "ChangeType": "Deletion",
+          "Description": "Removed the **teamsUserActivityUserDetail** resource type.",
+          "Target": "teamsUserActivityUserDetail"
+        },
+		{
+		  "Id": "c21c4740-959e-4b70-bf0e-f3df67f38cea",
+		  "ApiChange": "Method",
+		  "ChangedApiName": "getemailactivityuserdetail",
+		  "ChangeType": "Change",
+		  "Description": "Update JSON return type of [getemailactivityuserdetail](https://docs.microsoft.com/en-us/graph/api/reportroot-getemailactivityuserdetail?view=graph-rest-beta) API to stream.",
```
JarbasHorst, line 146: GLOBAL: Please update the value of the "Description" property for all method entries in the changelog.

Please,

- replace Update with Updated
- add "the" before JSON return type
- add "the" before the API name, for example, "the [getEmailActivityUserDetail]"
- note that entity names in Microsoft Graph are lower camel case, so getemailactivityuserdetail in this case should be getEmailActivityUserDetail

Suggested revision,

- Updated the JSON return type of the [getEmailActivityUserDetail](https://docs.microsoft.com/en-us/graph/api/reportroot-getemailactivityuserdetail?view=graph-rest-beta) API to stream.

JarbasHorst (changes requested):

changelog/Microsoft.O365Reporting.json
kszb: Updated.

changelog/Microsoft.O365Reporting.json
```diff
@@ -1,165 +1,495 @@
...
+          "ApiChange": "Resource",
+          "ChangedApiName": "teamsUserActivityUserDetail",
+          "ChangeType": "Deletion",
+          "Description": "Removed the **teamsUserActivityUserDetail** resource type.",
+          "Target": "teamsUserActivityUserDetail"
+        },
+		{
+		  "Id": "c21c4740-959e-4b70-bf0e-f3df67f38cea",
+		  "ApiChange": "Method",
+		  "ChangedApiName": "getemailactivityuserdetail",
```
JarbasHorst, line 144: GLOBAL: Please update the values of the "ChangedApiName" and "Target" properties for all method entries in the changelog. 

Currently, we don't show the values of the properties ChangedApiName and Target in the UI but that can change in the future. Hence, please fix these values to be lower camel case, for example, getemailactivityuserdetail in this case should be getEmailActivityUserDetail. Suggested revision:

```json
"ChangedApiName": "getEmailActivityUserDetail",
...
"Target": "getEmailActivityUserDetail"
```

JarbasHorst (changes requested):

changelog/Microsoft.O365Reporting.json
kszb: Fixed

JarbasHorst:
@Lauragra this PR is ready for final review.

Lauragra (approved):

PR #14684, opened 2021-10-20
added two new values to lobbyBypassScope enum

frankpeng7:
added two new values to lobbyBypassScope enum


frankpeng7:
Hi @Lauragra this is ready to merge now and I've changed the tags. Thanks!

PR #14701, opened 2021-10-21
removing managed tenant risky user docs

RaajGovind:


idwilliams (approved):

idwilliams:
@Lauragra the conflicts have been resolved. Please let us know if there are other items that need to be resolved.

Lauragra (changes requested):
Please add changelog entries for the deleted APIs. Thanks.

RaajGovind:
> Please add changelog entries for the deleted APIs. Thanks.

Done, thanks!

changelog/Microsoft.MultiTenantPlatform.json
```diff
@@ -1,454 +1,480 @@
 {
-    "changelog": [
```
Lauragra, line 2: I'm not sure why we have a large number of deletions and additions here. I think all we need is a ChangeList entry to mention that the resource was deleted. Thanks.

idwilliams: @Lauragra this is a good catch. Looking through the change, I am not certain why so many items are flagged as deletions. That might have been caused by formatting update that was done in Visual Studio before the file was shared to be included in the request. I have verified the only real change is the addition of the change list entry for the entity type and navigation property.

Lauragra (approved):

PR #14832, opened 2021-11-03
Add deprecation notes

sethmanheim:
I don't know if there's a private repo for internal folks to contribute, but we would like to add this change.


sethmanheim:
#sign-off

Lauragra:
@sethmanheim , can you please check with the workload owner about deprecating the API docs? We should announce that the API is deprecated and add a note to all the API topics here: https://docs.microsoft.com/en-us/graph/api/resources/notifications-api-overview?view=graph-rest-beta. After the deprecation period is over we can then retire/remove the docs.

sethmanheim:
@Lauragra Thank you for following up -- I am trying to get this information.

sethmanheim:
@Lauragra I have added all the necessary notifications in all the right places, I think. Is it possible to merge this PR? I don't know why there is a warning in the TOC file, I didn't touch that file. Thanks! #sign-off

sethmanheim:
@Lauragra is it possible to get this PR merged?

PR #14865, opened 2021-11-05
User preferred presence

elvinyang-msft:


api-reference/beta/api/presence-setuserpreferredpresence.md
```diff
@@ -0,0 +1,99 @@
...
+| :----------------- | :------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| availability       | string   | The base presence information.                                                                                                                                                                                                                 |
+| activity           | string   | The supplemental information to availability.                                                                                                                                                                                                  |
+| expirationDuration | duration | The expiration of the app presence session. The value is represented in ISO 8601 format for durations.<br/>If not provided, a default expiration will be applied:<br/>DoNotDisturb or Busy: expires in 1 day<br/>All others: expires in 7 days |
+
+Supported combinations of `availability` and `activity` are:
```
frankpeng7, line 54: Denote property names in bold(**) instead of code(``)

mkhribech (approved):

api-reference/beta/api/presence-clearuserpreferredpresence.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "presence: clearPreferredPresence"
+description: "Clear the user preferred presence for a user"
+author: "mkhribech"
+ms.localizationpriority: medium
+doc_type: apiPageType
+ms.prod: "cloud-communications"
+---
+
```
angelgolfer-ms, line 9: @elvinyang-msft , please update presence.md to include the new clearPreferredPresence action in its Methods table.

api-reference/beta/api/presence-setuserpreferredpresence.md
```diff
@@ -0,0 +1,99 @@
+---
+title: "presence: setUserPreferredPresence"
+description: "Set the user preferred presence for a user"
+author: "mkhribech"
+ms.localizationpriority: medium
+doc_type: apiPageType
+ms.prod: "cloud-communications"
+---
+
```
angelgolfer-ms, line 9: @elvinyang-msft , please update presence.md to include the new setUserPreferredPresence action in its Methods table.

api-reference/beta/api/presence-setuserpreferredpresence.md
```diff
@@ -0,0 +1,99 @@
...
+# presence: setUserPreferredPresence
+
```
angelgolfer-ms, line 11: @elvinyang-msft , in your api.md file, you bring up why there are 2 separate methods: setting session-level presence ([setPresence](https://docs.microsoft.com/en-us/graph/api/presence-setpresence?view=graph-rest-beta) method) and the new setting user-level presence method (setUserPreferredPresence).

I think it would be helpful to point that out to customers too. 
Can you please add the relevant info in the presence.md parent resource topic?
If you're going to add the table in api.md, then please include all this content under a new H3 subsection, before the "Methods" H2 section, so that its purpose will be clearer, and, so that this content can be linked to from each of those two method topics (see next paragraph below).

Also, in each of the setPresence and setUserPreferredPresence topics, in the top section before the "Permissions" section, call out the existence of the other method topic. If you have added an H3 section to presence.md, then please link to that section as well.

api-reference/beta/api/presence-clearuserpreferredpresence.md
```diff
@@ -0,0 +1,73 @@
...
+# presence: clearPreferredPresence
```
angelgolfer-ms, line 10: @elvinyang-msft , is the method clearUserPreferredPresence, or just clearPreferredPresence?
Assuming it's the former, please update lines 2 and 10, and anywhere else you might have referred to the latter.

api-reference/beta/api/presence-clearuserpreferredpresence.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "presence: clearPreferredPresence"
+description: "Clear the user preferred presence for a user"
+author: "mkhribech"
+ms.localizationpriority: medium
```
angelgolfer-ms, line 5: Line 4 and GLOBAL:
@elvinyang-msft , can you please confirm if mkhribech should be the author for all the topics in your workload?
The [guidance](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103185/GitHub-owners) for setting the author attribute is there should be one consistent person per partner team. Let me know if we should talk about the reasons.

api-reference/beta/api/presence-clearuserpreferredpresence.md
```diff
@@ -0,0 +1,73 @@
...
+## Request headers
+| Name          | Description                 |
+| :------------ | :-------------------------- |
+| Authorization | Bearer {token}. Required.   |
+| Content-Type  | application/json. Required. |
+
+## Request body
+
+This request only contains an empty object.
+
```
angelgolfer-ms, line 41: @elvinyang-msft , you do mean an empty request body?
"This request does not require a body" is not the same, but that's not appropriate in this case either, is that right?

api-reference/beta/api/presence-setuserpreferredpresence.md
angelgolfer-ms: BTW, @elvinyang-msft , I posed a few questions on our Teams chat regarding setPresence which is currently not part of this PR. Your answers to those questions can help me determine a better way to address the above comment about differentiating setPresence and setUserPreferredPresence.
Meanwhile, you can still go ahead to add the differences between the two methods (in MD). We just might need to reorganize the content a bit afterwards.

api-reference/beta/api/presence-clearuserpreferredpresence.md
elvinyang-msft: Yes she should be the author for all topics of our workload.

api-reference/beta/api/presence-clearuserpreferredpresence.md
elvinyang-msft: It was changed to "This request only contains an empty object." in a later commit.

api-reference/beta/api/presence-setuserpreferredpresence.md
elvinyang-msft: What's in the api.md was just to explain to the Graph reviewers why we proposed two separate APIs instead of using one API to achieve both goals. I don't think we need to expose such internal tech decision process to customers.

api-reference/beta/api/presence-setuserpreferredpresence.md
```diff
@@ -13,11 +13,13 @@ Namespace: microsoft.graph
...
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Set the preferred availability and activity status for a user. If the preferred presence of a user is set, the user's presence will show as the preferred presence.
+Set the preferred availability and activity status for a user. If the preferred presence of a user is set, the user's presence is the preferred presence.
 
-Preferred presence will take effect only when there is at least one presence session of the user, otherwise, the user's presence stays as Offline.
+Preferred presence takes effect only when there is at least one [presence session](presence-setpresence.md#presence-sessions) of the user. Otherwise, the user's presence stays as Offline.
 
-A presence session can be created if the user is signed in on Teams client, or by calling the setPresence API. For details about presences sessions, see [setPresence](presence-setpresence.md#presence-sessions).
+A presence session is created as a result of a successful [setPresence](presence-setpresence.md) operation, or if the user is signed in on a Teams client.
```
elvinyang-msft, line 20: `A presence session is created`
This may suggest that every call creates a presence session but that's not the case.

angelgolfer-ms: Oh! You mean when one already exists, then setPresence of course applies to the existing session?

angelgolfer-ms: @elvinyang-msft , FYI I changed this back to "can be".
In general, we'd want to be more definite in docs. This is because customers often just search and land in specific topics, especially API topics, and do not necessarily navigate to related topics to gain more complete understanding.

api-reference/beta/api/subscription-post-subscriptions.md
```diff
@@ -80,7 +80,7 @@ Additional limitations apply for subscriptions on Outlook items. The limitations
 
 ### presence
 
-**presence** subscriptions require [encryption](/graph/webhooks-with-resource-data). Subscription creation will fail if [encryptionCertificate](../resources/subscription.md) is not specified.
+Subscriptions on **presence** require any resource data included in a change notification to be encrypted. Always specify the **encryptionCertificate** parameter when [creating a subscription](/graph/webhooks-with-resource-data#creating-a-subscription) to avoid failure. See more information about [setting up change notifications to include resource data](/graph/webhooks-with-resource-data).
```
angelgolfer-ms, line 83: @elvinyang-msft , FYI I reorganized this paragraph mainly because the link text in both links are totally off; in general we do want link text to either closely match the destination topic, or at least intuitive enough such that when the customer lands in the destination topic, they don't suspect the link is unrelated. In the case of these two links, they don't meet even the second criterion.

I did the same to the v1 topic.

concepts/webhooks-with-resource-data.md
```diff
@@ -27,15 +27,15 @@ In general, this type of change notifications include the following resource dat
 
 ## Supported resources
 
-Currently, the Microsoft Teams [chatMessage](/graph/api/resources/chatmessage) as well as the Microsoft Teams [presence](/graph/api/resources/presence) (preview) resources supports change notifications that include resource data. Specifically, you can set up a subscription that applies to one of the following:
+Currently, the Microsoft Teams [chatMessage](/graph/api/resources/chatmessage) as well as the Microsoft Teams [presence](/graph/api/resources/presence) resources supports change notifications that include resource data. Specifically, you can set up a subscription that applies to one of the following:
```
angelgolfer-ms, line 30: @elvinyang-msft , FYI in both lines 30 and 38, we should remove the "(preview)" tag for the presence resource, since you confirmed that the change notifications support for the presence resource is the same in v1 and in beta.

FYI in a concept topic which is typically not versioned like this one for change notifications, we apply the "(preview)" tag to a resource name when we want to specifically call out that the context applies to only the beta version of the resource. This can happen if the resource is only in beta but not in v1, or if the resource is in both versions but the behavior (context) in the sentence applies to only the beta version.

In the case of change notifications for the presence resource, since the support is the same in v1 and in beta, the "(preview)" tag is outdated.

api-reference/includes/teams-subscription-notes.md
```diff
@@ -27,7 +27,7 @@ You must use the `Prefer: include-unknown-enum-members` request header to get th
 > `/teams/getAllMembers` and `/chats/getAllMembers` support both `model=A` and `model=B` query parameters.
 > If no model is specified, [evaluation mode](/graph/teams-licenses#evaluation-mode-default-requirements) will be used.
 
-### Request example
+#### Request example
```
angelgolfer-ms, line 30: FYI I changed the section level in this note as it was extremely confusing in the subscription-post-subscriptions.md topics to see a "Request example" section, that uses the same level heading as other H3 sections like the "driveItem", "contact, event, and message", "presence" sections that follow it.

changelog/Microsoft.Skype.Calling.json
```diff
@@ -1,5 +1,47 @@
 {
   "changelog": [
+    {
```
angelgolfer-ms, line 3: @elvinyang-msft , FYI, I forgot to mention to you that the JSON you had added manually didn't follow how the changelog is designed (e.g., lines 7 and 15 should be "Method", not "Action", lines 23 and 31 should be "Permission" with the permission names in lines 24 and 32). I fixed them for you on your behalf so that we can publish the PR.

Anyway, you can see similar examples in the CDK topic [Changelog](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103153/Changelog).

changelog/Microsoft.Skype.Calling.json
```diff
@@ -1,5 +1,47 @@
...
+      "ChangeList": [
```
angelgolfer-ms, line 4: @elvinyang-msft , one thing I wanted to get clarification is, if your team never announced support for change notifications for the presence resource in v1, technically there should be a changelog item for that as well.
In general, even is an API or permission, etc., has been made available, but there has been no docs to illustrate the API or behavior, and it's not generally know by some other means (I can't really think of), when the docs eventually get updated, we consider that most customers are finding that out for the first time, and hence the need for a changelog item.

But just to be clear, that's not to say that the changelog should reflect purely doc updates. Indeed the changelog is meant to reflect API and permissions updates in the service. For instance, adding an example to the docs for an API that has already been well explained in the docs does not and should not entail a changelog item.

You can add a changelog item in a separate PR, and not let this hold up this PR.

angelgolfer-ms (commented):
I worked with Elvin directly on Teams to iron out most of the issues, and made fixes to the PR today to finish up the content review so that we can merge and publish.

changelog/Microsoft.Skype.Calling.json
angelgolfer-ms: I sent a similar message on Teams so that you can follow up on this as appropriate, when you return to work. Meanwhile this PR is being merged and closed, so you may or may not follow up on this comment here.

PR #14885, opened 2021-11-09
Windows updates APIs: Add unknownFutureValue to enums

Alice-at-Microsoft:


Alice-at-Microsoft:
@isvargasmsft Adding "unknownFutureValue" as we discussed

api-reference/beta/resources/enums-windowsupdates.md
```diff
@@ -28,6 +29,7 @@ Namespace: microsoft.graph.windowsUpdates
 |offeringByRequest|
 |pausedByRequest|
 |pausedByMonitoring|
+|unknownFutureValue|
 |faultedByContentOutdated|
```
isvargasmsft, line 33: This item has values after the unknownFutureValue, this means we should clarify to add the proper headers in the property description of all resources that use this enum. Guidance here: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/174164/Evolvable-enumerations?anchor=step-3%3A-add-clarification-if-one-or-more-members-follow-unknownfuturevalue

isvargasmsft (changes requested):
Let's check the description of properties that use these enums include `unknownFutureValue` and for those with values after this one, the proper mention about the headers.

Alice-at-Microsoft:
Thanks, @isvargasmsft! I knew I was missing something...

Some notes based on how I interpreted the instructions:
* I included `unknownFutureValue` even in property descriptions for read-only properties (i.e. even though the value is [never accepted in a request](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103144/Evolvable-enums?anchor=requests))
* I included `unknownFutureValue` in property descriptions for properties that only handle a subset of the enum values in the schema

isvargasmsft (approved):

PR #14893, opened 2021-11-09
Update objectidentity.md

JeffBley:
Adding info on how to query graph for this information. It isn't found anywhere else that I can find.

### DIFF
--- a/api-reference/beta/resources/objectidentity.md
+++ b/api-reference/beta/resources/objectidentity.md
@@ -27,6 +27,13 @@ The **identities** property of the [user](user.md) resource is an **objectIdenti
 
 >**Note:** When filtering on the **identities** property, you must supply both **issuer** and **issuerAssignedId**.
 
+## HTTP request
+For a specific user:
+<!-- { "blockType": "ignored" } -->
+```http
+GET /users/{id | userPrincipalName}/identities
+```
+
 ## JSON representation
 
 The following is a JSON representation of the resource.


Lauragra:
@FaithOmbongi , can you please review and identify a place to document this query? Thanks.

FaithOmbongi:
Thank you @JeffBley for raising this PR. Unfortunately, request syntaxes are included only in the operations (in this case LIST or GET), not in the resource articles. Also, using `/identities` in the URL path would limit the query from using other query parameters like `$select` and `$filter`. To address the gap you've identified, my recommendation is to update [Example 2 in GET User](https://review.docs.microsoft.com/en-us/graph/api/user-get?view=graph-rest-beta&tabs=http#example-2-get-the-properties-of-the-specified-user) operation and use `$select` as below:

**Request**
```
GET https://graph.microsoft.com/beta/users/6ea91a8d-e32e-41a1-b7bd-d2d185eed0e0?$select=id,displayName,identities
```

**Response**
```
{
    "@odata.context": "https://graph.microsoft.com/beta/$metadata#users(id,displayName,identities)/$entity",
    "id": "6ea91a8d-e32e-41a1-b7bd-d2d185eed0e0",
    "displayName": "Conf Room Adams",
    "identities": [
        {
            "signInType": "userPrincipalName",
            "issuer": "contoso.com",
            "issuerAssignedId": "Adams@contoso.com"
        }
    ]
}
```

Let me know what you think and if you'd like to make the update or if I should.

FaithOmbongi:
Hi @JeffBley , I've raised PR #14941 to add an example for retrieving the **identities** property using `$select`. Thank you for raising this PR and we look forward to your future contributions!

PR #14923, opened 2021-11-12
Update attendance report documentation

frankpeng7:


api-reference/beta/api/meetingattendancereport-get.md
```diff
@@ -0,0 +1,252 @@
...
+doc_type: apiPageType
+---
+
+# Get meetingAttendanceReport
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Get an [meetingAttendanceReport](../resources/meetingAttendanceReport.md) of an [onlineMeeting](../resources/onlinemeeting.md). Each time an online meeting is started and ended, an attendance report will be generated for that session.
```
Lauragra: In the second sentence, do you mean that the meeting has to both start and end, or either start or end? If the attendance report isn't generated until the meeting ends, you can revise to say "Each time an online meeting ends, an attendance report...".

api-reference/beta/api/meetingattendancereport-get.md
```diff
@@ -0,0 +1,252 @@
...
+# [HTTP](#tab/http)
+<!-- {
+  "blockType": "request",
+  "name": "get_attendance_report"
+}-->
+
+```msgraph-interactive
+GET https://graph.microsoft.com/beta/me/onlineMeetings/MSpkYzE3Njc0Yy04MWQ5LTRhZGItYmZiMi04ZdFpHRTNaR1F6WGhyZWFkLnYy/meetingAttendanceReport
+```
+# [C#](#tab/csharp)
```
Lauragra, line 163: Are these snippets accurate? Generally for new operation topics we don't include the snippet links; they get added as part of the weekly snippet automation process. So I think we want to remove the H1s here, including in line 154.

api-reference/beta/api/meetingattendancereport-list.md
```diff
@@ -0,0 +1,112 @@
...
+## Request body
+
+Do not supply a request body for this method.
+
+## Response
+
+If successful, this method returns a `200 OK` response code and a list of [meetingAttendanceReport](../resources/meetingAttendanceReport.md) objects in the response body.
+
+> [!TIP]
+> The **attendanceRecords** property is not returned in this API.
```
Lauragra, line 63: The property is shown in the response; please verify?

api-reference/beta/api/meetingattendancereport-list.md
```diff
@@ -0,0 +1,112 @@
...
+doc_type: apiPageType
+---
+
+# List meetingAttendanceReports
+
+Namespace: microsoft.graph
+
+[!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
+
+Get a list of [meetingAttendanceReport](../resources/meetingAttendanceReport.md) objects for an [onlineMeeting](../resources/onlinemeeting.md). Each time an online meeting is starts and ends, an attendance report is generated for that session.
```
Lauragra, line 16: Same question here - does the report get generated only after the meeting ends?

Lauragra, line 16: How is it that an online meeting has multiple attendance reports? Is it a recurring meeting and this API gets the attendance reports for each occurrence?

changelog/Microsoft.Skype.Calling.json
```diff
@@ -1,5 +1,71 @@
 {
   "changelog": [
+    {
+      "ChangeList": [
+        {
+          "Id": "78707d6f-571b-4a27-9e90-b290ce1c42ff",
+          "ApiChange": "Resource",
+          "ChangedApiName": "attendanceRecord",
+          "ChangeType": "Deletion",
+          "Description": "Removed the **attendanceRecord** resource type.",
```
Lauragra, line 10: This entity still exists in the metadata, and also the resource topic is updated but not deleted in this PR. Please check this ChangeList item.

changelog/Microsoft.Skype.Calling.json
```diff
@@ -1,5 +1,71 @@
...
+          "Target": "attendanceRecord"
+        },
+        {
+          "Id": "78707d6f-571b-4a27-9e90-b290ce1c42ff",
+          "ApiChange": "Resource",
+          "ChangedApiName": "attendanceRecord",
```
Lauragra, line 16: This is confusing; why do we have deleted and added entries for the same resource? What should customers understand about this API update?

api-reference/beta/api/meetingattendancereport-get.md
frankpeng7: Ok yea that's what I meant and I did find my description a bit ambiguous. I'll update this.

api-reference/beta/api/meetingattendancereport-get.md
frankpeng7: yes, they are moved from the onlinemeeting-get topic without any changes. This example of getting attendance report used to be in the onlinemeeting-get topic but now that attendance report has it's own get topic, I figure maybe I should move the example.

api-reference/beta/api/meetingattendancereport-list.md
frankpeng7: I see your point. I meant to say in the tip that the value of attendanceRecords property will not be returned - it will always be an empty list. Should I say that?

api-reference/beta/api/meetingattendancereport-list.md
frankpeng7: Same reply as above. Will update

frankpeng7: A meeting can be started multiple times (even after it has ended and it's past the scheduled end time) hence there will be multiple attendance reports. It not necessarily needs to be a recurring meeting but yes a recurring meeting can use this list api to get attendance reports for each sessions (even if we don't support schedule a recurring meeting in Graph API yet)

changelog/Microsoft.Skype.Calling.json
frankpeng7: these stubs are auto generated when sending out the CSDL update PR. I will update the changelist

api-reference/beta/api/meetingattendancereport-list.md
Lauragra: Yes, that would be more clear.

Lauragra (approved):

PR #14933, opened 2021-11-13
Add a note about using the DirectoryObject APIs

Penguinwizzard:
There are cases where a developer wanting to support groups in an
application doesn't intend to make a distinction between supporting
user accounts and service principals, but may accidentally introduce
a difference simply because the 'user' APIs are significantly further
up the search results, and don't reference the more general API which
works for both kinds of identities. Here I add a note and links to the
directoryObject APIs for two such pages.

### DIFF
--- a/api-reference/v1.0/api/user-checkmembergroups.md
+++ b/api-reference/v1.0/api/user-checkmembergroups.md
@@ -19,8 +19,8 @@ types of groups provisioned in Azure AD. Note that Microsoft 365 groups cannot c
 in a Microsoft 365 group is always direct.
 
 > **Note:** This API only works for user accounts, and not for service principal identities. If your application
-supports both, consider using the [DirectoryObject checkMemberGroups endpoint](/graph/api/directoryobject-checkmembergroups)
-instead (at `/directoryObjects/{id}/checkMemberGroups`).
+supports both, consider using the [directoryObject: checkMemberGroups method](/graph/api/directoryobject-checkmembergroups)
+instead.
 
 ## Permissions
 
--- a/api-reference/v1.0/api/user-checkmembergroups.md
+++ b/api-reference/v1.0/api/user-checkmembergroups.md
@@ -19,7 +19,7 @@ types of groups provisioned in Azure AD. Note that Microsoft 365 groups cannot c
 in a Microsoft 365 group is always direct.
 
 > **Note:** This API only works for user accounts, and not for service principal identities. If your application
-supports both, consider using the [directoryObject: checkMemberGroups method](/graph/api/directoryobject-checkmembergroups)
+supports both, consider using [directoryObject: checkMemberGroups method](/graph/api/directoryobject-checkmembergroups)
 instead.
 
 ## Permissions
--- a/api-reference/v1.0/api/user-checkmembergroups.md
+++ b/api-reference/v1.0/api/user-checkmembergroups.md
@@ -18,6 +18,10 @@ You can check up to a maximum of 20 groups per request. This function supports M
 types of groups provisioned in Azure AD. Note that Microsoft 365 groups cannot contain groups. So membership
 in a Microsoft 365 group is always direct.
 
+> **Note:** This API only works for user accounts, and not for service principal identities. If your application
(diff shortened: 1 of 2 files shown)


hpsin:
@jpettere / @yyuank - would you be appropriate folks to sign off on this helpful addition? We've had multiple teams fail to discover this API, and are frustrated that they have to call different APIs to get group membership for a user or app. This is a much more usable API for Azure teams that treat service principals and user principals the same (as "security principals")

yyuank (commented):
Instead of adding a note, should we just add an example of checking/getting group memberships for a directory object to these pages?

hpsin:
@yyuank - yeah, examples are great! What matters is the cross-linking between the easily-found-nut-not-as-useful API (/user and /serviceprincipal) and the harder-to-find-and-better API

yyuank:
Can you share the document you want to link to?

hpsin:
[These APIs](https://docs.microsoft.com/en-us/graph/api/directoryobject-checkmembergroups?view=graph-rest-1.0&tabs=http#http-request) - /users/oid/checkMembergroups, /groups/id/checkmembergroups, and /serviceprincipals/checkmembergroup should probably all be deprecated or at least linked to /directoryObjects/id/checkmembergroups  .  It's by far the better API to call but very difficult to discover.

yyuank:
Makes sense. Approved.

api-reference/v1.0/api/user-checkmembergroups.md
```diff
@@ -18,6 +18,10 @@ You can check up to a maximum of 20 groups per request. This function supports M
 types of groups provisioned in Azure AD. Note that Microsoft 365 groups cannot contain groups. So membership
 in a Microsoft 365 group is always direct.
 
+> **Note:** This API only works for user accounts, and not for service principal identities. If your application
```
Lauragra, line 21: I don't see a checkMemberGroups action defined in the metadata that has a binding type of user; I do see one for directoryObject. Can you help clarify why we have a topic for user and a separate topic for directoryObject? Adding @FaithOmbongi  as well to review. The same applies to the getMemberGroups action. Also, if you update the topic in the v1.0 folder, please apply the same update to the topic in the beta folder.

api-reference/v1.0/api/user-getmembergroups.md
```diff
@@ -18,6 +18,10 @@ This function supports Microsoft 365 and other types of groups provisioned in Az
 request can return is 11000. Note that Microsoft 365 groups cannot contain groups. So membership in a Microsoft 365 group is
 always direct.
 
+> **Note:** This API only works for user accounts, and not for service principal identities. If your application
```
Lauragra, line 21: Same comment here, the metadata doesn't define this action with a user binding parameter. Please clarify and apply this update to the /beta topic.

api-reference/v1.0/api/user-checkmembergroups.md
FaithOmbongi: @Penguinwizzard @hpsin @yyuank , I've referred to the metadata & agree with @Lauragra here. Technically, **checkMemberGroups** and **checkMemberObjects** actions return directoryObject IDs. So following the file naming guidance, user-checkmembergroups.md and user-getmemberobjects.md are invalid files; the correct file being [directoryobject-checkmembergroups.md](https://docs.microsoft.com/en-us/graph/api/directoryobject-checkmembergroups?view=graph-rest-1.0&tabs=http#http-request). All the required changes and notes should be documented in [directoryobject-checkmembergroups.md](https://docs.microsoft.com/en-us/graph/api/directoryobject-checkmembergroups?view=graph-rest-1.0&tabs=http#http-request). I'll start an email thread to discuss this and come to an agreement.

Penguinwizzard:
Internal discussion lead to a conclusion that @FaithOmbongi is going to throw up a PR further simplifying the documentation situation here (consolidating a few docs because the types returned are the same), which should make it even easier to understand the API situation. As such, the docs modified in this PR won't even exist anymore, so I'm closing the PR.

FaithOmbongi:
The changes are live via PR #14954. Removing the changes requested label to avoid unactionable notifications.

PR #14950, opened 2021-11-16
AttackSimulation (beta) - Removing unused members, unused enums

aspati:
Removing unused members, unused enums


angelgolfer-ms (changes requested):
Hi, please add the changelog JSON for these changes - removing the mode property from the simulation resource, removing the simulationMode enumeration, and removing the notCompleted enum member. You can add to each of their description something like they are not used or not returned by the API, to clarify these are not deprecation.

aspati:
> Hi, please add the changelog JSON for these changes - removing the mode property from the simulation resource, removing the simulationMode enumeration, and removing the notCompleted enum member. You can add to each of their description something like they are not used or not returned by the API, to clarify these are not deprecation.

@angelgolfer-ms , added changelog. Please check.

changelog/Microsoft.Security.AttackSimulation.json
```diff
@@ -3,276 +3,36 @@
...
       ],
-      "Id": "86bb6c22-23cb-4ed1-9903-7d5fb8d7ca76",
+      "Id": "d9cff973-8801-441f-996f-f075d3f4bf82",
       "Cloud": "Prod",
       "Version": "beta",
-      "CreatedDateTime": "2021-09-27T06:42:46.5889852Z",
+      "CreatedDateTime": "2021-11-18T05:14:42.2618253Z",
       "WorkloadArea": "Security",
-      "SubArea": ""
+      "SubArea": "Attack simulation and training"
```
angelgolfer-ms, line 35: Line 35:
FYI now that the "Attack simulation and training" sub-node exists in the TOC, you'd have to specify the SubArea value as well, so that the changelog can indicate in the section title something like "Security | Attack simulation and training". We do that when under a top level node (e.g., Security) in the TOC, there are multiple services.

You didn't have to do that in your the last (and the debut) PR in September because that "Attack simulation and training" didn't exist yet.

changelog/Microsoft.Security.AttackSimulation.json
```diff
@@ -3,276 +3,36 @@
     {
```
angelgolfer-ms, line 3: @aspati, I made some fixes directly in your PR for the changelog items for this PR.
However, please add back the entire changelog record that was specified in the last PR 13789. If we merge this PR as is with the last changelog record removed, the resultant changelog will no longer display those changelog items for that PR, and that would be confusing.
You can see an image of what I mean by a changelog record in this section of the changelog CDK topic: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/103153/Changelog?anchor=summary-process

angelgolfer-ms (changes requested):
@aspati , thanks for the changelog update.
Please see my comment in the changelog JSON file Microsoft.Security.AttackSimulation.json.

changelog/Microsoft.Security.AttackSimulation.json
aspati: @angelgolfer-ms , fixed changelog. Please check.

angelgolfer-ms (approved):
Thanks, the changelog JSON file looks complete now.

PR #14955, opened 2021-11-16
Update add-properties-profilecard.md

PollyNincevic:
Added instructions on how to delete a custom attribute per PM request


isvargasmsft:
@kevinbellinger, what's the status of this PR? Is the technical review complete?

concepts/add-properties-profilecard.md
```diff
@@ -148,6 +148,11 @@ Content-type: application/json
   ]
 }
 ```
+## Removing a custom attribute
```
angelgolfer-ms, line 151: Line 151:
Please title this section as "Deleting a custom attribute".
Reason: In our REST API in general and in this case with the profileCardProperties nav property, there's a subtle distinction between using "remove" vs "delete". Because ContainsTarget is specified as "true" for profileCardProperties, a nav propertu of organizationSettings, we should use "delete". See the definition of the nav property in the beta metadata:

<EntityType Name="organizationSettings" BaseType="graph.entity">
...
<NavigationProperty Name="profileCardProperties" Type="Collection(graph.profileCardProperty)" ContainsTarget="true"/>
</EntityType>

concepts/add-properties-profilecard.md
```diff
@@ -148,6 +148,11 @@ Content-type: application/json
...
+
+To remove a custom attribute, run the following query in [Graph Explorer](https://developer.microsoft.com/graph/graph-explorer):
```
angelgolfer-ms, line 153: Hi @PollyNincevic , the body of this section is extremely slim. I don't have all the relevant context, but maybe you can check with your PMs. I can imagine at the very least, you'd want to model the prior section "Adding custom attributes" and say something like:
"Following the same mapping between Azure AD custom extension attributes and profile card custom attributes (such as `customAttribute1`) as described in the preceding section Adding a custom attribute (link to that section), you can delete a custom attribute using the [delete](/graph/api/profilecardproperty-delete?view=graph-rest-beta) operation as shown in the example below.

### Example
The following example deletes the custom attribute `customAttribute5` from the organization settings. A successful deletion returns `HTTP 204`.

#### Request


```http
DELETE https://graph.microsoft.com/beta/organization/{organizationId}/settings/profileCardProperties/customAttribute5
```

#### Response

```http
HTTP/1.1 204 No Content
```




BTW, I'd suggest renaming the existing section title "Adding custom attributes" as "Adding a custom attribute". Both that operation and this delete operation work on only one attribute at a time.

concepts/add-properties-profilecard.md
```diff
@@ -148,6 +148,11 @@ Content-type: application/json
   ]
 }
 ```
```
angelgolfer-ms, line 150: To make this new section for deletion parallel with the existing section in line 76, and to make the request and response in that example clearer, please add the following subsection titles:

- After line 107, add an H4 subsection with the title "Request".
- After line 131 before the line ```http```, add an H4 subsection with the title "Response".

concepts/add-properties-profilecard.md
```diff
@@ -148,6 +148,11 @@ Content-type: application/json
   ]
```
angelgolfer-ms, line 148: Since you're adding an example for deleting a customer profile card property, you should please:
- Retitle the topic from "Add additional properties to the profile card..." to "Add or delete custom properties from the profile card using the profile card API in Microsoft Graph (preview)".
- After line 17, add a sentence to say something like 
  "You can also delete custom attributes from profile cards of the organization."

concepts/add-properties-profilecard.md
angelgolfer-ms, line 153: Wanted to point out explicitly, in our REST API docs, when we show how to make a REST operation call, we just describe what it does and show the REST request and REST response. We don't explicitly tell customers to use Graph Explorer. Graph Explorer was primarily developed as a tool to let customers try out APIs quickly without the overhead of creating an app, for learning or prototyping purpose.

Instead of using Graph Explorer to carry out production work, we'd like customers to create apps and make the calls from their apps, either as REST API calls directly or by using an SDK.

angelgolfer-ms (changes requested):
@PollyNincevic , please see my comments and follow-up suggestions in my email (to be sent shortly).

concepts/add-properties-profilecard.md
PollyNincevic: Section renamed to "Deleting a custom attribute"

concepts/add-properties-profilecard.md
PollyNincevic: Thank you for the detailed feedback! I have updated per your suggestions.

concepts/add-properties-profilecard.md
PollyNincevic: Updated

concepts/add-properties-profilecard.md
PollyNincevic: Updated per suggestions

concepts/add-properties-profilecard.md
PollyNincevic: Thank you. Section replaced with what you had suggested in comments above.

PollyNincevic:
Changes made per Angela comments. @isvargasmsft fyi

isvargasmsft (approved):

PR #14961, opened 2021-11-16
Updating Identity Reporting limits

besiler:
Changed Identity reporting limits from 60 requests per 60 seconds to 5 requests per 10 seconds to match the actual policy being applied for the service


FaithOmbongi (approved):
Headsup @besiler , approving this PR and adding the **ready to merge** label. If there's any reason we should hold back this PR, please replace the label with the **do not merge** label and just leave a note for tracking.

PR #14963, opened 2021-11-17
Add maximum values to permissions reference doc

Jackson-Woods:
Added maximum values each app type can request and consent

### DIFF
--- a/concepts/permissions-reference.md
+++ b/concepts/permissions-reference.md
@@ -27,11 +27,22 @@ The _constraint_ element of the name determines the potential extent of access y
 * **AppFolder** grants permission for the app to read and write files in a dedicated folder in OneDrive. This constraint is only exposed on [Files permissions](#files-permissions) and is only valid for Microsoft accounts.
 * If **no constraint** is specified the app is limited to performing the operations on the resources owned by the signed-in user. For example, _User.Read_ grants privileges to read the profile of the signed-in user only, and _Mail.Read_ grants permission to read only mail in the mailbox of the signed-in user.
 
-> **Note**: In delegated scenarios, the effective permissions granted to your app may be constrained by the privileges of the signed-in user in the organization.
+> **Note**: In delegated scenarios, the effective permissions granted to your app are constrained by the privileges of the signed-in user in the organization.
 
 ## Microsoft accounts and work or school accounts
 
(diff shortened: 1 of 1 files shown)


concepts/permissions-reference.md
```diff
@@ -27,12 +27,23 @@ The _constraint_ element of the name determines the potential extent of access y
...
+| AzureADMyOrg                       | Users from the organization where the app is registered  | 400                                       | 400                                                       | About 155 delegated permissions and about 300 application permissions |
+| AzureADMultipleOrgs                | Users from any Azure AD organization                     | 400                                       | 400                                                       | About 155 delegated permissions and about 300 application permissions |
+| PersonalMicrosoftAccount           | Consumer users (Outlook.com or Live.com accounts, etc.)  | 30                                        | 30                                                        | 30                                                                    |
+| AzureADandPersonalMicrosoftAccount | Consumer users and users from any Azure AD organization  | 30                                        | 30                                                        | 30                                                                    |
```
FaithOmbongi, line 45: Suggestion: is there a chance of the max. limits being less than 155/300 or can we avoid the uncertainty of using "about"?
```suggestion
| AzureADMyOrg                       | Users from the organization where the app is registered  | 400                                       | 400                                                       | 155 delegated permissions and 300 application permissions |
| AzureADMultipleOrgs                | Users from any Azure AD organization                     | 400                                       | 400                                                       | 155 delegated permissions and 300 application permissions |
| PersonalMicrosoftAccount           | Consumer users (such as Outlook.com or Live.com accounts)  | 30                                        | 30                                                        | 30                                                                    |
| AzureADandPersonalMicrosoftAccount | Consumer users and users from any Azure AD organization  | 30                                        | 30                                                        | 30                                                                    |
```

FaithOmbongi (approved):
Thanks @Jackson-Woods for working on this change.

Left a comment on the limits, otherwise approved.

I think we should also link from the [application](https://docs.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0) resource doc (descriptions of requiredResourceAccess and signInAudience) to the section you've added.

concepts/permissions-reference.md
Jackson-Woods: 155 and 300 are relatively conservative values. Usually the limit will be slightly higher. It depends on the specific permissions passed, unfortunately.

FaithOmbongi:
Heads up @Jackson-Woods, I believe we'd agreed that the PR was ready to merge. I've added the **ready to merge** label to move it into the merge queue. If there's any reason we should hold back this PR, please replace **ready to merge** with **do not merge** label.

PR #14988, opened 2021-11-18
Add methods for endpoints nav property

FaithOmbongi:


FaithOmbongi:
FYI @xamarsinx @eketo-msft, this PR contains the proper API docs for the **endpoints** navigation property on service principals.

PR #14995, opened 2021-11-19
Update teamsasyncoperation.md

KoenZomers:
Updated explanation and added sample how the operation status can be queried. I found the current documentation confusing as the Location header returns something like this:

```
{
    "client-request-id": "8fb1a7d6-a791-79c3-ab93-7982ba8bb5d7",
    "location": "/teams('8d665b38-7659-4b57-b33d-8be31d696bed')/operations('d9b1da42-0d12-4040-8e7d-720b0e22748b')",
    "request-id": "d863fac7-4f47-40fb-8abc-cabc6dc5e469"
}
```

But to actually query the status, we need to reformat the location header and execute:

```
GET /v1.0/teams/8d665b38-7659-4b57-b33d-8be31d696bed/operations('d9b1da42-0d12-4040-8e7d-720b0e22748b')
```

This PR should clarify the documentation on how this can be used.


api-reference/v1.0/resources/teamsasyncoperation.md
```diff
@@ -17,10 +17,25 @@ A Microsoft Teams async operation is an operation that transcends the lifetime o
 These operations are long-running or too expensive to complete within the timeframe of their originating request.
 
 When an async operation is initiated, the method returns a 202 Accepted response code. 
-The response will also contain a Location header, which contains the location of the teamsAsyncOperation. 
+The response will also contain a Location header, which contains the location of the teamsAsyncOperation. Notice in the sample HTTP request below that the Location header needs to be slightly reformatted to use it to query for the actual status.
 Periodically check the status of the operation by making a GET request to this location; wait >30 seconds between checks.
 When the request completes successfully, the status will be "succeeded" and the targetResourceLocation will point to the created/modified resource.
 
+## HTTP request
```
Lauragra, line 22: This is a resource topic; it should not include the H2 headings that you've added here. Please remove, and remove the reference to the example in the description.

api-reference/v1.0/resources/teamsasyncoperation.md
```diff
@@ -47,14 +62,16 @@ The following is a JSON representation of the resource.
 
 ```json
 {
-    "id": "string",
-    "operationType": "archiveTeam",
-    "createdDateTime": "2018-01-01T00:00:00.0000000Z",
+    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#teams('8d665b38-7659-4b57-b33d-8be31d696bed')/operations/$entity",
+    "id": "d9b1da42-0d12-4040-8e7d-720b0e22748b",
```
Lauragra, line 64: The JSON block should not include fictitious data; it should just show the types. Please remove the values from the JSON section.

api-reference/v1.0/resources/teamsasyncoperation.md
KoenZomers: Removing this part would miss the entire point of clarification. The way its written today its not clear how it can be used. Took me several hours to figure out. This is the main thing to clarify this. If its not intended to be added here, what alternative do we have to clarify how to use it?

api-reference/v1.0/resources/teamsasyncoperation.md
KoenZomers: The sample already contains fictious data today. Don't see how that's any different from what I'm proposing here? This is just an updated one showing an actual response. I believe having a real response here helps in understand what kind of response and data to can expect to be returned.

api-reference/v1.0/resources/teamsasyncoperation.md
Lauragra: The HTTP request information goes in an API topic, not the resource topic. For example, it looks like this information is covered in this topic for chats: https://docs.microsoft.com/en-us/graph/api/teamsasyncoperation-get?view=graph-rest-beta&tabs=http. Do we need to also cover the same information for teams?

api-reference/v1.0/resources/teamsasyncoperation.md
Lauragra: The original content has fictitious data but in resource topics, we show the types in the JSON block. In the API topics we include examples with fictitious values to help illustrate.

api-reference/v1.0/resources/teamsasyncoperation.md
KoenZomers: I would really love that, yes! Would you like me to make a copy of the channel page you reference and change it for the Teams operations and add it to this PR for you to review?

api-reference/v1.0/resources/teamsasyncoperation.md
KoenZomers: Still not sure if I fully understand what you mean. That would imply that the sample currently on this topic page was also incorrect, right?

Look at the before and after, nothing really changed, other than the value of the dates going from 2018 to 2021?

image

api-reference/v1.0/resources/teamsasyncoperation.md
Lauragra: Yes, please, then we can have it reviewed. Thanks!

PR #15077, opened 2021-11-30
Fix incorrect example in user-list.md

kksrivatsa:
Example (7) was incorrect. We do not match real contains in $search, we do a tokenized starts with.

displayName: ContosoAdmin1 is broken into: "ContosoAdmin1", "Contoso", "Admin" and "1". The input search string is also tokenized and we do a startswith match on each of these tokens.
This was reported in: https://stackoverflow.com/questions/70165966/microsoft-graph-api-search-function-doesnt-work-as-expected

### DIFF
--- a/api-reference/v1.0/api/user-list.md
+++ b/api-reference/v1.0/api/user-list.md
@@ -410,7 +410,7 @@ Content-type: application/json
 }
 ```
 
-### Example 7: Use $search to get users with display names that contain the letters 'wa' or the letters 'to' including a count of returned objects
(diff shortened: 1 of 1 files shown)


davidmu1 (approved):

FaithOmbongi (approved):
Thank you @kksrivatsa for raising this PR. If you don't mind, kindly also make the same changes to the example in the beta doc. It's [Example 10](https://review.docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&branch=pr-en-us-15077&tabs=http)

PR #15078, opened 2021-11-30
Fix incorrect example in user-list.md

kksrivatsa:
Example (10) was incorrect. We do not match real contains in $search, we do a tokenized starts with.

displayName: ContosoAdmin1 is broken into: "ContosoAdmin1", "Contoso", "Admin" and "1". The input search string is also tokenized and we do a startswith match on each of these tokens.
This was reported in: https://stackoverflow.com/questions/70165966/microsoft-graph-api-search-function-doesnt-work-as-expected

### DIFF
--- a/api-reference/beta/api/user-list.md
+++ b/api-reference/beta/api/user-list.md
@@ -621,7 +621,7 @@ Content-type: application/json
 
 ```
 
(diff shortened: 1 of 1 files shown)


FaithOmbongi (approved):
Thanks @kksrivatsa !

PR #15083, opened 2021-11-30
Add more examples for $search

kksrivatsa:
Added a non-English example and explicitly stated that contains isn't supported.

### DIFF
--- a/concepts/search-query-parameter.md
+++ b/concepts/search-query-parameter.md
@@ -120,7 +120,7 @@ To learn more about the People API, see [Get information about relevant people](
 
 ## Using $search on directory object collections
 
-Azure AD resources and their relationships that derive from [directoryObject](/graph/api/resources/directoryobject) support the `$search` query parameter only in advanced queries. The search implementation does **not** support `contains`. Instead it uses a tokenization approach which works by extracting words from the property value and the search string using spaces, numbers, different casing, and symbols to separate the words, as follows:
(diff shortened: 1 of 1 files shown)


concepts/search-query-parameter.md
```diff
@@ -120,7 +120,7 @@ To learn more about the People API, see [Get information about relevant people](
 
 ## Using $search on directory object collections
 
-Azure AD resources and their relationships that derive from [directoryObject](/graph/api/resources/directoryobject) support the `$search` query parameter only in advanced queries. The search uses a tokenization approach which works by extracting words from your input and output string, using spaces, numbers, different casing, and symbols to separate the words, as follows:
+Azure AD resources and their relationships that derive from [directoryObject](/graph/api/resources/directoryobject) support the `$search` query parameter only in advanced queries. The search implementation does **not** support `contains` due to performance reasons. Instead it uses a tokenization approach which works by extracting words from your input and output string, using spaces, numbers, different casing, and symbols to separate the words, as follows:
```
FaithOmbongi, line 123: Recommending that we avoid highlighting this reason
```suggestion
Azure AD resources and their relationships that derive from [directoryObject](/graph/api/resources/directoryobject) support the `$search` query parameter only in advanced queries. The search implementation does **not** support `contains` but uses a tokenization approach which works by extracting words from your input and output string, using spaces, numbers, different casing, and symbols to separate the words, as follows:
```

FaithOmbongi (approved):
Thank you @kksrivatsa for raising this PR.

I've approved it with a minor suggestion.

FaithOmbongi (approved):
Thank you @kksrivatsa for raising this PR.

I've approved it with a minor suggestion.

concepts/search-query-parameter.md
kksrivatsa: Fixed.

PR #15085, opened 2021-12-01
updated mezzo supported properties

Licantrop0:
updated samples


FaithOmbongi (approved):
Thank you @Licantrop0  for raising this PR. If you don't mind, I've pushed a few commits to address the following before approving:

- Changed "Contacts" => "Organizational contacts". This is to differentiate the **orgContact** resource which is the directory object from the **contact** resource which is in the Microsoft 365 workload.
- Changed a few titles to H3

PR #15114, opened 2021-12-02
GA Deeplink for Chats [do not merge before 12/9]

jessiechan27:


api-reference/v1.0/resources/chat.md
```diff
@@ -92,7 +92,8 @@ Here is a JSON representation of the resource.
   "topic": "string",
   "createdDateTime": "dateTimeOffset",
   "lastUpdatedDateTime": "dateTimeOffset",
-  "chatType": "String"
+  "chatType": "String",
```
AkJo, line 95: Lets fix the casing over here.

api-reference/v1.0/resources/chat.md
```diff
@@ -57,7 +57,7 @@ not all scenarios are possible. It is possible to get chat IDs with delegated pe
 | id| String| The chat's unique identifier. Read-only.|
 | lastUpdatedDateTime| dateTimeOffset|  Date and time at which the chat was renamed or list of members were last changed. Read-only.|
 | topic| String|  (Optional) Subject or topic for the chat. Only available for group chats.|
-
+|  webUrl          |   string                  |  A hyperlink that will go to the chat in Microsoft Teams. This URL should be treated as an opaque blob, and not parsed. Read-only.     |
```
AkJo, line 60: @Lauragra , does the current structure sound clumsy?

AkJo (approved):

api-reference/v1.0/resources/chat.md
```diff
@@ -57,7 +57,7 @@ not all scenarios are possible. It is possible to get chat IDs with delegated pe
 | id| String| The chat's unique identifier. Read-only.|
```
AkJo, line 57: Can you fix the casing while you are touching this file.

api-reference/v1.0/resources/chat.md
```diff
@@ -57,7 +57,7 @@ not all scenarios are possible. It is possible to get chat IDs with delegated pe
...
 | lastUpdatedDateTime| dateTimeOffset|  Date and time at which the chat was renamed or list of members were last changed. Read-only.|
 | topic| String|  (Optional) Subject or topic for the chat. Only available for group chats.|
```
AkJo, line 59: Can you fix the casing while you are touching this file.

AkJo: If the change is needed in /beta file as well, then please apply those too.

AkJo (approved):

api-reference/v1.0/resources/chat.md
jessiechan27: It's the same as what's in beta, but we can change both.

api-reference/beta/resources/chat.md
```diff
@@ -65,13 +65,13 @@ not all scenarios are possible. It is possible to get chat IDs with delegated pe
 |:---------------|:--------|:----------|
 | chatType| [chatType](../resources/chat.md#chattype-values) | Specifies the type of chat. Possible values are: `group`, `oneOnOne`, `meeting`, `unknownFutureValue`.|
 | createdDateTime| dateTimeOffset|  Date and time at which the chat was created. Read-only.|
-| id| String| The chat's unique identifier. Read-only.|
+| id| string| The chat's unique identifier. Read-only.|
```
Lauragra, line 68: Why are we lowercasing these? In general the guidance is to capitalize the first letter in a table cell. The exception is for resource type names, which are lower camel case.

AkJo: It is still a resource, right - just a primitive type - 'string' instead of the custom types that we create. Or are we Pascal-Casing primitives?

I don;t have a preference either way. Whatever is the pattern, I am fine with it.

api-reference/v1.0/resources/chat.md
```diff
@@ -57,7 +57,7 @@ not all scenarios are possible. It is possible to get chat IDs with delegated pe
...
-
+| webUrl | String| A hyperlink that will go to the chat in Microsoft Teams. This URL should be treated as an opaque blob, and not parsed. Read-only. |
```
Lauragra, line 60: If you wanted to revise this here and in the beta file, suggested text:
"The URL for the chat in Microsoft Teams. The URL is treated as an opaque blob, and not parsed. Read-only."

api-reference/beta/resources/chat.md
```diff
@@ -71,7 +71,7 @@ not all scenarios are possible. It is possible to get chat IDs with delegated pe
 | tenantId| String | The identifier of the tenant in which the chat was created. Read-only.|
 | topic| String|  (Optional) Subject or topic for the chat. Only available for group chats.|
 | viewpoint|[chatViewpoint](../resources/chatviewpoint.md)|Represents caller-specific information about the chat, such as last message read date and time. This property is populated only when the request is made in a delegated context.|
-| webUrl| String | A hyperlink that will go to the chat in Microsoft Teams. This URL should be treated as an opaque blob, and not parsed. Read-only.|
+| webUrl| String | The URL for the chat in Microsoft Teams. The URL is treated as an opaque blob, and not parsed. Read-only.|
```
AkJo, line 74: On second thought, I would prefer the former form. 

"The URL should be treated as an opaque blob, and not parsed."
This message is meant for the developers aka consumer of this field. It is to warm them that *they* should not try to parse the URL. We, Teams, on the other hand, of course, does parse the URL.

In its current form, the phrase is reading like it is about how we treat the URL instead of how the developer should be treating it.

Lauragra (approved):

PR #15300, opened 2021-12-20
Update enums.md

ebasseri:
adding new enums ahead of a 1/31/21 public preview annoucement


FaithOmbongi (commented):
Thank you @ebasseri for raising this PR. 

**Recommendation:** While these changes affect a different workload & therefore changelog file (**Microsoft.AAD.Reporting.json**, I believe, instead of **Microsoft.IdentityProtectionServices.json**), I recommend they be part of the main PR which documents features that use these enum members. That'll allow the changes to be published together, and when you add the changelog description, the customer will know the (new) API which relates to these changes.

FaithOmbongi (commented):
Left a recommendation for consideration.

PR #15307, opened 2021-12-21
[Doc issue][Update filter limitations]

v-abirade:


api-reference/beta/api/schedule-list-shifts.md
```diff
@@ -33,8 +33,12 @@ GET /teams/{teamId}/schedule/shifts
 ```
 
 ## Optional query parameters
+
 This method supports the `$filter` [OData query parameter](/graph/query-parameters) to help customize the response.
 
+> [!NOTE]
+> $filter doesn't support using the same property more than once in the query, which prevents bounding queries, such as, sharedShift/startDateTime ge 2019-05-09T00:00:00Z and sharedShift/startDateTime le 2019-05-09T23:59:59Z.
```
Lauragra, line 40: Please use the tic marks (``) to style the parameter name and the parameter example.

Suggested revision: "The `$filter` parameter doesn't support the use of the same property more than once in a query. For example, the following query will not work: `<example>`

v-abirade: Updated as per comment.

api-reference/beta/api/schedule-list-shifts.md
```diff
@@ -33,8 +33,12 @@ GET /teams/{teamId}/schedule/shifts
 ```
 
 ## Optional query parameters
+
 This method supports the `$filter` [OData query parameter](/graph/query-parameters) to help customize the response.
 
+> [!NOTE]
+> The `$filter` parameter doesn't support the use of the same property more than once in a query. For example, the following query will not work: `sharedShift/startDateTime ge 2019-05-09T00:00:00Z` and `sharedShift/startDateTime le 2019-05-09T23:59:59Z`.
```
Lauragra, line 40: One minor fix here - the backticks (`) should go around the whole query, including the word "and", rather than around each property separately. You're trying to illustrate that a query with the same property twice won't work, but the way it's formatted, it looks like it can be two separate queries. Please fix in both topics.

v-abirade: Thanks @Lauragra, resolved as suggested.

Lauragra (approved):

PR #15360, opened 2021-12-29
Update M365 usage report API list

kszb:


Lauragra:
@kszb , looks good, my only comment is that we might want to call out in the changelog that support for these APIs was added to the US Govt cloud. Thanks.

Lauragra (approved):

PR #15377, opened 2022-01-03
Fix address examples in Microsoft Graph Documentation : Bookings API v1.0

razortbone:


isvargasmsft (approved):

kanchm (commented):
URLs are still not fixed in this file.  Line 29 shows /bookingBusinesses whereas it should be /solutions/bookingBusinesses

PR #15421, opened 2022-01-06
Fix broken link

haisi:


Lauragra:
Thanks for your contribution. The link might not work in GitHub, but it does work on the live site: https://docs.microsoft.com/en-us/graph/api/resources/mail-api-overview?view=graph-rest-1.0. I'll close this PR.

PR #15448, opened 2022-01-10
Fix compile error in Java example

haisi:
`MessageCollectionPage messagesPage` cannot be `final`, as on line 111 the variable is re-assigned.


andrueastman (approved):
Thanks for this @haisi

PR #15454, opened 2022-01-10
Update intune-mam-managedapppolicy.md

Hani-Microsoft:
removed json representation as this type is abstract base and thus will never have a concrete json example


Lauragra:
@Hani-Microsoft , FYI, the JSON representation is a standard section for resource topics. I'm going to close this PR. Thanks.

PR #14390, opened 2021-09-27
Adding docs and changelog for new limits on requiredResourceAccess

psignoret:

PR #15515, opened 2022-01-14
Update resourceaccess.md

jiasli:
Update https://docs.microsoft.com/en-us/graph/api/resources/resourceaccess?view=graph-rest-1.0

Per https://docs.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0, `appRoles` property contains `appRole` collection:

image

so `oauth2PermissionScope` should also use its singular form.

### DIFF
--- a/api-reference/v1.0/resources/resourceaccess.md
+++ b/api-reference/v1.0/resources/resourceaccess.md
@@ -17,7 +17,7 @@ Object used to specify an OAuth 2.0 permission scope or an app role that an appl
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
 |id|GUID|The unique identifier for one of the [oauth2PermissionScopes](permissionscope.md) or [appRole](approle.md) instances that the resource application exposes.|
-|type|String|Specifies whether the **id** property references an [oauth2PermissionScopes](permissionscope.md) or an [appRole](approle.md). The possible values are: `Scope` (for OAuth 2.0 permission scopes) or `Role` (for app roles).|
+|type|String|Specifies whether the **id** property references an [oauth2PermissionScope](permissionscope.md) or an [appRole](approle.md). The possible values are: `Scope` (for OAuth 2.0 permission scopes) or `Role` (for app roles).|
 
 ## JSON representation
 
--- a/api-reference/v1.0/resources/resourceaccess.md
+++ b/api-reference/v1.0/resources/resourceaccess.md
@@ -16,8 +16,8 @@ Object used to specify an OAuth 2.0 permission scope or an app role that an appl
 ## Properties
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
-|id|GUID|The unique identifier for one of the [oauth2PermissionScope](permissionscope.md) or [appRole](approle.md) instances that the resource application exposes.|
-|type|String|Specifies whether the **id** property references an [oauth2PermissionScope](permissionscope.md) or an [appRole](approle.md). The possible values are: `Scope` (for OAuth 2.0 permission scopes) or `Role` (for app roles).|
+|id|Guid|The unique identifier of an [app role](approle.md) or [delegated permission](permissionScope.md) exposed by the resource application. For delegated permissions, this should match the **id** property of one of the [delegated permissions](permissionscope.md) in the **oauth2PermissionScopes** collection of the resource application's [service principal](serviceprincipal.md). For app roles (application permissions), this should match the **id** property of an [app role](approle.md) in the **appRoles** collection of the resource application's [service principal](serviceprincipal.md).|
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/resources/resourceaccess.md
```diff
@@ -16,8 +16,8 @@ Object used to specify an OAuth 2.0 permission scope or an app role that an appl
 ## Properties
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
-|id|GUID|The unique identifier for one of the [oauth2PermissionScopes](permissionscope.md) or [appRole](approle.md) instances that the resource application exposes.|
-|type|String|Specifies whether the **id** property references an [oauth2PermissionScopes](permissionscope.md) or an [appRole](approle.md). The possible values are: `Scope` (for OAuth 2.0 permission scopes) or `Role` (for app roles).|
+|id|GUID|The unique identifier for one of the [oauth2PermissionScope](permissionscope.md) or [appRole](approle.md) instances that the resource application exposes.|
+|type|String|Specifies whether the **id** property references an [oauth2PermissionScope](permissionscope.md) or an [appRole](approle.md). The possible values are: `Scope` (for OAuth 2.0 permission scopes) or `Role` (for app roles).|
```
jiasli, line 20: Or perhaps we should simply use `permissionScope`?

FaithOmbongi: Thank you @jiasli for raising catching that and raising this PR. Yes, you're right, should be **permissionScope**. But for the sake of clarity (because oauth2PermissionScopes is the other term for delegated permissions), I also recommend that we add the word "oauth2PermissionScope" in brackets to remove ambiguity. I've updated your comment to a change suggestion.

psignoret: We should use "permissionScope" if we're referring to the type, and "oauth2PermissionScopes" if we're referring to the property. We should avoid using the singular form "oauth2PermissionScope", as that is neither a property nor a type name. We can use the terms "delegated permission" and "app role" when we are referring to these things generally, rather than to a type or property name.

```suggestion
|id|Guid|The unique identifier of an [app role](approle.md) or [delegated permission](permissionScope.md) exposed by the resource application. For delegated permissions, this should match the **id** property of one of the [delegated permissions](permissionscope.md) in the **oauth2PermissionScopes** collection of the resource application's [service principal](serviceprincipal.md). For app roles (application permissions), this should match the **id** property of an [app role](approle.md) in the **appRoles** collection of the resource application's [service principal](serviceprincipal.md).|
|type|String|Specifies whether the **id** property references a [delegated permission](permissionscope.md) or an [app role](approle.md) (application permission). The possible values are: `Scope` (for delegated permissions) or `Role` (for app roles).|
```

FaithOmbongi (approved):
Hi @jiasli, approved with @psignoret's change suggestion.

api-reference/v1.0/resources/resourceaccess.md
jiasli: Great suggestion! This is crystal clear!

jiasli:
@psignoret, how is the swagger synched with the document? 

https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/2d2ce81e3f093a9c6e8642beac37c868017b9c63/openApiDocs/v1.0/Applications.yml#L7757

```
          description: The unique identifier for one of the oauth2PermissionScopes or appRole instances that the resource application exposes.
```

api-reference/v1.0/resources/resourceaccess.md
jiasli: > We can use the terms "delegated permission" and "app role" when we are referring to these things generally, rather than to a type or property name.

On the second thought, won't this add understanding cost to the reader? Like using `app role` for `appRole`, `delegated permissions` for `permissionScope`. These terms becomes unclear especially when used in a context where hyperlink is not available, like CLI in-tool help.

jiasli:
@psignoret, thanks! So everything I made in this repo will finally be synchronized to the swagger?

psignoret:
@jiasli Yes. For example, [this change](https://github.com/microsoftgraph/microsoft-graph-docs/pull/14390/files#diff-5f23b7ddb3ca0f935792589d50914b89bc55f4f40251cf6883103066edc73279R96) I made to in this repo showed up in [cleaned and annotated metadata](https://github.com/microsoftgraph/msgraph-metadata/commit/d6087757362f53742403faf159a4646d1209aff9#diff-874b16b956e291a0700568376552444493dd8c126a492065fd8cbe5dd00144e7) a few days later, and then [in the OpenAPI spec](https://github.com/microsoftgraph/msgraph-sdk-powershell/commit/f6a28e313fee77971bfa5c7a7e792bae43e70d3e#diff-3d1d486b782406677b8943f90aa126fbc12744680cfd3c35e43f9a91531a8e60R6792) a few days after that. This is all described in the article I linked to in my previous comment.

FaithOmbongi (approved):

PR #15523, opened 2022-01-14
Remove known issue because it's by design

FaithOmbongi:
@lramosvea, FYI.


Lauragra:
@FaithOmbongi , I want to discuss this a bit more before merging. I want to better understand why we think some things aren't a  known issue if they're "by design", and whether it still might be helpful to include them in the Known issues list for discoverability, of the issue and workaround.

PR #15528, opened 2022-01-14
[Azure ABAC] Custom security attribute with allowed values

rolyon:


api-reference/beta/api/customsecurityattributedefinition-update.md
```diff
@@ -117,7 +117,57 @@ Content-Type: application/json
...
++ Attribute set: `Engineering`
++ Attribute: `Project`
++ Attribute data type: Collection of Strings
++ Update predefined value: `Baker`
++ New predefined value: `Skagit`
+
+> [!NOTE]
+> For this request, you must add the following to the header:<br/>
+> Key: odata-version<br/>
+> Value: 4.01
```
FaithOmbongi, line 133: Hi @rolyon, is there a reason why only this scenario requires the header?
Also, should we then add this info to the **Request headers** sections of both the Update and POST operations?

FaithOmbongi (changes requested):
Thank you @rolyon for your work on this PR. Let a Q about the purpose of the header and scenarios where it's required. Let's clarify this internally.

FaithOmbongi (changes requested):
Review feedback left here: https://github.com/microsoftgraph/microsoft-graph-docs/pull/15528#pullrequestreview-862122095

api-reference/beta/api/customsecurityattributedefinition-update.md
loradule: Hi Faith,

The update implementation leverages a feature called delta update that requires OData 4.01. This is specific to update, the POST does not need this particular header. Hope this helps!

FaithOmbongi (commented):
Thank you @rolyon, @loradule for the clarification. Left inline suggestions based on the feedback.

rolyon:
Hi @FaithOmbongi 
See doc updates.
thanks

FaithOmbongi (approved):
Thank you @rolyon for your work on this PR and for bringing this new capability to the attention of our customers.

PR #15563, opened 2022-01-19
Permission reference update

idwilliams:
Updating the permission reference update page to include details for the managed tenant scopes. Address issue #15011


Jackson-Woods (approved):

angelgolfer-ms (approved):
@idwilliams , I reversed the order of the 2 tables to be consistent with the rest of the topic. Also added a line to separate your section with the next, to follow the convention in the topic and for better scanning.
(I am still trying to recall how we missed that from PR 13285. :-|)

PR #15581, opened 2022-01-20
Update extended-properties-overview.md

abheek-das:
Minor fix in property value in the example.


abheek-das:
@isvargasmsft This PR has a minor fix in the doc page. However, there's 2 links which are coming as a warning while build. While this is not related to my changes, but looks like the link is anyways broken in the public docs as well that's already published today. Could you please help?

isvargasmsft:
Hi @abheek-das, 

As far as I understand, those links are referencing the Graph metadata. I am checking if there is a way to reference the metadata using relative links. If not, I'll change both to the absolute link to fix this. 

Please bear with me while I figure this out. 

Kind regards, 

Isaac

isvargasmsft (approved):

PR #15620, opened 2022-01-24
Update example request

SkylerStearns:
Documentation shows options "text" / "html" but example shows type of "String" which fails on usage.


SkylerStearns:
Hi @Lauragra, I appreciate you taking the time to look into this issue and respond, however I believe that this JSON representation is somewhat misleading, as it appears to be an example of the API usage. The solution I provided may not be the most ideal, but it would alleviate some confusion in my opinion. Respectfully speaking, I would recommend having someone on your team take another look at this documentation and making it a bit more clear that it does not intend to be an example with actual values as you suggest. Again, thank you for the response.

PR #15626, opened 2022-01-25
Added full URLs for token POST requests

Jackson-Woods:
Added two https://login.microsoftonline.com/ segments to URLs for POST requests to get tokens on this page.

### DIFF
--- a/concepts/auth-v2-user.md
+++ b/concepts/auth-v2-user.md
@@ -106,7 +106,7 @@ Your app uses the authorization `code` received in the previous step to request 
 ```
 // Line breaks for legibility only
 
-POST /{tenant}/oauth2/v2.0/token HTTP/1.1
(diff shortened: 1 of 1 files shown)


Jackson-Woods:
@hpsin Minor change but just double checking that these URL paths weren't truncated on purpose.

hpsin:
They were truncated on purpose for standard formatting.  This is how HTTP requests look on the wire - it's a bit funny looking but that's how HTTP works when aking direct requests (and through a proxy they look different).

Jackson-Woods:
> They were truncated on purpose for standard formatting. This is how HTTP requests look on the wire - it's a bit funny looking but that's how HTTP works when aking direct requests (and through a proxy they look different).

Ah, I see - thank you. This confused a user who raised an issue asking for the full path. Is there a clarification we could add to the page instead?

PR #15765, opened 2022-02-07
Update create-onlinemeeting-user-token-csharp-snippets.md

ravi16a87:
There is no documentation available on "AdditionalData" property and how to use this to override default OnlineMeeting Properties. Hence proposing these changes.

### DIFF
--- a/api-reference/v1.0/includes/snippets/csharp/create-onlinemeeting-user-token-csharp-snippets.md
+++ b/api-reference/v1.0/includes/snippets/csharp/create-onlinemeeting-user-token-csharp-snippets.md
@@ -6,11 +6,17 @@ description: "Automatically generated file. DO NOT MODIFY"
 
 GraphServiceClient graphClient = new GraphServiceClient( authProvider );
 
+//To override onlineMeeting Properties, like For E.g. You need to disable the chat for this OnlineMeeting.
+ IDictionary<string, object> additionalData = new Dictionary<string, object>();
+ additionalData.Add("allowMeetingChat", "disabled");
+                  
+
 var onlineMeeting = new OnlineMeeting
 {
 	StartDateTime = DateTimeOffset.Parse("2019-07-12T21:30:34.2444915+00:00"),
 	EndDateTime = DateTimeOffset.Parse("2019-07-12T22:00:34.2464912+00:00"),
-	Subject = "User Token Meeting"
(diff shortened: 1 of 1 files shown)


ravi16a87:
Hello @andrueastman @Lauragra , did you get a chance to look at this pull request?

andrueastman:
Hey @ravi16a87,

Apologies for the delayed response. 
As the C# snippets are autogenerated from the HTTP example, merging this PR will lead to the changes eventually being overridden. 

Since this is an issue with a missing property in the HTTP example, can you only update the HTTP example in the docs with the missing example? The snippet generator will pick this up and update the sample on its next run after this.

ravi16a87:
Hi @andrueastman , Thanks for your reply. Sure I will do it and create a new PR.

andrueastman:
No worries @ravi16a87. If it's okay, I'll close this for now as you will create a separate PR for this.

ravi16a87:
Hello @andrueastman , With respect to HTTPS you can directly override like this :
{
    "startDateTime": "2022-07-12T14:30:34.2444915-07:00",
    "endDateTime": "2022-07-12T15:00:34.2464912-07:00",
    "subject": "User Token Meeting",
    "allowMeetingChat": "disabled"
}

but when it comes to c# , we cannot pass object like this, we need to have additional data object and pass properties in that object to override the properties

andrueastman:
Thanks @ravi16a87.

I believe the C# SDK has the property for "allowMeetingChat" as seen [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/5da2eb3285030ee1ece1c0d7870a19e164c71bc2/src/Microsoft.Graph/Generated/model/OnlineMeeting.cs#L50). So, it might not be necessary to use the additionalData bag in this scenario. 

The snippet generator will therefore be able to generate the correct snippet as below once the PR is merged. 

```cs
            var onlineMeeting = new OnlineMeeting
            {
                StartDateTime = DateTimeOffset.Parse("2022-07-12T21:30:34.2444915+00:00"),
                EndDateTime = DateTimeOffset.Parse("2022-07-12T22:00:34.2464912+00:00"),
                Subject = "User Token Meeting",
                AllowMeetingChat = MeetingChatMode.Disabled
            };

            await graphClient.Me.OnlineMeetings
                .Request()
                .AddAsync(onlineMeeting);
```

PR #15782, opened 2022-02-08
Update bookingbusiness-list-calendarview.md

bonnyjohn123:
Based on the review, it appears that we only support filling customer info in "customers" property in GET appointment Call and not List appointments/calendarView as the "customers" is a list of complex type bookingCustomerInformation which is bigger payload.
This public documentation shows contradicting information in the Example section, which needs to be corrected with updated response.

### DIFF
--- a/api-reference/v1.0/api/bookingbusiness-list-calendarview.md
+++ b/api-reference/v1.0/api/bookingbusiness-list-calendarview.md
@@ -90,12 +90,12 @@ Content-type: application/json
             "joinWebUrl": null,
             "customerTimeZone": null,
             "serviceId": "d1e3db10-e679-4aac-aad4-1a6dccc77777",
-            "serviceName": "Initial consult",
+            "serviceName": "Catered bento",
             "duration": "PT1H",
             "preBuffer": "PT0S",
             "postBuffer": "PT0S",
-            "priceType": "notSet",
-            "price": 0.0,
+            "priceType": "FixedPrice",
+            "price": 10,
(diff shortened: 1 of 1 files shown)


Lauragra:
@bonnyjohn123 , if you're updating a topic in the /v1.0 folder, please also apply the same update to the comparable topic in the /beta folder. Please also address the validation error:

image

api-reference/v1.0/api/bookingbusiness-list-calendarview.md
```diff
@@ -80,68 +80,40 @@ HTTP/1.1 200 OK
...
-            "serviceId": "57da6774-a087-4d69-b0e6-6fb82c339976",
-            "serviceName": "Catered bento",
-            "duration": "PT30M",
-            "preBuffer": "PT5M",
-            "postBuffer": "PT10M",
-            "priceType": "fixedPrice",
-            "price": 10,
+            "additionalInformation": "",
+            "isLocationOnline": false,
+            "joinWebUrl": null,
```
smallick60, line 90: Same comment here too, as previous PR:

Any reson why value for these properties are changed too?

I think you could have done only these:

Remove "customers" property from top and add customers as empty array just below reminders, as already done here.
Have staffMemberIds filled with one Id, like you have already done.
And changing @odata.context as you have already done.
Have additionalInformation property too, as already done by you.

Lauragra:
Closing due to merge conflict and lack of response.

PR #15783, opened 2022-02-08
Update bookingbusiness-list-appointments.md

bonnyjohn123:
Based on the review, it appears that we only support filling customer info in "customers" property in GET appointment Call and not List appointments/calendarView as the "customers" is a list of complex type bookingCustomerInformation which is bigger payload.
This public documentation shows contradicting information in the Example section, which needs to be corrected with updated response.


Lauragra:
@bonnyjohn123 , if you are updating a topic in the v1.0 folder, please also apply the same update to the topic in /beta. Also, your PR has a validation error, please address it before we merge:

image

api-reference/v1.0/api/bookingbusiness-list-appointments.md
```diff
@@ -66,68 +66,40 @@ HTTP/1.1 200 OK
...
+                "timeZone": "America/Chicago"
             },
             "endDateTime": {
-                "dateTime": "2018-04-30T13:30:00.0000000Z",
-                "timeZone": "UTC"
+                "dateTime": "2022-01-27T22:00:00-06:00",
+                "timeZone": "America/Chicago"
             },
             "serviceLocation": {
-                "displayName": "Customer location (987 Third Avenue, Buffalo, NY 98052, USA)",
```
smallick60, line 130: Any reson why value for these properties are changed too?

I think you could have done only these:

1. Remove "customers" property from top and add customers as empty array just below reminders, as already done here.
2. Have staffMemberIds filled with one Id, like you have already done.
3. And changing @odata.context as you have already done.

bonnyjohn123: Thanks, if you take a look at the old response in the article and compare it side by side with the new response, there are much more changes in the metadata. So instead of patching the old response by editing one by one, its better to use a fresh response which is up to date (And then if we need to make any additional adjustments, we can do that)

Lauragra:
Closing due to merge conflict and lack of response.

PR #15784, opened 2022-02-08
Go SDK updates

jasonjoh:
- Added Go SDK dependencies to installing doc
- Added Go examples to paging doc


concepts/sdks/paging.md
```diff
@@ -116,6 +116,47 @@ while(messagesPage != null) {
...
+
+result, err := client.Me().Messages().Get(&options)
+
+// Initialize iterator
+pageIterator, err := msgraphcore.NewPageIterator(result, adapter.GraphRequestAdapterBase,
+    func() serialization.Parsable {
+        return messages.NewMessagesResponse()
+    })
+
+pageIterator.SetHeaders(map[string]string{"Content-Type": "application/json"})
```
baywet, line 149: @jasonjoh it should work without requiring you to set that header. What happens if you remove it?

concepts/sdks/paging.md
```diff
@@ -116,6 +116,47 @@ while(messagesPage != null) {
...
+    "github.com/microsoftgraph/msgraph-sdk-go/models/microsoft/graph"
+)
+
+query := messages.MessagesRequestBuilderGetQueryParameters{
+    Select: []string{"body", "sender", "subject"},
+}
+
+options := messages.MessagesRequestBuilderGetOptions{
+    H: map[string]string{
+        "Prefer": "outlook.body-content-type=\"text\"",
```
baywet, line 136: this header probably needs to be passed to the page iterator as well to get consistent results across pages.

baywet (changes requested):
Thanks for putting this together, I think you might have discovered a new issue with the implementation.

concepts/sdks/paging.md
jasonjoh: Yes you're right. I copied this from the original PR that added the iterator to the SDK, and assumed it was needed. I meant to try it without but honestly forgot it was there :D

concepts/sdks/paging.md
```diff
@@ -116,6 +116,49 @@ while(messagesPage != null) {
...
+
+// Initialize iterator
+pageIterator, err := msgraphcore.NewPageIterator(result, adapter.GraphRequestAdapterBase,
+    func() serialization.Parsable {
+        return messages.NewMessagesResponse()
+    })
+
+// Any custom headers sent in original request should also be added
+// to the iterator
+pageIterator.SetHeaders(map[string]string{"Prefer": "outlook.body-content-type=\"text\""})
```
baywet, line 151:
```suggestion
pageIterator.SetHeaders(options.H)
```

I'm not dead set on that if you think that's bad for readability, but we could promote reusability this way

baywet (approved):

concepts/sdks/paging.md
jasonjoh: Yeah, I think this is a good change.

baywet (approved):

Lauragra (approved):

PR #15808, opened 2022-02-10
[Doc Improvement][Update note in list tabs]

v-abirade:


api-reference/beta/api/channel-list-tabs.md
```diff
@@ -47,14 +47,17 @@ This method supports the $filter, $select, and $expand [OData query parameters](
 Do not supply a request body for this method.
 
 ## Response
+
 If successful, this method returns a `200 OK` response code and collection of [tabs](../resources/teamstab.md) objects in the response body.
 
+> **Note:** Files tab is native to channel or chat, and is not part of the list tabs API.
```
Lauragra, line 53: I suggest rephrasing this to say: "The Files tab is native to a channel or chat and is not returned by this API." Is there a need to clarify this in the summary description as well - i.e. "Retrieve the list of installed tabs..."?

Separate comment - if this also applies to the List tabs in chat operation, please make the same update to that topic: https://docs.microsoft.com/en-us/graph/api/chat-list-tabs?view=graph-rest-1.0&tabs=http

PR #15851, opened 2022-02-15
Added missing & to expand series master event example

0xVesion:


0xVesion:
Thank you :) @Lauragra

PR #15867, opened 2022-02-15
Adding Conversation Member Role Validation for Tenant Gust User

evanypark:
Adding Conversation Member Role Validation for Tenant Guest User when Chat is created.
Request will fail when tenant guest user is marked as "owner".

### DIFF
--- a/api-reference/beta/api/chat-post.md
+++ b/api-reference/beta/api/chat-post.md
@@ -48,7 +48,7 @@ The following table lists the properties that are required to create a chat obje
 |:---|:---|:---|
 |topic|(Optional) String|The title of the chat. The chat title can be provided only if the chat is of `group` type.|
 |chatType|[chatType](../resources/chat.md#chattype-values)| Specifies the type of chat. Possible values are: `group` and `oneOnOne`. |
-|members|[conversationMember](../resources/conversationmember.md) collection|List of conversation members that should be added. Every user, including the user who initiates the create request, who will participate in the chat must be specified in this list. Each member must be assigned a role of `owner` or `guest`. Guest tenant users must be assigned the `guest` role.|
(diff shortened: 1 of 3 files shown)


api-reference/beta/api/chat-post.md
```diff
@@ -93,6 +93,43 @@ Content-Type: application/json
   ]
 }
 ```
+
+### Example 2: Create a group chat with a guest tenant user
```
Lauragra, line 97: This example is added to the wrong location in the file; it currently appears as part of the request of Example 1. Please move it to follow the last example in the topic, Example 4 (or renumber the examples if you want this to be the second example in the topic). Please also add the response for this example.

api-reference/v1.0/api/chat-post.md
```diff
@@ -85,6 +85,43 @@ Content-Type: application/json
   ]
 }
 ```
+
+### Example 2: Create a group chat with a guest tenant user
```
Lauragra, line 89: Please make the same fix to this topic - the example is currently in the wrong location.

changelog/Microsoft.Teams.Core.json
```diff
@@ -1,6 +1,42 @@
 {
 
   "changelog": [
+    {
```
Lauragra, line 4: I don't think that you need a changelog entry for this PR because you aren't documenting an API-level change; this is just a clarification of one property and a new example, right?

evanypark: Yes, and clarification on that guest users needs to be noted as guest when creating chat. there wasn't any API feature change that caused this doc update.

Lauragra (approved):

Lauragra (approved):

PR #15905, opened 2022-02-18
Making Note to requests using UPN that UPN is not same as user email.

evanypark:
Adding Warning Notes to Examples with UPN that "User principal name does not equal to email"

### DIFF
--- a/api-reference/beta/api/channel-post-members.md
+++ b/api-reference/beta/api/channel-post-members.md
@@ -216,6 +216,8 @@ Content-type: application/json
     "user@odata.bind": "https://graph.microsoft.com/beta/users('jacob@contoso.com')"
 }
 ```
+>**Note:** User principal name is not the same as email.
+
 # [C#](#tab/csharp)
 [!INCLUDE [sample-code](../includes/snippets/csharp/channel-add-member-3-csharp-snippets.md)]
 [!INCLUDE [sdk-documentation](../includes/snippets/snippets-sdk-documentation-link.md)]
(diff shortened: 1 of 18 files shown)


api-reference/beta/api/channel-post-members.md
```diff
@@ -216,6 +216,8 @@ Content-type: application/json
     "user@odata.bind": "https://graph.microsoft.com/beta/users('jacob@contoso.com')"
 }
 ```
+>**Note:** User principal name is not the same as email.
```
Lauragra, line 219: Global: I think this needs further clarification. Can you provide the definition of a UPN and explain how it is different than an email address? I suggest moving the note to directly follow the H3 heading, formatted as regular text, with a clear description of what the UPN is and how it differs from a user's email address.  

Also, separate comment - please replace the John Doe fictitious name in the examples here with an approved fictitious name from the fictitious names list.

Lauragra (changes requested):

api-reference/beta/api/channel-post-members.md
evanypark: @Lauragra 

These updates were to point out that UPN values are often in format of email (johndoe@microsoft.com) but not always equal to email value for the user. 

Where can I find the list of fictitious names?

Lauragra: @evanypark , you can generate names here: https://microsoft.sharepoint.com/sites/CELAWeb-Copyrights-Trademarks-And-Patents/SitePages/trademarks-fictitious-names.aspx

isvargasmsft (changes requested):
Hello, thank you very much for the work in this PR and the clarification note. 

Please add a sentence or a link for readers to consult if they need more information to find what the UPN is. The note just says what it is not.

Thank you!

PR #15919, opened 2022-02-21
Update application-post-onlinemeetings.md

ravi16a87:
With C# we need to use "AdditionalData" to override default properties like For E.g. Disabling the chat for a meeting. Hence proposing one small example on how to use and override properties which one can use to override multiple properties for an online meeting.

### DIFF
--- a/api-reference/v1.0/api/application-post-onlinemeetings.md
+++ b/api-reference/v1.0/api/application-post-onlinemeetings.md
@@ -263,6 +263,93 @@ Content-Type: application/json
 }
 ```
 
+### Example 3: Create an online meeting with user token and override default meeting properties
+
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/api/application-post-onlinemeetings.md
```diff
@@ -263,6 +263,93 @@ Content-Type: application/json
...
 ```
 
+### Example 3: Create an online meeting with user token and override default meeting properties
+
+The following example creates an online meeting with a user token and override default meeting properties "allowMeetingChat" to disabled.
+
+#### Request
+
+
+# [C#](#tab/csharp)
```
Lauragra, line 273: The examples in the documentation show the REST API call. The SDK snippets are added via our snippet generation process. Please update this to include only the HTTP example. Thanks.

ravi16a87:
Not required

PR #15924, opened 2022-02-21
Indicating required properties in profile

clq:


api-reference/beta/resources/educationalactivitydetail.md
```diff
@@ -23,12 +23,14 @@ Represents additional detail about an undergraduate, graduate, postgraduate degr
...
 |awards        |String       |Any awards or honors associated with the program.              |
 |description   |String       |Short description of the program provided by the user.         |
-|displayName   |String       |Long-form name of the program that the user has provided.      |
+|displayName<sup>*</sup>|String       |Long-form name of the program that the user has provided.      |
 |fieldsOfStudy |String       |Majors and minors associated with the program. (if applicable) |
 |grade         |String       |The final grade, class, GPA or score.                          |
 |notes         |String       |Additional notes the user has provided.                        |
 |webUrl        |String       |Link to the degree or program page.                            |
 
+<sup>*</sup> Required property
```
Lauragra, line 32: Is there a way that we can just say "Required." in the description? Also, what do we mean by the fact that the property is required? In what contexts, i.e. in a POST operation? Would it make more sense to mention what properties are required in the applicable method topics?

Lauragra:
@clq , do you still want to publish this PR? If not, I will close. Thanks.

PR #15959, opened 2022-02-23
Jupetter licensing

jpettere:
Updated author for user-list-licenseDetails to be team that owns licensing


FaithOmbongi (approved):
Thank you @jpettere for correcting this ownership detail. Approved.

PR #15993, opened 2022-02-25
Remove extraneous "Device management" items in TOC

andyrobbins:
The same "Device management" item is listed 5 times in a row with no differences between each item. This removes 4 of those items, leaving just 1.


Lauragra:
@Erikre , can you please work with Jai to make sure the TOC for v1.0 and beta is updated? Thanks.

Erikre (approved):

PR #16057, opened 2022-03-03
EM: update case in filter used in role assignment query

markwahl-msft:
Address issue with OData case sensitivity constraint reported by customer.


FaithOmbongi (approved):
Thank you @markwahl-msft for raising this PR. Hope this case sensitivity can be removed in the future.

PR #16063, opened 2022-03-04
Use Graph Query to retrieve Service Principal added as members in Azu…

HAMATHAR:
Use Graph Query to retrieve Service Principal added as members in Azure AD group

We have observed that Service Principals added to Azure AD Group are not returned by PowerShell, however can be retrieved using graph query mentioned in the proposed change. If updated on the document this would help us deflect small amount of customer cases created for Azure Group Management Support Topic.

### DIFF
--- a/api-reference/beta/api/group-list-members.md
+++ b/api-reference/beta/api/group-list-members.md
@@ -325,4 +325,44 @@ Content-type: application/json
 }
 -->
 
+### Example 6: Use OData value to retrieve Service Principal added as members in Azure AD group.
 
+#### Request
+
+The following is an example of the request.
+
+<!-- {
+  "blockType": "request",
+  "name": "get_pr_count"
+}-->
+```msgraph-interactive
+GET https://graph.microsoft.com/beta/groups/{GroupId}/members/microsoft.graph.servicePrincipal
+ConsistencyLevel: eventual
+```
+
+#### Response
+
+The following is an example of the response.
+>**Note:** The response object shown here might be shortened for readability.
+
+<!-- {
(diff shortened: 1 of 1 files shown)


HAMATHAR:
I have replaced the group ID with a dummy value {GroupId} mentioned in the sample query for propose change.

api-reference/beta/api/group-list-members.md
```diff
@@ -325,4 +325,45 @@ Content-type: application/json
...
+#### Request
+
+The following is an example of the request.
+
+<!-- {
+  "blockType": "request",
+  "name": "get_members_serviceprincipals"
+}-->
+```msgraph-interactive
+GET https://graph.microsoft.com/beta/groups/{GroupId}/members/microsoft.graph.servicePrincipal
```
FaithOmbongi, line 339: Hi @HAMATHAR, please replace this placeholder **{GroupId}** with a real GUID.

FaithOmbongi (approved):
Thank you @HAMATHAR for raising this PR. I've left a suggestion to replace the placeholder value **{GroupId}** with a real GUID. Customers tell us they find examples useful when real values are used.

Approved with suggestions.

api-reference/beta/api/group-list-members.md
HAMATHAR: I have replace placeholder {GroupId} with a real GUID on the proposed change

FaithOmbongi: Thank you. I've added the PR to Tuesday's merge queue.

PR #16086, opened 2022-03-07
Additional documentation for Excel Graph throttling

chiz-ms:
Adding more detailed explanations and instructions for Excel Graph throttling.


concepts/throttling.md
```diff
@@ -492,4 +494,4 @@ The following limits apply to any type of requests for service communications un
 | Request type |  Limit per app per tenant |
 | ------------ | ------------------------ |
 | Any | 240 requests per 60 seconds |
-|Any | 800 requests per hour |
```
chiz-ms, line 495: It seems GitHub web editor changed this automatically, but I'm not sure what is changed. Maybe line-ending?

concepts/workbook-error-codes.md
```diff
@@ -64,9 +64,9 @@ The following are some additional errors that your app might encounter within th
 | **payloadTooLargeUncategorized**              | The request payload exceeds the size limit.
 | **rangeExceedsLimit**         | The cell count in range has exceeded the maximum supported number.
 | **requestAborted**         | The request was aborted during run time.
-| **serviceUnavailableUncategorized**      | The service is temporarily unavailable or is overloaded. Please try your request again.
-| **tooManyRequestsUncategorized**             | The client app or user has been throttled. Please try your request again.
-| **transientFailure**           | The request failed due to a transient error. Please try your request again.
+| **serviceUnavailableUncategorized**      | The service is temporarily unavailable or is overloaded. Please try your request again after the cooldown duration specified by `Retry-After` header.
+| **tooManyRequestsUncategorized**             | The client app or user has been throttled. Please try your request again after the cooldown duration specified by `Retry-After` header. Refer to https://docs.microsoft.com/graph/workbook-best-practice#throttling for best practices to reduce throttling.
```
chiz-ms, line 68: This corresponds to the error message in response body, so we couldn't use `[title](link)` grammar.

concepts/throttling.md
isvargasmsft, line 495: Removed line 495: Was this intended or should we revert this change?

concepts/workbook-best-practice.md
```diff
@@ -283,3 +283,19 @@ Content-type: application/json
...
+
+## Throttling
+
+### Why not simple limit numbers?
+
+We monitor the resource usage of multiple dependency services when the APIs are called, and the impact varies a lot between different requests. For example, you may easily imagine that updating a short string in a single cell of a small workbook would consume much less resources than adding a huge table with complex formulas to a large workbook. Even with the same API, parameters and target workbooks may introduce significant differences. Using simple and universal numbers would result in more restrictive limits; in contrast, following the best practices here would help Graph clients to complete tasks fast with fewer failures.
+
+### Retry-After header
+
+In many cases, a throttling response will include a `Retry-After` header. Respecting the value of the header and delaying similar requests would help the client to recover from throttling. See [Error handling for Excel APIs in Microsoft Graph](workbook-error-handling.md) for detailed instructions on how to handle error responses from Excel Graph.
```
isvargasmsft, line 295: Very last sentence: "... handle responses from Excel Graph".

Can we rephrase to "Excel in Microsoft Graph"?

isvargasmsft (approved):

isvargasmsft:
Hi @Lauragra, this one is ready for final review :)

Thanks!

concepts/throttling.md
```diff
@@ -418,6 +418,8 @@ The preceding limits apply to the following resources:
 
 ### Excel service limits
 
+See [Throttling](workbook-best-practice.md#throttling) for explanations and best practices related to Excel service throttling. The following numbers represent part of the limits.
```
Lauragra, line 421: The standard construction for cross-references is to lead with the subject matter and then provide the link. Please revise as:
For explanations and best practices related to Excel service throttling, see [Throttling](workbook-best-practice.md#throttling).

concepts/workbook-best-practice.md
```diff
@@ -283,3 +283,19 @@ Content-type: application/json
 ```
 
 >**Note:** Acquire session information depends on the initial request. You don't need to acquire the result if the initial request doesn't return a response body.
+
+## Throttling
+
+### Why not simple limit numbers?
```
Lauragra, line 289: This heading assumes that the reader has the context that the throttling numbers for the Excel service are not simple limits. Have you set that context for the reader? It might be helpful to add an introductory sentence after the Throttling H2 that explains that and leads to this heading for further explanation.

concepts/workbook-best-practice.md
```diff
@@ -283,3 +283,19 @@ Content-type: application/json
...
+
+We monitor the resource usage of multiple dependency services when the APIs are called, and the impact varies between different requests. For example, you may easily imagine that updating a short string in a single cell of a small workbook would consume much less resources than adding a big table with complex formulas to a large workbook. Even with the same API, parameters and target workbooks can introduce significant differences. Using simple and universal numbers would result in more restrictive limits; in contrast, the following practices will help Microsoft Graph consumers complete tasks quickly with fewer failures.
```
Lauragra, line 291: We don't use first person plural in our content, we use second person, addressing the reader as "you". For details, see this topic in the Writing Style Guide: https://docs.microsoft.com/en-us/style-guide/grammar/person. Please revise the first sentence. This might be a case when you would use passive voice, i.e. "When you call Excel service APIs, it can affect the resource usage of multiple dependent services, and the impact can vary between different requests."

Lauragra, line 291: For readability for a global audience, please revise the second sentence as follows:
For example, you might expect that updating a short string in a single cell of a small workbook would consume fewer resources than adding a big table with complex formulas to a large workbook.

Lauragra, line 291: In the last sentence, it might help to clarify that you're referring to throttling-related errors/failures? These best practices are specifically to prevent throttling, let's make sure that that's clear.

concepts/workbook-best-practice.md
```diff
@@ -283,3 +283,19 @@ Content-type: application/json
...
+
+### Retry-After header
+
+In many cases, a throttling response includes a `Retry-After` header. Respecting the value of the header and delaying similar requests would help the client recover from throttling. See [Error handling for Excel APIs in Microsoft Graph](workbook-error-handling.md) for detailed instructions on how to handle error responses from Excel API in Microsoft Graph.
```
Lauragra, line 295: Revise the second sentence as:
For details about handling error responses from Excel APIs in Microsoft Graph, see [Error handling for Excel APIs in Microsoft Graph](workbook-error-handling.md) .

concepts/workbook-best-practice.md
```diff
@@ -283,3 +283,19 @@ Content-type: application/json
...
+
+### Throttling and concurrency
+
+Another factor related to throttling is request concurrency. It isn't recommended to increase concurrency when using the Excel API (i.e. parallelizing the requests to the same workbook), especially for write requests. Instead, unless there is a specific concern, such as significant networking overhead compared to very short request execution time; we recommend sequential usage in the most common case: for each workbook, only send the next request after receiving a successful response to the current request.
```
Lauragra, line 299: In the second sentence, this is an exception to the first person rule, where we can revise it to say "We don't recommend increasing...". Please avoid using "i.e.", you can replace with "for example". In the last sentence, replace the semicolon with a comma.

concepts/workbook-best-practice.md
```diff
@@ -283,3 +283,19 @@ Content-type: application/json
...
+
+Concurrent write requests to the same workbook don’t usually execute in parallel (although in some cases they do); rather, they are often the cause of throttling, timeout (when requests are queued on servers), merge conflict (when concurrent sessions are involved) and other types of failures. They also complicate error handling, for example, when receiving a failure response, there is no way to confirm the status of other pending requests, making it difficult to determine or to recover the state of the workbook.
```
Lauragra, line 301: For global language, please change "execute" to "run".

Lauragra, line 301: Revise the last sentence as:
They also complicate error handling; for example, when you receive a failure response, there is no way to confirm the status of other pending requests, which makes it difficult to determine or to recover the state of the workbook.

concepts/workbook-error-codes.md
Lauragra, line 68: Revise the final sentence as:
For best practices to reduce throttling, see [Throttling](/graph/workbook-best-practice#throttling). Note that we don't expose full URLs for accessibility reasons, and we use relative links.

concepts/workbook-error-handling.md
```diff
@@ -105,7 +105,7 @@ For both the long-running operation pattern and the regular pattern, we recommen
...
 | **requestAborted**         | The request was aborted during run time, which was usually caused by long time calculation from functions in the workbook. The Microsoft Graph client is not expected to resend the failed request.
 | **serviceUnavailableUncategorized**      | The service is temporarily unavailable or is overloaded. The Microsoft Graph client is not expected to resend the failed request until the specified cooldown duration passes.
-| **tooManyRequestsUncategorized**             | The failed request exceeds certain frequency limitation. The Microsoft Graph client is not expected to resend the failed request until the specified cooldown duration passes.
+| **tooManyRequestsUncategorized**             | The failed request exceeds certain frequency limitation. The Microsoft Graph client is not expected to resend the failed request until the specified cooldown duration passes. Also refer to [Throttling](workbook-best-practice.md#throttling) for best practices to reduce throttling.
```
Lauragra, line 108: Revise as:
For best practices to reduce throttling, see [Throttling](workbook-best-practice.md#throttling).

concepts/workbook-best-practice.md
chiz-ms: Not only throttling errors; in the concurrency section (at the end) I gave a number of examples including throttling, timeout,  merge conflict, etc.. What should we do in this case?

concepts/workbook-error-codes.md
chiz-ms: This "message" column is in fact a copy of the strings we return as part of error responses. In error responses we're not able to use `[]()` syntax, and they're not part of the documentation, so I'm not aware of an alternative of full URLs. But if it's not appropriate here, I'll just remove all the messages.
By the way, please still review the message strings as they'll become a string resource change in our production. What's the case when a customer couldn't access this URL? Because the build report says
> Absolute link 'https://docs.microsoft.com/graph/workbook-best-practice#throttling' will be broken in isolated environments.

concepts/throttling.md
```diff
@@ -418,6 +418,8 @@ The preceding limits apply to the following resources:
 
 ### Excel service limits
 
+For explanations and best practices related to Excel service throttling, see [Throttling](workbook-best-practice.md#throttling). The following numbers represent part of the limits.
```
Lauragra, line 421: For the second sentence, this would be more clear: "The following are some of the limits." If these are only part or some of the limits though, where do we define the others?

chiz-ms: The other limits are complex and dynamic which could not be represented by simple numbers (explained in workbook-best-practice.md). What should we do in this case?

chiz-ms:
@Lauragra could you take a look at my replies to the active comments? Thanks.

chiz-ms:
@Lauragra The active comments disappeared from "Files changed" tab due to GitHub limitation, but you could find them in "Conversation" tab.

concepts/throttling.md
Lauragra: For the second sentence, I recommend the following revision: "In addition, the following are some throttling limits."

Lauragra:
> @Lauragra The active comments disappeared from "Files changed" tab due to GitHub limitation, but you could find them in "Conversation" tab.

I think I've replied on the one remaining active comment, just a minor update and then we'll be ready to merge. Thanks.

concepts/workbook-best-practice.md
chiz-ms: @Lauragra this is an active comment. Please see my reply above. I've updated this paragraph but I'm not sure whether that is ideal. The current version is:
> In contrast, we recommend the following practices for Excel API consumers to complete tasks quickly with fewer throttling errors.

concepts/workbook-error-codes.md
chiz-ms: @Lauragra this is another active comment. Please see my reply above.

Lauragra: It looks like you've deleted the descriptions from the detailed error codes from the topic, and hence the link is also no longer there. If you want to restore the error descriptions, I can review them. For the link, maybe we can style it with tic marks. So my suggestion for the updated text would be the same except that you have to use the full URL:

For best practices to reduce throttling, see `https://docs.microsoft.com/graph/workbook-best-practice#throttling`.

concepts/workbook-best-practice.md
Lauragra: Do we need to say "In contrast"? Suggested revision for that sentence:
"The following best practices will help you complete tasks more quickly with fewer throttling errors."

chiz-ms:
@Lauragra please see if we could complete this PR. Thanks.

Lauragra (approved):

PR #16147, opened 2022-03-11
[Doc Improvement][Update tenant id]

v-abirade:


api-reference/v1.0/api/application-post-calls.md
```diff
@@ -470,7 +470,8 @@ Content-Type: application/json
   "mediaConfig": {
     "@odata.type": "#microsoft.graph.serviceHostedMediaConfig",
     "removeFromDefaultAudioGroup": false
-  }
+  },
+  "tenantId": "aa67bd4c-8475-432d-bd41-39f255720e0a"
```
Lauragra, line 474: If the request includes a JSON representation of a **call** object, I don't see this property defined in the call resource topic. Can you help to clarify for the reader why this property is added to the JSON body for these two example?

PR #16168, opened 2022-03-14
Adding Note to provide context for property values.

adarshgh:
Based on @bivasmaiti26 comments, adding Note to include properties value as null.


bivasmaiti26 (commented):
Please delete the same sentence from line 250. Since we are already mentioning in the top, we don't need to mention this again.

isvargasmsft (approved):

PR #16315, opened 2022-03-25
Update user.md

expiscornovus:
Links were missing in the mail and proxyAddresses properties section. Added those for the Get User and Update User.


Lauragra:
Thank you!

PR #16323, opened 2022-03-28

David-Barrett-MS:
Group.Read.All and Group.ReadWrite.All application permissions cannot be used to access the calendar (per https://docs.microsoft.com/en-us/graph/api/calendar-get?view=graph-rest-1.0 ).  I've updated the description to remove the text that stated this was supported.


Lauragra:
@FaithOmbongi , can you please verify and then I'll publish? Thanks.

FaithOmbongi (changes requested):
Thank you @David-Barrett-MS for raising this PR and flagging this issue. Unfortunately, this display string should match what is presented for admin consent, even though the Group permissions may currently not allow you to use the Calendar APIs.

I propose that you pivot your PR to indicate the correct display string regardless of this erroneous mention of the Calendar permissions. I've added suggestions to your change to tweak the display string to what's currently shown for admin consent as per the screenshot below.

image

However, I've raised this issue internally and the team is working to resolve this display string before we can make the doc update with the new string. cc @psaffaie for visibility.

mrkeiths:
Hello,

@FaithOmbongi Thank you for catching the need to update admin consent verbiage.  I agree this should be updated too.

I believe the following must also be updated since it is very specific to Group.Read.All and Group.ReadWrite.All **_Application_** assigned permissions:

[https://docs.microsoft.com/en-us/graph/permissions-reference#application-permissions-29](https://docs.microsoft.com/en-us/graph/permissions-reference#application-permissions-29)
group_app_assigned_permissions

It's good that we plan to update admin consent verbiage to reflect limitations of Application assigned permission but the information above still suggests Application assigned permissions have the ability to read Group calendar events which is not accurate.  May we please have this documentation updated too?

Thank you, Keith

mrkeiths:
@FaithOmbongi   Can you please help us by merging this pull request for the doc update?

mrkeiths:
@FaithOmbongi there seems to be an issue that is preventing doc updates from being merged so they can be published.  Is anyone planning to resolve?  Is there a reason why this update shouldn't proceed still?

Thank you, Keith

FaithOmbongi:
Hi @mrkeiths - Unfortunately, only this PR is currently on hold - because of internal processes that need to be started first. I've escalated this request to the product team so we can remove all references of _Group.*_ permissions claiming to support calendar APIs. Please note that this "wrong" description is also in the API and in the Azure Portal so we are working get fixes for those underway first, then proceed with this docs PR, and make all three references consistent.

cc @psaffaie for visibility.

FaithOmbongi:
Hi @David-Barrett-MS, @mrkeiths - The `Group.Read.All` and `Group.ReadWrite.All` permission descriptions have now been corrected/fixed, and references to the "manage calendar" capability removed. I've updated the suggestion I shared in this PR to reflect the new descriptions. Please check and share any concerns.

Once you apply the suggestion, I can approve the PR for merge. It's taken a while to finally resolve the issue and I thank you for your patience and for pushing for this correction.

mrkeiths:
Hello @FaithOmbongi, I apologize for taking too long to review and share feedback as requested.

I reviewed Delegated and Application Group.Read.All and Group.ReadWrite.All changes in the following doc and everything looks good: https://docs.microsoft.com/en-us/graph/permissions-reference#group-permissions

I also see the same changes reflected in Delegated and Application Group.Read.All and Group.ReadWrite.All descriptions in the Azure Portal.

I think we should be good now.

Thank you everyone for your help,
Keith

PR #16333, opened 2022-03-29
Validating table columns

kanchm:
Method tables validated with "Return type" as the column header.
Excluding Intune* files at the moment.

No content change other than column header rename.

Review:
- driveitem -- "REST Path" is renamed to "Return type"
- site -- similar rename.


kanchm:
Abandoning this in favor of splitting by workload.

PR #16350, opened 2022-03-30
Enabling Expand Query Option for Channel List Message API

evanypark:


RamjotSingh (approved):

api-reference/beta/api/channel-list-messages.md
```diff
@@ -64,7 +70,9 @@ If successful, this method returns a `200 OK` response code and a collection of
 
 ## Example
```
isvargasmsft, line 71: Line 71: please revise to Examples, in plural.

api-reference/beta/api/channel-list-messages.md
```diff
@@ -271,6 +279,200 @@ Content-type: application/json
...
 ```
 
+### Example 2: Request with $expand query option on replies
+
+#### Request
+
+
+
+# [HTTP](#tab/http)
```
isvargasmsft, line 289: Line 289: please remove this line. Tabs are automatically generated.

api-reference/beta/api/channel-list-messages.md
```diff
@@ -271,6 +279,200 @@ Content-type: application/json
...
+<!-- {
+  "blockType": "request",
+  "name": "get_listchannelmessages_2"
+}-->
+```msgraph-interactive
```
isvargasmsft, line 294: Line 294: Please replace `msgraph-interactive` with `http`. These tabs are automatically generated.

api-reference/beta/api/chatmessage-delta.md
```diff
@@ -69,7 +69,10 @@ In subsequent requests, copy and apply the `nextLink` or `deltaLink` URL from th
...
-- `$filter` allows returning messages that meet a certain criteria. The only property that supports filtering is `lastModifiedDateTime`, and only the **gt** operator is supported. For example, `../messages/delta?$filter=lastModifiedDateTime gt 2019-02-27T07:13:28.000z` will fetch any messages created or changed after the specified date time.
+- `$filter` allows returning messages that meet a certain criteria. The only property that supports filtering is `lastModifiedDateTime`, and only the **gt** operator is supported. For example, `../messages/delta?$filter=lastModifiedDateTime gt 2019-02-27T07:13:28.000z` will fetch any **reply chain (each channel post message and associated reply messages)** created or changed after the specified date time.
+- `$expand` allows expanding properties for each channel message. Only **replies** is supported. If a channel messsage contains more than 1000 replies, `replies@odata.nextLink` will be provided for pagination. 
+
+> **Note:** For `$expand` query parameter, please refer to example 3 in [GET /teams/{team-id}/channels/{channel-id}/messages](channel-list-messages.md).
```
isvargasmsft, line 75: Line 75: please revise for simplicity to: "... please refer to [List channel messages] (insert here the link with # to example 3)

Is it example 3 or 2?

api-reference/v1.0/api/channel-list-messages.md
```diff
@@ -449,6 +455,201 @@ Content-type: application/json
 }
 ```
 
+### Example 3: Request with expand query option on replies
```
isvargasmsft, line 458: Why isn't example 2 of this topic reflected in the beta topic as well?

api-reference/beta/api/channel-list-messages.md
isvargasmsft: Please apply to the v1 topic as well.

api-reference/beta/api/channel-list-messages.md
isvargasmsft: Please apply to the v1 topic as well.

api-reference/beta/api/chatmessage-delta.md
isvargasmsft: Please apply to the v1 topic as well.

isvargasmsft (changes requested):
Hello @evanypark, 

Thank you very much for your work in this PR. 

Please take a look at a few comments regarding consistency and simplicity. Please apply the comments to both beta and v1 files. 

Feel free to ping me on the side if you have any questions. 

Kind regards, 

Isaac

api-reference/beta/api/chatmessage-delta.md
evanypark: @isvargasmsft What would the link look like for a section inside external file?

api-reference/v1.0/api/channel-list-messages.md
evanypark: was probably left out , adding to beta

isvargasmsft (approved):

changelog/Microsoft.Teams.Core.json
```diff
@@ -1,5 +1,41 @@
 {
   "changelog": [
+    {
+      "ChangeList": [
+        {
+          "Id": "fef256b6-cd3f-4ee8-89c6-7063f7068cab",
```
angelgolfer-ms, line 6: Lines 7-11:
The ChangedApiName should be an API element in the Microsoft Graph API. In this case, the API element that's impacted by this change is the replies nav property of chatMessage, when listing channel messages that are replies.

You can specify those lines of the changelist record as follows:
ApiChange: "Relationship",
ChangedApiName: "replies",
ChangeType: "Addition",
Description: "Added support for `$expand` on the **replies** relationship when [listing channel messages](https://docs.microsoft.com/en-us/graph/api/channel-list-messages?view=graph-rest-1.0) that are replies.",
Target: "chatMessage"

You can take a look at example 17 in this CDK topic: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog?anchor=changelist-record-examples

changelog/Microsoft.Teams.Core.json
```diff
@@ -1,5 +1,41 @@
...
+      "Cloud": "prd",
+      "Version": "beta",
+      "CreatedDateTime": "2022-04-01T13:04:49.407Z",
+      "WorkloadArea": "Teamwork",
+      "SubArea": ""
+    },
+    {
+      "ChangeList": [
+        {
+          "Id": "3580c1e4-e1bb-4611-843c-42e851863c3f",
```
angelgolfer-ms, line 24: Lines 25-29:
Please see my comment for the similar update in the beta version (lines 7-11).

api-reference/beta/api/channel-list-messages.md
```diff
@@ -43,7 +43,13 @@ GET /teams/{team-id}/channels/{channel-id}/messages
 
 ## Optional query parameters
 
-You can use the [$top](/graph/query-parameters#top-parameter) query parameter to control the number of items per response. Maximum allowed value for `$top` is 50.
+This method supports the following [OData query parameters](/graph/query-parameters).
+
+| Name      | Description          |
+|:----------|:---------------------|
+| [$top](/graph/query-parameters#top-parameter)| Controls the number of items per response. Maximum allowed `$top` value is 50. |
+| [$expand](/graph/query-parameters#expand)  | Currently supports **replies** only. If a channel message contains more than 1000 replies, replies@odata.nextLink will be provided for pagination.|
```
angelgolfer-ms, line 51: Line 51:
Please rephrase for clarity - use active voice to make it clearer who/what provides the nextLink and what the customer can do in that case.
For instance:
"You can apply `$expand` on only channel messages that are replies. By default, a response can include up to 1000 replies. For an operation that expands channel messages with more than 1000 replies, use the request URL returned in `replies@odata.nextLink` to get the next page of replies."

api-reference/beta/api/channel-list-messages.md
```diff
@@ -43,7 +43,13 @@ GET /teams/{team-id}/channels/{channel-id}/messages
 
 ## Optional query parameters
 
-You can use the [$top](/graph/query-parameters#top-parameter) query parameter to control the number of items per response. Maximum allowed value for `$top` is 50.
+This method supports the following [OData query parameters](/graph/query-parameters).
+
+| Name      | Description          |
+|:----------|:---------------------|
+| [$top](/graph/query-parameters#top-parameter)| Controls the number of items per response. Maximum allowed `$top` value is 50. |
```
angelgolfer-ms, line 50: Line 50:
Please clarify "items" and the behavior of $top.
I think "items" here means the number of channel messages returned per page?
FYI Microsoft docs use the Microsoft Voice which should be friendly, clear, and direct. Rather than just saying "this is how we designed this", in the docs, we want to phrase our descriptions as "this is what / how you can use our design to get your task done".
So I suggest you can rephrase this way please:
"Apply `$top` to specify the number of channel messages returned per page in the response. The maximum page size is 50 channel messages."

angelgolfer-ms: BTW, is there a default page size? It would be helpful to state it here, because specifying `$top` should be optional.

api-reference/beta/api/channel-list-messages.md
```diff
@@ -43,7 +43,13 @@ GET /teams/{team-id}/channels/{channel-id}/messages
```
angelgolfer-ms, line 43: This comment applies to how in general, you should please document support for a query parameter - in addition to documenting the support in the related operation topics like this one, you should update the description of the property or nav property in the parent resource topic. 
In this case for $expand, please update the description for the replies nav property in the chatmessage.md topic and add a separate sentence like "Supports `$expand`."
Take a look at the descriptions in the "Relationships" section of user.md.

api-reference/beta/api/channel-list-messages.md
```diff
@@ -271,6 +279,371 @@ Content-type: application/json
...
+            },
+            "channelIdentity": {
+                "teamId": "fbe2bf47-16c8-47cf-b4a5-4b9b187c508b",
+                "channelId": "19:4a95f7d8db4c4e7fae857bcebe0623e6@thread.tacv2"
+            },
+            "attachments": [],
+            "mentions": [],
+            "reactions": [],
+            "replies@odata.count": 3,
+            "replies@odata.nextLink": "https://graph.microsoft.com/v1.0/teams/fbe2bf47-16c8-47cf-b4a5-4b9b187c508b/channels/19:4a95f7d8db4c4e7fae857bcebe0623e6@thread.tacv2/messages/1616963377068/replies?$skiptoken=MSwwLDE2NDQ0MzkzODAxNDU",
```
angelgolfer-ms, line 525: Line 525:
Can you confirm if this nextLink line should be there?
Your line 51 in the same topic says:
"If a channel message contains more than 1000 replies, replies@odata.nextLink will be provided for pagination."

Here, the example response includes so far 3 replies. Where does the example request specify some page size for the number of replies per page?

angelgolfer-ms (changes requested):
Hi @evanypark ,
@isvargasmsft requested me to clarify how to specify the changelog for the updates in this PR. I took a quick look at what your PR does, and added a few other comments as well. Please take a look. Please apply the same comments to both v1 and beta versions of your updates.
And to clarify, @isvargasmsft is the reviewer to cover the entire PR.

changelog/Microsoft.Teams.Core.json
evanypark: shouldn't the target be channelMessage?

api-reference/beta/api/channel-list-messages.md
evanypark: Default is set to 20 for channel messages

api-reference/beta/api/channel-list-messages.md
evanypark: I think that might be misleading.
Expand is only available when applied on replies property for GET channel messages API. 
It isn't enabled anywhere else

api-reference/beta/api/channel-list-messages.md
evanypark: Correct, the actual response shouldn't contain next link unless it has more than 1000 replies, but I wanted to still provide an example with nextlink, even if the replies count was less than that. 
Any idea on how I should do that?

api-reference/beta/api/channel-list-messages.md
evanypark: I've added "Supports $expand for channel messages"

api-reference/beta/api/channel-list-messages.md
evanypark: Added line saying default is set to 20

api-reference/beta/api/channel-list-messages.md
angelgolfer-ms: Yes, that's right, specific to this PR and the current support for $expand, we should point out the support is only for channel messages.

api-reference/beta/api/channel-list-messages.md
angelgolfer-ms: I see. For example 3, I updated the descriptions in the "Request" and "Response" sub-sections. I also updated the request URL to include $top for the # of channel messages (of 1).
I made the similar updates in both the beta and v1 versions of the example.
Please take a look at my (upcoming) commit. If you see anything I overlooked, please let me know.

changelog/Microsoft.Teams.Core.json
angelgolfer-ms: No, the Target and ChangedApiName properties should be names of API elements defined in the Microsoft Graph API. You can see in the changelog CDK topic (https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog) for the kinds of API elements that you can assign to each of these properties.

In this case, we're adding further capability (being able to expand the properties of chatMessage resources) to chatMessage resources that are accessible through the replies navigation property of a channel message. So the ApiChange is "Relationship", the ChangedApiName is the name of that relationship "replies", and Target is the parent resource of the replies nav property "chatMessage". 

These properties in the changelist record JSON were not designed to sufficiently define every type of change that can happen to the APIs. In your case, you need to further describe the support as currently applicable to only chatMessage resources that are replies to channel messages.

angelgolfer-ms (changes requested):
Hi @evanypark , thanks for incorporating some feedback from the last round.
I responded to some of your comments and questions. I also updated the descriptions of example 3 directly in this PR. Please take a look.

api-reference/beta/api/channel-list-messages.md
```diff
@@ -62,9 +68,11 @@ Do not supply a request body for this method.
...
 
-## Example
+## Examples
 
-### Request
+### Example 1: Request without optional query parameters
+
+#### Request
 
 The following is an example of the request.
```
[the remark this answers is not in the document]
evanypark: @angelgolfer-ms, Not necessarily, if there are only 3 messages. it will return 3 messages even without specifying $top=3.
Also, the first example was for request without optional query parameters.

For example 2 and 3 it does makes sense, so I will add them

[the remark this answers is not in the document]
angelgolfer-ms: @evanypark , then what is the @odata.nextLink for on line 131 of the response for example 1? More channel messages to return, is that right?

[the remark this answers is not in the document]
evanypark: I've added $top query options to Example title.
It will match with the actual response users will be getting.

evanypark:
@angelgolfer-ms I've applied identified updates from the comments.
Let me know if you find any other changes to be need :)

angelgolfer-ms (changes requested):
@evanypark , thanks for incorporating the updates to the examples, per the last review iteration.
Please also update the changelog json, also per my comments in the last review iteration.
After that, I think we should be complete with content review.

evanypark:
> @evanypark , thanks for incorporating the updates to the examples, per the last review iteration.
> Please also update the changelog json, also per my comments in the last review iteration.
> After that, I think we should be complete with content review.

evanypark:
s

api-reference/v1.0/api/channel-list-messages.md
```diff
@@ -62,11 +68,11 @@ If successful, this method returns a `200 OK` response code and a collection of
 
 ## Examples
```
angelgolfer-ms, line 69: @evanypark , as I try to wrap up my part of the review so that you can publish, there's this build warning for line 277 of the v1 version of channel-list-messages.md, which is about the language-based snippets for example 2:
"Line 277, Column 1: [Warning: invalid-tab-group - [See documentation](https://review.docs.microsoft.com/help/contribute/validation-ref/invalid-tab-group?branch=main)] Tab group with different tab id set." I suspect this build warning has been around before this PR, but nonetheless, in general, we should minimize build warnings whenever we check in.

If you look at the linked docs for that warning, it seems the cause for the warning is, you have only 4 (HTTP, C#, JavaScript, and Java) languages specified for the code snippets, but example 1 has all 7, and all these "tab groups" should support the same number of languages and in the same order. Do you see a reason why for example 2, there would be problem generating code snippets for Objective-C, Go, and Powershell?

So since we're updating the request for example 2 by adding ?$top=3 (btw I am adding it on your behalf to the v1 version, you added it to the beta version but not the v1 version of examples 1 and 2), the code snippets should be regenerated anyway on the weekly update after you publish this PR. In that case, on your behalf, I am going to just delete the  lines for the language-based code snippets so to remove the build warning.

If on the other hand, there's a real reason that code snippets in Objective-C, Go, and Powershell would not work, then after publishing this PR and the code snippet generator tool inserts (non-working) code snippets for these languages for example 2, the next thing you should do is to remove the lines for the code snippets inserted for these languages, for example 2. By then, the tool will have generated the code snippets in the beta topic as well, so you'll have to delete the lines for example 2 in the beta topic too.

The CDK topic about code snippets is here: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211752/Generate-code-snippets

Let me know on Teams if you have further questions on this. Meanwhile, I'm going to delete these lines in the v1 topic, wrap up my review, and approve this PR.

angelgolfer-ms (approved):
@evanypark , thanks for updating the changelog JSON.
Please take a look at a comment and a commit I did today. I have approved this PR, but let me know if you have any further questions especially regarding today's comment.
Please follow step 8 in the [content workflow](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211773/Content-workflow) to get this PR published.

PR #16376, opened 2022-03-31
List of call end reasons.

navali-msft:
Added new block for sharing all the different codes/sub codes send to bot in case of call termination.

### DIFF
--- a/api-reference/v1.0/api/call-delete.md
+++ b/api-reference/v1.0/api/call-delete.md
@@ -162,14 +162,14 @@ These are some of the common error codes received via notification when call is 
 | 200  | 5000     | Removed from the conversation by another participant.                                              |
 | 200  | 5001     | The conversation has ended.                                                                        |
 | 200  | 5002     | The conversation has ended as all other participants have left the call.                           |
-| 200  | 5003     | The conversation has ended (Removing modality controller as this conversation has ended.)          |
-| 200  | 5007     | This conversation has ended as the initiator of the group call left the conversation.              |
-| 200  | 5010     | This conversation has ended as only one participant remained in the conversation.                  |
-| 200  | 5012     | This conversation has ended as there are no participants in the incoming roster.                   |
-| 200  | 5013     | This conversation has ended as no one else has joined the group call.                              |
-| 200  | 5014     | This conversation has ended since we were unable to determine a potential host for the group call. |
-| 200  | 5020     | This conversation has ended as there are no non-hidden participants in the incoming roster.        |
-| 200  | 5030     | This conversation has ended as the breakout room duration has elapsed.                             |
+| 200  | 5003     | The conversation has ended                                                                         |
+| 200  | 5007     | The conversation has ended as the initiator of the group call left the conversation.              |
+| 200  | 5010     | The conversation has ended as only one participant remained in the conversation.                  |
+| 200  | 5012     | The conversation has ended as there are no participants in the incoming roster.                   |
+| 200  | 5013     | The conversation has ended as no one else has joined the group call.                              |
+| 200  | 5014     | The conversation has ended since we were unable to determine a potential host for the group call. |
+| 200  | 5020     | The conversation has ended as there are no non-hidden participants in the incoming roster.        |
+| 200  | 5030     | The conversation has ended as the breakout room duration has elapsed.                             |
 | 200  | 5300     | Participant was removed from the conversation by another participant.                              |
 | 200  | 5855     | Participant waiting in lobby was removed from the conversation after lobby inactivity timeout.     |
 | 200  | 7000     | The conversation has ended by bot.                                                                 |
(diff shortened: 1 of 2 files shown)


api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
   ]
 }
 ```
+# Call End Reasons
```
Lauragra, line 154: We can't have more than one H1 per topic. Before I suggest the right heading level, can you please clarify, is this code returned in the Notification - terminated response? If so, then this section might be an H4 under that heading. (Incidentally, we need to adjust the heading levels in the Examples section, so Request, Response, and the notification headings are H3s).

Note that headings follow sentence style capitalization, so the heading should be "Call end reasons". Would it be clearer to use "Call end reason codes", since we're providing the list of codes? And if the code is in the termination notification, please also update the example to use one of the codes.

Lauragra, line 154: Please make the same update to the topic in the /beta folder, after incorporating any changes.

api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
...
+These are some of the common error codes received via notification when call is terminated.
+
+| Code | Sub-code | Termination reason                                                                                 |
+| :--- | :------  | :------------------------------------------------------------------------------------------------  |
+| 200  | 4097     | Call ended by the other participant in the call.                                                   |
+| 200  | 4521     | Call ended by the other participant in peer-to-peer call.                                          |
+| 200  | 5000     | Removed from the conversation by another participant.                                              |
+| 200  | 5001     | The conversation has ended.                                                                        |
+| 200  | 5002     | The conversation has ended as all other participants have left the call.                           |
+| 200  | 5003     | The conversation has ended                                                                         |
```
Lauragra, line 165: Please add a period here.

api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
...
+| 200  | 5007     | The conversation has ended as the initiator of the group call left the conversation.              |
+| 200  | 5010     | The conversation has ended as only one participant remained in the conversation.                  |
+| 200  | 5012     | The conversation has ended as there are no participants in the incoming roster.                   |
+| 200  | 5013     | The conversation has ended as no one else has joined the group call.                              |
+| 200  | 5014     | The conversation has ended since we were unable to determine a potential host for the group call. |
```
Lauragra, line 170: Please change "since" to "because" for clarity (since connotes time).

api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
...
+| 200  | 5020     | The conversation has ended as there are no non-hidden participants in the incoming roster.        |
+| 200  | 5030     | The conversation has ended as the breakout room duration has elapsed.                             |
+| 200  | 5300     | Participant was removed from the conversation by another participant.                              |
+| 200  | 5855     | Participant waiting in lobby was removed from the conversation after lobby inactivity timeout.     |
+| 200  | 7000     | The conversation has ended by bot.                                                                 |
+| 200  | 7015     | Call ended as the transfer completed successfully.                                                 |
+| 200  | 10550    | The conversation has ended by bot.                                                                 |
+| 200  | 18503    | The other participant in peer-to-peer call, left the conversation.                                 |
```
Lauragra, line 178: Add "the" before peer-to-peer and remove the comma.

api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
...
+| 200  | 540000/560000   | Call ended by the Pstn User.                                                                |
```
Lauragra, line 179: Change Ptsn User to PTSN user.

api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
...
+| 408  | 8537     | Keep Alive timeout, cleaned up inactive call.                                                      |
+| 408  | 1106     | An acknowledgement was not received for the call acceptance in the allotted time.                  |
+| 408  | 10057    | Call timed out due to no signaling from callee endpoints.                                          |
+| 410  | 301005   | Media connectivity failure.                                                                        |
+| 480  | 10037    | No callee endpoints were found.                                                                    |
+| 480  | 10076    | Callee couldnot be reached.                                                                        |
```
Lauragra, line 185: Add a space between could and not.

api-reference/v1.0/api/call-delete.md
```diff
@@ -151,6 +151,42 @@ Content-Type: application/json
...
+| 480  | 10134    | Call rejected due to inability to route the call.                                                  |
+| 480  | 10199    | Call rejected as private calling is disabled for the user.                                         |
+| 500  | 1005     | Media disconnected in the bot side 	Check media connectivity between bot and Microsoft.            |
```
zhengni-msft, line 188: based on the text it seems a client error while the code is server error. Can we confirm the behavior and either update code or update text to make them consistent?

zhengni-msft (commented):
Overall, can we review the subcode list with CC/CS/Media owners to ensure they are aware of that such code will be documented publicly and the definition should not be changed later?

api-reference/v1.0/api/call-delete.md
navali-msft: Yes, these are some of the example error code/subcode, we send to client with call termination notification.

api-reference/v1.0/api/call-delete.md
navali-msft: @zhengni-msft  I have updated the comments.

api-reference/v1.0/api/call-delete.md
navali-msft: Added in the beta folder as well.

navali-msft: I have fixed the heading levels and updated title as per your suggestion.

Lauragra (approved):

PR #16388, opened 2022-04-01
entitlement management: add example of creating a request with requestType userAdd

markwahl-msft:
In entitlement management v1.0, the article for creating a new access package assignment request has 2 examples, for administrator direct assignment, and administrator removing an assignment.   The beta article by contrast has a few more examples. This PR adds a further example to the v1.0 article that we previously had in beta, of a user already in tenant requesting for themselves.   Subsequent PRs will add examples to v1.0 of supplying answers/attributes, once that goes GA and becomes part of v1.0, and an administrator requesting for a user not yet in tenant.  There is no new API surfaces in this PR.


FaithOmbongi (approved):
Thank you @markwahl-msft for following this through to close this gap.

PR #16394, opened 2022-04-02
Updated online meeting doc and added a warning

frankpeng7:
Updated online meeting doc and added a warning


api-reference/beta/resources/meetingattendancereport.md
```diff
@@ -24,6 +24,9 @@ Meeting attendance reports are online meeting artifacts. For details, see [Onlin
 |[List meetingAttendanceReports](../api/meetingattendancereport-list.md)|[meetingAttendanceReport](../resources/meetingattendancereport.md) collection|Get a list of  [meetingAttendanceReport](../resources/meetingattendancereport.md) objects and their properties.|
 |[Get meetingAttendanceReport](../api/meetingattendancereport-get.md)|[meetingAttendanceReport](../resources/meetingattendancereport.md)|Read the properties and relationships of a [meetingAttendanceReport](../resources/meetingattendancereport.md) object.|
 
+> [!WARNING]
+> All Get/List methods of attendance report do not support channel meetings.
```
Lauragra, line 28: I think it would be more discoverable to users if we add a note to the Get and List operation topics rather than here. I suggest the following text for the note:

> [!NOTE]
> This operation does not return attendance reports for channel meetings.

Does this fall into the category of a known issue, or is it by design?

frankpeng7: It's by design.
I can move them to Get and List method topics

PR #16419, opened 2022-04-05
Update GDAP ms.prod to customer-relationship-management

FaithOmbongi:


adtangir (approved):

adtangir:
Looks good to me. So the tenant-relationship resource would still belong to the ms.prod: "microsoft-365-lighthouse" ?

FaithOmbongi:
> Looks good to me. So the tenant-relationship resource would still belong to the ms.prod: "microsoft-365-lighthouse" ?

For now, yes. Let me bring this for discussion in the docs v-team as we also discuss making "Customer relationship management" APIs a higher-level node.

PR #16493, opened 2022-04-11
accessReviewHistoryDefinition removing reference to unreleased scheduleSettings feature

leherpel:
Removing references to the scheduleSettings object as the feature has not yet been released.


FaithOmbongi (approved):
Thank you @leherpel for updating the docs so customers don't wonder when they can't use this property. Let's remember to remove the note when the feature is finally implemented.

PR #16510, opened 2022-04-12
Doc updates related to enabling resource data in user-context subscriptions for Microsoft Teams

eddie-lee-msft:
Doc updates for allow resource data to be included in user context subscriptions for Microsoft Teams


api-reference/includes/teams-subscription-notes.md
```diff
@@ -9,9 +9,7 @@ ms.localizationpriority: medium
...
-
-**chatMessage** subscriptions with application permissions include resource data, and require [encryption](/graph/webhooks-with-resource-data). The subscription creation fails if an [encryptionCertificate](/graph/api/resources/subscription) isn't specified. Before you can create a **chatMessage** subscription, you must request access. For details, see [Protected APIs in Microsoft Teams](/graph/teams-protected-apis).
+**chatMessage** subscriptions can be specified to include resource data. If specified to include resource data, [encryption](/graph/webhooks-with-resource-data) is required. The subscription creation fails if an [encryptionCertificate](/graph/api/resources/subscription) isn't specified for such subscriptions. Before you can create a **chatMessage** subscription, you may need to request access. For details, see [Protected APIs in Microsoft Teams](/graph/teams-protected-apis).
```
AkJo, line 12: You should perhaps add the following here to make it more obvious what a person must do to include resource data.

(**includeResourceData** must be `true`)

AkJo: If specified to include resource data (**includeResourceData** set to true), then [encryption](/graph/webhooks-with-resource-data) is required. T

eddie-lee-msft: Done

api-reference/includes/teams-subscription-notes.md
```diff
@@ -9,9 +9,7 @@ ms.localizationpriority: medium
...
-
-**chatMessage** subscriptions with application permissions include resource data, and require [encryption](/graph/webhooks-with-resource-data). The subscription creation fails if an [encryptionCertificate](/graph/api/resources/subscription) isn't specified. Before you can create a **chatMessage** subscription, you must request access. For details, see [Protected APIs in Microsoft Teams](/graph/teams-protected-apis).
+**chatMessage** subscriptions can be specified to include resource data. If specified to include resource data (**includeResourceData** set to true), [encryption](/graph/webhooks-with-resource-data) is required. The subscription creation fails if an [encryptionCertificate](/graph/api/resources/subscription) isn't specified for such subscriptions. Before you can create a **chatMessage** subscription, you may need to request access. For details, see [Protected APIs in Microsoft Teams](/graph/teams-protected-apis).
```
angelgolfer-ms, line 13: @eddie-lee-msft , related to the changes in line 12:
Am I correct in deducing that for chatMessage, you want to set straight both delegated and application permissions support notifications with resource data now?
If that is the supported behavior, then can you confirm ONLY creating subscription (with resource data) to chatMessage resources **using application permission** falls into "protected API" category (i.e., need to request prior permission), but using delegated permission does not? 

(I see your sentence is "... you MAY need to request access".)

If delegated permissions do NOT require requesting prior access, then I strongly suggest you clarify this in line 12. Something like:
"Before creating a chatMessage subscription with application permissions, make sure to request access to the chatMessage data. For details, see Protected APIs in..."
The reason is these conditions that are up to the customer to decide for their scenarios - such as do they want resource data, can they just use delegated permission, or do they need to use application permission - they further add burden to the customer when creating a subscription. So it would be helpful for the docs to be more prescriptive / clear, than using "may..."

On the other hand, if it turns out delegated permissions also require requesting prior access, then please remove the "may", and update the "Protected APIs in Microsoft Teams" topic, which currently lists scenarios using only application permissions. 

Or, is there a more complicated decision tree involved here other than delegated vs application permissions? If so, please elaborate here and we can decide whether to bring it up here or in the "Protected API in Microsoft Teams" topic.



to create subscription to chatMessage "Protected APIs in Microsoft Teams"?

changelog/Microsoft.Teams.Core.json
```diff
@@ -4941,6 +4941,42 @@
       "CreatedDateTime": "2021-04-25T10:11:41.458Z",
       "WorkloadArea": "Teamwork",
       "SubArea": ""
+    },
+    {
+      "ChangeList": [
+        {
+          "Id": "e473f528-d700-4803-9985-6db1bde5c1aa",
+          "ApiChange": "Resource",
+          "ChangedApiName": "notifications",
```
angelgolfer-ms, line 4950: Lines 4950 and 4968:
FYI when we specify the "ApiChange" as "Resource", we are saying that the change is applied to a resource that is / should be named in the "ChangedApiName" property. In this changelog update, that's not appropriate because "notifications" isn't a resource name for our subject in question (notifications for resource changes).

FYI the appropriate way to update the changelog is to note the behavioral change to the POST operation of the subscription entity. Something like this, for each of beta and v1.0 version.

          "Id": "e473f528-d700-4803-9985-6db1bde5c1ab",
          "ApiChange": "Method",
          "ChangedApiName": "POST",
          "ChangeType": "Addition",
          "Description": "Added support for [subscribing to notifications](https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-beta) with resource data in delegated context for [chatMessage](https://docs.microsoft.com/en-us/graph/api/resources/chatmessage?view=graph-rest-beta) resources in a [chat](https://docs.microsoft.com/en-us/graph/api/resources/chat?view=graph-rest-beta) or [channel](https://docs.microsoft.com/en-us/graph/api/resources/channel?view=graph-rest-beta).",
          "Target": "subscription"

angelgolfer-ms: BTW, I noticed your changelog description (lines 4952 and 4970) say "teams, channels, and chats".
I thought the change is in subscribing to change notifications for chatMessage, just from looking at the other updates currently in this PR.
Do you think the docs are clear for what you listed ("teams, channels, and chats")?
I just wanted to have you confirm first before we trace through the docs to see where clarification or update is necessary.

angelgolfer-ms (changes requested):
Hi @eddie-lee-msft , @Danielabom who's your content reviewer asked me a question on the changelog for this PR, hence I entered a few comments, some outside of the context of the changelog. Please take a look. If you have questions, please ping me on Teams.

AkJo (approved):

api-reference/includes/teams-subscription-notes.md
eddie-lee-msft: You are right. Requesting Protected API access is only relevant in application context. I will make a change with your suggestion.

changelog/Microsoft.Teams.Core.json
eddie-lee-msft: @angelgolfer-ms, I probably should have given more context in the description of this PR.

We are enabling change notifications with resource data in delegated context for every Microsoft Teams resource type that is supported in user context today. You can see those here: https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-beta&tabs=http. Any resource that has "chats", "teams", or "channels" in it that has support in delegated context is affected by this enablement. One example is "/teams/{id}/channels/{id}/messages". One non-example is "/chats/getAllMessages" because it is not supported in delegated context.

Now, you might be curious why this PR has such few changes. I looked through all the docs in the current state, and the large majority of them **already mention** that resource data is supported for those resources, without **any** mention of a specific context. Rather it is mentioned in a generic manner. One such example is here: https://docs.microsoft.com/en-us/graph/teams-changenotifications-teammembership. You will see that there is mention that resource data is allowed, but nothing about specific contexts.

If there is any place I have missed, please do let me know.

eddie-lee-msft: Also given above context, please let me know what the appropriate changelog addition should be. I wasn't sure myself.

api-reference/includes/teams-subscription-notes.md
eddie-lee-msft: Added the clarification

changelog/Microsoft.Teams.Core.json
angelgolfer-ms: Thanks, @eddie-lee-msft , for the clarification.
I traced through the current 5 Teams concept topics for change notifications and agree for the purpose of this PR, we've documented all the use cases (resources) which allow subscribing to rich notifications using delegated permissions (and application permissions).

Please update your changelist records for this PR as follows, for each of beta and v1:

```
     "Id": "e473f528-d700-4803-9985-6db1bde5c1ab",
      "ApiChange": "Method",
      "ChangedApiName": "POST",
      "ChangeType": "Addition",
      "Description": "Added support for [subscribing to notifications](https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-beta) with resource data in delegated context for [chatMessage](https://docs.microsoft.com/en-us/graph/api/resources/chatmessage?view=graph-rest-beta) resources in a specified [channel](https://docs.microsoft.com/en-us/graph/api/resources/channel?view=graph-rest-beta) or [chat](https://docs.microsoft.com/en-us/graph/api/resources/chat?view=graph-rest-beta).",
      "Target": "subscription"
```

As described in my earlier comment in this thread, the appropriate angle is to focus on the additive (we're supporting delegated perms in addition to the previous support of application perms) change to the behavior of the POST operation of the subscription entity.

angelgolfer-ms (changes requested):
Hi @eddie-lee-msft , I verified the explanation in your reply, thanks. I also adjusted my comment about the changelist records, please take a look.
As a side effect of going through the related topics, I noticed there are a few things that over time, authors have been missing some facts about Teams notifications. I included a list in an email. It would be good if you or your team can align the various topics for Teams.

changelog/Microsoft.Teams.Core.json
eddie-lee-msft: Changelog has been updated with your suggestion.

changelog/Microsoft.Teams.Core.json
```diff
@@ -4964,11 +4964,11 @@
...
-          "ApiChange": "Resource",
-          "ChangedApiName": "notifications",
-          "ChangeType": "Update",
-          "Description": "Added rich notification support (includeResourceData set to true) in delegated context for [notifications](https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-beta) for Teams teams, channels, and chats.",
-          "Target": "notifications"
+          "ApiChange": "Method",
+          "ChangedApiName": "POST",
+          "ChangeType": "Addition",
+          "Description": "Added support for [subscribing to notifications](https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-beta) with resource data in delegated context for [chatMessage](https://docs.microsoft.com/en-us/graph/api/resources/chatmessage?view=graph-rest-beta) resources in a specified [channel](https://docs.microsoft.com/en-us/graph/api/resources/channel?view=graph-rest-beta) or [chat](https://docs.microsoft.com/en-us/graph/api/resources/chat?view=graph-rest-beta).",
+          "Target": "subscription"
```
angelgolfer-ms, line 4971: Oh, @eddie-lee-msft , please make the links in line 4970 to point to v1.0, not beta.

eddie-lee-msft: Whoops, fixed.

angelgolfer-ms (approved):
PR looks good, thanks!

PR #16543, opened 2022-04-15
Update conditionalaccessroot-post-policies.md

davidspooner:
Changed permissions statement as it was confusing and misleading.

### DIFF
--- a/api-reference/v1.0/api/conditionalaccessroot-post-policies.md
+++ b/api-reference/v1.0/api/conditionalaccessroot-post-policies.md
@@ -15,13 +15,16 @@ Create a new [conditionalAccessPolicy](../resources/conditionalaccesspolicy.md).
 
 ## Permissions
 
-The following permissions are required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+
+>[!NOTE]
+>This API requires all the permissions listed in the table.
 
 | Permission type                        | Permissions (from least to most privileged) |
 |:---------------------------------------|:--------------------------------------------|
-| Delegated (work or school account)     | Policy.Read.All, Policy.ReadWrite.ConditionalAccess and Application.Read.All |
+| Delegated (work or school account)     | Policy.Read.All, Policy.ReadWrite.ConditionalAccess, and Application.Read.All |
 | Delegated (personal Microsoft account) | Not supported. |
-| Application                            | Policy.Read.All, Policy.ReadWrite.ConditionalAccess and Application.Read.All |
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/api/conditionalaccessroot-post-policies.md
```diff
@@ -15,7 +15,7 @@ Create a new [conditionalAccessPolicy](../resources/conditionalaccesspolicy.md).
 
 ## Permissions
 
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+The following permissions are required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
```
Lauragra, line 18: This is standard boilerplate text that appears in all API topics, so we shouldn't change it. Are you trying to convey that not one but all of these permissions are required?

davidspooner: Laura, that is correct. the user needs ALL the permissions to be able to call the API. Having only some of the permissions is not enough.

FaithOmbongi: Hi @davidspooner - As Laura's mentioned, this is a standard boilerplate text to help customers apply least privilege while selecting permissions. 

**Suggestion:** How about leaving the boilerplate text as-is but updating the list to `Policy.Read.All and Policy.ReadWrite.ConditionalAccess and Application.Read.All`

The `and` will also help the tooling translate that the permission requirement is a combination of 3 permissions.

Lauragra:
@Jackson-Woods , can you please take a look at the proposed update to the boilerplate text for permissions? Is this a standard scenario where multiple and not the least privileged permission is required? Thanks.

api-reference/v1.0/api/conditionalaccessroot-post-policies.md
```diff
@@ -17,11 +17,14 @@ Create a new [conditionalAccessPolicy](../resources/conditionalaccesspolicy.md).
 
 One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
 
+>[!NOTE]
+>This API requires all the permissions listed in the table.
```
Lauragra, line 21: @davidspooner , what if we do this - leave the standard boilerplate text, because it's possible that users won't notice the slight difference between the proposed and the standard boilerplate - and have a note that prominently calls out that all permissions are required?

Lauragra (approved):

Lauragra:
@FaithOmbongi , @Jackson-Woods , are you okay with publishing this PR? Thanks.

api-reference/v1.0/api/conditionalaccessroot-post-policies.md
Jackson-Woods: I would call this out more clearly. Something like a vertical list, or even using *and* or AND between each one.

PR #16567, opened 2022-04-19
[Azure AD roles] principalOrganizationId property

rolyon:


FaithOmbongi:
Hi @rolyon - I see this property in both the public schema and API payload so reopening this PR. Kindly action so we can get this merged as soon as possible - unless there's a reason we may not want this property documented for now?

PR #16579, opened 2022-04-20
Corrected the PowerShell redirects to make the source path an in-repo…

msewaweru:
… link


msewaweru:
@CelesteDG, I realized that the PowerShell redirects were not working after merging that PR yesterday. I have corrected the source path to make it an in-repo link. Please review.

PR #16634, opened 2022-04-24
Update toc.yml

jiasli:
These docs are for deleted **items**, not **users**.

image


Lauragra:
@jiasli , can you please update the toc.yml file in the /beta folder with the same changes? Thanks.

PR #16621, opened 2022-04-22
Edited custom connector tutorial

Linda-Editor:
High-level copyedit to fix formatting and punctuation.

PR #16647, opened 2022-04-25
validation: multiple-fixes-3

MelanieHom:


jasonjoh:
This is not needed. https://github.com/microsoftgraph/microsoft-graph-docs/pull/16621 moved these files into an "includes" subfolder, which signals to OPS they are include files and metadata is not required. This resolved the validation issues.

PR #16678, opened 2022-04-26
Add user securityIdentifier property

ekirapaMsfte:


franqq (approved):

ekirapaMsfte:
> 

@FaithOmbongi I mentioned in the property description that it is returned by default. Not sure if is necessary to show how it is generated in the background.

FaithOmbongi (approved):

PR #16711, opened 2022-04-28
Update teams-list-all-teams.md

EMarcais:
Modified line 58, Note to reflect that  resourceProvisioningOptions filter is now available in both version of the API


concepts/teams-list-all-teams.md
```diff
@@ -55,7 +55,7 @@ Using the beta APIs, you can use $filter to return only the groups that have tea
 GET /groups?$filter=resourceProvisioningOptions/Any(x:x eq 'Team')
 ```
 
-> **Note**: Filtering groups by resourceProvisioningOptions is only available through the beta endpoint. resourceProvisioningOptions is available in v1.0 and beta.
+> **Note**: Filtering groups by resourceProvisioningOptions is now available in v1.0 and beta.
```
Lauragra, line 58: Would it make more sense to remove the note entirely, since the note was basically to point out that the functionality wasn't available in v1.0 at the time? I assume that the promotion to v1.0 was covered in the changelog, and we might not want this note to continue to persist.

PR #16751, opened 2022-05-03
Team Template Graph API

shthakkar:


Lauragra:
@shthakkar , please contact me internally to get set up to be a contributor to the docs. You'll want to use generated topics. It looks like this PR might just be a placeholder?

PR #16752, opened 2022-05-03
Update migrate-azure-ad-graph-configure-permissions.md

scottwtang:
Issues with the code sample under **Option 4: Use the Microsoft Graph PowerShell SDK** -> **Step 2: Add Azure AD Graph permissions to your app** -> **Request**

The sample code only runs when the app registration has no existing permissions from the resource app (Azure AD Graph). There's no else statement for an example where the app registration already has permissions from the resource app.

Although the example uses the soon-to-be deprecated Azure AD Graph as the resource app, the example code is using the MS Graph PowerShell module, and there doesn't seem to be any other documentation providing an example for this scenario (using MS Graph to add additional permissions when permissions already exist)


FaithOmbongi (approved):
Thank you @scottwtang for this helpful contribution. Approved and should be live at the end of the day.

PR #16760, opened 2022-05-03
AAD best practices info for CRI reduction

besiler:
Getting lots of CRIs from customers who don't know how to avoid throttles in Azure AD reporting - hoping to help with some suggestions while we investigate other ways to improve performance

### DIFF
--- a/concepts/throttling.md
+++ b/concepts/throttling.md
@@ -304,9 +304,9 @@ The preceding limits apply to the following resources:
 [!INCLUDE [Azure AD identity and access reports throttling documentation](../includes/throttling-aad-reports.md)]
 
 #### Identity and access reports best practices
-Azure AD reporting APIs are throttled when Azure AD receives too many calls during a given timeframe from a tenant or app. Calls may also be throttled if the service takes too long to respond. If you have issues with 429 responses despite applying the best practices above, try reducing the number of events returned as a result of your query. Try these approaches first:
-- Use filters to target your query to just the data you need. If you only need a certain type of event or a subset of users, for example, filter out other events to reduce the size of your response and the risk of throttling.
-- If you need a broad set of Azure AD reporting data, use filters on createdDateTime to limit the amount of sign-in events you query in a single call. Then, iterate through the next timespan until you have all the records you need. For example, if you are being throttled, you can begin with a call that requests 3 days of data and iterate with shorter timespans until your requests are no longer throttled.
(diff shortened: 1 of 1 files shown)


concepts/throttling.md
```diff
@@ -303,6 +303,11 @@ The preceding limits apply to the following resources:
 
 [!INCLUDE [Azure AD identity and access reports throttling documentation](../includes/throttling-aad-reports.md)]
 
+#### Identity and access reports best practices
+Azure AD reporting APIs are throttled when Azure AD receives too many calls during a given timeframe from a tenant or app. Calls may also be throttled if the service takes too long to respond. If you have issues with 429 responses despite applying the best practices above, try reducing the number of events returned as a result of your query. Try these approaches first:
+- Use filters to target your query to just the data you need. If you only need a certain type of event or a subset of users, for example, filter out other events to reduce the size of your response and the risk of throttling.
+- If you need a broad set of Azure AD reporting data, use filters on createdDateTime to limit the amount of sign-in events you query in a single call. Then, iterate through the next timespan until you have all the records you need. For example, if you are being throttled, you can begin with a call that requests 3 days of data and iterate with shorter timespans until your requests are no longer throttled.
```
FaithOmbongi, line 309: Minor suggestions mostly formatting & referencing the OData query parameters so the customer knows what we mean why "use filters".
```suggestion
Azure AD reporting APIs are throttled when Azure AD receives too many calls during a given timeframe from a tenant or app. Calls may also be throttled if the service takes too long to respond. If your requests still fail with a `429 Too Many Requests` error code despite applying the [best practices above](#best-practices-to-handle-throttling), try reducing the amount of data returned. Try these approaches first:
- Use filters to target your query to just the data you need. If you only need a certain type of event or a subset of users, for example, filter out other events using the `$filter` and `$select` query parameters to reduce the size of your response object and the risk of throttling.
- If you need a broad set of Azure AD reporting data, use `$filter` on the **createdDateTime** to limit the amount of sign-in events you query in a single call. Then, iterate through the next timespan until you have all the records you need. For example, if you are being throttled, you can begin with a call that requests 3 days of data and iterate with shorter timespans until your requests are no longer throttled.
```

FaithOmbongi (approved):
Thank you @besiler for raising this PR. Approved with suggestions.

besiler:
Thanks @FaithOmbongi  for your help here! Marking ready to merge after committing your suggestions

PR #16763, opened 2022-05-03
[Azure AD roles] allowedToReadBitlockerKeysForOwnedDevice property

rolyon:


loradule (commented):
Could we please change the value of the new property to be true instead of Boolean, to match the others?

loradule (approved):

api-reference/beta/toc.yml
```diff
@@ -12697,6 +12697,10 @@ items:
               href: api/identityapiconnector-uploadclientcertificate.md
             - name: Delete
               href: api/identityapiconnector-delete.md
+        - name: User
+          items:
+          - name: Default user role permissions
```
Lauragra, line 12702: FYI, we don't expose complex types in the TOC; I recommend removing this entry.

FaithOmbongi: +1

rolyon: Sure, I removed TOC entry

PR #16802, opened 2022-05-06
Removed redundancy

Linda-Editor:
To make the page more consistent, I removed the labels in the limit column because they are defined in the description.

### DIFF
--- a/concepts/connecting-external-content-api-limits.md
+++ b/concepts/connecting-external-content-api-limits.md
@@ -15,22 +15,22 @@ This article describes implementation and operational limits for Microsoft Graph
 
 | **Limit** | **Description** |
 | --------- | --------------- |
-| **10 connections** | The maximum number of [connection](/graph/api/resources/externalconnectors-externalconnection?view=graph-rest-1.0&preserve-view=true) resources per Microsoft 365 tenant. |
-| **700,000 items** | The maximum number of [items](/graph/api/resources/externalconnectors-externalitem?view=graph-rest-1.0&preserve-view=true) per connection. |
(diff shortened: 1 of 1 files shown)


concepts/connecting-external-content-api-limits.md
```diff
@@ -15,22 +15,22 @@ This article describes implementation and operational limits for Microsoft Graph
 
 | **Limit** | **Description** |
 | --------- | --------------- |
-| **10 connections** | The maximum number of [connection](/graph/api/resources/externalconnectors-externalconnection?view=graph-rest-1.0&preserve-view=true) resources per Microsoft 365 tenant. |
-| **700,000 items** | The maximum number of [items](/graph/api/resources/externalconnectors-externalitem?view=graph-rest-1.0&preserve-view=true) per connection. |
+| **10** | The maximum number of [connection](/graph/api/resources/externalconnectors-externalconnection?view=graph-rest-1.0&preserve-view=true) resources per Microsoft 365 tenant. |
```
Lauragra, line 18: @Linda-Editor , I think we can look for a different way to make these tables more clear. I don't think removing the labels makes these tables scannable/easy to read. What if we do something like this:

|Limit type | Maximum number |
|-----------|----------------------|
|Connection resources per Microsoft 365 tenant | 10 |
| Items per connection | 700,000 |
| Connection byte size | 70 GB |

Linda-Editor: You're right. That's much better. I'll make these changes.

Linda-Editor:
FYI: I changed "Maximum number" to "Maximum" because sometimes it was a number and sometimes it was a size, and then I changed it to "Limit" (which was the original heading) because "Maximum" didn't really work by itself. Let me know if you have any questions.

PR #16805, opened 2022-05-06
Update user-exportpersonaldata.md

riro1010:
In a recent Teams channel post a caller to /user/{id}/exportPersonalData pointed out an inconsistency in the documentation. Essentially, the text states the User.Read.All is required when it is not.  Only User.Export.All is required. My team owns the Export workflow via Privacy Experience service and the text change has been validated with our Identity team partner.

I have removed the incorrect text 'User.Read.All' from the User Personal data Export. Kindly update the changes at your earliest

### DIFF
--- a/api-reference/beta/api/user-exportpersonaldata.md
+++ b/api-reference/beta/api/user-exportpersonaldata.md
@@ -18,9 +18,9 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type      | Permissions (from least to most privileged)              |
 |:--------------------|:---------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


alvarorahul:
We should also make the same fix in beta version of the API

mmcadamsms:
+1 to Alvaro's comment above ... need to also do the beta one

mmcadamsms (approved):
Looks like the change I was wanting to see.

riro1010:
Thanks, Laura, for approving the request on a short notice

alvarorahul (approved):

PR #16808, opened 2022-05-06
Updating roles/permissions for CA data

besiler:
Due to an MSRC security incident, we had to update the RBAC required to see CA data in sign ins. This change will roll out next week (week of May 9)

### DIFF
--- a/api-reference/beta/api/signin-get.md
+++ b/api-reference/beta/api/signin-get.md
@@ -38,22 +38,7 @@ In addition to the delegated permissions, the signed-in user needs to belong to 
 + Security Operator
 + Security Reader
 
-### Viewing applied conditional access (CA) policies in sign-ins
-The applied CA policies listed in **appliedConditionalAccessPolicies** property are only available to users and apps with roles that allow them to read [conditional access data](../resources/appliedconditionalaccesspolicy.md). If a user or app has permissions to read sign-in logs but not permission to read conditional access data, the **appliedConditionalAccessPolicies** property in the response will be omitted. The following Azure AD roles grant users permissions to view conditional access data:
-
-+ Global Administrator
-+ Global Reader
-+ Security Administrator
-+ Security Reader
-+ Conditional Access Administrator
-
-Applications must have at least one of the following permissions to see [appliedConditionalAccessPolicy](../resources/appliedconditionalaccesspolicy.md) objects in the sign-in logs. 
-
-+ Policy.Read.All
-+ Policy.ReadWrite.ConditionalAccess
-+ Policy.Read.ConditionalAccess
-
-**Note:** Azure AD users with any permissions can read sign-in logs in which their user is the actor signing in. This feature helps users spot unexpected activity in their accounts. Users cannot read CA data from their own logs unless they have one of the CA permissions identified above. 
+[!INCLUDE [signins-roles-for-ca-data](../../includes/signins-roles-for-ca-data.md)]
 
 ## HTTP request
 
--- a/api-reference/beta/api/signin-get.md
+++ b/api-reference/beta/api/signin-get.md
@@ -17,7 +17,7 @@ Get a [signIn](../resources/signin.md) object that contains a specific user sign
 
 ## Permissions
 
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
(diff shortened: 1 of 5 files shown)


api-reference/beta/api/signin-get.md
```diff
@@ -17,7 +17,7 @@ Get a [signIn](../resources/signin.md) object that contains a specific user sign
 
 ## Permissions
 
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
+Any Azure AD user can read their own sign-ins. In order to read sign-ins beyond those where the signed in user is the actor, one of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
```
FaithOmbongi, line 20: Same suggestions/Qs apply to the other 3 files
- Suggestion: Let's keep the intro sentence as-is and instead add a note after the table to cover this first sentence, "Any Azure AD user can read their own sign-ins."
- Q: By saying "the permissions are only required to read sign-ins beyond those where the signed in user is the actor, ", do we mean a user with just _User.Read_ should be able to call this API?

```suggestion
One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
```

api-reference/beta/api/signin-get.md
```diff
@@ -38,6 +38,21 @@ In addition to the delegated permissions, the signed-in user needs to belong to
...
++ Security Administrator
++ Security Reader
++ Conditional Access Administrator
+
+Applications must have at least one of the following permissions to see [appliedConditionalAccessPolicy](../resources/appliedconditionalaccesspolicy.md) in the logs. 
+
++ Policy.Read.All
++ Policy.ReadWrite.ConditionalAccess
++ Policy.Read.ConditionalAccess
+
```
FaithOmbongi, line 55: - Given the signIns reference the conditionalAccessPolicy resource only through the **appliedConditionalAccessPolicies** property, I've suggested some changes in where we reference the property vs where we reference the resource.
- Same suggestion applies in the other 3 files.
- I recommend you use an include file to make maintenance easier.

```suggestion
### Viewing applied conditional access (CA) policies in sign-ins
The applied CA policies listed in **appliedConditionalAccessPolicies** property are only available to users and apps with roles that allow them to read [conditional access data](../resources/appliedconditionalaccesspolicy.md). If a user or app has permissions to read sign-in logs but not permission to read conditional access data, the **appliedConditionalAccessPolicies** property in the response will be omitted. The following Azure AD roles grant users permissions to view conditional access data:

+ Global Reader
+ Company Administrator
+ Security Administrator
+ Security Reader
+ Conditional Access Administrator

Applications must have at least one of the following permissions to see [appliedConditionalAccessPolicy](../resources/appliedconditionalaccesspolicy.md) objects in the sign-in logs. 

+ Policy.Read.All
+ Policy.ReadWrite.ConditionalAccess
+ Policy.Read.ConditionalAccess

```

FaithOmbongi (commented):
Hi @besiler - I've shared some suggestion and posted a Q about the permission requirements for a signed-in user to call the signIns APIs.

besiler:
@FaithOmbongi added notes to cover the users being able to see their own sign in logs as you suggested. Please let me know if you have any other questions!

Also, for posterity, here's a description of the CA permissions change:



Before:

All signIn admins, users viewing their own signIns (guest and member) and those with signIns RBAC permission could view appliedConditionalAccessPolicies and all apps with auditLog.Read.All



Now:

Only a subset of those admins (so Global admin, global reader, security admin, security reader, and any other signIn admin role with conditional access admin) can view the condtional access data (but can still view the rest of the signins) Normal users cna still view their own signIns, just not the appliedCA data unless they also have a conditional access role.



AND

apps also have this same restriction

- if the app only had auditlog.read.all and not ca permission, then they only will get back data without appliedCApolicies for app only auth.

- (ADIbiza has both so this doen'st matter for the UX)

besiler:
@FaithOmbongi I sent some team chats on the latest here - I think this is a Graph Explorer based bug because calling my own sign-ins is working in Postman and the Azure AD UX. So we'll figure out the bug, but we want to move forward with this PR because customers are starting to notice the change in Graph CA permissions and support needs a public doc to point them to. 

Customer reporting incident here: https://portal.microsofticm.com/imp/v3/incidents/details/311853265/home

besiler:
I'm just going to mark this review as complete because the bug with calling one's own sign ins with Graph Explorer was the only thing holding it up. @FaithOmbongi and docs team please let me know if you disagree

PR #16846, opened 2022-05-11
Add certificateUserIds attribute for users

FaithOmbongi:
cc @vimrang


FaithOmbongi:
@vimrang - Bumpijg this in your inbox.

vimrang:
@FaithOmbongi what is the ask on me? Is there something I need to fix?

PR #16851, opened 2022-05-11
Edit of Toolkit overview and get started pages

Linda-Editor:


Linda-Editor:
@Lauragra Quick question: for the overview topic, do you think we should change it to just "Microsoft Graph Toolkit overview" instead of having components and providers in the title? I shortened it a bit and made it sentence-cased, but it's still kind of clunky. I vote we change it to just overview. Let me know what you think.

Lauragra (approved):

PR #16854, opened 2022-05-11
removes unecessary odata context annotations for clarity

baywet:
At the API council level today we discussed the fact these annotations are unnecesarry and should probably not be here at the first place. This PR removes them to discourage their usage until further action is taken on the service side.


mikepizzo (approved):

XVincentX (approved):
Goodbye, bad properties.

PR #16891, opened 2022-05-13
Edited two Postman docs

Linda-Editor:


concepts/use-postman.md
```diff
@@ -7,129 +7,131 @@ ms.localizationpriority: high
...
-1. In the left menu, select **Overview**. From here, you can get the **Application (client) ID** and **Directory (tenant) ID**. You will need these in step 4.
-1. In the left menu, select **Certificates & secrets**.
-1. Select **New client secret**, enter a description, and select **Add**. Hover over the new client secret **Value** and copy it. You will need this in step 4.
+1. On the horizontal menu, select **Grant admin consent for**, and then select **Yes**.
+1. On the left menu, select **Overview**. From here, you can get the **application (client) ID** and **directory (tenant) ID**. You'll need these in step 4.
+1. On the left menu, select **Certificates and secrets**.
+1. Select **New client secret**, enter a description, and then select **Add**. Hover over the new client secret **Value** and copy it; you'll need this in step 4.
 
-The Azure AD application now has permissions to make requests on behalf of a user to call Mail.Read and as an application for User.Read.All.
+The Azure AD application now has permissions to make requests on behalf of a user to call `Mail.Read` and as an application for `User.Read.All`.
```
Linda-Editor, line 61: @jasonjoh This sentence doesn't read well; is there a missing or incorrect word in the second half of this sentence? For example, should "and as an application for..." be "and HAS an application for..."? It doesn't seem to make sense, so would like your input. Thanks. There's a similar sentence in the other doc, too.

concepts/use-postman.md
```diff
@@ -7,129 +7,131 @@ ms.localizationpriority: high
...
+1. From the dropdown menu, select **Web**.
 1. Set the **Redirect URI** to `https://oauth.pstmn.io/v1/browser-callback`.
 1. Select **Register**.
 1. On the left menu, select **API Permissions**.
-1. In the horizontal menu, select **Add a permission**, select **Microsoft Graph**, and then select **Delegated Permissions**.
-1. Type `Mail.`, expand the **Mail** options, and check **Mail.Read**.
-1. Select **Application permissions** and type `User.`, and check **Application Permissions**.
-1. Expand the **User** options and check **User.Read.All**.
+1. On the horizontal menu, select **Add a permission**, select **Microsoft Graph**, and then select **Delegated Permissions**.
+1. Type `Mail.`, expand the **Mail** options, and then select **Mail.Read**.
```
Linda-Editor, line 52: @jasonjoh In step 11, we say "type Mail." (with a period), and in step 12, we say "type User." (with a period). Are these periods needed?

In the other doc, it just says "type User" (no period). 

Which is right?

jasonjoh: Basically it's narrowing the search even more. I think if you just type `Mail` you'll see more results, but `Mail.` will eliminate matches like `MailboxSettings.Read`, etc. Same with `User.`. Definitely leave the periods.

The other doc is wrong - you can tell they copied this one and repurposed the steps. `User` isn't what they mean there, it should be `ExternalItem` based on what they say to select.

concepts/use-postman.md
Linda-Editor: Thanks, Jason! On the other doc, it says the following: 
"The Azure AD application now has permissions to make requests on behalf of a user to call `ExternalItem.ReadWrite.All`, and as an application for `ExternalItem.ReadWrite.All`."

Would we apply the same type of updates there? Or is this one slightly different?

concepts/use-postman.md
Linda-Editor: Okay, in the other doc, I updated it to the following (changed User to ExternalItem):
11. Select **Application permissions**, type `ExternalItem`, and then select **Application Permissions**.
12. Expand the **ExternalItem** options, and then select `ExternalItem.ReadWrite.All`.

concepts/use-postman.md
jasonjoh: Yeah the same changes should be made.

jasonjoh: The permission names are correct - they use the same name for delegated and application permissions here. The descriptions need to change. Something like:

- The application can read/write external items on behalf of the user.
- The application can read/write external items on its own behalf.

Linda-Editor:
Made changes and refreshed stage. I think this is ready to go now.

jasonjoh (approved):

Lauragra (approved):

PR #16953, opened 2022-05-18
Update proxyaddresses limitation on B2C user

juntakata:
proxyAddresses attribute has a uniqueness constraint and we should not update it on B2C user account. I worked with Jas Sri, and he concluded that we should mark it as unsupported for b2c entirely. He also suggested we should use otherMails or extension attributes instead.


Lauragra:
@juntakata , if this PR is approved, please apply the same update to the topic in the /beta folder. Thanks.

FaithOmbongi:
Hi @juntakata, this change is currently in review.

Cc @yyuank for visibility.

yyuank:
The behavior of proxyAddresses for AAD users and B2C users is the same. We don't support update of proxyAddresses directly on MS graph, but customer can update mail property, which will update proxyAddresses. We should change the wording as "Updating proxyAddresses on MS Graph is not directly supported. You can update mail property, which will update proxyAddresses".

juntakata:
Thanks Yuan, Jas says we should not use mail property on B2C. The issue is, every time you write to mail, it will copy the value to proxyAddresses. proxyAddresses cannot be modified, and maintains a history of the last 10 email you put in mail attribute. As the proxyAddresses has uniqueness constraint, users cannot use their old mail on a different account on B2C. Here is the scenario. 

User A has abc@contoso.com. The user A changes mail to xyz@contoso.comuser. Then user B tries to use abc@contoso.com -> they cannot as user A's proxyAddress contains stale value of abc@contoso.com.

The row I am updating is "mail". So, we do not need to mention proxyAddresses (sorry, I mentioned to proxyAddresses in the PR message.) I think my original change is still valid.

> For Azure AD B2C accounts, updating this property is not supported. Use otherMails or other properties for B2C accounts.

yyuank:
This behavior is the same with AAD users. We can't recommend customers not to use the mail property because of this edge case. Our experience tells us if mail property is empty, it'll break lots of apps, as apps have dependency on the mail property. Not sure if Admins can manage proxyAddresses for B2C users with Exchange powershell or M365 Admin Center. For AAD users, that's the recommendation we've given to customers.

api-reference/v1.0/resources/user.md
```diff
@@ -171,7 +171,7 @@ This resource supports:
...
-|mail|String|The SMTP address for the user, for example, `jeff@contoso.onmicrosoft.com`.<br>Changes to this property will also update the user's **proxyAddresses** collection to include the value as an SMTP address. For Azure AD B2C accounts, this property can be updated up to only ten times with unique SMTP addresses. This property cannot contain accent characters.<br><br>Returned by default. Supports `$filter` (`eq`, `ne`, `not`, `ge`, `le`, `in`, `startsWith`, `endsWith`, and `eq` on `null` values).|
+|mail|String|The SMTP address for the user, for example, `jeff@contoso.onmicrosoft.com`.<br>Changes to this property will also update the user's **proxyAddresses** collection to include the value as an SMTP address. For Azure AD B2C accounts, updating this property is not supported. Use otherMails or other properties for B2C accounts. This property cannot contain accent characters.<br><br>Returned by default. Supports `$filter` (`eq`, `ne`, `not`, `ge`, `le`, `in`, `startsWith`, `endsWith`, and `eq` on `null` values).|
```
FaithOmbongi, line 174: @juntakata - suggestion after the internal discussion. Please make this change to the `beta` doc as well.
```suggestion
|mail|String|The SMTP address for the user, for example, `jeff@contoso.onmicrosoft.com`. Changes to this property will also update the user's **proxyAddresses** collection to include the value as an SMTP address. This property cannot contain accent characters. <br/> **NOTE:** We do not recommend updating this property for Azure AD B2C user profiles. Use the **otherMails** property instead. <br><br>Returned by default. Supports `$filter` (`eq`, `ne`, `not`, `ge`, `le`, `in`, `startsWith`, `endsWith`, and `eq` on `null` values).|
```

FaithOmbongi (approved):
Hi @juntakata - I've added a suggestion based on the discussion we had internally. Please make the same change to the `beta` doc.

Approved with suggestions.

juntakata:
@FaithOmbongi I applied your suggestions and also updated beta document. Please review the changes and merge the updates Thank you!

PR #16974, opened 2022-05-19
Add new properties to email activity reports v1.0 APIs

kszb:
Add new properties to email activity reports v1.0 APIs.
There is some missing in the beta API doc before. But no new change in the beta API response.


isvargasmsft (approved):
@FaithOmbongi FYI, I reviewed this one. Please double check in case I overlooked something.

PR #16994, opened 2022-05-20
Update to Microsoft Graph documentation hub

Linda-Editor:
Added link to feedback page per Dana's request

### DIFF
--- a/concepts/index.yml
+++ b/concepts/index.yml
@@ -128,7 +128,7 @@ additionalContent:
               url: https://developer.microsoft.com/graph/support
             - text: Microsoft Graph Q&A community support
               url: /answers/products/graph
-            - text: Microsoft Graph feedback
+            - text: Microsoft Graph feedback portal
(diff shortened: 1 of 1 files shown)


concepts/index.yml
```diff
@@ -128,5 +128,7 @@ additionalContent:
               url: https://developer.microsoft.com/graph/support
             - text: Microsoft Graph Q&A community support
               url: /answers/products/graph
+            - text: Microsoft Graph feedback
```
Lauragra, line 131: It might be more clear to refer to this as the "Microsoft Graph feedback portal", so they understand that it's a destination for providing feedback; what do you think?

Linda-Editor: Yes, that's a good idea. I was wondering how to best phrase it. I thought about "Provide Microsoft Graph feedback" but then it wouldn't be parallel with the other entries, so "portal" works.

PR #16997, opened 2022-05-20
Delete synchronization-synchronizationtemplate-post.md

ArvindHarinder1:


ArvindHarinder1:
This is not a supported operation / doesn't work. Removing as there was recently a support case where a customer found the documentation and started trying it.

PR #17026, opened 2022-05-24
Purview rebrand update

Saisang:


LauraWi (changes requested):
Requires human review. Reference to "Microsoft 365 Security & Compliance portal" needs updating. Need developer SME to confirm any API updates.

Lauragra (approved):

PR #17052, opened 2022-05-25
EM create policy: include example for review settings

markwahl-msft:
In entitlement management in v1.0 API, I had included `reviewSettings` as a property of the accessPackageAssignmentPolicy resource, but didn't mention it in the crreate article's list of properties as one of the properties that could be included in a new policy when its created.  This PR adds the missing row to that table to clarify that a new policy can have reviewSettings, and includes an example of its syntax in the second example.


FaithOmbongi (approved):
Thank you @markwahl-msft for your work in this PR. Approved.

Issue #8760, opened 2020-06-17
Calendar Throttling when batching


svrooij:
Where can I find the throttling specifications for bulk changing calendars?

We created an multi-tenant application with client credentials that uses the batch endpoint to change user calendars. Until recently we could combine up to 15 changes in one batch and they would all complete successfully.

This 15-requests limit is discussed multiple times:
- https://github.com/microsoftgraph/microsoft-graph-docs/issues/4366
- https://stackoverflow.com/questions/56873802/microsoft-graph-api-batch-limit?noredirect=1#comment109103171_56873802

But recently we found that we can only combine up to 4 change requests when changing calendars.
All other requests in the batch fail with a 429 error (indicating we are throttled).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8df32e5f-8ccb-0c93-28d9-02392f9c7256
* Version Independent ID: 4aadd084-fb61-e89b-f63f-3b71622636b7
* Content: [Microsoft Graph throttling guidance - Microsoft Graph](https://docs.microsoft.com/en-us/graph/throttling?view=graph-rest-1.0)
* Content Source: [concepts/throttling.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/concepts/throttling.md)
* Product: **non-product-specific**
* Technology: **microsoft-graph**
* GitHub Login: @baywet
* Microsoft Alias: **MSGraphDocsVteam**

svrooij:
Hi Vincent (@baywet),
do you know if we can expect feedback on this documentation issue? 
Thanks!

baywet:
Hey Stephan,
I don't have a definitive answer at the moment but I've been looking into it.
It seems that the [deployment of throttling limits for Outlook services to v1](https://developer.microsoft.com/en-us/graph/blogs/beta-outlook-api-throttling-limits-deploying-to-v1-0-on-microsoft-graph-may-2020/) is what impacted the behavior and this is why you are hitting the 4 concurrent requests behavior.
Can you confirm you are using the v1 endpoint? And that the behavior changed around the beginning of June?

svrooij:
We can confirm both.
It was unclear to us that this also affected batching.

Going from EWS (up to 80 changes in one request) to Outlook api (20 requests) to Graph api (first 15) to graph api (4 changes) seems like a limit going the wrong direction.
[...]

PR #17078, opened 2022-05-27
docs: Batch limits on Batching page

### LINKED ISSUE #4366: Batch limit incorrect
By trial and error we found out the batch limit is actually 15 instead of the mentioned 20.
If we send batch requests with more then 15 they start failing.

It would be very nice if this could be extended to a more reasonable amount (or at least support the 20 mentioned here).

Apart from that would it be possible to extend this number to 50 (with the restriction that all the requests should be to the same resource type). So it would be possible to add or remove 50 events from a single users calendar.

---
#### Document Details

* ID: ac7ca25a-037d-5202-debc-92c06a2c8480
* Version Independent ID: 8ebb4dbc-76a1-00fd-e740-87eda29a40f0
* Content: [Known issues with Microsoft Graph - Microsoft Graph](https://docs.microsoft.com/en-us/graph/known-issues#feedback)
* Content Source: [concepts/known-issues.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/concepts/known-issues.md)
* GitHub Login: @jthake-msft

svrooij:
In my opinion the limits of batching should be on the page describing batching.

There have been lengthy discussions on this topic several times [here](https://stackoverflow.com/questions/56873802/microsoft-graph-api-batch-limit) and in the issue below

Fixes #4366

This PR doesn't fix the [actual discussion](https://github.com/microsoftgraph/microsoft-graph-docs/issues/8760#issuecomment-648878458) on the batching limits, it just clarifies them to all users.

### DIFF
--- a/concepts/json-batching.md
+++ b/concepts/json-batching.md
@@ -196,15 +196,15 @@ If an individual request fails, any request that depends on that request fails w
 
 An additional use case for JSON batching is to bypass URL length limitations. In cases where the filter clause is complex, the URL length might surpass limitations built into browsers or other HTTP clients. You can use JSON batching as a workaround for running these requests because the lengthy URL simply becomes part of the request payload.
 
-## Batch size is limited
+## Batch size limitations
 
-JSON batch requests are currently limited to **20** individual requests.
+JSON batch requests are currently limited to 20 individual requests, in addition to the following limitations:
 
-* Depending on the APIs part of the batch request, the underlying services impose their own throttling limits that affect applications that use Microsoft Graph to access them.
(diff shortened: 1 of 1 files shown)


svrooij:
@baywet was the original author of the limits, but in my opinion they should be on the batching page. Also the outlook part isn't specified anywhere and takes guess work.

baywet (approved):

baywet:
Thanks for the contribution. I'm not working on that area anymore, but from I can remember the limits you documented look correct, assuming they haven't changed since.

Lauragra (approved):

PR #17130, opened 2022-06-02
Update choose-authentication-providers.md

ejazhussain:
scopes were wrong under Client credentials provider with certificate approach.


jasonjoh:
@ejazhussain thanks for catching this!

PR #17143, opened 2022-06-03
Update displayName in Microsoft Graph TOC

Linda-Editor:
PROBLEM: The displayName tag was attached to folders, but it doesn't work on empty folders. Therefore, users could not get to the Overview topics for many sections when using the search filter. 

FIXES:
- Moved displayName to first href reference in a category; usually, this was the Overview topic
- Added missing displayName elements where needed
- Updated some of the Overview topic TOC titles to be more descriptive so as to provide a better search experience. For example, in the Use the APIs section, there were several Overview topics that would come up in the same search, and they all had the same search result: Overview - Develop > Use the API. So it was impossible to tell what they were the overview for. Adding a more descriptive title fixed that issue.
- Made other tweaks where needed

### DIFF
--- a/concepts/toc.yml
+++ b/concepts/toc.yml
@@ -12,7 +12,7 @@ items:
       - name: Services in Microsoft Graph
         href: overview-major-services.md
       - name: What's new
-        expanded: false
+        expanded: true
         items:
           - name: What's new highlights
             href: whats-new-overview.md
@@ -28,6 +28,8 @@ items:
             href: deployments.md
           - name: Hybrid deployments (deprecated)
             href: hybrid-rest-support.md
+      - name: Quick start
+        href: https://developer.microsoft.com/graph/quick-start
       - name: Tutorials
         items:
           - name: Overview
@@ -51,8 +53,6 @@ items:
             href: /graph/tutorials/python
           - name: TypeScript
             href: /graph/tutorials/typescript
-      - name: Quick start
-        href: https://developer.microsoft.com/graph/quick-start
       - name: Versioning and support
         href: versioning-and-support.md
       - name: Terms of use
@@ -103,9 +103,9 @@ items:
             href: data-connect-filtering.md
           - name: Integrate with PAM
             href: data-connect-pam.md
-          - name: Frequently asked questions
+          - name: Data Connect FAQ
             href: data-connect-faq.md
-            displayName: Microsoft Graph Data Connect, FAQ
+            displayName: Microsoft Graph Data Connect, frequently asked questions
       - name: Calendar
         items:
           - name: Overview
@@ -377,10 +377,10 @@ items:
         items:
           - name: Overview
             href: notifications-concept-overview.md
-            displayName: deprecated
+            displayName: deprecated, notifications
           - name: Get started
             items:
-              - name: Get started overview
+              - name: Integrate with notifications
                 href: notifications-integration-e2e-overview.md
               - name: App registration
                 href: notifications-integration-app-registration.md
@@ -677,9 +677,9 @@ items:
                     href: migrate-azure-ad-graph-client-libraries.md
               - name: "4: Deploy, test, and extend"
                 href: migrate-azure-ad-graph-deploy-test-extend.md
-              - name: Migration FAQ
+              - name: Azure AD Graph migration FAQ
                 href: migrate-azure-ad-graph-faq.md
-                displayName: migrate, Azure AD Graph, frequently asked questions
+                displayName: migrate, frequently asked questions
               - name: Configure Azure AD Graph permissions
                 href: migrate-azure-ad-graph-configure-permissions.md
           - name: Exchange Web Services
@@ -824,7 +824,7 @@ items:
               - name: Right-to-left support
                 href: toolkit/customize-components/right-to-left.md
       - name: Resources
-        expanded: false
+        expanded: true
         items:
           - name: Best practices
             href: best-practices-concept.md
--- a/concepts/toc.yml
+++ b/concepts/toc.yml
@@ -105,7 +105,7 @@ items:
             href: data-connect-pam.md
           - name: Frequently asked questions
             href: data-connect-faq.md
-            displayName: Microsoft Graph Data Connect
+            displayName: Microsoft Graph Data Connect, FAQ
       - name: Calendar
         items:
           - name: Overview
@@ -679,7 +679,7 @@ items:
                 href: migrate-azure-ad-graph-deploy-test-extend.md
               - name: Migration FAQ
                 href: migrate-azure-ad-graph-faq.md
-                displayName: migrate, Azure AD Graph
+                displayName: migrate, Azure AD Graph, frequently asked questions
               - name: Configure Azure AD Graph permissions
                 href: migrate-azure-ad-graph-configure-permissions.md
           - name: Exchange Web Services
--- a/concepts/toc.yml
+++ b/concepts/toc.yml
@@ -532,6 +532,7 @@ items:
         items:
           - name: Overview
             href: auth/index.yml
+            displayName: authentication, authorization
           - name: Basics
             href: auth/auth-concepts.md
           - name: Register your app
@@ -552,31 +553,33 @@ items:
         items:
           - name: Overview
             href: use-the-api.md
+            displayName: use the API, Microsoft Graph API
           - name: Access data and methods
             href: traverse-the-graph.md
           - name: Paging
             href: paging.md
           - name: Query parameters
-            displayName: OData query parameters
             items:
               - name: Use query parameters
                 href: query-parameters.md
+                displayName: OData query parameters, OData
               - name: Use the search query parameter
                 href: search-query-parameter.md
               - name: Advanced query capabilities
                 href: aad-advanced-queries.md
           - name: Batching
-            displayName: Batch requests
             href: json-batching.md
+            displayName: Batch request
           - name: Throttling
             href: throttling.md
           - name: Change notifications
-            displayName: Webhooks, cloud printing, Universal Print, Outlook mail, Outlook calendar, Outlook contacts, Outlook contact, OneDrive, OneDrive for Business, OneDrive business, OneDrive personal, SharePoint, Security, Users, Groups
             items:
               - name: Overview
                 href: webhooks.md
+                displayName: WebHooks, cloud printing, Universal Print, Outlook mail, Outlook calendar, Outlook contacts, Outlook contact, OneDrive, OneDrive for Business, OneDrive business, OneDrive personal, SharePoint, Security, Users, Groups
               - name: Tutorial
                 href: /learn/modules/msgraph-changenotifications-trackchanges
+                displayName: change notification
               - name: Missing change notifications
                 href: webhooks-lifecycle.md
               - name: Notifications with resource data
@@ -589,6 +592,7 @@ items:
                 items:
                   - name: Overview
                     href: teams-change-notification-in-microsoft-teams-overview.md
+                    displayName: change notifications, Teams, Microsoft Teams
                   - name: Teams and channels
                     href: teams-changenotifications-team-and-channel.md
                   - name: Team and channel membership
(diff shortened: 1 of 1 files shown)


concepts/toc.yml
```diff
@@ -12,16 +12,18 @@ items:
       - name: Services in Microsoft Graph
         href: overview-major-services.md
       - name: What's new
-        expanded: true
+        expanded: false
```
Lauragra, line 15: Please don't collapse any nodes for now because specific nodes are expanded by design.

concepts/toc.yml
```diff
@@ -48,6 +51,8 @@ items:
             href: /graph/tutorials/python
           - name: TypeScript
             href: /graph/tutorials/typescript
+      - name: Quick start
```
Lauragra, line 54: Maybe we should move the Quick start to come right before the Tutorials node, just because this might be the "first" thing they do before completing a tutorial walkthrough?

concepts/toc.yml
```diff
@@ -75,17 +82,19 @@ items:
...
+            displayName: applications, apps
           - name: Automate SAML-based SSO configuration
             href: application-saml-sso-configure-api.md
           - name: Configure application proxy
             href: application-proxy-configure-api.md
       - name: Accessing data at scale
-        displayName: Microsoft Graph Data Connect
         items:
-          - name: Overview of Data Connect
+          - name: Data Connect overview
```
Lauragra, line 92: For consistency with the other nodes, I suggest we make the TOC entry "Overview".

concepts/toc.yml
```diff
@@ -649,13 +679,14 @@ items:
                 href: migrate-azure-ad-graph-deploy-test-extend.md
               - name: Migration FAQ
                 href: migrate-azure-ad-graph-faq.md
+                displayName: migrate, Azure AD Graph, frequently asked questions
```
Lauragra, line 682: For the other FAQ topic we have FAQ in the displayName; is it better to use the same one consistently?

concepts/toc.yml
```diff
@@ -96,11 +105,12 @@ items:
             href: data-connect-pam.md
           - name: Frequently asked questions
```
Lauragra, line 106: We should probably change this entry here to FAQ, and update it in the topic title as well. FAQ is a recognized acronym and is a much more commonly searched term.

concepts/toc.yml
```diff
@@ -152,16 +163,19 @@ items:
       - name: Compliance
         href: compliance-concept-overview.md
       - name: Connecting external content
-        displayName: connectors
         items:
-          - name: Overview of Microsoft Graph connectors
+          - name: Connectors overview
```
Lauragra, line 167: This is another place where, for consistency, we might just use Overview.

concepts/toc.yml
```diff
@@ -173,30 +187,35 @@ items:
...
           - name: Business rules
             href: bookingsbusiness-business-rules.md
+            displayName: Bookings, Microsoft Bookings, booking
       - name: Device and app management
         expanded: true
         items:
           - name: Cloud printing
             items:
-              - name: Overview
+              - name: Cloud printing overview
```
Lauragra, line 216: Not sure why the change in some places to include the node name in the TOC entry while in other nodes we just use Overview. Let's make all the nodes consistent and just use Overview for the overview topic in each node.

concepts/toc.yml
```diff
@@ -350,14 +374,13 @@ items:
               - name: Error codes
                 href: onenote-error-codes.md
       - name: Notifications (deprecated)
-        displayName: Notifications
         items:
           - name: Overview
             href: notifications-concept-overview.md
+            displayName: deprecated
```
Lauragra, line 380: Are we sure that deprecated is what we want for the displayName? Maybe we include Notifications and deprecated? It's possible that even though the API is deprecated, someone will still search for it in the filter.

concepts/toc.yml
```diff
@@ -350,14 +374,13 @@ items:
...
           - name: Get started
-            displayName: Notifications
             items:
-              - name: Overview
+              - name: Get started overview
```
Lauragra, line 383: This might be a better TOC title: Integrate with notifications

concepts/toc.yml
```diff
@@ -567,8 +596,9 @@ items:
                 href: universal-print-webhook-notifications.md
               - name: Change notifications for Microsoft Teams
                 items:
-                  - name: Overview
+                  - name: Change notifications for Microsoft Teams resources
```
Lauragra, line 599: Seems like it's better to keep this as Overview for readability/scannability since it basically duplicates that parent node title.

concepts/toc.yml
```diff
@@ -595,10 +625,10 @@ items:
               - name: Get changes to users (example)
                 href: delta-query-users.md
           - name: Add custom data
-            displayName: Outlook calendar, Groups, Outlook mail, Outlook contacts, Outlook contact, People, Users
             items:
-              - name: Overview
+              - name: Add custom data to resources using extensions
```
Lauragra, line 629: I think we can keep this as Overview.

concepts/toc.yml
```diff
@@ -787,17 +824,22 @@ items:
               - name: Right-to-left support
                 href: toolkit/customize-components/right-to-left.md
       - name: Resources
-        expanded: true
+        expanded: false
```
Lauragra, line 827: We should keep everything expanded for now.

concepts/toc.yml
Linda-Editor: Done.

concepts/toc.yml
Linda-Editor: I had thought about that too, but the difference between the Data Connect and Connectors nodes and other overview nodes is that the folders don't say Data Connect or Connectors; they say "accessing data at scale" and "connecting external content".  And both of these were spelled out previously, probably by design. So that's why I didn't change them, and that's why I feel like we shouldn't change them.

concepts/toc.yml
Linda-Editor: I thought about that too. It could go either way. But since you noticed it too, I'll go ahead and make them consistent.

concepts/toc.yml
Linda-Editor: See my reply for Data Connect.

concepts/toc.yml
Linda-Editor: There is a method to my madness, and I'm going to push back on this suggestion. There are multiple instances where the search results show the exact same wording because of multiple overview topics in the same top node. The only way the user can know which topic they're looking at is if we spell it out. One reason this is happening is because someone in the past has added the same displayName entries for multiple overview topics, mostly in the API section. But it also happens because in a given folder, there is more than one overview, but the search results don't differentiate. While it may make the TOC inconsistent, it improves the search functionality. And my guess is that users use the search function more than they browse through the TOC.

concepts/toc.yml
Linda-Editor: Done

concepts/toc.yml
Linda-Editor: I'm pushing back for the same reasons I mentioned before.

concepts/toc.yml
Linda-Editor: For example, if you search on **Change notifications**, there are 6 results from the Develop > Use the API section. If we don't change the display name to be more specific, most of these results show up as Overview - Develop > Use the API. There's no way for the user to know which topic they're looking at, other than to painstakingly click each one, then go back and filter again, etc. And it also makes it look like the same result is showing up 6 times. I'm optimizing for the search experience, not the browse experience.

concepts/toc.yml
Linda-Editor: Done

concepts/toc.yml
Lauragra: Am I misunderstanding the purpose of the displayName attribute? If we apply a unique and specific displayName attribute to a TOC entry, won't that come up in the search filter regardless of the name of the TOC label?

Linda-Editor: Yes. However, under Use the API, where there are multiple Overviews, there are also multiple displayName entries for each TOC overview, meaning that all of the overview topics under Develop > API might show up in a single keyword search (such as Users). If we don't change the title for these overviews to be more specific, they show up identically in the search result as Overview - Develop > API. If you don't want this is to happen, we need to scope the keywords for each of these overviews to not include all the product/service names they're connected to. Getting identical search results for a given keyword is not helpful.

concepts/toc.yml
```diff
@@ -94,13 +105,14 @@ items:
             href: data-connect-filtering.md
           - name: Integrate with PAM
             href: data-connect-pam.md
-          - name: Frequently asked questions
+          - name: Data Connect FAQ
```
Lauragra, line 108: The context is probably clear here; can this just be FAQ?

concepts/toc.yml
```diff
@@ -173,30 +189,35 @@ items:
...
           - name: Business rules
             href: bookingsbusiness-business-rules.md
+            displayName: Bookings, Microsoft Bookings, booking
       - name: Device and app management
         expanded: true
         items:
           - name: Cloud printing
             items:
-              - name: Overview
+              - name: Cloud printing overview
```
Lauragra, line 218: I still recommend that we keep this node as Overview, to improve the browsing experience.

concepts/toc.yml
```diff
@@ -208,11 +229,11 @@ items:
             displayName: Intune
           - name: Cloud PC (preview)
             href: cloudpc-concept-overview.md
-          - name: Device updates (preview)
-            displayName: Windows updates
+          - name: Device updates (preview)           
             items:
-              - name: Overview
+              - name: Device updates overview
```
Lauragra, line 234: Overview here, too.

concepts/toc.yml
```diff
@@ -248,40 +270,44 @@ items:
...
                 href: tutorial-riskdetection-api.md
           - name: Governance
             items:
               - name: Manage access to resources
                 href: tutorial-access-package-api.md
+                displayName: governance, identity
               - name: Review access to your resources
                 items:
-                  - name: Overview of the access reviews API
+                  - name: Access reviews API overview
```
Lauragra, line 308: Let's make this just Overview.

concepts/toc.yml
```diff
@@ -684,24 +719,26 @@ items:
...
       - name: Use the toolkit
         items:
-          - name: Toolkit overview
+          - name: Overview
             href: toolkit/overview.md
+            displayName: Microsoft Graph Toolkit, toolkit
           - name: Get started
             items:
-              - name: Overview
+              - name: Get started overview
```
Lauragra, line 739: Let's just make this "Get started".

concepts/toc.yml
Linda-Editor: Only if we change the folder name to Data Connect. It's currently Accessing data at scale. Someone decided before my time that instead of using Data Connect for the folder name, we should put the intent here, which is fine. But we've received feedback that no one can find Data Connect resources.  Everything that comes up in search for this folder has the path **Learn > Accessing data at scale**.  People might not know that this means Data Connect. So I'd prefer to keep it for visibility.

concepts/toc.yml
Linda-Editor: There are two overviews under Device and app management, so if they appear in search results together, the user has no way of knowing which is which. This is why I made this change.

Improving the browse experience = diminishing the search experience (in some cases). I'm trying to fix the search experience; your priority is the browse experience. It all depends on which you think the user will use more. If most of our users get to our site via Google search, might they not also search the TOC rather than browse it? I know that when I've wanted to find something in this TOC, I've always used the search filter because there are a lot of files in this docset. 

I'll make these changes because you insist on them even though I've tried to explain the method to my madness. But I don't agree with them.

concepts/toc.yml
Linda-Editor: Done, but this also could be Access reviews API.

concepts/toc.yml
Linda-Editor: Done

PR #17170, opened 2022-06-07
Fixed shared channel doc nconsistencies

devjha-ms:
Fix shared channel doc inconsistencies as pointed out by Angela Chu-Hatoun in https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2374934/


akhilkohlimicrosoft (approved):

api-reference/beta/resources/team.md
```diff
@@ -90,6 +90,7 @@ For a POST request example, see [Request (create team in migration state)](/micr
 |:---------------|:--------|:----------|
 |allChannels|[channel](channel.md) collection|List of channels either hosted in or shared with the team (incoming channels).|
 |channels|[channel](channel.md) collection|The collection of channels and messages associated with the team.|
```
angelgolfer-ms, line 92: @devjha-ms , can you add the missing group navigation property to the table as well?

angelgolfer-ms (changes requested):
Hi @devjha-ms , thanks for following up with this PR. Please see my one comment.

api-reference/beta/resources/team.md
devjha-ms: Not added "group" in relationship as we don't expose /teams/{teamId}/group graph API.

angelgolfer-ms: @devjha-ms , further question to you offline on Teams, please take a look.

angelgolfer-ms:
@devjha-ms , per our Teams chat, I can approve this PR, but we should follow up on the group nav property of the team entity separately.

angelgolfer-ms (approved):

PR #17195, opened 2022-06-09
Update metadata for Learn docs: Users, Groups, Applications

Linda-Editor:
Updated meta title and description


concepts/application-proxy-configure-api.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: Configure Application Proxy using Microsoft Graph APIs
-description: Configure Application Proxy using the Microsoft Graph APIs to provide remote access and single sign-on to on-premises applications.
-author: FaithOmbongi
-ms.topic: conceptual
+title: "Configure Application Proxy using Microsoft Graph APIs"
+description: "Provide remote access and single sign-on to on-premises applications by configuring Application Proxy using the Microsoft Graph APIs."
```
Lauragra, line 3: Let's either make API singular, or remove the "the" before Microsoft Graph APIs.

concepts/application-proxy-configure-api.md
```diff
@@ -430,13 +430,13 @@ No Content - 204
...
-- [applicationTemplate: instantiate](/graph/api/applicationtemplate-instantiate?view=graph-rest-1.0)
-- [appRoleAssignment](/graph/api/resources/approleassignment?view=graph-rest-beta)
-- [connector](/graph/api/resources/connector?view=graph-rest-beta)
-- [connectorGroup](/graph/api/resources/connectorGroup?view=graph-rest-beta)
-- [implicitGrantSettings](/graph/api/resources/implicitgrantsettings?view=graph-rest-1.0)
-- [on-premises publishing profiles](/graph/api/resources/onpremisespublishingprofile-root?view=graph-rest-beta)
-- [servicePrincipal](/graph/api/resources/serviceprincipal?view=graph-rest-1.0)
-- [singleSignOnSettings](/graph/api/resources/onpremisespublishingsinglesignon?view=graph-rest-beta)
-- [user](/graph/api/resources/user?view=graph-rest-1.0)
+- [application](/graph/api/resources/application?view=graph-rest-1.0&preserve-view=true)
```
Lauragra, line 433: I wonder if it would be better to remove the view parameter here instead of adding the preserve parameter. That way, if in the future there is ever a "v2.0", the links will automatically point to the latest topic without our needing to manually update them.

concepts/azuread-users-concept-overview.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: "Overview of users in Microsoft Graph"
-description: "Users are the representation of an Azure Active Directory (Azure AD) work or school user account or a Microsoft account in Microsoft Graph. The **user** resource in Microsoft Graph is a hub from which you can access the relationships and resources that are relevant to your users."
+title: "Users in Microsoft Graph"
```
Lauragra, line 2: In general the title attribute matches the H1 title in the topic; can we keep the H1 title as-is here?

concepts/application-proxy-configure-api.md
Linda-Editor: Done

concepts/application-proxy-configure-api.md
Linda-Editor: Done for all articles in this PR.

concepts/azuread-users-concept-overview.md
Linda-Editor: Actually, the meta title is not necessarily supposed to match the H1 exactly (I read up on all the guidelines previous to starting this scrub). It's supposed to be shorter when possible. In this case, "overview" doesn't add much value, so I left it out.

Lauragra (approved):

Issue #5755, opened 2019-09-20
Unable to subscribe when user email contains apostrophe


AFrankowicz:
When trying to subscribe using Microsoft Graph Java SDK (version 1.6.0) I get an error:

```
{
  "error": {
    "code": "ExtensionError",
    "message": "Operation: Create; Exception: [Status Code: ServiceUnavailable; Reason: Mailbox info is stale.]",
    "innerError": {
      "request-id": "2b5bda0c-5df6-4dfc-804e-6d59eb015cf5",
      "date": "2019-09-20T08:47:59"
    }
  }
}
```

This occurs when user email contains apostrophe.

piotrci:
Pani Alicjo,

currently we have a bug affecting the Outlook resources; when a UPN is used in the resource path, and it contains an apostrophe, we do not correctly handle that case. we recommend developers use the guid user id instead of the UPN. when using the SDK, pass the guid id instead of UPN and everything should work. let us know if you run into more problems around this.

dziekujemy!

vanta:
Panie Piotrze,

Can you reference this bug here? Any ETA when it's fixed?

macmaxbh:
As a heads up--this workaround is no longer necessary; you should be able to subscribe to email resources by user email address even if the address contains apostrophes. I'm working to get the workaround information removed from the docs.

PR #17217, opened 2022-06-10
Remove apostrophe workaround as it is no longer needed

macmaxbh:
Webhooks documentation includes a workaround regarding notification subscriptions to Outlook resources when using an email with an apostrophe (originally added due to #5755). This workaround is no longer necessary as I have fixed the underlying problem; so I'm removing it.


Danielabom (approved):
@macmaxbh, based in our chat, this change doesn't affect any CSDLs. It looks good to me. Thanks.

PR #17278, opened 2022-06-16
Update chatmessage-update.md

BassamMSFT:
The change to clearly state the operation works only with Teams and not Skype as per the confirmation from the Engineering Team on the ICM : https://portal.microsofticm.com/imp/v3/incidents/details/312679767/home

### DIFF
--- a/api-reference/v1.0/api/chatmessage-update.md
+++ b/api-reference/v1.0/api/chatmessage-update.md
@@ -11,6 +11,8 @@ ms.prod: "microsoft-teams"
 
 Update a [chatMessage](../resources/chatMessage.md) object. Only the **policyViolation** property of a **chatMessage** can be updated.
 
+The update only works for chats where are members are Teams users. If one of the participants is using Skype, the operation will fail. This is a known limitation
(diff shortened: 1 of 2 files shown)


api-reference/v1.0/api/chatmessage-update.md
```diff
@@ -11,6 +11,8 @@ ms.prod: "microsoft-teams"
 
 Update a [chatMessage](../resources/chatMessage.md) object. Only the **policyViolation** property of a **chatMessage** can be updated.
 
+The update only works for chats where are members are Teams users. If one of the participants is using Skype, the operation will fail. This is a known limitation
```
Lauragra, line 14: Please revise as follows:
The update only works for chats where members are Microsoft Teams users. If one of the participants is using Skype, the operation will fail. 

Please also update the beta version of the topic. If there is no plan to fix this limitation, we should document it as the default behavior. If it's something you plan to fix later, you will want to add this to the Known issues concept topic.

Lauragra (approved):

PR #17289, opened 2022-06-16
Update metadata for Learn docs: Cloud communications & Compliance & Cross-device

Linda-Editor:
Updated meta page titles and descriptions.

Updated H1s and H2s and intro paragraphs as needed.

Removed view= from most links.

Fixed formatting and typos as needed.

### DIFF
--- a/concepts/activity-feed-concept-overview.md
+++ b/concepts/activity-feed-concept-overview.md
@@ -1,29 +1,33 @@
 ---
-title: "Using the activity feed API in Microsoft Graph to enable cross-device experiences"
-description: "When you use activities, these Microsoft experiences can start driving engagement with your app. You can also display activities in your apps to help users get back to what they were doing on any device, on any platform, including Windows, Android, and iOS."
+title: "Use the activity feed API to enable cross-device experiences"
+description: "Using Microsoft Graph, you can enable experiences that flow seamlessly between devices, create richer activities with Adaptive Cards, and help drive app usage."
 ms.localizationpriority: medium
 ms.prod: "project-rome"
 ---
 
-# Using the activity feed API in Microsoft Graph to enable cross-device experiences
+# Use the activity feed API in Microsoft Graph to enable cross-device experiences
 
-Activities make users more productive by helping them resume important tasks in your app quickly across devices. Microsoft helps drive user productivity with your apps through experiences like Windows Timeline, Windows Sets, Cortana Pick up Where I left off and Microsoft Launcher, which are all powered by the activity feed. When you use activities, these Microsoft experiences can start driving engagement with your app. You can also display activities in your apps to help users get back to what they were doing on any device, on any platform, including Windows, Android, and iOS.
+Activities make users more productive by helping them resume important tasks in your app quickly across devices. Microsoft helps drive user productivity with your apps through experiences like Windows Timeline, Windows Sets, Cortana Pick up Where I left off, and Microsoft Launcher, which are all powered by the activity feed. When you use activities, these Microsoft experiences can start driving engagement with your app. You can also display activities in your apps to help users get back to what they were doing on any device, on any platform, including Windows, Android, and iOS.
 
-## Why integrate with activities?
-### Enable experiences that flow seamlessly between Windows, Android, Linux, and iOS devices 
-Great applications help users do great things — enabling a wide range of creativity, productivity, and entertainment scenarios. Returning to a task can be a challenge, especially when a person wants to continue the task on a different device or platform. By incorporating activities into applications, you can help users return to recent tasks quickly using whatever screen is handy — they can move from web to mobile to desktop and back again. With history items, users can easily see which activities they used most recently, when, and for how long.   
+The following sections describe the benefits of integrating with activities.
 
-Each user activity represents a single destination within your app: a product page, TV show, document, or current campaign in a game. All you need is a deep link to resume the activity in your app. Use [get recent activities](/graph/api/projectrome-get-recent-activities?view=graph-rest-1.0) to create a list of recently viewed products for your shopping app or a current reading list for books & news articles. 
+## Enable experiences that flow seamlessly between Windows, Android, Linux, and iOS devices
 
-### Create richer activities for any experience with Adaptive Cards
-When activities are rendered in Microsoft experiences such as Windows Timeline, they're displayed using the [Adaptive Card](https://adaptivecards.io/) framework, which allows you to create beautiful, rich cards to showcase your app's activities. You can use the Adaptive Card SDK to render rich cards in your own app too. If you do not provide an adaptive card for each activity, we'll automatically create a simple activity card based on your application name and icon, the required Title field, and optional Description field. 
+Great applications help users do great things&mdash;enabling a wide range of creativity, productivity, and entertainment scenarios. Returning to a task can be a challenge, especially when a person wants to continue the task on a different device or platform. By incorporating activities into applications, you can help users return to recent tasks quickly using whatever screen is handy&mdash;they can move from web to mobile to desktop and back again. With history items, users can easily see which activities they used most recently, when, and for how long.
 
-### Let Microsoft help drive app usage with features that reach hundreds of millions of customers
-Integrating with user activities not only enables users to resume activities in your app seamlessly, it means tapping into a growing set of Microsoft experiences for Windows, iOS, and Android - experiences designed to drive user productivity and help users engage with your app on all devices. Using Microsoft Graph, you can integrate with user activities just once and reach hundreds of millions of consumers, and tens of millions of customers in organizations who use Windows as well as Microsoft products for iOS and Android devices.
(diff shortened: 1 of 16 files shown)


concepts/activity-feed-concept-overview.md
```diff
@@ -1,29 +1,33 @@
...
+Activities make users more productive by helping them resume important tasks in your app quickly across devices. Microsoft helps drive user productivity with your apps through experiences like Windows Timeline, Windows Sets, Cortana Pick up Where I left off, and Microsoft Launcher, which are all powered by the activity feed. When you use activities, these Microsoft experiences can start driving engagement with your app. You can also display activities in your apps to help users get back to what they were doing on any device, on any platform, including Windows, Android, and iOS.
 
-## Why integrate with activities?
```
Lauragra, line 12: Do you feel strongly about removing these Why headings? It's part of the template that was established for these conceptual overview topics: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211743/Conceptual-topics

I did have an issue initially with the stacked headings, but we followed up with the style guide team at the time and their feedback was that stacked headings can be okay. Notice now that it does say not to insert "filler text" between headings: https://docs.microsoft.com/en-us/style-guide/scannable-content/headings.

concepts/cloud-communications-callrecords.md
```diff
@@ -1,18 +1,19 @@
...
-description: "Call records give you insights into the calls and meetings that occur within your organization."
+title: "Call records in the cloud communications API"
+description: "Get insights into the calls and online meetings that occur within your organization when using Microsoft Teams or Skype for Business."
 author: "williamlooney"
 ms.localizationpriority: medium
 ms.prod: "cloud-communications"
 ---
 
-# Call records overview
+# Get call records
```
Lauragra, line 9: Let's update this H1 heading so the topic doesn't get confused with a GET operation topic: https://docs.microsoft.com/en-us/graph/api/callrecords-callrecord-get?view=graph-rest-1.0&tabs=http. This is more of an overview topic; can we use the same or similar string that we have for the title for the H1?

concepts/cloud-communications-calls.md
```diff
@@ -1,16 +1,17 @@
...
-description: "Learn about the supported call types and how they're used for the signaling process."
+title: "Supported call types in the cloud communications API"
+description: "Learn about the supported call types in the cloud communications API in Microsoft Graph and how they're used for the signaling process."
 author: "ananmishr"
 ms.localizationpriority: medium
 ms.prod: "cloud-communications"
 ---
 
-# Calls overview
+# Use supported call types
```
Lauragra, line 9: Let's not make this action-oriented since it's a conceptual article; can we just use "Supported call types"?

concepts/cloud-communications-online-meeting-artifacts.md
```diff
@@ -1,13 +1,13 @@
...
+title: "Get online meeting artifacts and permissions"
+description: "Find out about online meeting artifacts, such as attendance reports and recordings, and the permissions needed to fetch them."
 author: "mkhribech"
 ms.localizationpriority: medium
 ms.prod: "cloud-communications"
 ms.date: "09/20/2021"
 ---
 
-# Online meeting artifacts and permissions
+# Get online meeting artifacts and permissions
```
Lauragra, line 10: Maybe we don't use Get here as well? There is a Get onlineMeeting method topic. Also, the permissions are required to perform the operation, but the title sort of implies they get both artifacts and permissions.

Also, the Get onlineMeeting reference topic happens to link to this topic; we should do a search and replace for any links to topics that we're updating the titles for and update the link text as well: https://docs.microsoft.com/en-us/graph/api/onlinemeeting-get?view=graph-rest-1.0&tabs=http.

concepts/cross-device-concept-overview.md
```diff
@@ -1,37 +1,37 @@
...
+Microsoft is building a platform for creating experiences that transcend a single device so they can harmonize across devices&mdash;empowering you to create **human-centric** scenarios that move with the user and blur the lines between their devices, regardless of form factor or platform. Microsoft Graph provides a single unified endpoint to give you access to data from Azure Active Directory and Microsoft 365. Now through Microsoft Graph, you can also access the activities and devices that belong to your customers and enable rich human-centric experiences that cross devices and platforms.
 
-## Why invest in cross-device experiences?
```
Lauragra, line 17: Again, I suggest we keep the H2 and remove any filler text between it and the H3.

concepts/activity-feed-concept-overview.md
Linda-Editor: Thanks for the background. I knew you were going to ask about this, and I was planning to give you a heads up. :-) I think I only ended up doing this on a few topics before I realized that it was probably a template change that writers were required to use. However, I think that SEO was not a factor in the decision to design this template because this style is really bad for SEO and navigation. Basically, the right nav for every overview page says, "Why integrate...", "API reference", "Next steps". The headings for the bulk of the content are not present in the right nav. Search engines now scan documents for H2s to determine the structure of the article. In fact, sometimes the list of H2s is what appears in search descriptions. If they can't find a good structure, they don't index the page. So that's why I changed these. For the few that I've done, I'd like to leave them as is for now and perhaps you can test their discoverability in SERPs in a few months vs. the pages that I didn't change. Also, it might be worthwhile to revisit these templates with SEO in mind.

concepts/cloud-communications-callrecords.md
Linda-Editor: Good idea. I changed it to "Obtain call records" for the H1 and TOC entry, and I added "Obtain" to the meta title. This way it keeps the original intent without using "Get".

concepts/cloud-communications-calls.md
Linda-Editor: Okay. I changed the H1 to "Supported call types" but I'm leaving "Use..." in the TOC for consistency with the rest of the TOC.

concepts/cloud-communications-concept-overview.md
```diff
@@ -1,94 +1,96 @@
...
+description: "Use the cloud communications API in Microsoft Graph to build bots that handle incoming calls, collaborate via group calls, send reminders, and set up meetings."
 author: "ananmishr"
 ms.localizationpriority: medium
 ms.prod: "cloud-communications"
 ---
 
 # Cloud communications API overview
-The cloud communications APIs in Microsoft Graph add a new dimension to how your apps and services interact with users through various communications related features, such as calling and online meetings. Grow your business by expediting how you respond to your customers’ needs and how your employees collaborate with each other.
 
-## Why integrate with the cloud communications APIs?
```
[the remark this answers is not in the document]
Linda-Editor: I'm going to disagree with you on this one (see my reasoning on the Calendar PR). All the potential search juice we can get from having the H2s present will disappear if we revert this topic.

concepts/cross-device-concept-overview.md
Linda-Editor: I ended up changing this one back because the Integrate section on this doc is short.

Linda-Editor:
@Lauragra I just noticed that the API is actually the Communications API and not the Cloud communications API, so I updated all the docs. Do you want me to also update the TOC entry?

Lauragra:
> @Lauragra I just noticed that the API is actually the Communications API and not the Cloud communications API, so I updated all the docs. Do you want me to also update the TOC entry?

For sure, the TOC entry should be Cloud communications. Let me follow up on what we should refer to it in the articles, and why the discrepancy. FYI, the team is looking into combining this node with Teamwork, so we shouldn't spend a lot of time on renaming nodes and such right now anyway.

concepts/activity-feed-concept-overview.md
Lauragra: That makes sense; we can probably update the template to make that section optional and make sure that the headings that follow are elevated to H2s. I want to hold off on merging this one though until we determine whether we want to refer to this as the cloud communications API vs. communications.

concepts/cloud-communications-callrecords.md
Lauragra: I actually think we don't need a verb here since this is more of an overview-type topic and covers more than one type of operation, not just GET/obtain. Can we just use "Call records" for the H1 title? Then the title attribute might make more sense as "Call records in the {cloud} communications API".

concepts/cloud-communications-calls.md
Lauragra: Why wouldn't we want the TOC entry to match the title/intent of the article? If we add "Use", that might imply that the article provides a discussion/recommendation about which call types to use, i.e. comparing supported vs. "unsupported" call types, but really it just provides an overview of the call types.

Linda-Editor: The TOC is basically a bulleted list, and in lists we try to keep the entry words consistent (all verbs, all nouns, etc.). That's what I'm trying to do in the TOC. It makes it harder to scan when there are verb entries and noun/gerund entries mixed together. But if that's something that this team doesn't mind, I'll change them back for now. This is something that your team might want to consider down the line.

Lauragra (approved):

PR #17356, opened 2022-06-20
Mail api doc updates

abheek-das:
1. Added filtering based on _changeType_ for /delta query on message resource.
2. Added missing details on _Prefer: outlook.timezone_ for createReply, reply, replyAll docs pages.
3. Added new concept page (concepts/outlook-things-to-know-about-send-mail.md) on actual steps involved in the backend before a mail is delivered to recipients.
4. Clarified that the _contentType_ of the attachment is optional and will be returned as null by default, when not set explicitly for _referenceAttachment_ and _itemAttachment_ resource types.
5. ~~Added a known issue with `$top` and the _/messages_ endpoint.~~


abheek-das:
Hi @angelgolfer-ms, hope you are doing well! Could you please help review this PR and provide your feedback? I also wrote a new concept page (outlook-things-to-know-about-send-mail.md). Could you also help suggest what a good hierarchy would be, to put this under?

api-reference/v1.0/api/message-delta.md
```diff
@@ -28,12 +28,26 @@ One of the following permissions is required to call this API. To learn more, in
...
+GET /users/{id}/mailfolders/{id}/messages/delta?changeType=created
+
+GET /me/mailfolders/{id}/messages/delta?changeType=updated
+GET /users/{id}/mailfolders/{id}/messages/delta?changeType=updated
+
+GET /me/mailfolders/{id}/messages/delta?changeType=deleted
+GET /users/{id}/mailfolders/{id}/messages/delta?changeType=deleted
+```
+
 ## Query parameters
```
angelgolfer-ms, line 52: Since changeType is another custom query options (aka query parameter in our docs' lingo), please add that to after line 64 in the query parameter table.

api-reference/v1.0/api/message-delta.md
```diff
@@ -28,12 +28,26 @@ One of the following permissions is required to call this API. To learn more, in
 |Application | Mail.ReadBasic.All , Mail.Read, Mail.ReadWrite |
 
 ## HTTP request
```
angelgolfer-ms, line 30: This comment is for the beta version of message-delta.md - since your changelog included the support for the beta version of message-delta.md, can you make the similar changes in that beta topic as well?

api-reference/v1.0/api/message-delta.md
```diff
@@ -28,12 +28,26 @@ One of the following permissions is required to call this API. To learn more, in
...
+GET /me/mailfolders/{id}/messages/delta?changeType=updated
+GET /users/{id}/mailfolders/{id}/messages/delta?changeType=updated
+
+GET /me/mailfolders/{id}/messages/delta?changeType=deleted
+GET /users/{id}/mailfolders/{id}/messages/delta?changeType=deleted
+```
+
 ## Query parameters
 
 Tracking changes in messages incurs a round of one or more **delta** function calls. If you use any query parameter
```
angelgolfer-ms, line 53: Line 53:
Let's be a bit clearer since now we're adding another custom query option "changeType":
Can you update the second sentence in this paragraph to say:
"... If you use any query parameter that is an OData system query option or the custom query option `changeType` , you must specify it in the initial **delta** request."

api-reference/v1.0/api/message-delta.md
abheek-das: Added

api-reference/v1.0/api/message-delta.md
abheek-das: Added.

api-reference/v1.0/api/message-delta.md
abheek-das: Added.

api-reference/beta/api/message-delta.md
```diff
@@ -50,6 +60,7 @@ includes the encoded, desired parameters.
 |:---------------|:--------|:----------|
 | $deltatoken | string | A [state token](/graph/delta-query-overview) returned in the `@odata.deltaLink` URL of the previous **delta** function call for the same message collection, indicating the completion of that round of change tracking. Save and apply the entire `@odata.deltaLink` URL including this token in the first request of the next round of change tracking for that collection.|
 | $skiptoken | string | A [state token](/graph/delta-query-overview) returned in the `@odata.nextLink` URL of the previous **delta** function call, indicating there are further changes to be tracked in the same message collection. |
+| changeType | string | A custom query filter to filter the delta response based on the changeType. Supported values are `created`, `updated` or `deleted`.|
```
angelgolfer-ms, line 63: Line 63:
At the end of the first sentence in the description, you mention "changeType". Do you mean "change type" (as in natural language) trying to explain what the query option "changeType" means? 
If we spell "changeType" as one word, we'd be implicitly saying this is some programmability term. I only know of 2 such usages - one is "changeType" property of the changeNotification complex type, which is technically a separate feature from delta query, the other is here the changeType customer query option.

api-reference/beta/api/message-delta.md
```diff
@@ -29,18 +29,28 @@ One of the following permissions is required to call this API. To learn more, in
 |Delegated (personal Microsoft account) | Mail.ReadBasic, Mail.Read, Mail.ReadWrite    |
 |Application | Mail.ReadBasic.All, Mail.Read, Mail.ReadWrite |
 
+To get all changes in messages in the specified mailFolder:
 ## HTTP request
```
angelgolfer-ms, line 33: Line 32:
FYI I flipped the order of lines 32 and 33 for you (to save us time reviewing this issue again).

concepts/delta-query-messages.md
```diff
@@ -53,6 +53,14 @@ See the [example](#example-synchronize-messages-in-a-folder) to learn how to use
   - The only supported `$orderby` expression is `$orderby=receivedDateTime+desc`. If you do not include an `$orderby` expression, the return order is not guaranteed.
 - There is no support for `$search`.
```
angelgolfer-ms, line 55: @abheek-das , I have a question and a comment, depending on the answer to the question.

Question: does this changeType query option assume there is an initial round of sync where all the messages in the folder are sync'd down?
It seems to me only that way would there be a point of reference that indicates whether a message is considered new, updated, or removed, right? 
Like, if there's no base line (no initial sync), what is being compared to such that a message can be considered updated?

angelgolfer-ms: I think this is relevant because if a customer takes a holistic approach and starts from the concept overview topic, the customer may be aware that AAD resources that support delta support something called "[sync from now](https://docs.microsoft.com/en-us/graph/delta-query-overview?view=graph-rest-beta#use-delta-query-to-track-changes-in-a-resource-collection)".  
Sync from now saves customers the initial round of sync-ing data to the local store from the remote store. In the case of messages, even if one wants only changes of a certain type, the initial round still syncs down all the data from the remote store, is that right?

angelgolfer-ms (changes requested):
Hi @abheek-das , I have only reviewed the content updates in this PR pertaining to delta query. In the interest of time, you can take a look at those comments first. I'll continue reviewing in the afternoon tomorrow.

api-reference/beta/api/message-replyall.md
```diff
@@ -49,6 +49,7 @@ POST /users/{id | userPrincipalName}/mailFolders/{id}/messages/{id}/replyAll
 |:---------------|:--------|:----------|
 | Authorization  | string  | Bearer {token}. Required|
 | Content-Type | string  | Nature of the data in the body of an entity. Required <br/> Use `application/json` for a JSON object and `text/plain` for MIME content |
+| Prefer: outlook.timezone | string | Sets the timezone in the email thread to the desired format while replying to an email. Values can be any of the outlook [supportedTimeZones](outlookuser-supportedtimezones.md). If the header is not specified, by default UTC timezone is followed. Optional.|
```
angelgolfer-ms, line 52: This comment is about support customizing the time zone for those 3 message methods that return a message resource - where, which property/properties would setting the time zone take effect?

Your description says "timezone in the email thread". That comes down to individual message objects in the email thread, because (unlike the COM object model) we don't expose a resource that represents an email thread. If that's the case, there are only 3 date/time-related properties in message - createDateTime, lastModifiedDateTime, sentDateTime - each of which we say the time zone is UTC: 
"The date and time information uses ISO 8601 format and is always in UTC time..."

If that's changing, then you'd please modify accordingly the descriptions of those 3 properties in message.md as well?

angelgolfer-ms: BTW, assuming indeed that those properties no longer have to be in UTC, and, we're referring using the Prefer header to set the time zone of the 3 date/time properties in each message of an email thread, then, I'd suggest rephrasing the Prefer header description as follows:
"Sets the time zone for **createdDateTime**, **lastModifiedDateTime**, and **sentDateTimeproperties** of each message in the email thread. Values can be any of the [supportedTimeZones](outlookuser-supportedtimezones.md) configured for the user. If not specified, those date-time values are returned in UTC. Optional."

angelgolfer-ms: Please make sure indeed we mean each message in the email thread? I just deduced from your existing description, I don't know the actual behavior.

Also, please note that we should use "time zone" as two words. I just checked how the style guide presents that term.

api-reference/beta/resources/itemattachment.md
```diff
@@ -30,7 +30,7 @@ Derived from [attachment](attachment.md).
 ## Properties
 | Property	   | Type	|Description|
 |:---------------|:--------|:----------|
-|contentType|String|The content type of the attachment.|
+|contentType|String|The content type of the attachment. Optional. Returned as `null` by default, when not set explicitly.|
```
angelgolfer-ms, line 33: This comment applies to the identical changes for contentType in itemattachment.md and referenceattachment.md:
Please leave the term "Optional." to the last thing in a description.

changelog/Microsoft.Exchange.json
```diff
@@ -1,5 +1,41 @@
 {
   "changelog": [ 
+    {
+      "ChangeList": [
+        {
```
angelgolfer-ms, line 5: This comment applies to both changelist records for v1 and for beta:
Please see example 18 in the [changelog](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog) topic to fix your changelist JSON.

concepts/delta-query-messages.md
abheek-das: > Synchronizing message items in a local store can use the **delta** query for both the initial full synchronization along with subsequent incremental synchronizations. Typically, you would do an initial full synchronization of all the messages in a folder (for example, the user's Inbox), and then, get incremental changes to that folder periodically.
> 
> To get incremental changes of only a certain type - created, updated, or deleted since the initial sync - do an initial round of synchronization of all the messages in the folder, and then get incremental changes of the specific type in subsequent rounds. Specify the desired change type as a query option in the initial **delta** request; Microsoft Graph automatically encodes any OData and custom query options into the @odata.nextLink or @odata.deltaLink provided in the response.

Yes, that's the correct understanding - the initial round of sync is needed to establish baseline.


> So assuming that there still should _always_ be an initial sync, like before, then I think we need to clarify that at the top of this topic, specifically the second paragraph. Let's change the second paragraph to the following second and third paragraphs (the third paragraph is new):
> 
> "Synchronizing message items in a local store can use the **delta** query for both the initial full synchronization along with subsequent incremental synchronizations. Typically, you would do an initial full synchronization of all the messages in a folder (for example, the user's Inbox), and then, get incremental changes to that folder periodically.
> 
> To get incremental changes of only a certain type - created, updated, or deleted since the initial sync - do an initial round of synchronization of all the messages in the folder, and then get incremental changes of the specific type in subsequent rounds. Specify the desired change type as a query option in the initial **delta** request; Microsoft Graph automatically encodes any OData and custom query options into the @odata.nextLink or @odata.deltaLink provided in the response. "

I agree this makes things much clearer, I have updated the doc as per above.

> I think this is relevant because if a customer takes a holistic approach and starts from the [conceptual overview](https://docs.microsoft.com/en-us/graph/delta-query-overview?view=graph-rest-beta#use-delta-query-to-track-changes-in-a-resource-collection) topic, the customer may be aware that AAD resources that support delta support something called "[sync from now](https://docs.microsoft.com/en-us/graph/delta-query-overview?view=graph-rest-beta#use-delta-query-to-track-changes-in-a-resource-collection)".
> Sync from now saves customers the initial round of sync-ing data to the local store from the remote store. In the case of messages, even if one wants only changes of a certain type, the initial round still syncs down all the data from the remote store, is that right?

Yes, the initial round of sync cannot be avoided even when only syncing changes of a certain type.

api-reference/beta/api/message-delta.md
abheek-das: Makes sense. I actually just changed "...based on the changeType..." to "...based on the type of change..." to make it a bit more clearer.

concepts/known-issues.md
```diff
@@ -507,6 +507,9 @@ The following limitations apply to query parameters:
 * `$count`:
   * Not supported in Azure AD B2C tenants.
   * When using the `$count=true` query string when querying against directory resources, the `@odata.count` property will be present only in the first page of the paged data.
```
angelgolfer-ms, line 509: Lines 511-512:
This comment is about listing this behavior as a known issue.
The behavior that the item types in an Exchange Online folder are not homogenous (not all items are of the same type) has existed for at least two decades. Assuming there's no plan to change that behavior, then this behavior should not be in the known issues topic; known issues are described as such only if there's a plan to fix the issue in a semester. Otherwise the behavior should be considered as default behavior and described in the corresponding topics where the behavior crops up.

api-reference/beta/resources/itemattachment.md
abheek-das: Fixed for both intemattachment & referenceattachment and in both beta & v1.0 version of docs.

concepts/known-issues.md
angelgolfer-ms: About the following sentence in line 512:
"To avoid this, it's recommended to use $top with a specific mailFolder (like `/me/mailFolders/Inbox/messages`) which *usually* contains only messages as opposed to `/me/messages` which tries to get messages from the entire mailbox."
Respectfully, I don't think we should suggest that usually message folders contain only email messages. It depends. 
Because Outlook/Exchange allows folders to contain more than one message type, for example, some organizations can run add-ins take advantage of this characteristic and can create heterogeneous folders.

changelog/Microsoft.Exchange.json
abheek-das: Fixed.

concepts/known-issues.md
angelgolfer-ms: Instead, I suggest to add a paragraph to an "include" file, and expose that paragraph in a few topics where this is relevant:

- query-parameters.md (under the [top](https://docs.microsoft.com/en-us/graph/query-parameters#top-parameter) parameter section)
- paging.md (in the last bullet list of the topic)
- user-list-messages.md (after the 4th paragraph)
- message-delta.md

The paragraph goes like this:
"If you include `$top` to control the page size when getting multiple messages, it's possible that the response contains fewer messages than the page size. This is because item types are not necessarily homogenous in folders on Exchange Online. The same folder can contain messages and other item types. If the operation returns an @odata.nextlink, continue calling the operation using the URL in the latest @odata.nextlink until the operation no longer returns an @odata.nextlink."

api-reference/beta/api/message-replyall.md
abheek-das: @angelgolfer-ms Hmm, yeah I get the ambiguity here. So - 
- None of the createdDateTime, lastModifiedDateTime, sentDateTime, receivedDateTime is impacted because of adding this header & are unchanged.
- However, in the Reply's "Send Date: XXXX" is modified as per the specified time zone. See highlighted in the examples below:
  - With Prefer: outlook.timezone=**"IndianStandard Time"**
    
image

   - With Prefer: outlook.timezone=**"Eastern Standard Time"**
 
image

- So I think I should probably change the description as - "Sets the _Sent Date_ in the desired time zone format while replying to a message. Values can be any of the outlook [supportedTimeZones](outlookuser-supportedtimezones.md). If the header is not specified, by default UTC timezone is followed. Optional."

concepts/known-issues.md
abheek-das: So, this is a bit interesting. 

1. You are right that the Exchange folders are not homogenous, and that behavior will not change.
2. The issue here was that, because of this behavior when someone specifies let's say, $top=500 and the mailbox contain a lot of non-message items, the request used to time out because MS Graph times out before Exchange can complete the request. So as per the new behavior, in such cases we'll just send a next link to avoid the timeout and exit gracefully. 
3. In parallel, I checked with the engineering team, to solve for this we are trying another fix which would send the query to the Exchange store with appropriate filters to get only a certain item type. This would potentially solve this issue, but it's not confirmed yet.

So, for now - I am reverting this change in the known issues page. Based on #2 or #3, whichever way we decide to go, we will make the doc updates at a later stage.

api-reference/beta/api/message-replyall.md
angelgolfer-ms: Hi @abheek-das , if that's what the Prefer header does - sets the "Sent" field of the reply draft or reply message (depending on the API whether it's createReply, createReplyAll vs reply, replyAll) - should we say:

"Sets the time zone for the `Sent` field of the reply message (use "reply draft" if this description appears in createReply or createReplyAll) that this API creates based on the request body. The value can be any of the [supportedTimeZones](https://github.com/microsoftgraph/microsoft-graph-docs/pull/17356/outlookuser-supportedtimezones.md) configured for the user. If not specified, that `Sent` field is in UTC. Optional."

angelgolfer-ms: Pointing out a few things:

1. Would createReplyAll be in scope for this change too?
2. The Prefer header sets the time zone, not time zone format, right? The format is whatever ISO 8601 supports.

3. I wanted to make the description clear that the Prefer header acts on the one message (or draft) that the API call creates as a reply. Setting that Prefer header doesn't change the time zone in the Sent fields of any earlier messages in an email thread that the API is replying to, is that the behavior?

If we look at the [first example in createReplyAll](https://docs.microsoft.com/en-us/graph/api/message-createreplyall?view=graph-rest-beta&tabs=http#example-1-create-a-draft-in-json-format-to-reply-all-to-an-existing-message) by opening the file message-createreplyall.md in VS Code, and search for "sent", the example response body has a body/content property specifying an email thread consisting of 2 messages. Only the first Sent field would be impacted by the Prefer header, is that right?

4. And is it true that if the Content/type of the API is set to text/plain, then the Prefer header doesn't do anything?
The response has the body/content property in text format, not HTML, so there's no Sent field to be impacted?
If this is true, then the description for the Prefer header should be 2 paragraphs like below:

"Sets the time zone for the `Sent` field of the reply message in HTML (use "reply draft" if this description appears in createReply or createReplyAll) that this API creates based on the request body. The value can be any of the [supportedTimeZones](https://github.com/microsoftgraph/microsoft-graph-docs/pull/17356/outlookuser-supportedtimezones.md) configured for the user. If not specified, that `Sent` field is in UTC. 

Use this header only if you're specifying the `Content-Type: application/json` header to create the reply message (or "reply draft" if this description appears in createReply or createReplyAll) in HTML. If you use the `Content-Type: text/plain` header, this `Prefer` header does not have any effect. Optional."

concepts/known-issues.md
angelgolfer-ms: #3 in your last message sounds good!
Do nothing in the docs for now sounds good too!

concepts/delta-query-messages.md
```diff
@@ -12,10 +12,9 @@ Delta query lets you query for additions, deletions, or updates to messages in a
...
-and incremental synchronization that retrieves all of the messages that have changed in that folder since
-the last synchronization. Typically, you would do an initial full synchronization of all the messages in a folder, and
-subsequently, get incremental changes to that folder periodically.
+Synchronizing message items in a local store can use the **delta** query for both the initial full synchronization along with subsequent incremental synchronizations. Typically, you would do an initial full synchronization of all the messages in a folder (for example, the user's Inbox), and then, get incremental changes to that folder periodically.
+
+To get incremental changes of only a certain type - created, updated, or deleted since the initial sync - do an initial round of synchronization of all the messages in the folder, and then get incremental changes of the specific type in subsequent rounds. Specify the desired change type as a query option in the initial **delta** request; Microsoft Graph automatically encodes any OData and custom query options into the @odata.nextLink or @odata.deltaLink provided in the response.
```
angelgolfer-ms, line 18: @abheek-das , thanks for incorporating my feedback to the above two paragraphs.

Per our Teams chat, yes, it would be helpful for customers if you can add one more example to show using the changeType header. You can please model the existing example, like using $select to get fewer properties, and a Prefer: odata.maxpagesize=2, so that it's easier for customers to see the big picture (nextLink, deltaLink, "borders" of each message), and not be distracted by extraneous data.

I suggest a simple case where we'll state the folder starts with having 4 messages, and after the first round, 2 more messages get created. Albeit other updates or deletions, the second round would show only what's been created. :-)

Round 1 - syncs down entire folder of 4 messages:
Initial call specifying the changeType, returning a nextLink. Syncs down 2 messages.
Second call using the last nextLink, returning a deltaLink. Syncs down remaining 2 messages.
Round 2:
One call using the last deltaLink, returning only the 2 newly created messages.

Please enhance where you see fit :-)
(All

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Things to know about sendMail"
+description: "Different steps involved in sending email using Microsoft Graph API till delivery."
+author: "abheek-das"
+ms.localizationpriority: high
+ms.prod: "outlook"
+---
+
+# Things to know about sendMail
+Sending email is a multi-step process. Most of the steps take place after the sendMail Microsoft Graph call has returned. Here is a simple summary of what typically
```
angelgolfer-ms, line 10: Lines 9-21:
Just want to be sure - we say "sendMail" and not "sending mail" to exclude the other APIs that sends a message or a draft, like the send, reply, replyAll, and forward actions?

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+ms.localizationpriority: high
+ms.prod: "outlook"
+---
+
+# Things to know about sendMail
+Sending email is a multi-step process. Most of the steps take place after the sendMail Microsoft Graph call has returned. Here is a simple summary of what typically 
+happens.
+
+## 1. Create a new message in sender's mailbox
+Microsoft Graph creates a new message (typically in the sender's Drafts folder), copies message content, recipients, and attachments from the JSON request to it, and
```
angelgolfer-ms, line 14: Line 14:
It would be correct if I say Outlook instead of "Microsoft Graph" as the subject, right?
Microsoft Graph is a developer platform which exposes a common REST API endpoint for developers to access Microsoft 365 services, including Outlook. It's Outlook that creates the message in the Drafts folder, and so on. Is that right?
Please adjust the text if that's not correct.

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+---
+
+# Things to know about sendMail
+Sending email is a multi-step process. Most of the steps take place after the sendMail Microsoft Graph call has returned. Here is a simple summary of what typically 
+happens.
+
+## 1. Create a new message in sender's mailbox
+Microsoft Graph creates a new message (typically in the sender's Drafts folder), copies message content, recipients, and attachments from the JSON request to it, and 
+saves it. If successful, the sendMail API returns a 2xx status.
+
```
angelgolfer-ms, line 16: Line 15:
For the 4 methods - sendMail, reply, replyAll, forward - that create a message and send it, we document in each of those topics that they return 202 Accepted if successful. So why do we say 2xx status here?
Is it in reality, they can return some other "successful" status code other than 202? If this is the case, that should be clarified in the docs.
Or, should line 15 say specifically 202 Accepted?

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+## 3. The outbound message is copied to the transport pipeline
+Next, the transport process reads message content from the sender's mailbox, converts it to MIME format, and stores it into the transport pipeline. If the sender 
+provided MIME content, then it is copied more or less intact; otherwise, transport serializes the message properties to construct MIME content.
+
+If step 3 fails, transport constructs a non-delivery report message and places it in the sender's inbox.
+
+## 4. The original message is moved to the Sent Items folder
+After all of this succeeds, transport calls back to store to accept responsibility for the message. In response, the Exchange store updates the message and (typically) 
+moves it from the Drafts folder to the Sent Items folder. (Depending on optional message properties, it may move to a different folder or delete the message instead.)
```
angelgolfer-ms, line 36: Line 35:
"... Exchange store updates the message and..." - update how? If we mention update, then we should be clearer please.

Also, if this article is specifically about sendMail, then the next phrase about moving the message from the Drafts folder to the Sent Items folder should not use the indefinite "typically".
Is it true that (only) if the saveToSentItems parameter of sendMail is set to false would the store deletes the message?

As for "... move to a different folder", what optional message properties are we referring to? For the properties we expose, I don't see anyone we have described that would save the message in some other folder after sending. If we're not pointing out which optional properties, then we should just not mention moving to a folder other than Sent Items or Deleted Items.

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+# Things to know about sendMail
+Sending email is a multi-step process. Most of the steps take place after the sendMail Microsoft Graph call has returned. Here is a simple summary of what typically 
+happens.
+
+## 1. Create a new message in sender's mailbox
+Microsoft Graph creates a new message (typically in the sender's Drafts folder), copies message content, recipients, and attachments from the JSON request to it, and 
+saves it. If successful, the sendMail API returns a 2xx status.
+
+If the sender provided MIME content, it is copied to a single property in the new message. Additionally, MIME content is parsed and copied to message properties, plus 
+the recipient and attachment tables. When complete, the sendMail API returns a 2xx status.
```
angelgolfer-ms, line 18: Line 17:
What is the significance of mentioning that Outlook copies the MIME content to a "single property", if we're not naming it? Can we skip that and just say "Outlook parses the MIME content and copies the corresponding content to message properties, and to recipients and attachments tables.

Also, about the "2xx" status code again - we have documented success returning 202 Accepted. Now the vagueness of "2xx" can make customers doubt our docs (if that's indeed a reality and some customers don't know yet). Can you clarify what the actual behavior can be and we can see what needs to be changed?

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+
+This step may fail if (for example) the sender's mailbox is full, or the network connection to the sender's server is down, or for other more exotic reasons. In that 
+case, the sendMail API returns a 4xx or 5xx status.
```
angelgolfer-ms, line 21: Line 20:
"exotic" as an adjective for reasons for messaging to fail doesn't seem to be an appropriate word, and besides, it's vague. 
I think we should leave that out. Stating the first two as _examples_ (implying there are other reasons not listed here) is sufficient.

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+
+Once step 1 is complete, your app's direct interaction with Microsoft Graph is over.
+
+## 2. Transport is notified of new outbound mail
+Next, M365 notifies its transport service that a new message is available for pickup.
+
```
angelgolfer-ms, line 27: Line 26:
Can I say "Outlook" instead of "Microsoft 365" (we don't use "M365" in external communication, that's breaking the brand)?

concepts/outlook-things-to-know-about-send-mail.md
```diff
@@ -0,0 +1,52 @@
...
+## 5. The transport pipeline performs policy evaluation and routing
+The next steps taken by transport include policy enforcement, routing, and next-hop delivery. Transport examines recipient email addresses and buckets them according 
+to what the initial routing hop has to be. Invalid recipient addresses may be are detected at this point; for those, a non-delivery report will be mailed back to the 
+sender. Transport then applies policies configured by tenant administrators. Such policies may reject the message based on its content, store additional copies, and 
+so on. After applying policy, transport fans out a copy of the message to each next-hop destination. More detail about the transport flow may be found at [mail flow 
+and the transport pipeline](https://docs.microsoft.com/en-us/Exchange/mail-flow/mail-flow?view=exchserver-2019&viewFallbackFrom=exchonline-ww).
+
+## 6. The message is delivered to its recipient(s)
+M365 transport may or may not be responsible for final delivery to all recipients. That depends on whether or not those recipients have M365 mailboxes.
+
```
angelgolfer-ms, line 47: Line 46:
Again, is it accurate to refer to "Microsoft 365 transport" as Exchange Online?
To me, the Marketing messaging for "Microsoft 365" is a collection of products and services. Are customers supposed to visualize transport as a layer across all services in Microsoft 365?

angelgolfer-ms: For the second instance of "M365 mailbox", can we be more specific and say "Exchange Online mailboxes"?

angelgolfer-ms (changes requested):
HI @abheek-das , FYI I finished reviewing all the topics in this PR. Please see my latest comments and direct commits in the concept topic.

concepts/known-issues.md
abheek-das: Okay great. Reverted the changes!

concepts/delta-query-messages.md
abheek-das: @angelgolfer-ms  - Added another example. Please take a look!

concepts/outlook-things-to-know-about-send-mail.md
abheek-das: We can make it `202 Accepted`. Made the change!

concepts/outlook-things-to-know-about-send-mail.md
abheek-das: Agreed with above. `202 Accepted` makes sense to me.

concepts/outlook-things-to-know-about-send-mail.md
abheek-das: I made it "Exchange Online". I've changed "Outlook"/"M35=65" to "Exchange Online" as well throughout the doc. Please review!

concepts/outlook-things-to-know-about-send-mail.md
abheek-das: `"After all of this succeeds..." - is it accurate if we say "if step 3 succeeds"? Or if you're not just referring to step 3, and step 2 can fail, then should we make it clear in step 2 what happens if step 2 fails?`
Changed it to - "After step 3 succeeds, ..."

`"... Exchange store updates the message and..." - update how? If we mention update, then we should be clearer please.`
Changed it to - "In response, the Exchange store updates the draft state and internet message headers associated with the message and moves it from the Drafts folder to the Sent Items folder."

`Is it true that (only) if the saveToSentItems parameter of sendMail is set to false would the store deletes the message?`
Yes, if saveToSentItems is explicitly specifed as false, then a copy is not saved in the sent items folder. Else it is, even if this property is not mentioned explicitly.

`As for "... move to a different folder", what optional message properties are we referring to? For the properties we expose, I don't see anyone we have described that would save the message in some other folder after sending. If we're not pointing out which optional properties, then we should just not mention moving to a folder other than Sent Items or Deleted Items.` 
Modified as suggested.

concepts/outlook-things-to-know-about-send-mail.md
abheek-das: @angelgolfer-ms  - I think this will be applicable to all the Graph APIs that can send mail. The backend process should be similar for all such Graph actions. In that case, what do you suggest should be an appropriate title for this article? Something like - 
_"How does sending mail works using Microsoft Graph APIs?"_

api-reference/beta/api/message-replyall.md
abheek-das: `And is it true that if the Content/type of the API is set to text/plain, then the Prefer header doesn't do anything?
The response has the body/content property in text format, not HTML, so there's no Sent field to be impacted?`

I think the header only impacts when the "comment" property is used instead of "body" property to reply to the message. When using "message" --> "body" --> "content", the entire body is constructed as part of the API response, so this won't have any effect!

abheek-das: #1, #2, #3 & #4 are all the correct understanding! :)
Made the suggested change in all the files!

concepts/delta-query-messages.md
```diff
@@ -342,6 +342,209 @@ The response contains a `@odata.deltaLink`. This indicates that all changes in t
...
+The first request specifies the following:
+
+- A `changeType` parameter to return only the _created_ messages in the subsequent delta response.
+- A `$select` parameter to return the `subject`, `sender`, and `isRead` properties for each message in the response.
+- The [optional request header](#optional-request-header), _odata.maxpagesize_, returning 2 messages at a time.
+
+<!-- {
+  "blockType": "ignored",
+  "sampleKeys": ["AQMkADNkNAAAgEMAAAA"],
+  "name": "get_messages_delta_4"
```
angelgolfer-ms, line 369: Line 368 and other sampleKeys lines in this example:
FYI sampleKeys is to specify ID values in the example request.
In this case, line 373 contains only one such value that indexes into the mailFolders collection.
The line should be:
"sampleKeys": ["AAMkAGUwNc4LTMzAAA="],

I'm fixing these values on your behalf to try to cut down on the review iterations.

angelgolfer-ms: More info on sampleKeys: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211782/Set-up-for-API-Doctor-validation?anchor=id-values-in-example-requests

concepts/outlook-things-to-know-about-send-mail.md
angelgolfer-ms: HI @abheek-das ,
You seem to have made a change in line 17 but left line 15 alone, i.e., it's still the following:
"If successful, the **sendMail** method returns an HTTP response status code in the 2xx category."

I'm changing that sentence in line 15:
"If successful, the **sendMail** method returns an HTTP response `202 Accepted` status code."

angelgolfer-ms (approved):
I think the PR is fine for publishing. Thanks!

angelgolfer-ms (commented):
Made adjustments, completed content review, approving PR.

PR #17400, opened 2022-06-22
Deprecate ediscovery topics under microsoft.graph.ediscovery subnamespace

SeunginLyu:
# Deprecate ediscovery topics under microsoft.graph.ediscovery subnamespace

## [api.md](https://microsoftgraph.visualstudio.com/onboarding/_git/onboarding?path=%2Freviews%2F4094-Add-EDiscovery-Service-API-To-Graph%2Fapi.md&version=GBmaster&_a=preview)


api-reference/includes/ediscovery-api-deprecation.md
```diff
@@ -0,0 +1,12 @@
...
+author: SeunginLyu
+ms.topic: include
+ms.date: 06/22/2022
+ms.localizationpriority: medium
+---
+
+<!-- markdownlint-disable MD041-->
+
+> [!CAUTION]
+> The eDiscovery APIs in the microsoft.graph.ediscovery subnamespace are deprecated.
```
Lauragra, line 11: Do you have a timeline by which the APIs will be removed? If so, then you can add "... deprecated and will be removed on XX."

api-reference/includes/ediscovery-api-deprecation.md
```diff
@@ -0,0 +1,12 @@
...
+> Please use the [eDiscovery APIs in the microsoft.graph.security subnamepsace](/graph/api/resources/security-api-overview?view=graph-rest-beta#ediscovery-preview) for your existing and new apps.
```
Lauragra, line 12: We are also going to need a Changelog entry that announces the deprecation. We can use the one we did when we added the APIs to the security subnamespace as a template, but we'll need to list each individual resource that has been deprecated, in individual ChangeList records.

SeunginLyu:
Reopening PR as deprecation plan becoming more active

PR #17448, opened 2022-06-24
[Intune] Update author metadata

rolyon:


Lauragra:
@rolyon , I'll go ahead and merge this PR, but be aware that the changes may get reintroduced in a future drop of the Intune docs. Please follow up with Jaiprakesh to ensure that this change is pushed upstream to the docs generation process.

PR #17461, opened 2022-06-27
Updated profilephoto-update.md for size dimensions

fabianwilliams:
Users are citing errors while updating photos while adhering to the size limit however they are not honoring the dimensions that are supported and called out in the photos endpoint description. its worth calling out here, same language used in also placed here as well for completeness.


isvargasmsft:
@kevinbellinger would you be so kind to perform a quick technical review on this PR, please? Either you or assign someone in the team. I'd deeply appreciate it.

kevinbellinger (approved):

PR #17486, opened 2022-06-28
added warning for online meeting roles

frankpeng7:
added warning for online meeting roles behavior

### DIFF
--- a/concepts/known-issues.md
+++ b/concepts/known-issues.md
@@ -154,9 +154,9 @@ The beta version offers a workaround, where you can use the **onlineMeetingProvi
 
 The Microsoft Teams client does not show the **View Meeting details**  menu for channel meetings created via the cloud communications API.
 
-### Presenter role assigned to non-AAD participants
+### Presenter role cannot be assigned to non-Azure AD participants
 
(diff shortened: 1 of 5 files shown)


concepts/known-issues.md
```diff
@@ -154,6 +154,10 @@ The beta version offers a workaround, where you can use the **onlineMeetingProvi
 
 The Microsoft Teams client does not show the **View Meeting details**  menu for channel meetings created via the cloud communications API.
 
+### Presenter role cannot be assigned to non-Azure AD participants
+
+Assigning the `presenter` or `coorganizer` role to users who are not registered in Azure Active Directory is not currently supported. Such requests will be accepted by the [create onlineMeeting](/graph/api/application-post-onlinemeetings.md) method, but the role will not be applied when the participant joins the online meeting. The [create onlineMeeting](/graph/api/application-post-onlinemeetings.md) method will reject the request and return a `400 Bad Request` error.
```
Lauragra, line 159: FYI, I removed the recommendation to update their code. If there is a workaround to the known issue, we should document it, but it was unclear as to whether there was a workaround or what update they should make. If you want to provide a workaround, feel free to open a new new PR to update the known issue. This is temporary, though, right, and we can remove it when fixed?

PR #17508, opened 2022-06-29
Add link to samples in TOC

jasonjoh:


jasonjoh:
@angelgolfer-ms Steve and I wanted to try adding a samples link in the TOC. Take a look at where I put it and let me know if you think there's a better place for it.

concepts/toc.yml
```diff
@@ -55,6 +55,8 @@ items:
             href: /graph/tutorials/python
           - name: TypeScript
             href: /graph/tutorials/typescript
+      - name: Samples
```
Lauragra, line 58: I'm not sure if this is the right location to surface this, since this is the "Explore" part of the user journey. Do we think that the point when they will want to browse samples would be the "Develop" part of the journey? In which case, we would add this link somewhere in the Develop section. Also, it does take them out of the context of the docs. with no easy way back, so we might not want to send them away at this point in their journey through the docs.

angelgolfer-ms: +1 - similar opinion as Laura's regarding the location of the new TOC item "Samples". Let's not break the developer's journey laid out as Explore -> Learn -> Develop.

Related is the use of the term "Tutorials", even though the original purpose of this PR is not about that TOC node. Please see my next comment.

concepts/toc.yml
```diff
@@ -55,6 +55,8 @@ items:
             href: /graph/tutorials/python
           - name: TypeScript
```
angelgolfer-ms, line 56: This comment applies to line 35, regarding the developer's journey.

In the "Explore" section, can we qualify the node "Tutorials" as "Getting-started tutorials" or some other term we can brainstorm on, to better differentiate this set of content with Quick start? Doesn't seem to me at the TOC label level we're clearly differentiating between "Quiek start" and these tutorials.

In general, the term tutorials is not limited to the getting started "Explore" stage, but applicable to the "Learn" stage as well. We know there's a need to group and better expose existing tutorial-type content for discovery. That would also encourage API producers to grow this content type. However, to-date, we haven't done that partly because the "Tutorials" node is used in the "Explore" section.

angelgolfer-ms (changes requested):
Please see my comments. Thanks.

concepts/toc.yml
jasonjoh: How do you feel about "Build your first app"?

concepts/toc.yml
jasonjoh: If we push this down into the Develop node, we sort of lose what we were trying to do :D. Essentially we're seeing two things: 

- No one is clicking the "Code Samples" link in the top nav bar.
- Said link goes to the samples browser with no filters, so it's not just Graph samples.

The hypothesis that @SteveWilkins1123 and I had was that folks aren't looking to that top-level nav because it's too generic, and that the TOC gets more eyes on.

Lauragra: The Develop section in the reference TOCs is a little less consolidated, what if you add an entry there (in addition to in the concepts TOC but it will be more buried there), something like "Explore samples", next to Use SDKs, Use the toolkit?
image

PR #17560, opened 2022-07-05
Added samlMetadataUrl property to Application entity in v1

JohnMaina:


FaithOmbongi (approved):
Thank you @JohnMaina for your work in this PR.

One gap (please also fill it in the beta doc): Add the **samlMetadataUrl** property to the list of updatable properties in the [Update application](https://docs.microsoft.com/en-us/graph/api/application-update?view=graph-rest-beta&tabs=http#request-body) doc. Exclude the word "Nullable." from the description in the Update doc.

Approved with suggestions.

PR #17577, opened 2022-07-06
The type of API is POST and not GET.

Altamish-MS:
There are  2 errors here- 
1) This API is of type POST 
2) The solutions in path is only for v1 and not for beta


Lauragra (approved):

Danielabom (approved):
This is a small doc update. PR approved.

PR #17611, opened 2022-07-08
Update user-get.md

SteveWilkins1123:
Edited requests so that the tryit button will work


FaithOmbongi (approved):
Thank you @SteveWilkins1123 for handling these fixes. Approved.

PR #17700, opened 2022-07-15
Updated eDiscovery sub service value under compliance to deprecated

izaktumz:
-Creating this PR to update  the eDiscovery value under compliance to eDiscovery (deprecated)
Compliance
  eDiscovery (deprecated)

We want to show the value as deprecated for the time being until the APIs in that namespace are removed.


Lauragra:
@izaktumz , the eDiscovery value isn't actually deprecated, it just moved to the security subnamespace. We have a deprecation PR for eDiscovery open already so I'll go ahead and close this one.

PR #17701, opened 2022-07-15
Concept changes for viva learning

hemashar:

PR #17709, opened 2022-07-16
Update teams-concept-overview.md

SteveWilkins1123:
Updated teams overview topic to include viva learning information


angelgolfer-ms:
Incorporated the content in this PR in PR #17701 . Made some additions there to first introduce Viva Learning from the end user perspective, and then lead in to the value of the debut API set to developers.

PR #17732, opened 2022-07-19
Minor doc fixes

FaithOmbongi:


concepts/known-issues.md
```diff
@@ -15,6 +15,10 @@ For information about the latest updates to the Microsoft Graph API, see the [Mi
 
 ## Applications
 
+### Application.ReadWrite.OwnedBy permission allows an app to read apps it doesn't own
+
+The _Application.ReadWrite.OwnedBy_ permission allows an app to call `GET /applications` and `GET /servicePrincipals` to list all applications and service principals in the tenant. This is a known issue that won't be resolved.
```
Lauragra, line 20: @FaithOmbongi , if this known issue won't be resolved, we should document it in the topics as the default behavior instead of including it in the Known issues topic. The known issues topic is for listing temporary issues that we plan to resolve in the future.

FaithOmbongi: Ah, got it. I'll remove references to the Known issues then and instead document this behavior in the Permissions reference and API topics.

PR #17786, opened 2022-07-20
Added clarification on transitive managers query

Licantrop0:
`$expand=manager($levels=n)` only works on a single user and not on the entire list of users. Need to make this clear in the doc.

### DIFF
--- a/api-reference/v1.0/api/user-list-manager.md
+++ b/api-reference/v1.0/api/user-list-manager.md
@@ -36,8 +36,7 @@ GET /users/{id | userPrincipalName}/manager
 Get the management chain:
 <!-- { "blockType": "ignored" } -->
 ```http
-GET /me?$expand=manager
-GET /users?$expand=manager($levels=n)
+GET /users?$expand=manager
 GET /users/{id | userPrincipalName}/?$expand=manager($levels=n)
 ```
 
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/api/user-list-manager.md
```diff
@@ -133,12 +134,10 @@ Content-type: application/json
 
 ### Example 2: Get manager chain up to the root level
 
-The following example shows a request to get the manager chain up to the root level. This request requires the **ConsistencyLevel** header set to `eventual` because `$count=true` query string is in the request. For more information about the use of **ConsistencyLevel** and `$count`, see [Advanced query capabilities on Azure AD directory objects](/graph/aad-advanced-queries).
+The following example shows a request to get the manager chain up to the root level.
```
FaithOmbongi, line 137: From past issues, customers don't seem to always check the "Optional query parameters" section first so sometimes they miss the advanced queries statement. And if we don't indicate it in this intro statement, they think *ConsistencyLevel: eventual* is redundant and obviously their query fails. This statement helps with self-service resolution when that happens.

FaithOmbongi (approved):
Hi @Licantrop0 - Thanks for highlighting the advanced query condition for using $levels. I've left suggestions for minor fixes. Otherwise, approved with suggestions.

PR #17787, opened 2022-07-21
Fixed issues with the try it experience across several API ref topics

SteveWilkins1123:


api-reference/v1.0/api/authentication-list-temporaryaccesspassmethods.md
```diff
@@ -76,7 +76,7 @@ If successful, this method returns a `200 OK` response code and a collection of
   "name": "list_temporaryaccesspassauthenticationmethod"
 }
 -->
-```msgraph-interactive
+```
```
FaithOmbongi, line 79: Hmm, I wonder if the snippets generator and other downstream tools will automatically infer this as HTTP? Maybe we should  use ```http instead of empty backticks to be on the safe side?

PR #17805, opened 2022-07-21
update link for protected APIs 

MariaWissler:


Danielabom (approved):
@MariaWissler thank you for clearing out my questions via chat this morning. The changes look good.

PR #17813, opened 2022-07-22
Update accessPackageAssignments to assignments

rjbooden:
In v1.0 the endpoint is assignments instead of the beta accessPackageAssignments


FaithOmbongi:
Thank you @rjbooden for raising this PR to fix this doc error. I didn't see your PR on time, but the doc was updated recently with these fixes and is now correct. I'll close your PR.

PR #17814, opened 2022-07-22
EDU Assignments: Adding Graph API best practices

sharad-sharma-msft:
This is initial draft of the article. I have put together pieces of information that will allow developers to save both latency, and bandwidth. There are also options on how to handle/pass cancellation token etc.

We might need to iterate over this article frequently to park all good information that will benefit our customers.


cristobal-buenrostro:
verified to solve warnings on url links

cristobal-buenrostro (approved):

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+
+This helps performance in the following ways:
+1.
+Can reduce the calls we make to our underlying storage to retrieve the assignment (especially if the application takes advantage of large properties and omits them from $select query parameter in certain scenarios).
+2.
+Reduces COGs for JSON deserialization in the application calling assignments.
+3.
+Less bits to transmit over the wire.
+
+### When using $filter, also specify $orderby
```
ssriniva-msft, line 176: Shouldn't we have a section that recommends using $filter first before asking to use $orderby (while using $filter)? In this section, we could offer examples like applying a createdDateTime filter in a call to GET ~/assignments and other common calls and explain the rationale for these filters

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+When a GET call with a $filter query parameter also has a $orderby, and the $orderby contains at least one property from the $filter, we will generate an index in our underlying database so subsequent GETs will return more quickly. So, if an application has a filter clause like this:
```
ssriniva-msft, line 177: Is this something substrate does? I am not aware of any index building in our service

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+.../assignments/?$filter=status ne 'draft'
+
+You can generate an index for that filter by adding the ordering clause:
+
+.../assignments/?$filter=status ne 'draft'&$orderby=displayName
+
+The performance gain is amortized; the first time an application makes the $filter/$orderby query, the index needs to be generated, so the request will take longer than if just the $filter query parameter is used. However, all subsequent calls will be faster than if just the filter was passed.
+
+We allow filtering and ordering on select properties.
+
```
ssriniva-msft, line 188: It seems like we are missing many additional best practice topics. Shouldn't we talk about additional best practices related to perf/latency such as:

- use delta queries (instead of making full queries all the time)
- use client side paging (esp. if you don't need all the data at once), and don't misuse client side paging (by asking for a large number of entities at once)
- picking between ~/users/id/assignments and ~/classes/id/assignments depending on scenario (in the past, partners made wrong choices like calling ~/classes first and then calling ~/classes/id/assignments on each class i.e. fanout instead of just 1 call to ~/users/id/assignments)
- $expand best practices if any?

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+## Throttling
+Your application calls the assignments service. In some cases, you might need to make calls in the context of e.g., say a lot of students and/or resources. This can cause a spike in service traffic. In order to safeguard itself against such spikes and to make sure service remains available to all callers, the service will throttle this specific caller.
```
ssriniva-msft, line 190: Add link to our throttling limits here -- https://docs.microsoft.com/en-us/graph/throttling-limits#assignment-service-limits

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+
+In order to make sure you build smooth user experience; the application need to ensure it is not generating lot of calls in the short span of time. You can implement a concurrency control using one of the following mechanisms-
```
ssriniva-msft, line 192: Again seems like we could point to some generic guidance elsewhere (this is not Assignments specific) unless repeating here is the best option

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+  Use a SemaphoreSlim with an initialCount equal to the maximum number of threads. Then you use WaitAsync to wait until it's OK to queue up another.
+- [Parallel.ForEach](/dotnet/api/system.threading.tasks.parallel.foreach)
+   You can simply specify the `MaxDegreeOfParallelism` and then provide an Action to perform on each item in your `IEnumerable`
+- [Polly Bulkhead Policy](https://github.com/App-vNext/Polly/wiki/Bulkhead)
+  A bulkhead policy restricts the number of concurrent calls that can be made, and optionally allows you to queue up calls that exceed that number.
+
+These patterns allow you to use a constrained number of threads to work through a batch of jobs.
+
+### How to avoid excessive throttling?
+Often times excessive throttling is self-inflicted injury i.e. you are calling the service more often than service would want you to call it. The assignment service will send Retry-After header in the response with recommended delay in making subsequent calls. The best practice here is to listen to this header value and no make request earlier that the recommended value.
```
ssriniva-msft, line 205: Maybe the content folks can look at this, but I don't think we could use language like this :)

ssriniva-msft, line 205: Typo "no" should be "not"? Better to phrase it as "Please consume the value in this header and wait for the recommended interval before making the next API call from your application"

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+title: "Education Graph API Best Practices"
+description: "Provides best practices for developers on how best to use EDU assignments graph apis."
+ms.localizationpriority: medium
+author: "sharad-sharma-msft"
+ms.prod: "education"
+doc_type: conceptualPageType
+---
+
+# Education Graph API Best Practices
+This article walks through the best practices for using the Assignments .NET SDK. Following these practices, will help improve your latency, availability, and boost overall performance.
```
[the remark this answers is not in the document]
ssriniva-msft: Would it help to add at the top a link to some of the concept documents we have previously written such as the description of assignment and submission states? I consider understanding these concepts central to understanding the best practices documented here

concepts/education-graph-api-best-practices.md
```diff
@@ -0,0 +1,280 @@
...
+
+These patterns allow you to use a constrained number of threads to work through a batch of jobs.
+
+### How to avoid excessive throttling?
+Often times excessive throttling is self-inflicted injury i.e. you are calling the service more often than service would want you to call it. The assignment service will send Retry-After header in the response with recommended delay in making subsequent calls. The best practice here is to listen to this header value and no make request earlier that the recommended value.
+
+## CancellationToken
+Whether you’re doing async work or not, accepting a CancellationToken as a parameter to your method is a great pattern for allowing your caller to express lost interest in the result.
+
+Supporting cancelable operations comes with a little bit of extra responsibility on your part.
```
ssriniva-msft, line 210: I am confused why cancelable operations are part of this document. This isn't in anyway related to Assignments Graph API right? Or is this specific to Graph .NET SDK? Is this document for callers using the .NET SDK only? What about callers that want to skip and not use the SDK for whatever reason?

ssriniva-msft (changes requested):

concepts/education-graph-api-best-practices.md
ssriniva-msft: I would consider best practices to be independent of whether SDK is used or not. There might be some best practices that are specific to SDK which could be left to the SDK team?

concepts/education-graph-api-best-practices.md
cristobal-buenrostro: done

cristobal-buenrostro: revised based on your suggestion

concepts/education-graph-api-best-practices.md
cristobal-buenrostro: added

Danielabom:
This PR was labeled as "ready for content review"; however, I reached out to @cristobal-buenrostro and it seems that it is not ready yet.

PR #17815, opened 2022-07-22
Fix Tenant Info TOC hierarchy

jkdouglas:


jkdouglas:
The tenantInfo API was under the wrong TOC hierarchy. This PR just changes the TOC to make sure its a top level entry under 'identity and access -> directory management'.

PR #17837, opened 2022-07-25
Supporting Application Context for List Chats API in v1.0

evanypark:
Application Context has been enabled for v1.0 List Chats API.


RamjotSingh (approved):

api-reference/v1.0/api/chat-list.md
```diff
@@ -21,7 +21,7 @@ One of the following permissions is required to call this API. To learn more, in
 |:--------------------|:---------------------------------------------------------|
 |Delegated (work or school account) | Chat.ReadBasic, Chat.Read, Chat.ReadWrite |
 |Delegated (personal Microsoft account) | Not supported.    |
-|Application | Not supported. |
+|Application | Chat.ReadBasic.All*, Chat.Read.All*, Chat.ReadWrite.All* |
```
angelgolfer-ms, line 25: I have a compound comment that applies to lines 14, 24, and 30-32:
(a) For line 14, the description says "... that the user is part of". That implies somewhere the REST query specifies a user. However, line 32 doesn't specify a user.
Please confirm if line 14 is accurate with respect to line 32. If line 14 is accurate, I think it's still worthwhile to call out that line 32 returns all the chats that (which?) user is part of - signed in user, specified user?

Can you apply any clarification to both the beta topic and v1 topic?

(b) Line 24:
What is that asterisk for? In the beta topic, topic, there's a line that follows the permissions table that says:
>>
* This scenario is only supported for the following call: GET /users/{user-id | user-principal-name}/chats

angelgolfer-ms (changes requested):
@evanypark , please see my comment/questions.

api-reference/v1.0/api/chat-list.md
evanypark: (a) When accessing this API using application context, it will return all chats in tenant that the AAD application can access. With Application context being enabled, I think it should be like "retrieve the list of chats that the user is part of or application has access to"

(b) For asterisk, we should add below, similar to what we are showing for /beta API, where application context has been enabled earlier.
* This scenario is only supported for the following call: GET /users/{user-id | user-principal-name}/chats

angelgolfer-ms: @evanypark , sorry, I still don't get it. If the note for application permissions is accurate and says:
>>
* This scenario is only supported for the following call: GET /users/{user-id | user-principal-name}/chats

Then that means application permissions _don't_ apply to this syntax line (line 32):
GET /chats

So that means only delegated permissions are supported for GET /chats.
But there's no user indicated there.
So what would GET /chats return?

evanypark: @angelgolfer-ms Sorry about the confusion. you're correct.
Application context is only supported when user id is specified.
I think it should have description like, "List chat operation in application-only context is only allowed when listing chats for a specific user.", which is the error message we throw when user id is not specified with application context

evanypark: I think with the added notes under asterisk, it should make sense to the users?

angelgolfer-ms: @evanypark , let's try again. My question is what does GET /chats do.

There are 3 syntax lines for the list chats operation:
(1) GET /me/chats
Get the chats for the signed-in user in a tenant. Supports delegated or application permissions.

(2) GET /users/{user-id | user-principal-name}/chats
Get the chats for the specified user in a tenant. Supports delegated or application permissions.

(3) GET /chats
What does this query do?
Does this query support only delegated permissions (because there's this note:
This scenario is only supported for the following call: GET /users/{user-id | user-principal-name}/chats)?
But which would be the user for delegated permissions in this case? There's no user specified in the query URL.

angelgolfer-ms: Per Teams chat with @evanypark , GET /chats returns the signed-in user's chats in the organization. It supports only delegated permissions.

evanypark:
> @evanypark , please see my comment/questions.

angelgolfer-ms (approved):
Updates are good for publishing, thanks.

PR #17880, opened 2022-07-27
entitlement management: add GET for accesspackageassignment

markwahl-msft:
In entitlement management, when we launched the iniital beta APIs, we didn't really have a scenario where a client would need to do a GET of an accessPackageAssignment, so it wasn't in the docs.  With the new features we're adding to v1.0, it will be more useful.  Everything else in EM in v1.0 has a GET, but accessPackageAssignment only had documented how to do List. 
 This PR documents the GET operation to retrieve a single accessPackageAssignment to the v1.0 API.


api-reference/v1.0/api/accesspackageassignment-get.md
```diff
@@ -0,0 +1,87 @@
...
+  "blockType": "ignored"
+}
+-->
+``` http
+GET /identityGovernance/entitlementManagement/assignments/{accessPackageAssignmentId}
+```
+
+## Optional query parameters
+
+This method supports the `$expand` OData query parameter to expand the relationships, to retrieve the `accessPackage`, `target` and `assignmentPolicy`.  For general information, see [OData query parameters](/graph/query-parameters).
```
FaithOmbongi, line 38: Suggestion only if $select is supported
```suggestion
This method supports the `$expand` OData query parameter to expand the `accessPackage`, `target` and `assignmentPolicy`. It also supports the `$select` query parameter.  For general information, see [OData query parameters](/graph/query-parameters).
```

FaithOmbongi (approved):
Thank you @markwahl-msft for your work in this PR. Approved with suggestions.

PR #17887, opened 2022-07-28
update: Add a new property to mgt-person

musale:
Add the `usage` property to the person component.


gavinbarron:
@musale given that this is a change targeted for the v3 release could you change the target branch for this PR to mgt/v3 

That will allow us collect the docs changes here as we close out the PRs on the v3 release work for mgt

PR #17895, opened 2022-07-28
EM: suppress generation of example payloads in v1.0 which have derived types 

markwahl-msft:
There is currently an issue with the Graph SDK client, seen with PSh but probably affecting others, in generating payloads in which there is a derived type. These are easy to spot as there is usually a mention of @odata.type within the payload.  See email thread of Feb 15 2022 with Darrel Miller "RE: named location, Graph case sensitive and generated PSh fragments".  EM uses derived types in policy and connected organization, and the second example in the v1.0 create of an assignment policy illustrates this in the fallback approver selection. Unfortunately this example does not work as shown,  New-MgEntitlementManagementAssignmentPolicy in 1.10.0 with
```
 FallbackPrimaryApprovers = @(
                     @{
                         "@odata.type" = "#microsoft.graph.singleUser"
                         UserId = "3b5d4c83-0517-4f44-9d6d-ff9624f28aa3"
```
does not convert from PSh case to OData case and sends .  

```
"fallbackPrimaryApprovers": [
          {
            "UserId": "3b5d4c83-0517-4f44-9d6d-ff9624f28aa3",
```
which results in invalid model error.

It would be possible to make a change to the generated example PSh snippet itself to make it legal, but then that PSh might get overwritten the next time the article is edited.  So for now suppressing the generation of examples where the v1.0 payload has @odata.type for now until this can be resolved.

### DIFF
--- a/api-reference/v1.0/api/entitlementmanagement-post-assignmentpolicies.md
+++ b/api-reference/v1.0/api/entitlementmanagement-post-assignmentpolicies.md
@@ -164,9 +164,8 @@ The following example shows a more complex policy with two stages of approval an
 #### Request
 
 
-# [HTTP](#tab/http)
 <!-- {
-  "blockType": "request",
+  "blockType": "ignored",
   "name": "create_accesspackageassignmentpolicy_2"
 }
 -->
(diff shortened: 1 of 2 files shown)


markwahl-msft:
(supression of auto generated samples for beta APIs will be a separate PR)

PR #17917, opened 2022-08-01
adding revocation permissions

Vishaal-Mahtani:


FaithOmbongi (commented):
Hi @Vishaal-Mahtani - Ideally, we recommend that the permissions are published at the same time as the APIs that need them - and if possible, in the same PR. Can we hold back this PR until you're ready to publish the APIs? At that time, you can pivot the PR to introduce both docs and the permissions. Thanks.

FaithOmbongi:
Hi @Vishaal-Mahtani - As I mentioned internally, there are APIs related to this permission. You have at least one entity (userRevocationEvent) and a new relationship (in directory resource). You should also update the [subscription resource docs](https://docs.microsoft.com/en-us/graph/api/resources/subscription?view=graph-rest-1.0) and its associated methods with the requests for revocation.

isvargasmsft (approved):

isvargasmsft:
Noticed Faith is taking this one.

PR #16957, opened 2022-05-18
[Teamwork] Add `orderBy` query option when listing chats

jessiechan27:

PR #17924, opened 2022-08-02
Update chat-list.md

JarbasHorst:
Fix for PR #16957
Changed formatting because lastMessagePreview/createdDateTime are properties and descending isn't a property.


maxsatula:
Does it really work? Querying `https://graph.microsoft.com/v1.0/me/chats?$expand=members&$orderBy=lastMessagePreview/createdDateTime+desc`
yields:
```
"code": "BadRequest",
"message": "Could not find a property named 'lastMessagePreview' on type 'microsoft.graph.chat'.",
```

Issue #18012, opened 2022-08-11
completedDate is returned instead of completedDateTime - completedDate cannot be used in OData filter


tjsawyer:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 58329c0d-50b2-2b69-b968-e36e4d7c6ad9
* Version Independent ID: 0369ad98-97b9-c193-eff8-3bc1aec1c267
* Content: [accessPackageAssignmentRequest resource type - Microsoft Graph v1.0](https://docs.microsoft.com/en-us/graph/api/resources/accesspackageassignmentrequest?view=graph-rest-1.0)
* Content Source: [api-reference/v1.0/resources/accesspackageassignmentrequest.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/main/api-reference/v1.0/resources/accesspackageassignmentrequest.md)
* Product: **governance**
* Technology: **microsoft-graph**
* GitHub Login: @markwahl-msft
* Microsoft Alias: **MSGraphDocsVteam**

tjsawyer:
Documentation is incorrect.  completedDateTime is not returned - it is returned as completedDate.  In addition, completedDate cannot be used as an OData filter.   This is highly inefficient.

......
            "requestType": "UserAdd",
            "requestState": "Delivered",
            "requestStatus": "FulfilledNotificationTriggered",
            "createdDateTime": "2022-08-08T15:07:25.147Z",
            "completedDate": "2022-08-09T20:05:56.523Z",
........


When trying to use as a filter ...
  "error": {
        "code": "InvalidFilter",
        "message": "OData query is invalid: Filter 'completedDate GreaterThan 8/1/2022 11:34:13 AM +00:00 is not supported. .", ........................

markwahl-msft:
You are correct, completedDateTime cannot currently be used as a filter when listing assignmentRequests. It is possible to use createdDateTime, however.

FaithOmbongi:
Thank you @tjsawyer for reporting the error. The doc has been updated. Closing this issue.

tjsawyer:
> You are correct, completedDateTime cannot currently be used as a filter when listing assignmentRequests. It is possible to use createdDateTime, however.

[...]

PR #18013, opened 2022-08-11
correct completedDateTime in filterbycurrentuser

markwahl-msft:
In Graph v1.0, filterByCurrentUser() returns completedDateTime rather than completedDate, updating the example response.
Add an example of using a date filter in the list article. Related to 18012.

https://github.com/microsoftgraph/microsoft-graph-docs/issues/18012

### DIFF
--- a/api-reference/v1.0/api/entitlementmanagement-list-assignmentrequests.md
+++ b/api-reference/v1.0/api/entitlementmanagement-list-assignmentrequests.md
@@ -34,12 +34,13 @@ GET /identityGovernance/entitlementManagement/assignmentRequests
 
 ## Query parameters
 
-This method supports the `$select`, `$expand` and `$filter` OData query parameters to help customize the response.
+This method supports the `$select`, `$expand` and `$filter` OData query parameters to help customize the response.  Not all attributes of an [accessPackageAssignmentRequest](../resources/accesspackageassignmentrequest.md) are supported for filtering.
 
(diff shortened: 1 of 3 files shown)


api-reference/v1.0/api/entitlementmanagement-list-assignmentrequests.md
```diff
@@ -34,12 +34,13 @@ GET /identityGovernance/entitlementManagement/assignmentRequests
 
 ## Query parameters
 
-This method supports the `$select`, `$expand` and `$filter` OData query parameters to help customize the response.
+This method supports the `$select`, `$expand` and `$filter` OData query parameters to help customize the response.  Not all attributes of an [accessPackageAssignmentRequest](../resources/accesspackageassignmentrequest.md) are supported for filtering.
```
FaithOmbongi, line 37: Hi @markwahl-msft - I recommend you update the property descriptions in the [accessPackageAssignmentRequest resource doc](https://docs.microsoft.com/en-us/graph/api/resources/accesspackageassignmentrequest?view=graph-rest-1.0) to indicate those that support `$filter` and which operators can be used. You can also indicate in the relationships those that support `$expand`, including nested `$select`. Please see the [user properties](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0) for a sample.

This comment applies across all EM docs.

FaithOmbongi (approved):
Approved with suggestions for all of EM docs.

api-reference/v1.0/api/entitlementmanagement-list-assignmentrequests.md
markwahl-msft: Added to two properties in the resource article that I recall supports $filter.  In a subsequent PR we'll need to make a pass through _all_ the resource articles which have list methods.

PR #18039, opened 2022-08-15
Adding description about `$filter` support

hishikitetsuya:
We confirmed that O365 SCC alert does not support $filter by alert status from past issue [here](https://github.com/microsoftgraph/security-api-solutions/issues/56). It looks like they don't support $filter as of now but there are no descriptions about this restriction. Thus I propose a description of `$filter` support for this API.

Any comments and suggestions are welcomed.

### DIFF
--- a/api-reference/v1.0/api/alert-list.md
+++ b/api-reference/v1.0/api/alert-list.md
@@ -58,9 +58,9 @@ The following table lists the `$filter` keywords by each vendor name. Even thoug
 | Microsoft Defender for Endpoint | Microsoft Defender ATP |
 | Office 365 | Not currently supported. |
 
-To return an alternative property set, use the OData `$select` query parameter to specify the set of **alert** properties that you want.  For example, to return the **assignedTo**, **category**, and **severity** properties, add the following to your query: `$select=assignedTo,category,severity`.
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/api/alert-list.md
```diff
@@ -60,6 +60,8 @@ The following table lists the `$filter` keywords by each vendor name. Even thoug
 
 To return an alternative property set, use the OData `$select` query parameter to specify the set of **alert** properties that you want.  For example, to return the **assignedTo**, **category**, and **severity** properties, add the following to your query: `$select=assignedTo,category,severity`.
 
+> **Note:** Some providers may not support `$filter` keywords.
```
Lauragra, line 63: In the table that precedes this note, there's an entry for Office 365 that indicates that it's not currently supported. Is that sufficient, or do you still think that a more general note is warranted?

hishikitetsuya: Thanks. Yes I believe a more general note is valuable for customers since the table you mentioned just describes how to specify each vendor name in `$filter` keyword (e.g., `$filter=vendorInformation/provider eq 'IPC'`). The table never says O365 SCC does not support every `$filter`. Thus I believe this clarification is very helpful.

PR #18107, opened 2022-08-19
Support Bookmark And Acronym Searching

ZChristine:


concepts/search-concept-acronym.md
```diff
@@ -0,0 +1,84 @@
...
+    }
+   ]
+  }
+ ]
+}
+```
+
+## Known issues
+
+Sorting, aggregation, and pagination are not supported for acronym searches.
```
Lauragra, line 80: Just to verify, this is the "default"/expected behavior, and not something you plan to fix or update in the future?

concepts/toc.yml
```diff
@@ -447,6 +447,10 @@ items:
           - name: Overview
             href: search-concept-overview.md
             displayName: search
+          - name: Search bookmarks
```
Lauragra, line 450: Is this the order that you want the topics to be surfaced in the TOC? Consider applying some logic to the ordering, such as most common to least common search types, or alphabetical, instead of adding the new topics at the top.

concepts/search-concept-acronym.md
ZChristine: yes. this is by design right now. we will not update it in the near future.

ZChristine: > Just to verify, this is the "default"/expected behavior, and not something you plan to fix or update in the future?

yes. this is by design right now. we will not update it in the near future.

concepts/toc.yml
ZChristine: > Is this the order that you want the topics to be surfaced in the TOC? Consider applying some logic to the ordering, such as most common to least common search types, or alphabetical, instead of adding the new topics at the top.

order it by most common to least common search types

Lauragra (approved):

PR #18149, opened 2022-08-23
Update insights-list-shared.md

berkantelovski:
Add external user interaction clarification.
Change has been internally approved by API owning team.

### DIFF
--- a/api-reference/v1.0/api/insights-list-shared.md
+++ b/api-reference/v1.0/api/insights-list-shared.md
@@ -13,9 +13,7 @@ Namespace: microsoft.graph
 
 Calculated insight that includes the list of documents shared with a user.
 
-For documents hosted on OneDrive/Sharepoint, this insight will only include documents shared within the user's own Microsoft 365 tenant.
-
-For documents that are attached as a file, this insight will include all such documents shared with a user.
+This insight includes documents hosted on OneDrive/SharePoint in the user's Microsoft 365 tenant that are shared with the user, and documents that are attached as files and sent to the user.
 
 ## Permissions
 One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/api/insights-list-shared.md
```diff
@@ -13,6 +13,10 @@ Namespace: microsoft.graph
 
 Calculated insight that includes the list of documents shared with a user.
 
+For documents hosted on OneDrive/Sharepoint, this insight will only include documents shared within the user's own M365 tenant.
+
+For documents which are attached directly as a file, this insight will include all such documents shared with a user.
+
```
Lauragra, line 19: This text is unclear. Does it mean that the insight returns a list of both categories of documents - shared and attached as files? Would this be an accurate revision of the text:

This insight includes documents hosted on OneDrive/SharePoint in the user's Microsoft 365 tenant that are shared with the user, and documents that are attached as files and sent to the user.

berkantelovski: Hi @Lauragra , thanks for the feedback!

I feel like the revision of the text clarifies the point I wanted to make.
For additional context: the insight returns both documents hosted on OD/SP and documents attached as files, but with a caveat: in the scenario of sharing with an external user (a.k.a. a user outside of the current user's M365 tenant), OD/SP hosted documents will be returned by the insight, only ones attached as files. This is what I wanted to distinguish with the change.

Lauragra: @berkantelovski , okay, would this revised text work?

This insight includes documents hosted on OneDrive/SharePoint in the user's Microsoft 365 tenant that are shared with the user, and documents from external users that are attached as files and sent to the user.

berkantelovski: Reading through both revisions, I think the first revision is clearer, so I would prefer that to be the final version.
P.S. I apologize for just noticing that I had made an error in my previous elaboration - I meant to say OD/SP hosted documents will *not* be returned by the insight in case of sharing with an external user, while documents attached as files will be returned regardless of user status.

PR #18156, opened 2022-08-23
Entitlement management: access package assignment request requesttype values

markwahl-msft:
In EM, update beta and 1.0 resource type articles to add more details on the accessPackageRequestType that can be expected to be seen or that can be set on a request.


FaithOmbongi:
Hi @markwahl-msft:
1. For consistency across both versions of docs, can we apply the same camelCase in beta as well (instead of PascalCase)? Or are the values of this property case-sensitive?
2. Please also update the casing in the "Request body" sections of the Create accessPackageAssignmentRequest docs as well.

markwahl-msft:
@FaithOmbongi  added the POST article as well.  Yes, case is significant across these values at this time, the syntax changed between beta and v1.0 and so the value cases are different in each API.

FaithOmbongi (approved):

FaithOmbongi:
@markwahl-msft - Hopefully, the team can consider removing case sensitivity altogether. It's a bit of a pain point for customers.

PR #18261, opened 2022-08-31
EM: more details on originid of resource roles 

markwahl-msft:
This illustrates where the originId field comes from, particularly for SPO sites as resources, where the originId is an opaque integer provided from SPO.

This is additional guidance based on questions we've received; there is no API change.


isvargasmsft (approved):
@FaithOmbongi I took the liberty of reviewing this one :)

PR #18350, opened 2022-09-07
[BrokenLinksH2] Fix path in link

DhurataJ:
Fixing broken image

### DIFF
--- a/api-reference/beta/api/chart-image.md
+++ b/api-reference/beta/api/chart-image.md
@@ -76,7 +76,7 @@ You can display the base-64 string inside an HTML image tag: `<img src="data:ima
 
 For default behavior, use `Image(width=0,height=0,fittingMode='fit')`. Here is an example of a chart image returned with the default parameters.
 
-![Screenshot of an Excel chart image displayed with the default height and width.](https://cdn.graph.office.net/prod/GraphDocuments/concepts/images/GetChart-default.png)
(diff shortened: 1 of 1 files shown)


DhurataJ:
@Lauragra - The C+E Skilling team is fixing broken links on docs.microsoft.com. This will eliminate potential accessibility, security, and usability issues. This PR includes only link fixes/removals and does not change other content. This PR is ready for review and merge. Thank you!

api-reference/beta/api/chart-image.md
```diff
@@ -76,7 +76,7 @@ You can display the base-64 string inside an HTML image tag: `<img src="data:ima
 
 For default behavior, use `Image(width=0,height=0,fittingMode='fit')`. Here is an example of a chart image returned with the default parameters.
 
-![Screenshot of an Excel chart image displayed with the default height and width.](https://cdn.graph.office.net/prod/GraphDocuments/concepts/images/GetChart-default.png)
+![Screenshot of an Excel chart image displayed with the default height and width.](https://github.com/microsoftgraph/microsoft-graph-docs/blob/live/concepts/images/GetChart-default.png)
```
Lauragra, line 79: This image link doesn't currently work. We might need to add an images folder in the api-reference folder and then use a relative link to link to the image. The concepts and api-reference folders are two different content sets. Are you able to take care of doing that? Thanks.

DhurataJ: I thought so Laura, I was trying to test on view if it's going to work, and results are as you mention the image would not work. As a part of the C+E team and workflow process, we are fixing only broken links and replacing with right target. If there is someone more appropriate from your team to handle this case would be great. Thanks!

PR #18464, opened 2022-09-13
Delete Role Assignable Groups

psaffaie:
adding a note for role assignable group deletion. Role assignable groups can apparently only be deleted by Users who created the group or are a global admin or privileged role admin. This is documented internally in a troubleshooting wiki, can be validated using DS explorer, and is mentioned under the Group page for MS Graph but isn't mentioned on the MS Graph Delete Group documentation.

The reasoning for this is if I as a customer attempted to make a request to delete a group and a 403 was returned, a reasonable first instinct would be to check the documentation for the deletion action first, rather than the overview page defining everything about a group. Since this is a side effect caused by the property being set, it wouldn't be immediately clear that this property in the cause.


api-reference/v1.0/api/group-delete.md
```diff
@@ -15,6 +15,8 @@ Delete group.
 
 When deleted, Microsoft 365 groups are moved to a temporary container and can be restored within 30 days. After that time, they're permanently deleted. This isn't applicable to Security groups and Distribution groups which are permanently deleted immediately. To learn more, see [deletedItems](../resources/directory.md).
 
+Note: Role assignable groups can only be deleted by users who created the group, are a global administrator, or a privileged role administartor. The caller must be assigned the RoleManagement.ReadWrite.Directory permission to delete a role assignable group.
```
FaithOmbongi, line 18: - Suggestion to move this note after the permissions table because it's privilege-related.
- Q: For delegated, it is creator or owner? Please confirm.
- Also recommending that we call out privileges for app-only

```suggestion
The following conditions apply for apps to delete role-assignable groups:
+ For delegated scenarios, the app must be assigned the _RoleManagement.ReadWrite.Directory_ delegated permission, and the calling user must be the creator of the group or a global administrator or a privileged role administrator.
+ For app-only scenarios, the calling app must be the owner of the group or be assigned the _RoleManagement.ReadWrite.Directory_ application permission or be assigned the Global Administrator or Privileged Role Administrator Azure AD role.
```

FaithOmbongi (changes requested):
Hi @psaffaie - I've left a change suggestion inline. Please move the text after the permissions table.

psaffaie (commented):
@FaithOmbongi  let me know what else I need to do to implement the changes you recommended.

api-reference/v1.0/api/group-delete.md
```diff
@@ -15,6 +15,10 @@ Delete group.
 
 When deleted, Microsoft 365 groups are moved to a temporary container and can be restored within 30 days. After that time, they're permanently deleted. This isn't applicable to Security groups and Distribution groups which are permanently deleted immediately. To learn more, see [deletedItems](../resources/directory.md).
 
+The following conditions apply for apps to delete role-assignable groups:
++ For delegated scenarios, the app must be assigned the _RoleManagement.ReadWrite.Directory_ delegated permission, and the calling user must be the creator of the group or a global administrator or a privileged role administrator.
++ For app-only scenarios, the calling app must be the owner of the group or be assigned the _RoleManagement.ReadWrite.Directory_ application permission or be assigned the Global Administrator or Privileged Role Administrator Azure AD role.
+
```
FaithOmbongi, line 21: @psaffaie - Pending feedback: Please move this after the permissions table below where the context is privileges and permissions.

FaithOmbongi (approved):
Hi @psaffaie - Just one pending feedback. Please move the block of text after the permissions table where the context fits.

PR #18523, opened 2022-09-16
Update delegatedadminrelationships-api-overview.md

Saneesha-L:
Changed the position of the GDAP workflow above the GDAP use cases and added note section for creating invitation link.


FaithOmbongi (approved):
Hi @Saneesha-L - For completeness, I also recommend that you add the relevant note to the [example response object in the Create request API](https://review.learn.microsoft.com/en-us/graph/api/delegatedadminrelationship-post-requests?view=graph-rest-beta&branch=pr-en-us-18523&tabs=http#response-1). Maybe as an intro statement before the truncation note.

PR #18682, opened 2022-09-29
[Cloud PC]Remove description about deprecate patch addomainpassword API As deprecate period is over now.

Emma0727:
Roadmap ADO: 15290


RuiHou105 (approved):

isvargasmsft (approved):

Lauragra:
> @Lauragra Hi Laura, hope to hear your suggestion here do we need to add change-log for this?

No, I think you're good here.

PR #18730, opened 2022-10-04
Updated allowDeletedIdentitiesDataRemoval

suawat:
Description for allowDeletedIdentitiesDataRemoval is updated to Not Implemented yet to minimize customer friction.


api-reference/beta/resources/externalidentitiespolicy.md
```diff
@@ -26,7 +26,7 @@ Inherits from [policyBase](../resources/policybase.md).
 ## Properties
 |Property|Type|Description|
 |:---|:---|:---|
-|allowDeletedIdentitiesDataRemoval|Boolean|Notifies Azure AD whether to clean up the user information about the external identity, from the guest tenant, when the user is deleted in their home tenant. |
+|allowDeletedIdentitiesDataRemoval|Boolean|<!--Notifies Azure AD whether to clean up the user information about the external identity, from the guest tenant, when the user is deleted in their home tenant.--> **Not Implemented Yet.**|
```
mmcadamsms, line 29: Do we want to say its not implemented yet?  I mean its in the API, just the propagation of the delete is not implemented.

api-reference/beta/api/externalidentitiespolicy-get.md
```diff
@@ -108,7 +108,6 @@ Content-Type: application/json
     "id": "externalIdentityPolicy",
     "deletedDateTime": null,
     "allowExternalIdentitiesToLeave": true,
-    "allowDeletedIdentitiesDataRemoval": false,
```
mmcadamsms, line 111: So this is still in the API on Get

api-reference/beta/resources/externalidentitiespolicy.md
suawat: If there is a better way to convey that message, I am happy to update this.

alvarorahul: +1 there aren't plans yet to implement this.

Can we instead say "Reserved for future use.".

api-reference/beta/api/externalidentitiespolicy-update.md
```diff
@@ -45,7 +45,6 @@ PATCH /policies/externalIdentitiesPolicy
 
 |Property|Type|Description|
 |:---|:---|:---|
-|allowDeletedIdentitiesDataRemoval|Boolean|Notifies Azure AD whether to clean up the user information about the external identity, from the guest tenant, when the user is deleted in their home tenant. Required.|
```
alvarorahul, line 48: we shouldn't remove it entirely... instead have the same message as the other file... since customers will see this in the API response

api-reference/beta/api/externalidentitiespolicy-get.md
alvarorahul: let's leave it in the API response since we're still returning it.. we'll need to follow deprecation process in order to deprecate it

alvarorahul (changes requested):

api-reference/beta/resources/externalidentitiespolicy.md
suawat: yes agreed. Updated the description to say "Reserved for future use". I like that better.

api-reference/beta/api/externalidentitiespolicy-get.md
suawat: sounds good. No change here then

api-reference/beta/api/externalidentitiespolicy-update.md
mmcadamsms: |allowDeletedIdentitiesDataRemoval|Boolean|**Reserved for future use.**|

api-reference/beta/resources/externalidentitiespolicy.md
```diff
@@ -26,7 +26,7 @@ Inherits from [policyBase](../resources/policybase.md).
 ## Properties
 |Property|Type|Description|
 |:---|:---|:---|
-|allowDeletedIdentitiesDataRemoval|Boolean|Notifies Azure AD whether to clean up the user information about the external identity, from the guest tenant, when the user is deleted in their home tenant. |
+|allowDeletedIdentitiesDataRemoval|Boolean|<!--Notifies Azure AD whether to clean up the user information about the external identity, from the guest tenant, when the user is deleted in their home tenant.--> **Reserved for future use.**|
```
mmcadamsms, line 29: |allowDeletedIdentitiesDataRemoval|Boolean|**Reserved for future use.**|

mmcadamsms (approved):
Fine by me

PR #18802, opened 2022-10-10
Documentation for supported clouds

abshar-teams:


api-reference/beta/api/userteamwork-list-installedapps.md
```diff
@@ -18,6 +18,17 @@ Retrieve the list of [apps](../resources/teamsappinstallation.md) installed in t
 > [!NOTE]
 > The `id` of a **teamsAppInstallation** resource is not the same value as the `id` of the associated **teamsApp** resource.
 
+### Supported environments
+
+The following table shows the permission types that are supported for this API in different clouds.
```
Lauragra, line 23: For consistency can we refer to these as "national clouds"?

api-reference/beta/api/userteamwork-list-installedapps.md
```diff
@@ -18,6 +18,17 @@ Retrieve the list of [apps](../resources/teamsappinstallation.md) installed in t
...
+
+|Environment      | Supported permission types              |
```
Lauragra, line 25: Should this be "national cloud environment"? Also, the list of clouds should match the list in our national clouds topic; do we have a definitive list?

api-reference/beta/api/userteamwork-list-installedapps.md
```diff
@@ -18,6 +18,17 @@ Retrieve the list of [apps](../resources/teamsappinstallation.md) installed in t
 > [!NOTE]
 > The `id` of a **teamsAppInstallation** resource is not the same value as the `id` of the associated **teamsApp** resource.
 
+### Supported environments
```
Lauragra, line 21: Would we want to call this "Supported national cloud environments"? Also given the focus on permission types, would it make sense to move this H3 under the Permissions H2?

api-reference/beta/api/userteamwork-list-installedapps.md
Lauragra, line 25: How do we imagine that users would want to access this information - at the API level, or at the workload level, or both? For that latter, would it be useful to consider a National clouds table in the API overview topic, where we list the support for the APIs in a single, scannable location?

abshar-teams: Sure, we can. but why the word 'national'. We are going to add all the clouds supported globally.

Lauragra: Just for consistency with our existing content, unless there is a reason to change it. They were referred to as "national clouds" back when we published the topic.

api-reference/beta/api/userteamwork-list-installedapps.md
abshar-teams: yes, i will change it to national cloud environments.
for the section's placement, i think environment should take precedence over permissions types.
I expect clients to look at environments before looking at the permissions types supported for a given api.
Let me know your thoughts.

api-reference/beta/api/userteamwork-list-installedapps.md
abshar-teams: It's actually at both levels.

Like our workload teams graph is currently enabled in all national clouds except Gallatin, which we are working on.

If we specify this in api overview, we will still need a separate <cloud-permissions type> table and most of our api's are available in ALL the clouds, it's just a few api's which doesn't work in some clouds. To begin with, having just a section looks ok to me.
and it's not going to be as simple as putting down a table. For lot of api's, we will have a subset of functionalities disabled in a cloud. Example: (https://learn.microsoft.com/en-us/graph/api/chat-list-messages?view=graph-rest-1.0&tabs=http) works with Gov clouds, except the Example 3 (use of 'orderby' parameter), which is disabled in gov clouds. For these type of exceptions, we may have to put a note under the table.
Let me know your thoughts

api-reference/beta/api/userteamwork-list-installedapps.md
```diff
@@ -18,6 +18,18 @@ Retrieve the list of [apps](../resources/teamsappinstallation.md) installed in t
 > [!NOTE]
 > The `id` of a **teamsAppInstallation** resource is not the same value as the `id` of the associated **teamsApp** resource.
 
+### Supported national cloud environments
+
+
+The following table shows the permission types supported for this API in different national clouds.
```
Lauragra, line 24: Revise this text slightly: "The following table lists the permission types that are supported for this API in different national clouds."

api-reference/beta/api/userteamwork-list-installedapps.md
Lauragra: Depending on how common the exceptions are, we could have a note after the table, or a separate column in the table for notes/exceptions. 

If most APIs are supported in all the national clouds, do we need section and table in each API topic, or just in the overview topic, with notes calling out exceptions in the API topics? If the latter, we would still want a consistent place for this information to simplify discovery. If we have tables in each topic and most of them support all permission types, users will get used to glossing over the information and might miss the places where something isn't supported. 

It will be good to understand the scope of the exceptions.

api-reference/beta/api/userteamwork-list-installedapps.md
Lauragra: Permissions is the first section of the API topics because that is something that we wanted to be immediately discoverable to users; in fact, the section used to be named "Prerequisites". I still recommend that we have the national cloud information after the Permissions table. Especially if most APIs will be supported in all the national clouds, and users will generally only be looking for any exceptions.

api-reference/beta/api/userteamwork-list-installedapps.md
abshar-teams: Here is a list of all such exceptions for the teams graph workload, i have found so far. 

1)Where to put this information ? do we need table/section ?
- I think we don't need section or table, since there are very few exceptions compared to the number of api's we have. An exception note should be good.


Api's to manage the apps installed in the personal scope of a user doesn't work in app context in gcch cloud. 6 docs needs update (List, Add, Get(s), Upgrade, Remove)

Api to list messages in a chat (https://learn.microsoft.com/en-us/graph/api/chat-list-messages?view=graph-rest-beta&tabs=http) doesn't work with 'Orderby' parameter for gcc. 

Api to list user's chats (https://learn.microsoft.com/en-us/graph/api/chat-list?view=graph-rest-beta&tabs=http) doesn't work with 'Orderby' parameter in gcc/gcch.

Api to provision/unprovision email for a channel (https://learn.microsoft.com/en-us/graph/api/channel-provisionemail?view=graph-rest-beta&tabs=http) doesn't work for gcc/gcch

Meeting transcript api's (https://learn.microsoft.com/en-us/graph/api/resources/calltranscript?view=graph-rest-beta) are not supported for gcc/gcch clouds. 

Api to soft delete a message (https://learn.microsoft.com/en-us/graph/api/chatmessage-softdelete?view=graph-rest-beta&tabs=http)  doesn't work in gcch

Api's to manage tabs in the channels (https://learn.microsoft.com/en-us/graph/api/channel-list-tabs?view=graph-rest-beta&tabs=http) doesn't work for shared channels. 5 doc needs update.

Lauragra: Okay, so what do we think of the idea to have a table in the resource topic only, and then in the API topic, we only call out exceptions to national cloud support? Would we need to mention APIs that have exceptions in the table in the resource topic?

I can take a stab at mocking up what this might look like.

api-reference/beta/resources/chat.md
```diff
@@ -62,6 +62,17 @@ A chat is a collection of [chatMessages](chatmessage.md) between one or more par
 
 >**Note:** When using application permissions, be sure you know how to get the chat ID. Because listing chats with application permissions is not supported, not all scenarios are possible. It is possible to get chat IDs with delegated permissions, and from [change notifications for /chats/getAllMessages](../api/subscription-post-subscriptions.md) with application permissions.
 
+### Supported national cloud environments
+
+The following table lists the permission types that are supported for this API in different national clouds.
```
abshar-teams, line 67: it might be a bit confusing for an end user to read about the api exceptions on the resource page.
I think the API page is the best place to add this info

api-reference/beta/api/userteamwork-list-installedapps.md
abshar-teams: As per my limited experience around docs, having a table in the resource topic may not be the best thing. Lets have it only on the api topic page ?
Yes, for now, its only national clouds

api-reference/beta/resources/chat.md
Lauragra: We could have a note that states something like: "Any exceptions to API support in national clouds are specified in the method topic." Are there any cases where only application or only delegated context are supported, or is it generally always both, or none? We might change the table depending on the answer. Also is there an expectation that after we settle on a location/format, this information will be generated from the CSDL? And implemented broadly in the content set?

abshar-teams: We could have a note that states something like: "Any exceptions to API support in national clouds are specified in the method topic." 
- Sounds good to me. So, we will have to add this text to every api page in that case ? Does this require editing of every doc manually ?

Are there any cases where only application or only delegated context are supported, or is it generally always both, or none?
- In general, everything works for both contexts, but still there are exceptions to this, but rare (at least for teams graph api's).

Also is there an expectation that after we settle on a location/format, this information will be generated from the CSDL?
- not sure about this. it depends on what content we add. API schema folks can confirm this.

And implemented broadly in the content set?
- By broadly, if you mean other workloads, then yes, i expect other verticals to have similar requirements sometime in the future

PR #18824, opened 2022-10-11
Update the post content type in example response to HTML, which is the support type in service communications API

payiAzure:
update the post content type to HTML, which is the support type in service communications API

### DIFF
--- a/api-reference/beta/api/serviceannouncement-list-healthoverviews.md
+++ b/api-reference/beta/api/serviceannouncement-list-healthoverviews.md
@@ -211,7 +211,7 @@ Content-Type: application/json
                     "createdDateTime": "2020-11-12T07:07:38.97Z",
                     "postType": "Regular",
                     "description": {
-                        "contentType": "Text",
+                        "contentType": "Html",
                         "content": "Title: Exchange Online service has login issue. We'll provide an update within 30 minutes."
                     }
                   }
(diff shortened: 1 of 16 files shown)


simonwahub (approved):

api-reference/beta/api/serviceannouncement-list-healthoverviews.md
```diff
@@ -211,7 +211,7 @@ Content-Type: application/json
                     "createdDateTime": "2020-11-12T07:07:38.97Z",
                     "postType": "Regular",
                     "description": {
-                        "contentType": "Text",
+                        "contentType": "Html",
```
isvargasmsft, line 214: HI @payiAzure, I noticed the metadata states that this is a String. Is there a reason to specify HTML instead of String?

payiAzure: Hi @isvargasmsft , in metadata, the contentType is graph.bodyType type, and bodyType has two types, html and text.
   <ComplexType Name="itemBody">
        <Property Name="content" Type="Edm.String" />
        <Property Name="contentType" Type="graph.bodyType" />
      </ComplexType>

isvargasmsft: @payiAzure I understand, thank you for the clarification. In this case I believe it is worth it to mention it in the descriptions of the property. 

| Resource | Property |
|---|---|
| ServiceUpdateMessage | body |
| serviceHealthIssuePost | description | 

To the descriptions of the properties in the table above, please add: The supported value for the contentType property is `html`.

isvargasmsft (approved):

PR #18840, opened 2022-10-12
[Calling] add InvitationParticipantInfo options for contact center fe…

Hank-TNguyen:
* If InvitationParticipantInfo.RemoveFromDefaultAudioRoutingGroup is specified as true in the target of participant invitation request, the target user will be put out of default audio mixer once it joins the group call, which means it cannot talk to anyone in the group call.

* If InvitationParticipantInfo.Hidden is specified as true in the target of participant invitation request, the target user will be hidden to all other participants once it joins the group call.


isvargasmsft (approved):
Hello @Hank-TNguyen , @frankpeng7, 

Thank you very much for this updates. 

Please double check if a changelog entry was added for these properties in the past. If not, please add changelog entries following example 2 here: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog?anchor=changelist-record-examples.

If the changelog entries were added then this is approved. 

Thanks!

Hank-TNguyen:
#feedback-addressed

PR #18846, opened 2022-10-12
Changing domain to security

sseth-msft:


Danielabom (changes requested):
Thank you for working on this PR, @sseth-msft  
The changes seem okay. Just one question, what is the action for these permissions in particular? 
Are you no longer granting these specific permissions or deprecating them? Is this change, documentation-wise only?
Just in case, see example 14 in the changelog wiki: [Changelog - Overview (azure.com)](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog)

PR #18877, opened 2022-10-14
Remove RoleManagement.Read.All app-only for PIM

FaithOmbongi:


Fuco1:
The documentation claims I can use this endpoint with `RoleManagement.Read.Directory` but I get the following error

```json
{
    "instanceAnnotations": [],
    "message": "Authorization failed due to missing permission scope RoleAssignmentSchedule.ReadWrite.Directory,RoleManagement.ReadWrite.Directory.",
    "errorCode": "PermissionScopeNotGranted"
}
```

I don't see why I would need ReadWrite to simply list the active requests.

FaithOmbongi:
> The documentation claims I can use this endpoint with `RoleManagement.Read.Directory` but I get the following error
> 
> ```json
> {
>     "instanceAnnotations": [],
>     "message": "Authorization failed due to missing permission scope RoleAssignmentSchedule.ReadWrite.Directory,RoleManagement.ReadWrite.Directory.",
>     "errorCode": "PermissionScopeNotGranted"
> }
> ```
> 
> I don't see why I would need ReadWrite to simply list the active requests.

Hi @Fuco1 - Could you file this issue against the respective API page? It will help us track the problem for proper resolution. Thanks.

Hi @rkarim-ms - While @Fuco1's filing a GitHub issue, can you help to confirm the authZ issue with PIM READ operations in app-only context? Thanks.

Fuco1:
@FaithOmbongi I submitted a new issue with a back-link.  Note that I think this is a bug/misfeature and not only a documentation issue.  A read-only action really should only require read permissions because granting read-write is too much risk.

PR #18908, opened 2022-10-18
Adding SharePoint Premium API Docs

AdwoodMicrosoft:
Adding TOC and relevant docs for monetized APIs


concepts/onedrive-sharepoint-payment-overview
```diff
@@ -0,0 +1,29 @@
+# Payment requirements for the OneDrive and SharePoint APIs
+
+This article describes the licensing and payment requirements for the SharePoint and OneDrive APIs in Microsoft Graph.
+
+The following APIs have consumption charges:
+
+- [AssignSensitivityLabel API](http://https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-beta&tabs=http "AssignSensitivityLabel API") [Private Preview] – Asynchronously assigns a sensitivity label to a file hosted in OneDrive or SharePoint online. 
+
```
angelgolfer-ms, line 8: Line 7 and GLOBAL:
@AdwoodMicrosoft , from the Microsoft Graph API release perspective, APIs that exist in the beta metadata and not marked hidden are in public preview. And, Microsoft Graph docs in general do not include private preview APIs. For these reasons, we cannot represent this API here as in "private preview". That would confuse our internal folks and external customers.

concepts/onedrive-sharepoint-payment-overview
```diff
@@ -0,0 +1,29 @@
+# Payment requirements for the OneDrive and SharePoint APIs
```
angelgolfer-ms, line 1: This topic file needs to be specified with a .md extension. Otherwise it's not staging.

concepts/toc.yml
```diff
@@ -301,6 +301,8 @@ items:
             displayName: files, OneDrive, OneDrive for Business, OneDrive business, OneDrive personal, SharePoint
           - name: Addressing drive items
             href: onedrive-addressing-driveitems.md
```
angelgolfer-ms, line 303: Per our discussion during the walkthrough, since some customers have assumed "Files" <-> OneDrive APIs and "Sites and lists" <-> SharePoint APIs as exposed on Microsoft Graph, i.e., "sites and lists" are not only about sites and lists but also files that exist on SP document libraries. From that perspective, and from the perspective that the topic title and introduction mention both OneDrive and SharePoint, please add an entry point to this concept topic under the "Sites and lists" node as well in the conceptual section.

angelgolfer-ms (changes requested):
@AdwoodMicrosoft , I did a superficial review based on our discussions at the walkthrough. Please take a look at my comments.
Laura will be able to provide her feedback as the main content reviewer.

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
...
+ms.prod: "sites-and-lists"
+---
+
+# Payment requirements for the OneDrive and SharePoint APIs
+
+This article describes the licensing and payment requirements for the SharePoint and OneDrive APIs in Microsoft Graph.
+
+The following APIs have consumption charges:
+
+- [AssignSensitivityLabel API](http://https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-beta&tabs=http "AssignSensitivityLabel API") [Private Preview] – Asynchronously assigns a sensitivity label to a file hosted in OneDrive or SharePoint online.
```
Lauragra, line 15: What do we mean when we say that this API is in "private preview"? In the assignSensitivityLabel topic, we mention that it's a protected API and that users need to request access. However, the API is exposed in the public metadata and is not hidden, so it's not technically "private".

Is there a relationship between the fact that the API is protected/requires users to request access, and the fact that the API has consumption charges? 

I recommend that instead of labeling the API as Private Preview here, you add a note similar to callout in the API topic that mentions that the API is protected, and users must request access.

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
+---
+title: "Payment requirements for the OneDrive and SharePoint APIs"
+description: "Learn about the licensing and payment requirements for the SharePoint and OneDrive APIs in Microsoft Graph.."
+ms.localizationpriority: high
+author: AdwoodMicrosoft
+ms.prod: "sites-and-lists"
+---
+
+# Payment requirements for the OneDrive and SharePoint APIs
```
Lauragra, line 9: This title might convey that the OneDrive and SharePoint APIs require payment, when in fact only some/one API is monetized. Consider changing the title to "Licensing and payment model for OneDrive and SharePoint APIs".

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
...
+
+This article describes the licensing and payment requirements for the SharePoint and OneDrive APIs in Microsoft Graph.
```
Lauragra, line 11: We mention that this article describes licensing and payment requirements for SharePoint and OneDrive APIs. However, the article seems to be specific to the sensitivity label API, without providing the overall context around licensing and billing for use of APIs. Do you expect that there will be more monetized APIs for SharePoint and OneDrive in the future? If so, I recommend that you make this content general to the licensing requirements and payment model that applies/will apply to these APIs.

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
...
+
+The following APIs have consumption charges:
+
+- [AssignSensitivityLabel API](http://https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-beta&tabs=http "AssignSensitivityLabel API") [Private Preview] – Asynchronously assigns a sensitivity label to a file hosted in OneDrive or SharePoint online. 
+
+ Note: You must [associate](https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-how-subscriptions-associated-directory// "associate") or [create](https://docs.microsoft.com/en-us/azure/cost-management-billing/manage/create-subscriptionp:// "create") Azure Subscription for your tenant to call the AssignSensitivityLabel API.  During Private Preview, an Azure billing statement will be created you to see the details of the transactions, however, no charge will be associated with that bill. Please see“Enroll in the Preview" below for instructions on how to apply to be in the Assign Sensitivity Label Private Preview. 
+
+## Licensing model at General Availability
```
Lauragra, line 19: The API is currently only available in beta/preview. So are you saying that the billing model will only apply when the API reaches GA? And what do we mean by "licensing" model here when in the text you reference "billing" model? What is the licensing requirement for the API, and does the licensing requirement apply while the API is in beta or only when it reaches GA?

If the latter, does it make sense to wait to publish information about the licensing model until the API is GA? And in the meantime, mention in the API topic that "licensing and payment requirements will apply to the API when it reaches GA"?

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
...
+ms.localizationpriority: high
+author: AdwoodMicrosoft
+ms.prod: "sites-and-lists"
+---
+
+# Payment requirements for the OneDrive and SharePoint APIs
+
+This article describes the licensing and payment requirements for the SharePoint and OneDrive APIs in Microsoft Graph.
+
+The following APIs have consumption charges:
```
Lauragra, line 13: I recommend covering the payment model in general in this article before getting into the specific list of monetized APIs. For example, what are the charges per call, is there any seeded capacity, where can BDMs go to learn about what the charges will be? Then the article might have a list a bit later of the APIs that entail payment requirements or require some sort of a license.

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
...
+
+- [AssignSensitivityLabel API](http://https://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-beta&tabs=http "AssignSensitivityLabel API") [Private Preview] – Asynchronously assigns a sensitivity label to a file hosted in OneDrive or SharePoint online. 
+
+ Note: You must [associate](https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-how-subscriptions-associated-directory// "associate") or [create](https://docs.microsoft.com/en-us/azure/cost-management-billing/manage/create-subscriptionp:// "create") Azure Subscription for your tenant to call the AssignSensitivityLabel API.  During Private Preview, an Azure billing statement will be created you to see the details of the transactions, however, no charge will be associated with that bill. Please see“Enroll in the Preview" below for instructions on how to apply to be in the Assign Sensitivity Label Private Preview.
```
Lauragra, line 17: Will this apply to any/all monetized APIs in the future? If so, I recommend phrasing this as a general requirement for current and future APIs. This might fall under a heading like "Prerequisites for using monetized APIs", and the text can be along the lines of the following:

"Before you can use premium OneDrive or SharePoint APIs, you must associate your Azure subscription with your tenant, or create a new Azure subscription and associate it with your tenant. While the APIs are in preview, an Azure billing statement will be created for you to see the details of the transactions; however, you will not be charged for API calls until the APIs are generally available."

Will all monetized APIs for OneDrive and SharePoint require users to enroll in a private preview, or is that unique/specific to assignSensitivityLabel? If the former, then you might want to add a generalized statement such as:

"To start using premium APIs while they are in preview, you must first request access."

{I think we're referring to monetized APIs as "premium" APIs. If that is the public term that has been established, we should use it in this context.)

concepts/onedrive-sharepoint-payment-overview.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Licensing model at General Availability 
+
+The AssignSensitivityLabel API, will follow a per call billing model at General Availability. 
+
+## Enroll in the Preview
+
+For your application to enroll in the Preview to call [AssignSensitivityLabel](hhttps://learn.microsoft.com/en-us/graph/api/driveitem-assignsensitivitylabel?view=graph-rest-beta&tabs=httpttp:// "AssignSensitivityLabel"), we require you to complete this request form providing an active Azure subscription. Once we have processed your request form, we will follow up with next steps to onboard your application to billing.
```
Lauragra, line 25: Because we already call out that they must enroll/request access in the API topic, do we need to repeat that information here? Unless it applies more generally to future premium API, in which case, you might want to make a more general statement, per my previous comment.

concepts/onedrive-sharepoint-payment-overview.md
Lauragra, line 9: You'll want to update the assignSensititvityLabel API topic to include a cross-reference to this topic, which I imagine will have more general information about premium APIs for SharePoint and OneDrive. In that topic currently, we mention that the API is protected, and requires users to request access, but we don't specifically mention that it will be a premium (monetized) API, when it reaches GA.

PR #18923, opened 2022-10-19
EducationSubmission: new webUrl property

cristobal-buenrostro:
webUrl property added to EducationSubmission resource type,  this change is applicable to beta only

### DIFF
--- a/api-reference/beta/api/educationsubmission-get.md
+++ b/api-reference/beta/api/educationsubmission-get.md
@@ -64,7 +64,7 @@ The following is an example of the request.
   "name": "get_educationsubmission"
 }-->
 ```msgraph-interactive
-GET https://graph.microsoft.com/beta/education/classes/59069eb2-2a09-4d90-bb19-2089cc69d613/assignments/80da1069-a635-4913-813f-d775a5470a8f/submissions/869369de-3e5a-89eb-6f2d-83cd88f860b5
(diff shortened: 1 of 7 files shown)


api-reference/beta/api/educationsubmission-get.md
```diff
@@ -386,94 +390,39 @@ Content-length: 4492
             "lastModifiedBy": null,
             "points": null,
             "publishedPoints": null
-        },
+        }
+    ],
+    "resources@odata.context": "https://graph.microsoft.com/beta/$metadata#education/classes('2003c52e-807a-4186-9b49-60c573095461')/assignments('2be08d97-b140-4eec-8cbd-88238d571060')/submissions('d9c09b6e-a84d-139a-1b9d-0fbb0ab0de4a')/resources",
+    "resources": [
         {
-            "@odata.type": "#microsoft.graph.educationRubricOutcome",
```
bivasmaiti26, line 391: Can we try to find an example with these values? This response should be an example of a response with expand options, so we should not be deleting these

bivasmaiti26 (approved):

Jarbas-MSFT:
Hello @cristobal-buenrostro, I removed the schema live label because the webUrl property isn't yet part of the live schema.

Jarbas-MSFT:
@Lauragra, this PR is ready for final review.

Lauragra (approved):

PR #18945, opened 2022-10-21
Clarify which ChatMessage properties are read-only

nkramer:
@RamjotSingh , @sumanac , can you verify this? With the introduction of update message in delegated, there's been some questions about which properties can be updated, see also PR https://github.com/microsoftgraph/microsoft-graph-docs/pull/18944


dodaromike (approved):
Thank you.

PR #18959, opened 2022-10-24
Update file-list file-extensions sample to include filter of file extension and folder

aaclage:
Update documentation MGT File-List to include sample of filtering that include folder

https://github.com/microsoftgraph/microsoft-graph-toolkit/issues/1749
@sebastienlevert is the responsable for MGT repo and content revision. 

Thank you, 
AL


sebastienlevert:
Thanks for the PR @aaclage. This is a good example of a bug vs. a feature 😃 At this point, this is not a supported method as it could bring other artefacts, like files without extensions (which are valid in OneDrive and SharePoint). It can be used as a stop gap but it assumes that no file without extension are available in the scope. 

We are currently working on another feature to allow this specific need (and support better paging) that might come and fix this issue, and we would provide the required docs at the same time. https://github.com/microsoftgraph/microsoft-graph-toolkit/issues/1749

Thanks for your time crafting this PR though! Let us know your thoughts!

aaclage:
@sebastienlevert ok o woll close pr

PR #18997, opened 2022-10-26
Added new risk detail

tracyshi:
Added new risk detail m365DAdminDismissedDetection.


Jarbas-MSFT:
This PR is no longer needed because PR 18684 will introduce this change now.

PR #19001, opened 2022-10-26
Updated MESG description

psaffaie:
Updated the mail enabled security description to correct the statement of a shared mailbox to routed emails to group members.

### DIFF
--- a/api-reference/v1.0/resources/groups-overview.md
+++ b/api-reference/v1.0/resources/groups-overview.md
@@ -91,7 +91,7 @@ To learn more about Microsoft 365 groups and the administrator experiences, see 
 
 **Security groups** are for controlling user access to resources. By checking whether a user is a member of a security group, your app can make authorization decisions when that user is trying to access some secure resources in your app. Security groups can have users, other security groups, devices, and service principals as members.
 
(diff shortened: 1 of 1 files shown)


api-reference/v1.0/resources/groups-overview.md
```diff
@@ -91,7 +91,7 @@ To learn more about Microsoft 365 groups and the administrator experiences, see
 
 **Security groups** are for controlling user access to resources. By checking whether a user is a member of a security group, your app can make authorization decisions when that user is trying to access some secure resources in your app. Security groups can have users, other security groups, devices, and service principals as members.
 
-**Mail-enabled security groups** are used in the same way as security groups, but with the added feature of a shared mailbox. Mail-enabled security groups can't be created or updated through the API; instead, they're read-only. Learn more in the [Manage mail-enabled security groups Exchange article](/Exchange/recipients/mail-enabled-security-groups).
+**Mail-enabled security groups** are used in the same way as security groups, but have the ability to send group members email commnunications. Mail-enabled security groups can't be created or updated through the API; instead, they're read-only. Learn more in the [Manage mail-enabled security groups Exchange article](/Exchange/recipients/mail-enabled-security-groups).
```
FaithOmbongi, line 94: Suggestion
```suggestion
**Mail-enabled security groups** are used in the same way as security groups, but can be used to send emails to group members. Mail-enabled security groups can't be created or updated through the API; instead, they're read-only. Learn more in the [Manage mail-enabled security groups Exchange article](/Exchange/recipients/mail-enabled-security-groups).
```

FaithOmbongi (approved):
Thank you @psaffaie for fixing this reference. Approved with suggestions.

PS: Please also apply the fix to the beta doc.

PR #19022, opened 2022-10-27
Update contentsharingsession.md

satyakonmsft:


benlee-msft (approved):

api-reference/v1.0/resources/contentsharingsession.md
```diff
@@ -44,3 +45,98 @@ The following is a JSON representation of the resource.
   "id": "String (identifier)"
 }
 ```
+
+## Receiving contentSharingSession Notifications
```
Lauragra, line 49: This is a non-standard H2 heading. We don't introduce custom H2s in resource or API topics so that we maintain a consistent and predictable structure for users. Can we add this information somewhere else, such as a how-to/concept topic?

PR #19027, opened 2022-10-28
Added alternate key support to directory role

JohnMaina:


FaithOmbongi:
To validate queries once the API goes live.

The documented syntax violates the OData-defined syntax for alternate keys: https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#_Toc31360936

FaithOmbongi (changes requested):
Hi @franqq - Can you verify whether these documented syntaxes are correct and work? They are different from the OData alternate key syntax: https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#_Toc31360936

FaithOmbongi (approved):

PR #19058, opened 2022-11-01
Fixed app-only permissions for list and get methods

rkarim-ms:
Removed the app-only permission RoleManagement.Read.Directory from the get and list methods for Role Assignment Schedule Requests and Role Eligibility Schedule Request APIs for  PIM Graph v1.


FaithOmbongi (approved):
Hi @rkarim-ms - Please also apply these changes to the corresponding beta files. Thanks.

PR #19059, opened 2022-11-01
Add description for alert workflow

zhishending:


AshleyYangSZ (approved):

kennywangms (approved):

api-reference/beta/resources/devicemanagement-monitoring.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "monitoring resource type"
-description: "Represents the entry point entity type to access all alert-related resources."
+description: "Represents the entry point entity type to access all alert-related resources. The monitoring APIs provide a programatic alert experience in Microsoft Endpoint Manager(MEM) admin center. MEM admins could create alert rules with prefered notification channels, like email and MEM portal notification, and receive alerts when conditions set in alert rules are met. Each alert will be recorded as an alert record. Admins could review created alert records to learn about alert impact, severity, status and more. Only MEM global admin, Intune admin and Cloud PC admin have access to monitoring APIs."
```
isvargasmsft, line 3: Please move this description to line 16.

isvargasmsft, line 3: Please move this description to line 16.

isvargasmsft, line 3: Please move this description to line 16.

isvargasmsft (changes requested):
Hello @zhishending, 

Please move the description to line 16 where it is visible for customers. The description in the topic metadata (line 3 in this case) is used for search engines mostly.

Kind regards,

Isaac Vargas

api-reference/beta/resources/devicemanagement-monitoring.md
angelgolfer-ms, line 3: Line 3:
@zhishending , wanted to explain this - line 3 is metadata that is not displayed to customers on the docs site. Everything from lines 1-9 is not visible there.
In general, it's useful to check how your content looks to customers by going to the staging site. When your commit is built successfully, the OPS build info (prefixed with opbldxx, like "opbld32") would show a table of the topics that have been built. That table has a "Preview URL" column, and each topic that has been built has a "View" link. Click that link to see how the topic is staged on the staging site for your PR.

angelgolfer-ms: I have some more feedback in an email about the content that would be helpful for customers to gain enough context. I will send it in a short while.

angelgolfer-ms (changes requested):
Hi @zhishending , thank you for the follow-up. I entered a few comments, and emailed some more feedback for the content and the TOC issue we had discussed on Teams. Please take a look. Thank you!

api-reference/beta/api/devicemanagement-alertrecord-setportalnotificationassent.md
```diff
@@ -13,7 +13,7 @@ Namespace: microsoft.graph.deviceManagement
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Set a single portal notification status to published by modifying the **isPortalNotificationSent** property to `true` for the user specified in the request.
+Set a single portal notification status to published by modifying the **isPortalNotificationSent** property to `true` for the user specified in the request. Portal refers to Microsoft Endpoint Manager admin center.
```
angelgolfer-ms, line 17: Line 16:
@zhishending , I know the first  sentence of this line has already been published as is. I think we can clarify this better though - the portalNotification type doesn't characterize a notification with any "status" (there's no "status" property), and therefore saying something like setting some "status" to "published" can mislead customers to expect and look for some kind of status-related property, which they won't find. On the other hand, there's this explicit property isPortalNotificationSent. In this case, I think we should just say:
"Set the sent-status of a notification on the Microsoft EndPoint Manager admin center by modifying the **isPortalNotificationSent** property to `true` for the user specified in the request. "

api-reference/beta/api/devicemanagement-alertrecord-setportalnotificationassent.md
```diff
@@ -13,7 +13,7 @@ Namespace: microsoft.graph.deviceManagement
...
 
 ## Permissions
```
angelgolfer-ms, line 19: Still line 16:
I see you included "for the user specified in the request".
Can you clarify which request you're referring to?

If I look at the query URL of this methods, there's no segment that represents any user. Can you clarify which user specified where?

angelgolfer-ms: In the interest of time, FYI I went through the metadata and it doesn't seem to me a request like the following specifies any particular user:
`POST https://graph.microsoft.com/beta/deviceManagement/monitoring/alertRecords/{alertRecordId}/setPortalNotificationAsSent`

So I am going to remove the phrase "for the user specified in the request". If that's an oversight on my part, please let me know on Teams.

angelgolfer-ms: Related to the concept of some specified user - I see that in devicemanagement-portalnotification.md, you had specified as the description:
"Represents the portal notification associated with the alert record of a user." 

How is some user specified in this context of your API?

PR #19064, opened 2022-11-01
Change of wording due to upcoming change in behavior for trending APIs

einarmr:
Changed one paragraph and added one paragraph regarding the consequences of disabling item insights. The service change is already communicated to M365 enterprise customers and will roll out on December 1, 2022.


concepts/item-insights-overview.md
```diff
@@ -60,8 +60,9 @@ After disabling item insights, the following happens:
...
-* Experiences that show content trending around you is blocked for you and inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending-insights also affects your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
+* Experiences that show content trending around you is inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending insights also affect your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
 ![disabled item insights in Delve](images/disabled-item-insights-in-delve.PNG)
+* Note that _your_ access to experiences showing trending content is _not_ revoked, while _your activities_ are protected from being shared. In other words, when you have item insights disabled you are protected against undesirable discovery of your data, but you yourself are not locked out of seeing activities and data from users who have chosen to contribute.
```
elmakhmu, line 65: Maybe it's worth to highlight the change and specify the difference in experiences before/after 1Dec plus wrap it using
> [!NOTE]
>

what do you think @angelgolfer-ms ?

concepts/item-insights-overview.md
```diff
@@ -60,8 +60,9 @@ After disabling item insights, the following happens:
...
 * Your colleagues can still see files that you have shared with them from OneDrive for Business and SharePoint in experiences such as Delve and [the profile cards in Microsoft 365](https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501). To avoid seeing shared files in these places, remove the sharing permissions on the files.  
-* Experiences that show content trending around you is blocked for you and inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending-insights also affects your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
+* Experiences that show content trending around you is inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending insights also affect your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
 ![disabled item insights in Delve](images/disabled-item-insights-in-delve.PNG)
```
elmakhmu, line 64: we might need +1 screenshot displaying Delve->Me page changes

concepts/item-insights-overview.md
einarmr: Yes, thanks for the tip. I highlighted the upcoming change in a NOTE section now.

concepts/item-insights-overview.md
einarmr: I think the "NOTE" section that I added should be sufficient. Let's not make a bigger thing out of it.

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
...
-* Disabling item insights will stop [suggested meeting hours](https://support.microsoft.com/office/update-your-meeting-hours-using-the-profile-card-0613d113-d7c1-4faa-bb11-c8ba30a78ef1) from being calculated and shown to the user on their persona card. 
-* For a user who has disabled item insights, querying the [trending](/graph/api/resources/insights-trending) and [used](/graph/api/resources/insights-used) resources in Microsoft Graph API returns `HTTP 403 Forbidden`.
+1. Note that _your_ access to experiences showing trending content is _not_ revoked, while _your activities_ are protected from being shared. In other words, when you have item insights disabled you are protected against undesirable discovery of your data, but you yourself are not locked out of seeing activities and data from users who have chosen to contribute.
```
Lauragra, line 65: Remove "Note that" from the text here.

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
...
+1. Disabling item insights will stop [suggested meeting hours](https://support.microsoft.com/office/update-your-meeting-hours-using-the-profile-card-0613d113-d7c1-4faa-bb11-c8ba30a78ef1) from being calculated and shown to the user on their persona card.
```
Lauragra, line 66: We've already set the context that item insights is disabled, so please rephrase this as:
Suggested meeting hours will no longer be calculated and shown to the user on their persona card.

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
...
+1. Disabling item insights will stop [suggested meeting hours](https://support.microsoft.com/office/update-your-meeting-hours-using-the-profile-card-0613d113-d7c1-4faa-bb11-c8ba30a78ef1) from being calculated and shown to the user on their persona card. 
+
+> [!NOTE]
+> Item [4] in the list above is a change in behavior that was put into effect on December 1, 2022. Previously, users who had item insights turned off were not themselves able to use experiences with trending content, like Delve. As an example, when a user who had disabled item insights opened the Delve home page, it showed an empty page. After the behavior change, Delve _does_ show content when users with item insights disabled open the Delve home page, although the activities from these users are still protected from being shared with others. It should be noted that nothing has changed for organizations where the administrator has disabled item insights for the entire organization – since no users in the organization share their activities, experiences like Delve remain empty.
```
Lauragra, line 69: I'm not sure that we want to specify a date here as over time this will get outdated unless we remove it later. We can just publish this update on December 1. If you want to call attention to the change, you can publish a blog post that describes the change and links to this topic.

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
 Updating settings can take up to 24 hours to apply within the user’s organization across Microsoft 365 experiences.
 
 After disabling item insights, the following happens: 
-* Microsoft stops using your signals when computing recommendations for others. Your colleagues stop receiving item insights based on activities that you perform in shared workspaces, and your colleagues’ insights-based experience becomes less rich in places such as the **Recommended** files in Office.com and Delve.
+1. Microsoft stops using your signals when computing recommendations for others. Your colleagues stop receiving item insights based on activities that you perform in shared workspaces, and your colleagues’ insights-based experience becomes less rich in places such as the **Recommended** files in Office.com and Delve.
 ![disabled item insights in recommended in office](images/disabled-item-insights-in-office.PNG)
```
angelgolfer-ms, line 61: Line 60:
I just noticed this was from a previous update, not this PR - FYI I tweaked the last part of the sentence because technically, when we bold a term which appears in the UI ("Recommended"), we should refer to the term as a UI string so that it's not ambiguous for Loc. 

I changed the original phrase "such as the **Recommended** files in Office.com and Delve"
to "such as the **Recommended** section of files in Office.com and Delve".

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
...
-* Your colleagues can still see files that you have shared with them from OneDrive for Business and SharePoint in experiences such as Delve and [the profile cards in Microsoft 365](https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501). To avoid seeing shared files in these places, remove the sharing permissions on the files.  
-* Experiences that show content trending around you is blocked for you and inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending-insights also affects your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
+1. Your colleagues can still see files that you have shared with them from OneDrive for Business and SharePoint in experiences such as Delve and [the profile cards in Microsoft 365](https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501). To avoid seeing shared files in these places, remove the sharing permissions on the files.
```
angelgolfer-ms, line 62: Line 62:
Again, I know my next comment applies to something added in a previous PR, nonetheless, there might be room for clarification. 
Regarding the phrase "To avoid seeing shared files in these places", does replacing the phrase by "to prevent these places from displaying shared files" cover the behavior?
I find it a bit odd to say "to avoid seeing", because technically, one can just avert or close their eyes, not go to Delve, nor open any profile card. I also wonder how different languages have handled this (even though I'm not about to check every localized language now...).

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
 Updating settings can take up to 24 hours to apply within the user’s organization across Microsoft 365 experiences.
 
 After disabling item insights, the following happens: 
-* Microsoft stops using your signals when computing recommendations for others. Your colleagues stop receiving item insights based on activities that you perform in shared workspaces, and your colleagues’ insights-based experience becomes less rich in places such as the **Recommended** files in Office.com and Delve.
```
angelgolfer-ms, line 60: Lines 60-66:
FYI in docs, when we number the items in a list, we're saying there's an order for the items. In our developer docs, typically we use numbered lists to show steps in a process. For more info, see this topic in the Microsoft Writing Style Guide: https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=56575

So, please change this back to a bullet list.

concepts/item-insights-overview.md
angelgolfer-ms, line 66: Line 66:
1 comment and 2 questions: 
FYI I changed the first part of the sentence from future tense to present tense. That is "Disabling item insights will stop..." has become "Disabling item insights stops..."
This is because in docs, as long as the behavior is already implemented, even though there is a cause and effect relationshop between the 2 actions, we still use present tense. 

Questions:
(a) Regarding "... shown to the user...", are you referring to any user who is viewing a specific user's persona card, not just a user's viewing his/her own persona card and stop seeing suggested meeting hours, right?
If that's the case, then please rephrase as:
"Disabling item insights stops suggested meeting hours from being calculated and displayed on a user's persona card shown for other users."

(b) Regarding "persona card" - do you mean profile card as described here: https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501

FYI searching for "persona card" on the internet in the context of Microsoft 365 is not returning me any relevant results. I just want to make sure we're using a term that end users and admin users recognize, and is used in end user and admin docs as well.

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
...
-* Experiences that show content trending around you is blocked for you and inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending-insights also affects your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
+1. Your colleagues can still see files that you have shared with them from OneDrive for Business and SharePoint in experiences such as Delve and [the profile cards in Microsoft 365](https://support.microsoft.com/en-us/office/profile-cards-in-microsoft-365-e80f931f-5fc4-4a59-ba6e-c1e35a85b501). To avoid seeing shared files in these places, remove the sharing permissions on the files.  
+1. Experiences that show content trending around you is inaccessible to others, in places such as Delve and Outlook mobile’s **Discover** section. The lack of the trending insights also affect your personalized relevance in Microsoft Search since it cannot use your trending content as signals for relevance.
```
angelgolfer-ms, line 63: Line 63:
1 comment and 1 question:
FYI the verb "affect" would have to be "affects", because the subject is actually "the lack of ..." which is uncountable. The subject is not "trending insights". :-)

Question:
I know the use of the "affect" verb is from a previous PR. Nonetheless, is there another verb that is less vague than "affect", such as "reduces your personalized relevance..."?

concepts/item-insights-overview.md
angelgolfer-ms: May I tag on and suggest a further revision?
Perhaps we shouldn't associate _not_ sharing any data about oneself with the word "protected", allowing the room for customers to interpret that sharing would be a bad thing and the user is unprotected or less protected?  
After all, our uber message for these insights features is, there's sufficient control and customers should not be concerned as long as they use those controls depending on their preference. 

Also, in the first sentence, I think "activities" may be a bit broad, because in the context of Microsoft Graph, there are many other well-known activities, such as messaging, meeting, etc.

So, what about:
"Insights about your file-based activities are no longer shared with others. "
Or even:
"File-based insights about you are not longer shared with others."

Just picking one of the above two and continuing:
"Insights about your file-based activities are no longer shared with others, protecting you from undesirable discovery of your data. On the other hand, your access to experiences showing trending content is not revoked. You can see the activities and data of users who have not disabled item insights."

concepts/item-insights-overview.md
angelgolfer-ms: I agree with Lauragra's comment - this type of information about transitioning from one behavior to another behavior should go to a blog post. FYI, in general, documentation should show the current behavior with no ambiguity. This is the expectation that has been set up for customers about docs over the years.

angelgolfer-ms: Nonetheless, the following comment is for your blog draft:
You have used the terms "empty page" or "remain empty" for Delve. 
I don't have admin control to see what the experience actually is, perhaps indeed Delve doesn't display any trending documents anyway. In this case, at two levels:
- I would hope the Delve UI actually has a short message that says something to the effect of "there are no documents to display for your organization as your admin has disabled display of popular documents", or just some message. IMO a real empty page in Delve without any message may make customers think that Delve is broken or still trying to return results.
- In your eventual blog draft, instead of "empty", you can please say something to the effect of, "Delve doesn't display documents trending in the organization".

concepts/item-insights-overview.md
```diff
@@ -57,13 +57,19 @@ By default, item insights are enabled. Administrators use specific settings to c
...
+1. Disabling item insights will stop [suggested meeting hours](https://support.microsoft.com/office/update-your-meeting-hours-using-the-profile-card-0613d113-d7c1-4faa-bb11-c8ba30a78ef1) from being calculated and shown to the user on their persona card. 
+
+> [!NOTE]
+> Item [4] in the list above is a change in behavior that was put into effect on December 1, 2022. Previously, users who had item insights turned off were not themselves able to use experiences with trending content, like Delve. As an example, when a user who had disabled item insights opened the Delve home page, it showed an empty page. After the behavior change, Delve _does_ show content when users with item insights disabled open the Delve home page, although the activities from these users are still protected from being shared with others. It should be noted that nothing has changed for organizations where the administrator has disabled item insights for the entire organization – since no users in the organization share their activities, experiences like Delve remain empty.
+
```
angelgolfer-ms, line 70: Extending my last comment about not describing historical changes in docs, since it's been a while that the mechanism for admins to disable displaying insights has come  into effect, @elmakhmu or @einarmr , can you please look into removing this Background section in the other concept topic: https://review.learn.microsoft.com/en-us/graph/insights-customize-item-insights-privacy?branch=main#background

We let that get into the docs initially to call out actions taken to address customers' concerns about privacy. Since the change has been in place for a while now, as Lauragra and I mentioned above, we can snap back into consistency. 

If either of you can plan to do that in the near future, that would be good. Thanks in advance.

angelgolfer-ms (changes requested):
Thanks for the PR. I have a few comments, please take a look. As a heads-up, some of my comments apply to updates in the same section that came from a previous PR. I hope @einarmr, or an appropriate person on your team, doesn't mind addressing them. In general, even after initial publishing, docs do change to better support customers' needs.

concepts/item-insights-overview.md
einarmr: Very well, thanks

concepts/item-insights-overview.md
elmakhmu: Can I add my 2 cents, can we be even more precise saying "Insights about your file-based activities **from collaborative workspace** are no longer shared with others." 

maybe also worth to mention that ACLs are untouched and other users still will have access to the documents (in search for example)

concepts/item-insights-overview.md
einarmr: Thanks, done

concepts/item-insights-overview.md
einarmr: Thanks @elmakhmu, @Lauragra, @angelgolfer-ms. I merged your suggestions, I think it looks great now. 
@elmakhmu, regarding your point on ACLs, I think this is already very well described in the earlier paragraphs (especially in the paragraph with Alice, Robert, Kate), so I chose to skip that.

concepts/item-insights-overview.md
einarmr: Thanks @angelgolfer-ms, @Lauragra. I consulted with Melinda Jackson and Bill Bär earlier, and they instructed me to publish a message center post to tenants 30 days ahead of the change, as well as to update the documentation. Their view was that no additional communication was needed (like blog posts, etc.). Adding to that, I do not really _want_ to draw much attention towards this change. In case a customer ever wonders (and contacts Microsoft with their questions), we can refer to the updated documentation, and that should be sufficient.

With that context, I will simply remove this "NOTE" section. Please let me know if you (strongly) disagree. @elmakhmu FYI.

Please also note that users with item insights disabled will see an informational banner in Delve from December 1st, with wording along the lines of "Starting Dec 1, 2022, you are permitted to use Delve even though Item Insights is disabled. Your own activities are not being shared. For more information, see the [documentation](https://learn.microsoft.com/en-us/graph/item-insights-overview).

concepts/item-insights-overview.md
einarmr: Done, thanks

concepts/item-insights-overview.md
einarmr: Agree, thanks

concepts/item-insights-overview.md
einarmr: (a) Yes.
(b) Yes.
I also changed this to "**your** profile card", since all the other items refer to "your this" and "your that".

concepts/item-insights-overview.md
einarmr: I agree, this should be fixed. I will keep it outside this PR, though. 
(I see that there is a note in [Transition period](https://review.learn.microsoft.com/en-us/graph/insights-customize-item-insights-privacy?branch=main#transition-period) that should be removed too.)

concepts/item-insights-overview.md
```diff
@@ -56,14 +56,14 @@ By default, item insights are enabled. Administrators use specific settings to c
```
angelgolfer-ms, line 56: @einarmr , this question/comment is for the section that starts at line 46:
Is it true that the approaches - admin center, PowerShell, REST API - described in the [customize item insights privacy](https://learn.microsoft.com/en-us/graph/insights-customize-item-insights-privacy) topic are all still in public preview?
If so , then per our guidance to apply "(preview)" tag to content in the docs, I think the section describing disabling item insights (starting with line 46) should also be tagged with "(preview)".

I am adding that tagging to line 46 for now. Please confirm.

einarmr: Very good question. 😊 I believe the best person to answer this is @elmakhmu... Elvira, may I suggest that we just remove the word "preview" from that [other page](https://learn.microsoft.com/en-us/graph/insights-customize-item-insights-privacy)? We could do that in a separate PR to keep things clean, though.

angelgolfer-ms: Just to be clear, the docs use "preview" to identify APIs or behaviors that  apply to only beta but not v1. 

So are the capabilities described for the admin center, PowerShell, REST API in the [customize item insights privacy](https://learn.microsoft.com/en-us/graph/insights-customize-item-insights-privacy) topic all still in public preview?

einarmr: Yes, you are correct. This stuff is only exists on 'beta', not on v1.0.

concepts/item-insights-overview.md
```diff
@@ -56,14 +56,14 @@ By default, item insights are enabled. Administrators use specific settings to c
 
 Updating settings can take up to 24 hours to apply within the user’s organization across Microsoft 365 experiences.
 
-After disabling item insights, the following happens: 
-* Microsoft stops using your signals when computing recommendations for others. Your colleagues stop receiving item insights based on activities that you perform in shared workspaces, and your colleagues’ insights-based experience becomes less rich in places such as the **Recommended** files in Office.com and Delve.
+After disabling item insights, the following happens:
```
einarmr, line 59: @angelgolfer-ms, I changed this line back to the original again. The section above describes how a user can disable item insights herself, while the paragraph above describes that an admin can do so too (and that the admin control "wins" if they're in conflict). 

In other words, there are actually three ways (in total) to have item insights disabled. The admin can disable it for everyone in the organization, the admin can disable it for a security group within the organization, or the user can disable it herself.

concepts/item-insights-overview.md
```diff
@@ -56,7 +56,7 @@ By default, item insights are enabled. Administrators use specific settings to c
 
 Updating settings can take up to 24 hours to apply within the user’s organization across Microsoft 365 experiences.
 
-After an administrator disables your item insights, the following happens:
+After disabling item insights, the following happens:
```
angelgolfer-ms, line 59: @einarmr , I see, I guess you're referring to using REST API or PowerShell to change the insights settings for a user, right?

einarmr: Yes. (The screenshot in the section above (from MyAccount) is the user-consented switch.)

concepts/item-insights-overview.md
elmakhmu: Confirming that API is still only in beta, however UI toggles in MyAccount and MAC are on prod and even in some Sovereign rings. We are blocked by other service, this issue I hope will be solved in CY2023

angelgolfer-ms: Thanks for the confirmation, @elmakhmu .
FYI I left the "(preview)" tag in the topic and TOC titles for the [customize item insights privacy](https://learn.microsoft.com/en-us/graph/insights-customize-item-insights-privacy) topic. 

Technically, if the topic title doesn't include "in Microsoft Graph", I would just gauge the contents of the topic (which includes the admin center experience that is generally available to admins) and not tag the topic as "(preview)". However, in Microsoft Graph which offers PowerShell and REST APIs, indeed those APIs are in beta only, so it's accurate to tag that title as "preview". And we keep "in Microsoft Graph" in the topic title for SEO.

concepts/item-insights-overview.md
```diff
@@ -56,14 +56,14 @@ By default, item insights are enabled. Administrators use specific settings to c
 
 Updating settings can take up to 24 hours to apply within the user’s organization across Microsoft 365 experiences.
 
-After disabling item insights, the following happens: 
-* Microsoft stops using your signals when computing recommendations for others. Your colleagues stop receiving item insights based on activities that you perform in shared workspaces, and your colleagues’ insights-based experience becomes less rich in places such as the **Recommended** files in Office.com and Delve.
+After your item insights are disabled, the following happens:
+* Microsoft stops using your signals when computing recommendations for others. Your colleagues stop receiving item insights based on activities that you perform in shared workspaces. Your colleagues’ insights-based experience becomes less rich in places such as the **Recommended** section of files in Office.com and Delve, as shown in the following example, if you were signed in with the user name MeganB.
```
angelgolfer-ms, line 60: Line 59:
Thanks for reminding me that disabling item insights is not an admin-only task.
The following comments are FYI just in case anyone wonders why I changed from "After disabling item insights" to "After your item insights are disabled" :-)
- For reminding readers that as in line 54, we're talking about disabling "your item insights", as the section continues with that perspective beyond line 59.
- Grammatically, if we start a sentence with a gerund phrase, like "Skipping sleep last night", then the next phrase of the same sentence needs a subject, as in, "Skipping sleep last night, I am sleepy today", to clarify who "skipped sleep". So similarly, if we say "After disabling item insights", or "After disabling your item insights", then the next phrase must start with a subject to clarify who does the disabling in the first phrase, but the next phrase doesn't right now. 
- So if we say "After your item insights are disabled", the subject of that phrase becomes "your item insights" (even though it's in passive voice and we generally want to avoid passive voice), so the next phrase doesn't have to clarify the subject of the first phrase anymore.

einarmr: Thanks for the great advice 😊

angelgolfer-ms (approved):
I think the PR looks good now.
Please let me know if you notice any issues otherwise.

concepts/item-insights-overview.md
angelgolfer-ms: While closing my opened browsers for this PR, I discovered your last comment. At a point a couple of days ago, I did find the various ways weren't the clearest anymore especially when skimming the published topic. So FYI I made another commit. 
That would be it for the time being unless anyone finds any issues. Thanks.

angelgolfer-ms:
@einarmr , I made another [commit](https://github.com/microsoftgraph/microsoft-graph-docs/pull/19064/commits/59fc28094e4d70943d831d070103f2b6dbc3e751) today to make some more clarifications. I am done for my review unless anyone sees any more issues for now.
When you're ready to publish (the behavior goes live), please follow step 8 of the [content review workflow](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211773/Content-workflow) to set the PR label to ready to publish.

einarmr:
Thanks @angelgolfer-ms! I made another minor adjustment to your commit in insights-customize-item-insights-privacy.md. Feel free to make changes to this again 😊

PR #19092, opened 2022-11-02
Update LCW tasks to use cross-repo include

FaithOmbongi:
Added suppression for the cross-repo includes references because API Doctor doesn't understand cross-repo includes. Will follow up with @millicentachieng to get a long-term fix.


FaithOmbongi:
Rebuild after changes to Azure docs links

PR #19101, opened 2022-11-03
Removed extra resource file and updated query option support

vibatra:
Removed extra resource file and updated documentation for query option support at property level for various resources.

ADO content review work item: 2150282


api-reference/beta/resources/identitygovernance-lifecycleworkflowscontainer.md
```diff
@@ -36,9 +36,7 @@ Inherits from [entity](../resources/entity.md).
 
 ## Properties
 
-|Property|Type|Description|
-|:---|:---|:---|
-|id|String|Identifier used for individually addressing the lifecycle workflows objects. Inherited from [entity](../resources/entity.md).|
+None
```
FaithOmbongi, line 39: nit: id is actually an inherited property.

api-reference/beta/resources/identitygovernance-task.md
```diff
@@ -31,28 +31,28 @@ Inherits from [entity](../resources/entity.md).
...
 |Relationship|Type|Description|
 |:---|:---|:---|
 |taskProcessingResults|[microsoft.graph.identityGovernance.taskProcessingResult](../resources/identitygovernance-taskprocessingresult.md) collection|The result of processing the task.|
 
+### Supported tasks
+
+Lifecycle Workflows currently support the following tasks:
+
+[!INCLUDE [identitygovernance-lifecycleworkflows-tasks](../includes/lifecycle-workflows-tasks-table.md)]
+
```
FaithOmbongi, line 55: Move this after the "Properties" table and before the "Relationships" section

api-reference/beta/resources/identitygovernance-taskdefinition.md
```diff
@@ -28,12 +28,12 @@ Inherits from [entity](../resources/entity.md).
 
 |Property|Type|Description|
 |:---|:---|:---|
-|category|microsoft.graph.identityGovernance.lifecycleTaskCategory|The category of the HR function that the tasks created using this definition can be used with. The possible values are: `joiner`, `leaver`. This is a multi-valued enumeration whose allowed combinations are `joiner`, `joiner,leaver`, or `leaver`. Supports `$filter` (`eq`, `has`). |
+|category|microsoft.graph.identityGovernance.lifecycleTaskCategory|The category of the HR function that the tasks created using this definition can be used with. The possible values are: `joiner`, `leaver`. This is a multi-valued enumeration whose allowed combinations are `joiner`, `joiner,leaver`, or `leaver`.<br><br>Supports `$filter` and `$orderby`. |
```
FaithOmbongi, line 31: `has` operator should be supported to filter by this enum, but I stand corrected.
```suggestion
|category|microsoft.graph.identityGovernance.lifecycleTaskCategory|The category of the HR function that the tasks created using this definition can be used with. The possible values are: `joiner`, `leaver`. This is a multi-valued enumeration whose allowed combinations are `joiner`, `joiner,leaver`, or `leaver`.<br><br>Supports `$filter` (`has`) and `$orderby`. |
```

api-reference/beta/resources/identitygovernance-taskdefinition.md
```diff
@@ -28,12 +28,12 @@ Inherits from [entity](../resources/entity.md).
...
 |description|String|The description of the taskDefinition.|
-|displayName|String|The display name of the taskDefinition`.|
+|displayName|String|The display name of the taskDefinition.<br><br>Supports `$filter`(`eq`, `ne`) and `$orderby`.|
 |id|String|The unique identifier for the taskDefinition. Inherited from [entity](../resources/entity.md).|
-|parameters|[microsoft.graph.identityGovernance.parameter](../resources/identitygovernance-parameter.md) collection|The parameters that must be supplied when creating a workflow task object.|
-|version|Int32|The version number of the taskDefinition. New records are pushed when we add support for new parameters.|
+|parameters|[microsoft.graph.identityGovernance.parameter](../resources/identitygovernance-parameter.md) collection|The parameters that must be supplied when creating a workflow task object.<br><br>Supports `$filter`.|
```
FaithOmbongi, line 35: indicate operator(s).

.openpublishing.publish.config.json
```diff
@@ -73,6 +73,7 @@
     "redirects/.openpublishing.redirection.1121.json",
     "redirects/.openpublishing.redirection.1221.json",
     "redirects/.openpublishing.redirection.0522.json",
-    "redirects/.openpublishing.redirection.changedapi.json"
+    "redirects/.openpublishing.redirection.changedapi.json",
+    "redirects/.openpublishing.redirection.lcwapi.json"
```
FaithOmbongi, line 77: I recommend removing this redirect. Because the file was never surfaced on the TOC, it had zero hits and therefore no broken links to expect.

image

api-reference/beta/api/identitygovernance-userprocessingresult-summary.md
```diff
@@ -118,7 +118,7 @@ The following is an example of the response
 <!-- {
   "blockType": "response",
   "truncated": true,
-  "@odata.type": "microsoft.graph.identityGovernance.usersProcessingSummary"
+  "@odata.type": "microsoft.graph.identityGovernance.userSummary"
```
FaithOmbongi, line 121: Great catch!

redirects/.openpublishing.redirection.lcwapi.json
```diff
@@ -0,0 +1,9 @@
+{
```
FaithOmbongi, line 1: Delete this redirect file because the deleted file was undiscoverable anyway.

FaithOmbongi (approved):
Thank you @vibatra for your work on this PR, documenting the supported query capabilities. It's great all four relational operators are supported where one is supported - I don't think there's any entity that has that scope of support.


Good catch on identitygovernance-usersprocessingsummary.md! I recommend removing the redirects because the deleted file was almost undiscoverable anyway.

I've left other suggestions for minor fixes.

AlexFilipin (approved):

api-reference/beta/resources/identitygovernance-task.md
vibatra: I thought it was better to mention it after property and relationship sections but yeah don't know the pattern. will revert it back to the previous location after Property section.

.openpublishing.publish.config.json
vibatra: Without it, one of the build checks fails :(.

api-reference/beta/resources/identitygovernance-taskdefinition.md
vibatra: This supports any operator/functions available on a collection in OData, of which there are only 2 at the moment. Even then only one actually works in MS OData implementation. Added the 'any' operator.

api-reference/beta/resources/identitygovernance-taskdefinition.md
vibatra: Ahh. This is a mistake, didn't intent to remove the "eq" and "has" operators but add "ne" and "$orderby" option. Blame my butter fingers. Fixed.

api-reference/beta/resources/identitygovernance-lifecycleworkflowscontainer.md
FaithOmbongi: I support updating this in a separate PR after you actually break that inheritance. Reason is that the inheritance is already propagated downstream to these highlighted 👇🏾 objects and the objects that inherit from them. Removing this reference  now will require also you to also remove/change the "inherited from entity" statement for downstream objects.

image

vibatra: Sounds good. The objects contained inside are not inheriting from this type and thus don't require breaking the inheritance. You will notice that WorkflowBase (and subsequently Workflow and WorkflowVersion) doesn't inherit from Entity but Tasks still do.

OWinfreyATL (approved):

PR #19130, opened 2022-11-04
call out invalid characters are not validated

psaffaie:
Wanted to add a note that calls out the validate properties API does only validates the naming policy and does not take into consideration invalid characters. Invalid characters will fail during group creation but will not throw an error for validate properties


FaithOmbongi (approved):
Let me ask the API Doctor team to help investigate the source of the validation error.

PR #19171, opened 2022-11-08
Added topic on using SDKs with national clouds

jasonjoh:


concepts/sdks/national-clouds.md
```diff
@@ -0,0 +1,188 @@
...
+  clientId: 'YOUR_CLIENT_ID',
+  tenantId: 'YOUR_TENANT_ID',
+  // https://login.microsoftonline.us
+  authorityHost: AzureAuthorityHosts.AzureGovernment,
+  redirectUri: 'YOUR_REDIRECT_URI'
+});
+
+// Create the authentication provider
+const authenticationProvider = new TokenCredentialAuthenticationProvider(credential, {
+  scopes: ['https://graph.microsoft.us/.default']
```
sebastienlevert, line 80: Is this required? I'm thinking about delegated scenarios with incremental consent. Are the scopes the same or neec to be prefixed also?

jasonjoh: It seems you can do dynamic consent with scopes like `User.Read`, but you cannot do `https://graph.microsoft.com/.default` with pre-configured permissions, you'll get an "Invalid audience" error. For pre-configured with the `.default` scope, you need to use the correct domain.

Perhaps a callout above the code samples detailing this would be helpful.

jasonjoh: I just pushed a new commit adding a section on this.

sebastienlevert: Excellent, looks good to me!

isvargasmsft (approved):

PR #19178, opened 2022-11-09
Update team.md

JarbasHorst:
Fixed alphabetical order of the properties table and JSON representation object. Added missing properties to JSON representation. Added classSettings to Properties table based on the info from this note "If the team is of type class, a **classSettings** property is applied on the team.".


JarbasHorst:
@Lauragra, I fixed the API doctor issue by removing the **summary** property for now because it was actually not part of the currently published version of this article. The **summary** property is available in the live schema tough and I can add it properly to this resource topic if we want :)

Lauragra:
> @Lauragra, I fixed the API doctor issue by removing the **summary** property for now because it was actually not part of the currently published version of this article. The **summary** property is available in the live schema tough and I can add it properly to this resource topic if we want :)

Yes, sure, let's do that update as a separate PR. Thanks!

PR #19187, opened 2022-11-09
Change how training modules are referenced in docs

FaithOmbongi:


concepts/toc.yml
```diff
@@ -69,7 +69,7 @@ items:
             displayName: users
           - name: Default user permissions
             href: /azure/active-directory/fundamentals/users-default-permissions?context=graph/context
-          - name: Access user data
+          - name: "Training module: Access user data"
```
Lauragra, line 72: Do we have a pattern of surfacing training modules in the TOC? This takes them out of the context of the docs. Does it need to be surfaced here, or can it be a Next steps link in the overview topic, for example?

concepts/delta-query-overview.md
```diff
@@ -11,7 +11,7 @@ ms.custom: graphiamtop20
 Delta query enables applications to discover newly created, updated, or deleted entities without performing a full read of the target resource with every request. Microsoft Graph applications can use delta query to efficiently synchronize changes with a local data store.
 
 > [!div class="nextstepaction"]
```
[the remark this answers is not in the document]
FaithOmbongi: Gotcha - fixing this.

concepts/toc.yml
FaithOmbongi: Great question. And this loss of context is a big con.

I've asked the writers attached to the SCIM content and no one's aware of any guidance. But we have an ongoing conversation about this to discuss how best to handle referencing links that make readers lose context.

PR #19200, opened 2022-11-10
 Add ACS user type to Teams chats

jessiechan27:


abshar-teams (approved):

api-reference/beta/resources/azurecommunicationservicesuserconversationmember.md
```diff
@@ -0,0 +1,56 @@
...
+  "@odata.type": "#microsoft.graph.azureCommunicationServicesUserConversationMember",
+  "id": "String (identifier)",
+  "roles": [
+    "String"
+  ],
+  "displayName": "String",
+  "visibleHistoryStartDateTime": "String (timestamp)",
+  "azureCommunicationServicesId": "String"
+}
+```
```
dodaromike, line 56: Please add a TOC entry for this entity.

dodaromike (commented):
Thank you for your work on this addition PR.  We found that it needs a TOC entry for the new entity.

isvargasmsft (approved):

api-reference/beta/toc.yml
```diff
@@ -854,6 +854,19 @@ items:
           href: api/group-post-rejectedsenders.md
         - name: Remove rejected senders
           href: api/group-delete-rejectedsenders.md
+      - name: Conversation member
```
Lauragra, line 857: This resource is surfaced under Groups in the TOC but not surfaced under Teamwork and communications/Messaging. Is that the intent, or should we be surfacing it in both places? 

Note that this resource is also not included in the v1.0 TOC; please be sure to update that as well to replicate what we have in beta.

api-reference/beta/toc.yml
```diff
@@ -854,6 +854,19 @@ items:
...
+        href: resources/conversationmember.md
+        items:
+        - name: Anonymous guest
```
Lauragra, line 860: Is there a reason to surface these entities in the TOC? Generally we don't surface complex types or entities that don't have methods on them in the TOC. I wonder if this warrants more detailed explanation in the **conversationMember** topic, instead of the See also links that we have there currently. For example, are these the user types that can be members of a conversation?

api-reference/beta/toc.yml
jessiechan27: I'll do a separate PR for GA soon because I also need to add the ACS user for GA soon when it becomes live. Is that ok?

api-reference/beta/toc.yml
jessiechan27: I removed it

PR #19217, opened 2022-11-11
Update group-list-transitivememberof.md

psaffaie:
adding additional least privileged permissions per email thread "RE: scopes for GET /groups/transitiveMemberOf" the table does not make it clear to developers that they can use the transitivememberOf query without Directory Read All permissions, but may be limited in what is returned


api-reference/v1.0/api/group-list-transitivememberof.md
```diff
@@ -19,7 +19,7 @@ One of the following permissions is required to call this API. To learn more, in
 
 | Permission type                        | Permissions (from least to most privileged) |
 | :------------------------------------- | :------------------------------------------ |
-| Delegated (work or school account)     | Directory.Read.All, Directory.ReadWrite.All |
+| Delegated (work or school account)     | GroupMember.Read.All, Group.Read.All, Device.Read.All, AdministrativeUnit.Read.All, Directory.Read.All, Directory.ReadWrite.All|
```
FaithOmbongi, line 22: I understand why you want to add the Device* and Administrative* permissions but I recommend against it here. Reasons:
1. The include note in Line 26 already mentions that limited info is returned.
2. Device.Read.All or AdministrativeUnit.Read.All permissions by themselves won't allow you to read group memberships. Instead, they will work together with GroupMember.Read.All or Group.Read.All to allow you to read more properties of devices and AUs respectively, that would otherwise have been hidden/null, because of the limited info rule.
```suggestion
| Delegated (work or school account)     | GroupMember.Read.All, Group.Read.All, Directory.Read.All, Directory.ReadWrite.All|
```

FaithOmbongi (approved):
Hi @psaffaie - Please also apply the suggested change to the beta doc.

Approved with suggestions.

psaffaie:
Thank  you, Faith. I committed your suggestions and added the ready to merge label

PR #19220, opened 2022-11-11
Update team.md

JarbasHorst:
Added missing summary property based on live schema.


JarbasHorst:
Closed PR because I noticed that the resource topic for this new property doesn't exist yet.

PR #19247, opened 2022-11-15
[Doc Improvement][Update channel-post]

v-shalinirobinson:


ChetanSharma-msft (approved):

api-reference/v1.0/api/channel-post.md
```diff
@@ -12,7 +12,15 @@ doc_type: apiPageType
 Namespace: microsoft.graph
 
 Create a new [channel](../resources/channel.md) in a team, as specified in the request body.
-> **Note:** You can add a maximum of 200 members when you create a private channel.
+
+If you're creating a private channel, you can add a maximum of 200 members.
```
Lauragra, line 16: FYI, I changed this note to normal text so that we don't have two different types of alerts next to each other. I also made some edits to the text of the Important.

Please apply these changes to this topic in the beta folder as well.

api-reference/v1.0/api/channel-post.md
```diff
@@ -12,7 +12,15 @@ doc_type: apiPageType
...
+
+> [!IMPORTANT]
```
Lauragra, line 18: Is this a temporary issue that will be fixed in the future? If so, then we should treat it as a Known issue. If this is the general behavior, then we should keep the text as-is. Would it be helpful to mention the 400 Bad Request error and reason in the Response section of the Get filesFolder API topic as well?

For details about documenting known issues, see: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211754/Known-issues.

api-reference/v1.0/api/channel-post.md
v-shalinirobinson: Thank you, @Lauragra. I'll make the required change.

api-reference/v1.0/api/channel-post.md
v-shalinirobinson: Hello @Lauragra,

I checked with Prasad who'd raised this request for change. As per him, this is general behavior, and there's no indication if it's going to be fixed in near future. He's agreed that we could call it treat it as a Known Issue as you suggested.

I'm unable to access the link you've shared for Known-issues. Please let me know how to call it out as a Known Issue.

Thanks.

v-shalinirobinson:
#feedback-addressed

v-shalinirobinson:
#feedback-addressed

v-shalinirobinson:
@Lauragra I have applied the comment regarding known issue and updated the known issues file, and the Get filesFolder API file as well.

Lauragra (approved):

PR #19266, opened 2022-11-16
[Doc Issue][resource-id]

v-shalinirobinson:
Adding the missing id property for team resource type

### DIFF
--- a/api-reference/v1.0/resources/team.md
+++ b/api-reference/v1.0/resources/team.md
@@ -53,7 +53,7 @@ For more information about working with groups and members in teams, see [Use th
 
 | Property | Type | Description |
 |:---------------|:--------|:----------|
-| id | string | The unique identifier of the team. The group has the same ID as the team. |
+| id | string | The unique identifier of the team. The group has the same ID as the team. This property is read-only, and is inherited from the base entity type. |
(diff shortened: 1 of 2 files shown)


api-reference/v1.0/resources/team.md
```diff
@@ -53,6 +53,7 @@ For more information about working with groups and members in teams, see [Use th
 
 | Property | Type | Description |
 |:---------------|:--------|:----------|
+| id | string | The unique identifier of the team. The group has the same ID as the team. |
```
Lauragra, line 56: Are we adding this property because it's inherited from the entity type? Please note that if you update the v1.0 topic, you also need to update the same topic in the beta folder. Also, do you need to call out that the property is read only?

v-shalinirobinson: @Lauragra Thank you for the comment. I have raised the comment to the team that raised the issue. I'll get back to you after they respond.

AkJo: It is a yes to all of the following:
1. Inherited from the base entity type.
2. This is a read-only property.
3. The updates should be made in beta version of the doc as well.

v-shalinirobinson:
#feedback-addressed

isvargasmsft (approved):

Lauragra (approved):

PR #19292, opened 2022-11-17
Select_escaped is not found, there is Select which does the expected thing

hofq:
Reference:
https://github.com/microsoftgraph/msgraph-sdk-go/blob/main/users/item/calendarview/calendar_view_request_builder.go#L31


andrueastman (approved):
@Lauragra 
It should be okay to have this merged as this has been updated in the snippet generator. 
The next run should update and resolve to the same result.

hofq:
> @Lauragra It should be okay to have this merged as this has been updated in the snippet generator. The next run should update and resolve to the same result.

okay, that's interesting. How often do the updates happen? I think there are a few other things out of date in the Golang snippets.

jasonjoh (approved):

PR #19345, opened 2022-11-23
Update driveitem-createlink.md with available value, users, for scope

ns-calvinwu:
We noticed `users` is also an available option from the response data of [listing SharePoint permissions endpoint](https://learn.microsoft.com/en-us/graph/api/driveitem-list-permissions?view=graph-rest-1.0&tabs=http)


ns-calvinwu:
@microsoft-github-policy-service agree

PR #19372, opened 2022-11-24
Documentation and change list of peopleInsights settings

stealth-17:


PeterNjorogeMS:
@stealth-17 there are some changes that need to be added before we can push the documentation. 

Let's start from this menu option 
image

The page usersettings here seems to house all the settings a user has control over. This means it would work to edit the page and add peopleInsights to it as well and not replicate the page. 

Each setting has a get and update option to enable or disable. The pages usersettings-get and usersettings-update seem to cover item insights setting owned by @elmakhmu. 

If we are adding two new get and update pages for people insights and the same will follow for all other settings that will come after item and people Insights. Should we:

Option 1:
Have just two single get and update pages where the users can then have a table for the various settings. If we have a new setting it will be just to editing the pages to add the settings. 

Option 2:
Rename usersettings-get page and usersettings-update page to something similar to this usersettings-iteminsights-get and usersettings-iteminsights-update and create similar pages for peopleInsights user setting both get and update. usersetting page can be shared. 

Now option 2 means all other new settings will have their pages added meaning we will have multiple pages and maybe sub menus. The menu can be something like
User Setting
       Usersetting Overview
    >Item Insight user control
        >Get Iteminsight user setting
        >Update Iteminsight user setting
    >People Insight user control 
        >Get Peopleinsight user setting
        >Update PeopleInsight user setting
     >New Insight user control

The menu would look like that. 

@elmakhmu what are your thoughts?

elmakhmu:
@PeterNjorogeMS thanks for adding me to reviewing. I would create separate usersettings-peopleinsights-get/update pages and place them under separate "People Insights" sub menus similar to "item insights" 
Item insights:
- item insights 
- Org controls
- User controls
People insights:
- people insights
- Org controls
- User controls

This will avoid leapfrog in customers' minds and let the flexibility if we will want to add to IIS another level of granularity (if so)

@angelgolfer-ms what do you think would be the best approach?

PR #19435, opened 2022-11-30
REST API Call doesn't work as expected

robgrame:
I've tried to make use of the REST API call mentioned in the this page using Graph API Explorer and I was not able to perform the desired updates. Here are the issues I face with:
- The rest api call as stated won't work . I was able to make ti work only when I've replaced the DEP MacOS Enrollment Profile ID in place of defaultMacOsEnrollmentProfile. So the working call resemble like this: PATCH /deviceManagement/depOnboardingSettings/{depOnboardingSettingId}/{depMacOSEnrollmentProfileID}
- I was not able to change any setting related to: primaryAccountFullName, primaryAccountUserName, primaryUser, primaryUserFullName. Everytime I've tried to setup with a value, the outcome I've got with a following GET request was a null value in place of the expected desired value. The only attribute I was able to change was the Profile name.

I hope it's clear. Please reach me out with DM if required.

/roberto

### DIFF
--- a/api-reference/beta/api/intune-enrollment-depmacosenrollmentprofile-update.md
+++ b/api-reference/beta/api/intune-enrollment-depmacosenrollmentprofile-update.md
@@ -32,7 +32,7 @@ One of the following permissions is required to call this API. To learn more, in
 }
 -->
 ``` http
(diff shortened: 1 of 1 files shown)


chrichards:
Currently have the exact same problem and a case open with Microsoft about it; Case #:34166038
Case was opened 12/6/2022

PR #19436, opened 2022-11-30
Update chatmessage.md

erichui-ms:


api-reference/v1.0/resources/chatmessage.md
```diff
@@ -48,7 +48,7 @@ Represents an individual chat message within a [channel](channel.md) or [chat](c
 
 | Property   | Type |Description|
 |:---------------|:--------|:----------|
-|id|String| Read-only. Unique ID of the message.|
+|id|String| Read-only. ID of the message. IDs are unique within a chat/channel, but may conflict with message IDs in other chats/channels. |
```
Lauragra, line 51: For the second part of the sentence, would it make sense to revise as:
... but might be duplicated in other chats/channels.

If we use "conflict", that sounds like something that would need to be resolved somehow?

Also, please apply the same update to the beta version of the topic as well.

erichui-ms:
FYI, here is the pull request for the beta version of the doc: https://github.com/microsoftgraph/microsoft-graph-docs/pull/19550

Lauragra (approved):

PR #19462, opened 2022-12-02
Update message-reply.md

JarbasHorst:
Applied standards to the API topic.


Lauragra:
@JarbasHorst , FYI, just one small fixed, I changed the description for the comment parameter back to two sentences because they are two separate thoughts.

PR #19467, opened 2022-12-02
Added mail-related known issue

lramosvea:
Mail PM confirmed known issue

### DIFF
--- a/concepts/known-issues.md
+++ b/concepts/known-issues.md
@@ -370,6 +370,10 @@ does not become part of the body of the resultant message draft.
 
 In both the v1.0 and beta endpoints, the response to `GET /users/id/messages` includes the user's Microsoft Teams chats that occurred outside the scope of a team or channel. These chat messages have "IM" as their subject.
 
+### Change notifications missing in `/delta` calls to messages API using immutable IDs
+
+When you make `/delta` calls to the messages API using immutable IDs, you might miss some change notifications in some cases (e.g. when a message moves out of a folder and is then moved back in).
+
 ## Reports
 
 ### License check errors for Azure AD activity reports


concepts/known-issues.md
```diff
@@ -370,6 +370,10 @@ does not become part of the body of the resultant message draft.
 
 In both the v1.0 and beta endpoints, the response to `GET /users/id/messages` includes the user's Microsoft Teams chats that occurred outside the scope of a team or channel. These chat messages have "IM" as their subject.
 
+### Change notifications missing in `/delta` calls to messages API using immutable IDs
+
+When you make `/delta` calls to the messages API using immutable IDs, you might miss some change notifications in some cases (e.g. when a message moves out of a folder and is then moved back in).
```
Lauragra, line 375: Add a link to the API topic - I assume this one? https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-beta&tabs=http

Suggested revision:
When you call the [Get message delta](https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-beta&tabs=http) API using [immutable IDs](https://learn.microsoft.com/en-us/graph/outlook-immutable-id), you might miss some change notifications; for example, the API does not return messages that were moved out and then moved back in to a folder.

Does this also warrant a mention in the delta method topic? Is there a plan to fix this issue in the future?

lramosvea: @abheek-das will you please take a look at this? It is about the known issue we discussed last month and you confirmed it is indeed a known issue. Does the revision suggested by Laura look good to you (see below)?

When you call the [Get message delta](https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-beta&tabs=http) API using [immutable IDs](https://learn.microsoft.com/en-us/graph/outlook-immutable-id), you might miss some change notifications; for example, the API does not return messages that were moved out and then moved back in to a folder.

lramosvea:
@abheek-das friendly reminder:
Will you please take a look at this? It is about the mail-related issue we discussed a couple of months ago and you confirmed it is indeed a known issue. Does the revision suggested by Laura look good to you (see below)?

When you call the [Get message delta](https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-beta&tabs=http) API using [immutable IDs](https://learn.microsoft.com/en-us/graph/outlook-immutable-id), you might miss some change notifications; for example, the API does not return messages that were moved out and then moved back in to a folder.

PR #19509, opened 2022-12-06
Add limitation notes for create shared channel

devjha-ms:
https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2755442

Specifying more than one member while creating the shared channel is not allowed. The request will fail with similar error as below when more than one member is specified:

"message": "{"errors":[{"message":"In App-context, the specified Users in the request must be a team owner.","errorCode":"Unknown"}],"operationId":"db7fba4e777f7a499015cb5c464afe22"}",


api-reference/beta/api/channel-post.md
```diff
@@ -14,7 +14,9 @@ Namespace: microsoft.graph
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
 Create a new [channel](../resources/channel.md) in a team, as specified in the request body.
-> **Note:** You can add a maximum of 200 members when you create a private channel.
+> **Note**: You can add a maximum of 200 members when you create a private channel.
+
+> **Note**: Specifying more than one member while creating the shared channel is not supported.
```
Danielabom, line 19: @devjha-ms ,  there isn't technically a rule against having multiple notes in a section, but we would want to avoid that because having two notes like that will dilute the stand-out factor of the notes.
I have a couple of options for you: 
We could combine them in one single note.
Move the notes to different locations so they're separate. 
Convert one of the notes to normal text. 
Also, is the second note a Known issue or just the default behavior? We could turn it into the standard text if it's the default behavior.

api-reference/v1.0/api/channel-post.md
```diff
@@ -12,7 +12,8 @@ doc_type: apiPageType
 Namespace: microsoft.graph
 
 Create a new [channel](../resources/channel.md) in a team, as specified in the request body.
-> **Note:** You can add a maximum of 200 members when you create a private channel.
+> **Note**: You can add a maximum of 200 members when you create a private channel.
+> **Note**: Specifying more than one member while creating the shared channel is not supported.
```
Danielabom, line 16: @devjha-ms , there isn't technically a rule against having multiple notes in a section, but we would want to avoid that because having two notes like that will dilute the stand-out factor of the notes.
I have a couple of options for you:
We could combine them in one single note.
Move the notes to different locations so they're separate.
Convert one of the notes to normal text.
Also, is the second note a Known issue or just the default behavior? We could turn it into the standard text if it's the default behavior

Danielabom (changes requested):
@devjha-ms. Thank you for working on this PR. I have some comments and questions. Let me know what you think and I'll be happy to make the edits myself.

api-reference/beta/api/channel-post.md
```diff
@@ -14,7 +14,7 @@ Namespace: microsoft.graph
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
 Create a new [channel](../resources/channel.md) in a team, as specified in the request body.
-> **Note:** You can add a maximum of 200 members when you create a private channel.
+> **Note**: While creating a channel, only one members can be specified for shared channel however maximum of 200 members can be specified for private channel.
```
Danielabom, line 17:
```suggestion
> **Note**: While creating a channel, you can only add one member per shared channel; however, you can add up to 200 members per private channel.
```

api-reference/v1.0/api/channel-post.md
```diff
@@ -12,7 +12,7 @@ doc_type: apiPageType
 Namespace: microsoft.graph
 
 Create a new [channel](../resources/channel.md) in a team, as specified in the request body.
-> **Note:** You can add a maximum of 200 members when you create a private channel.
+> **Note**: While creating a channel, only one members can be specified for shared channel however maximum of 200 members can be specified for private channel.
```
Danielabom, line 15:
```suggestion
> **Note**: While creating a channel, you can only add one member per shared channel; however, you can add up to 200 members per private channel.
```

Danielabom (approved):
@devjha-ms, thank you for working on this PR. I made some small changes. The content looks good. Changes approved.

PR #19560, opened 2022-12-09
Add AzureAD Admin role information to permissions section

andyrobbins:
A service principal with the Organization.ReadWrite.All MS Graph app role can POST to this endpoint. Without that app role, the only AzureAD Admin role that is able to POST to this endpoint is Global Administrator.


FaithOmbongi:
Thanks @andyrobbins for highlighting this piece of missing information. In the spirit of encouraging least privilege access, allow me to first confirm whether this API supports a lesser privileged Azure AD admin role than the Global Admin. I'll get back to you and approve the PR then.

FaithOmbongi (approved):
Thank you @andyrobbins for raising this PR. I've confirmed that only the Global Admin can do POST/DELETE operations. I've approved this PR and it'll be merged soon.

PR #19578, opened 2022-12-12
Update search-concept-interleaving.md

yiwenwang-ms:
Fix the request


njerigrevious (approved):
Thanks! This looks great to me. Let's separately make sure that this is reflected in Graph Explorer sample.

isvargasmsft (approved):

PR #19595, opened 2022-12-13
Correction in update to-do list example and its description

lalamkumar:
**Correction in update to-do list example and its description**

As you can see the below documentation references on **p tag or (to-do item), **replace** action do not support **data-id** as **target** instead **generated id** must be used for updating p element or to-do item in a to-do list.** Hence target value in the example and example description needs correction.

https://learn.microsoft.com/en-us/graph/onenote-update-page#supported-elements-and-actions https://learn.microsoft.com/en-us/graph/onenote-update-page#update-a-to-do-item

Thanks


SrinivasaRaoDarna-MSFT (approved):
With reference to this PR https://github.com/microsoftgraph/microsoft-graph-docs/pull/19563

PR #19614, opened 2022-12-14
Moved misplaced SubscribedSkus to proper parent

ioamnesia:
The SubscribedSkus entry is currently a child of Organizational Contacts and should be a direct child of Directory Management. Simple indentation issue.


Lauragra:
@ioamnesia , thanks for your PR. It looks like this issue was already fixed on the live site so I'm going to close this. Thanks!

PR #19639, opened 2022-12-15
Alphabetize 12 15 22

dodaromike:
Alphabetizing API V1.0 for 12-15-2022


api-reference/v1.0/resources/assignedplan.md
```diff
@@ -30,11 +30,11 @@ The following table describes the possible statuses for the **capabilityStatus**
 
 | Member | Description	|
 |:---------------|:--------|
```
angelgolfer-ms, line 32: This is an example of an enum. Please don't reorder enum tables. Even though we want developer customers to refer to enum members by their string names, each of the members is mapped to a numerical value in ascending order. So please don't reorder the members in these tables.

dodaromike: Yes, you mentioned that and I have been leaving the enums alone.

dodaromike: On second look, there might be a few |Member | Description | items that I changed.  I'll revert them back to what they were.

api-reference/v1.0/resources/calendarpermission.md
```diff
@@ -44,14 +44,14 @@ Once a sharee or delegate has been set up for a calendar, you can [update](../ap
 
 | Member        | Description |
```
angelgolfer-ms, line 45: This is another example of an enum table. Please don't reorder enum tables.

api-reference/v1.0/resources/channel.md
```diff
@@ -56,24 +56,24 @@ where files are shared, and where tabs are added.
...
 ### channelMembershipType values
 
 | Member             | Description                                                                       |
 |:-------------------|:----------------------------------------------------------------------------------|
-| standard           | Channel inherits the list of members of the parent team.                          |
 | private            | Channel can have members that are a subset of all the members on the parent team. |
-| unknownFutureValue | Evolvable enumeration sentinel value. Do not use.                                 |
 | shared             | Members can be directly added to the channel without adding them to the team.     |
+| standard           | Channel inherits the list of members of the parent team.                          |
+| unknownFutureValue | Evolvable enumeration sentinel value. Do not use.                                 |
```
angelgolfer-ms, line 77: Yes, this is another example of an enum table. Please don't reorder.
In particular, newer enums that define "unknownFutureValue" are called evolvable enums. The location of  "unknownFutureValue" relative to the rest of the enum members is important, we can't move it anywhere else. This is explained here: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211749/Evolvable-enumerations

api-reference/v1.0/resources/chat.md
```diff
@@ -69,9 +69,9 @@ A chat is a collection of [chatMessages](chatmessage.md) between one or more par
```
angelgolfer-ms, line 69: This is another example of an enum table.  Please exclude enum tables from reordering please.

angelgolfer-ms (changes requested):
Per our chat on Teams, please do not reorder enum members in their tables.
Methods also follow a different order than plain alphabetical order. I would suggest leaving that from the main task. Further reasons in our Teams chat.
The ordering of the properties and nav properties look good. Caveat: I only browsed the diff, and not the staged topics, so if there's any oversight such as missing a nav properties table, I wouldn't have picked that out and would love to leave that to your eye for detail :-)

api-reference/v1.0/resources/applemanagedidentityprovider.md
```diff
@@ -22,12 +22,13 @@ Inherits from [identityProviderBase](../resources/identityproviderbase.md).
...
-|[Get](../api/identityproviderbase-get.md) |[appleManagedIdentityProvider](../resources/applemanagedidentityprovider.md) |Retrieve properties of the Apple identity provider configuration.|
-|[Update](../api/identityproviderbase-update.md)|None|Update the Apple identity provider configuration.|
 |[Delete](../api/identityproviderbase-delete.md)|None|Delete the Apple identity provider configuration.|
+|[Get](../api/identityproviderbase-get.md) |[appleManagedIdentityProvider](../resources/applemanagedidentityprovider.md) |Retrieve properties of the Apple identity provider configuration.|
+|[List](../api/identitycontainer-list-identityproviders.md)|[identityProviderBase](../resources/identityproviderbase.md) collection|Retrieve all identity providers configured in a tenant, including the Apple identity providers. There is no way to retrieve only the Apple identity providers in a tenant.|
 |[List available provider types](../api/identityproviderbase-availableprovidertypes.md)|String collection|Retrieve all available identity provider types available in the tenant.|
+|[Update](../api/identityproviderbase-update.md)|None|Update the Apple identity provider configuration.|
+
```
angelgolfer-ms, line 31: This is an example of a Methods table that you have updated. The original order is actually correct per our guidance. Please revert. 

In general:
Methods tables actually take a different order, not as straight forward as just alphabetizing. The order is - list the CRUD operations first, in the order of list, create, get, update, delete, where applicable. Then list any other actions or functions alphabetically. You can find this guidance in the CDK here: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211766/ToC-and-topic-title.

angelgolfer-ms: I'd strongly recommend leaving the Methods tables alone for this pass, so that completing the reordering of properties and nav properties in complex type and entity topics is attainable within a reasonable time. The scope creep would extend this project; there could be other tasks that our manager might prioritize higher. As described in our Teams chat, the ordering of methods (or lack of a consistent order) doesn't affect customers as much as for properties and nav properties.

angelgolfer-ms: Thanks for reverting.

dodaromike: Reverted the methods table.

Danielabom (approved):
No enums or methods were reordered. 
Alphabetial order of the properties and nav properties: OK
Tables' format: OK
PR approved.

PR #19660, opened 2022-12-19
Update group-list-members.md to add Example 6.

cjvandyk:
Example 6 shows how to use pagination to get all members in cases where the membership exceeds 100.


api-reference/v1.0/api/group-list-members.md
```diff
@@ -351,3 +351,28 @@ Content-type: application/json
...
   ]
 }-->
+
+### Example 6: Using pagination to get all Members when there are more than 100.
+
+#### Request
+
+The following is an example of the request.
+
+# [C#](#tab/csharp)
```
Lauragra, line 361: We only include the HTTP requests in the content; the language snippets are generated via our SDK snippet process. I'm going to close this PR. If you want to add this example, please add the HTTP request and response. Thanks.

PR #19682, opened 2022-12-21
Minor changes for sitePage related API

sangle7:
1.  add a new example in 'sitepage-get,md'
2.  add some notes in 'sitepage-create.md'
3. (newly updated) append changelogs and fix invalid links in changelog
4. update the ownership of sitePages related API

### DIFF
--- a/api-reference/beta/api/sitepage-create.md
+++ b/api-reference/beta/api/sitepage-create.md
@@ -1,5 +1,5 @@
 ---
-author: rahmit
+author: sangle7
 description: "Create a new sitePage in the site pages list in a site."
 ms.date: 05/07/2018
 title: Create a new page in a SharePoint site
@@ -38,11 +38,13 @@ POST /sites/{site-id}/pages
 
 In the request body, supply a JSON representation of the [sitePage](../resources/sitepage.md) resource to create.
 
+> Note: If you're using the response from [get-sitepage](../api/sitepage-get.md) to create a sitePage, we recommend that you add the HTTP header `Accept: application/json;odata.metadata=none`. This will remove all OData metadata from the response. You can also manually remove all OData metadata.
+
 ## Example
 
 The following example shows how to create a new page.
 
-<!-- { "blockType": "request", "name": "create-page", "scopes": "sites.readwrite.all" } -->
+<!-- { "blockType": "request", "name": "create_sitepage" } -->
 
 ```http
 POST /sites/{site-id}/pages
--- a/api-reference/beta/api/sitepage-create.md
+++ b/api-reference/beta/api/sitepage-create.md
@@ -38,7 +38,7 @@ POST /sites/{site-id}/pages
 
 In the request body, supply a JSON representation of the [sitePage](../resources/sitepage.md) resource to create.
 
-> Note: If you're using the response from [get-sitepage](../api/sitepage-get.md) to create a sitePage, we recommend that you add the HTTP header `Accept: application/json;odata.metadata=none`. This will remove all OData metadata from the response. You can also manually remove all OData metadata.
+> **Note:** If you're using the response from the [Get sitepage](../api/sitepage-get.md) operation to create a **sitePage**, we recommend that you add the HTTP header `Accept: application/json;odata.metadata=none`. This will remove all OData metadata from the response. You can also manually remove all OData metadata.
 
 ## Example
 
--- a/api-reference/beta/api/sitepage-get.md
+++ b/api-reference/beta/api/sitepage-get.md
@@ -1,5 +1,5 @@
 ---
-author: rahmit
+author: sangle7
 description: "Returns the metadata for a sitePage in the site pages list in a site."
 ms.date: 03/15/2018
 title: Get a page in a site
@@ -232,6 +232,15 @@ Content-type: application/json
     }]
 }
 ```
+
+### Example 3: Get a page with page content
+
+To access the page with page content, append `?expand=canvasLayout` to query string:
+
+```http
+GET /sites/{site-id}/pages/{page-id}?expand=canvasLayout
+```
+
 <!--
 {
   "type": "#page.annotation",
--- a/api-reference/beta/api/sitepage-get.md
+++ b/api-reference/beta/api/sitepage-get.md
@@ -124,6 +124,7 @@ Content-type: application/json
 ```
 
 ### Example 2: Get a page using select and expand
+
 #### Request
 
 With `select` and `expand` statements, you can retrieve sitePage metadata and page content in a single request.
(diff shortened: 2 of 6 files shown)


dodaromike (commented):
Looks good. One question about this sentence: "It can help to remove all OData metadata from the response."  I think you mean that "It will remove all OData metadata... " The way it stands now leaves some doubt about the removal.

api-reference/beta/api/sitepage-create.md
```diff
@@ -38,11 +38,13 @@ POST /sites/{site-id}/pages
 
 In the request body, supply a JSON representation of the [sitePage](../resources/sitepage.md) resource to create.
 
+Note: If the [sitePage](../resources/sitepage.md) you used here is retrieved from the [get-sitepage](../api/sitepage-get.md) API, we recommend to add an http header `Accept: application/json;odata.metadata=none` when calling the [get-sitepage](../api/sitepage-get.md) method. It will remove all OData metadata from the response. Otherwise, it would lead to an OData metadata mismatch issue when [creating a sitepage](../api/sitepage-create.md). You can also manually remove all OData metadata.
```
Lauragra, line 41: Please format this text as a Note. Also, please clarify, are they using the GET operation after they create the site page? If so, it would be more clear to revise as:

>**Note:** If you use the [Get sitePage](../api/sitepage-get.md) method to retrieve the site page, we recommend that you add the HTTP header `Accept: application/json;odata.metadata=none`. This will remove all OData metadata from the response. You can also manually remove all OData metadata.

The part about the OData mismatch is not clear to me because it's a recursive link to the existing topic. Can you please clarify whether your intent is that they call the get operation before the create operation, or after?

api-reference/beta/api/sitepage-get.md
```diff
@@ -30,12 +30,27 @@ One of the following permissions is required to call this API. To learn more, in
 | Delegated (personal Microsoft account) | Not supported.                              |
 | Application                            | Sites.Read.All, Sites.ReadWrite.All         |
 
-## HTTP request
```
Lauragra, line 33: Please don't remove or replace the standard H2 sections. We strive to keep the topic structure consistent across topics to simplify the user experience. Please preserve the HTTP request H2 heading and the standard first syntax. For the second syntax, please add that to the Examples section as a second example.

api-reference/beta/api/sitepage-get.md
```diff
@@ -30,12 +30,27 @@ One of the following permissions is required to call this API. To learn more, in
...
 
+## Get a page with page content
+
+To access the page with page content, append `?expand=canvasLayout` to query string:
+
+```http
+GET /sites/{site-id}/pages/{page-id}?expand=canvasLayout
+```
+
+> Note: If you're willing to use the response to [create a sitePage](../api/sitepage-create.md), we recommend to add an http header `Accept: application/json;odata.metadata=none` to remove all OData metadata from the response.
```
Lauragra, line 51: Why do we use the word "willing" here? Suggested revision:

>**Note:** If you're using the response to [create a sitePage](../api/sitepage-create.md), we recommend that you add the HTTP header `Accept: application/json;odata.metadata=none` to remove all OData metadata from the response.

changelog/Microsoft.FileServices.json
```diff
@@ -3016,6 +3016,32 @@
...
       "SubArea": ""
+    },
+    {
+      "ChangeList": [
+        {
+          "Id": "7626dec5-d667-45f6-9986-ee9b8aee2278",
+          "ApiChange": "Member",
+          "ChangedApiName": "neutral",
+          "ChangeType": "Addition",
+          "Description": "Added the `neutral` member to the **sectionEmphasisType** enumeration.",
```
Lauragra, line 3027: This changelog entry doesn't make sense, it's adding and removing the same enum member, and this enum isn't part of the updates in this PR. The changes in this PR don't require a changelog entry; unless something is missing, please remove this ChangeList record.

api-reference/beta/api/sitepage-get.md
sangle7: Sounds good to me! Addressed :)

changelog/Microsoft.FileServices.json
sangle7: Removed:) Thanks for pointing it out

api-reference/beta/api/sitepage-get.md
sangle7: Comment addressed

api-reference/beta/api/sitepage-create.md
sangle7: I've updated ths string with `> Note: If you're using the response from [get-sitepage](../api/sitepage-get.md) to create a sitePage, we recommend that you add the HTTP header `Accept: application/json;odata.metadata=none`. This will remove all OData metadata from the response. You can also manually remove all OData metadata.
`

api-reference/beta/api/sitepage-get.md
```diff
@@ -232,6 +232,15 @@ Content-type: application/json
     }]
 }
 ```
+
+### Example 3: Get a page with page content
+
+To access the page with page content, append the `?expand=canvasLayout` query string.
+
+```http
+GET /sites/{site-id}/pages/{page-id}?expand=canvasLayout
```
Lauragra, line 241: This example request is missing a response example.

api-reference/beta/api/sitepage-get.md
```diff
@@ -232,6 +232,15 @@ Content-type: application/json
     }]
 }
 ```
+
+### Example 3: Get a page with page content
+
+To access the page with page content, append the `?expand=canvasLayout` query string.
+
```
Lauragra, line 239: Add an H4 Request subheading here, and then add an H4 Response and include the response example. Also as an aside, I notice that this method topic doesn't conform to the prescribed structure. Please see this topic: [Get Virtual appointment join link](../api/virtualappointment-getvirtualappointmentjoinweburl.md). If we can't update it in this PR, we should have a follow-up PR to bring this topic into compliance.

sangle7: Thanks Lauragra. I didn't get it clear what do you mean by 'method topic'? Are you suggestion changing the page title from '# Get a page in the site pages list of a site' to '# Get SitePage'?

Lauragra: Yes, that change would be consistent with our topic title guidelines. Also, because this is a method (operation) topic, it needs to follow the prescribed structure. It is missing H2 sections for Request body and Response. You can get more details in the CDK topic I linked to.

Lauragra (approved):

PR #19723, opened 2022-12-27
[Doc Improvement][add-codesamples-teamsapp]

v-shalinirobinson:
w.r.t Issue #13065

### DIFF
--- a/api-reference/beta/resources/teamsapp.md
+++ b/api-reference/beta/resources/teamsapp.md
@@ -67,17 +67,13 @@ Users can see these apps in the Microsoft Teams Store, and these apps can be ins
 }
 ```
 
-## Code samples
-
-| Sample name | Description | C# | Node.js |
-| --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


dodaromike (approved):
Thanks for samples.

api-reference/beta/resources/teamsapp.md
```diff
@@ -67,6 +67,12 @@ Users can see these apps in the Microsoft Teams Store, and these apps can be ins
 }
 ```
 
+## Code samples
```
Lauragra, line 70: @v-shalinirobinson , @dodaromike , this is a non-standard H2 heading. Our reference topics, although hand-authored, adhere to a standard H2 structure. In this case, the resource topics include the optional See also section, so we can add a link to the code sample there. Note also that the use of View for the link text is not accessible. For users who use a screen reader, the link text should match or be similar to the target link. Multiple links with the same text make it difficult for sight-impaired users to navigate the links.

See this topic for details about our reference topic structure: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/344038/Reference-topic-templates

PR #19737, opened 2022-12-29
Add note regarding permissions

psaffaie:
When customer make a call to modify group member, they also need the read and write permission to the member object. If it's adding a user then the app needs User.ReadWrite.All, if it's adding a device, the app needs Device.ReadWrite.All. Same as service principals. This part is not stated clearly in the doc so partner team has difficulties in granting those high privileged permission for their first party apps as these requirements for users, devices, service principals are not stated in the doc.


api-reference/v1.0/api/group-post-members.md
```diff
@@ -27,6 +27,7 @@ One of the following permissions is required to call this API. To learn more, in
 
 > [!IMPORTANT]
 > To add members to a role-assignable group, the calling user must also be assigned the _RoleManagement.ReadWrite.Directory_ permission.
+> When an application queries a relationship that returns a directoryObject type collection, if it does not have permission to read or write to the directory object (i.e. users). With this behaviour applications can request the least privileged permissions they need (i.e. User.ReadWrite.All) to add members to a group, rather than rely on the set of Directory.* permissions. For details, see Limited information returned for inaccessible member objects.
```
isvargasmsft, line 30: Please review the first two sentences, it is not clear what is the behavior we are referring to. In addition, in the last sentence, please add a link to the topic you are mentioning.

psaffaie: Here is the link to the topic mentioned: https://learn.microsoft.com/en-us/graph/permissions-reference#limited-information-returned-for-inaccessible-member-objects

I am attempting to explain that a ReadWrite directory permission is not required if the user/app has ReadWrite permissions for the specific directory object they need to access i.e. a user object or device object

psaffaie (commented):
https://learn.microsoft.com/en-us/graph/permissions-reference#limited-information-returned-for-inaccessible-member-objects

psaffaie:
@FaithOmbongi  hoping you can help me understand if there is anything I am missing for this PR for Issac?

FaithOmbongi (approved):
Hi @psaffaie - Please apply these changes to the beta doc as well.

FaithOmbongi:
Hi @Lauragra - This is a case of a shared resource type. We've switched to using the four-column permissions table which seems to be the most suitable for now.

api-reference/v1.0/api/group-post-members.md
```diff
@@ -17,13 +17,15 @@ Add a member to a security or Microsoft 365 group through the **members** naviga
...
 ## Permissions
 
-One of the following permissions is required to call this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
-
-| Permission type                        | Permissions (from least to most privileged)                                |
-| :------------------------------------- | :------------------------------------------------------------------------- |
-| Delegated (work or school account)     | GroupMember.ReadWrite.All, Group.ReadWrite.All, Directory.ReadWrite.All    |
-| Delegated (personal Microsoft account) | Not supported.                                                             |
-| Application                            | GroupMember.ReadWrite.All, Group.ReadWrite.All and Directory.ReadWrite.All |
+The following table shows the least privileged permission that's required by each resource type when calling this API. To learn more, including how to choose permissions, see [Permissions](/graph/permissions-reference).
```
Lauragra, line 20: @FaithOmbongi , should we make the same change to the topic in beta before we merge this PR?

FaithOmbongi: @Lauragra - Yes, the author did via PR #20080.

PR #19738, opened 2022-12-30
identity governance: hide PSh in beta that is not correctly generated

markwahl-msft:
It appears the stub generator for PSh is leaving out some of the payload of requests.  For example in entitlementmanagement-post-accesspackageassignmentpolicies.md example 3, the multiple choice question within an access package assignment policy has several values in HTTP and C#, but in PSh, it's just shown as an empty array.

```powershell
...
	Questions = @(
		@{
...
			"@odata.type" = "#microsoft.graph.accessPackageMultipleChoiceQuestion"
			Choices = @(
			)
...
```
I noticed in one of the v1.0 APIs, the PSh stub had left out a number value.

Replace those with 'snippet not available'.
Also some previous PR had a spurious code comment in a HTTP payload, removing that.


FaithOmbongi (approved):
Hi @markwahl-msft - I understand the need for this but I wonder if this will be honored by the snippets generator. Let me approve and we'll see if this doesn't get overwritten in the next generation.

cc @msewaweru for visibility.

markwahl-msft:
#sign-off

PR #19808, opened 2023-01-09
Content freshness fixes

FaithOmbongi:
Fix the previous date format to mm/dd/yy
The last refresh date is when the snippets got generated. 

-------
cc: @FaithOmbongi

### DIFF
--- a/concepts/how-to-pim-update-rules.md
+++ b/concepts/how-to-pim-update-rules.md
@@ -5,7 +5,7 @@ author: "FaithOmbongi"
 ms.localizationpriority: medium
 ms.topic: how-to
 ms.prod: "governance"
-ms.date: 08/12/2022
+ms.date: 01/05/2022
 ---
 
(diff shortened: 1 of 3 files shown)


concepts/how-to-pim-update-rules.md
```diff
@@ -5,7 +5,7 @@ author: "FaithOmbongi"
 ms.localizationpriority: medium
 ms.topic: how-to
 ms.prod: "governance"
-ms.date: 08/12/2022
+ms.date: 01/05/2022
```
Lauragra, line 8: @FaithOmbongi , do you plan to maintain these dates? I general, in Microsoft Graph content, we rely on the date of the last commit to show on the article rather than setting the date manually, so we don't have to maintain this.

FaithOmbongi: Hi @Lauragra - Yes, I do. Celeste has asked me to add them for Identity conceptual articles so the Skilling org can track content freshness for Graph (Identity) articles as well.

PR #19825, opened 2023-01-10
Francismbui/default usage location v1

franqq:
Add default usage location information to organization resource

### DIFF
--- a/api-reference/v1.0/api/organization-get.md
+++ b/api-reference/v1.0/api/organization-get.md
@@ -120,6 +120,7 @@ Content-type: application/json
     "country": null,
     "countryLetterCode": "NL",
     "createdDateTime": "2021-08-02T10:30:06Z",
+    "defaultUsageLocation": "String",
     "displayName": "Contoso",
     "isMultipleDataLocationsForServicesEnabled": null,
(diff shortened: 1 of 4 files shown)


api-reference/v1.0/api/organization-get.md
```diff
@@ -120,6 +120,7 @@ Content-type: application/json
     "country": null,
     "countryLetterCode": "NL",
     "createdDateTime": "2021-08-02T10:30:06Z",
+    "defaultUsageLocation": "String",
```
FaithOmbongi, line 123: Replace "String" with an actual country code  the API may return

api-reference/v1.0/api/organization-list.md
```diff
@@ -108,6 +108,7 @@ Content-type: application/json
             "country": null,
             "countryLetterCode": "NL",
             "createdDateTime": "2021-08-02T10:30:06Z",
+            "defaultUsageLocation": "String",
```
FaithOmbongi, line 111: Replace "String" with an actual country code  the API may return

FaithOmbongi (approved):
Thank you @franqq for your work on this PR. Approved with suggestions.

PR #19849, opened 2023-01-11
2023 01 11 tasks app only auth updates

TarkanSevilmis:
Adding app-only permissions for Planner

### DIFF
--- a/changelog/Microsoft.Office.Tasks.json
+++ b/changelog/Microsoft.Office.Tasks.json
@@ -1335,7 +1335,7 @@
           "ApiChange": "Permission",
           "ChangedApiName": "Tasks.Read.All",
           "ChangeType": "Addition",
-          "Description": "Added support for application permission of `Tasks.Read.All` to the read operations of [plannerPlan](https://learn.microsoft.com/en-us/graph/api/resources/plannerplan?view=graph-rest-beta) resource.",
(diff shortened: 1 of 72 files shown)


changelog/Microsoft.Office.Tasks.json
```diff
@@ -1327,6 +1327,290 @@
...
       "SubArea": ""
+    },
+    {
+      "ChangeList": [
+        {
+          "Id": "c31056e4-ae34-4a4e-acd3-d660d4a5266c",
+          "ApiChange": "Permission",
+          "ChangedApiName": "Tasks.Read.All",
+          "ChangeType": "Addition",
+          "Description": "Added support for application permission of `Tasks.Read.All` to the read operations of all Planner resources.",
```
Lauragra, line 1338: Global: The descriptions all appear to be identical; this will end up looking like a repetitive list in the table on the Changelog. The Target resource name doesn't show in the table. I recommend that we update each entry to be specific; so, instead of "all Planner resources", this would say "the <plannerPlan> resource" (and link to the resource topic).

angelgolfer-ms: Yes, that's right. That would be similar to example 14 in the CDK [changelog](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog?anchor=changelist-record-examples) topic.

Lauragra (approved):

PR #19858, opened 2023-01-12
Add note to clarify how Planner relates to business scenario API.

angelgolfer-ms:


TarkanSevilmis (commented):
These changes look good to me

PR #19880, opened 2023-01-13
updated gcc docs for todo

ashokobhanmicrosoft:
We are releasing To Do Graph APIs for GCC H an DoD customers. As a result we have made the below changes to the documentation

Change 1 
Add Microsoft To Do to the list of supported features on Graph
Link - https://learn.microsoft.com/en-us/graph/deployments

Change 2
The Webhook "A todoTask ofa Microsoft To Do" is only available for WW. Hence, we need to mention that this webhook is only available in WW and none of the GCC environments.


Lauragra:
@ashokobhanmicrosoft , just a quick mention/reminder to set the **ready for content review** label on your PR when you are ready for a reviewer to take a look. Thanks.

concepts/deployments.md
```diff
@@ -87,6 +87,7 @@ The following Microsoft Graph features are generally available on the `/v1.0` en
 | SharePoint                        | ✔                                | ✔                                          |
 | Teams                             | ✔                                | ✔                                          |
 | Users                             | ✔                                | ✔                                          |
+| Microsoft To Do                   | ✔                                | ➖                                         |
```
Lauragra, line 90: Unless branding requires it, for consistency, please remove "Microsoft" from the product name and list To Do in alphabetical order, so following the entry for Teams. In general, we do precede product names with Microsoft on first mention but in this case, the context is clear and not doing so removes unnecessary repetition and also makes the feature list more scannable.

concepts/includes/change-notifications-supported-resources-flat.md
```diff
@@ -27,7 +27,7 @@ Currently, subscriptions are enabled for the following resources.
 - The [presence][] of a user in Microsoft Teams.
 - A [team][team] in Microsoft Teams.
 - A [printer][] (when a print job for the printer gets to JobFetchable state - ready to be fetched for printing) and a [printTaskDefinition][] in Universal Print. For more information, see [Subscribe to change notifications from cloud printing APIs](/graph/universal-print-webhook-notifications).
-- A [todoTask][] of a user in Microsoft To Do.
+- A [todoTask][] of a user in Microsoft To Do (webhook only available in WW and none of the GCC environments).
```
Lauragra, line 30: Please revise the parenthetical: "(webhooks are only available in the worldwide endpoint and no other national clouds)". You might want to add this mention in the subscription resource topic as well: https://learn.microsoft.com/en-us/graph/api/resources/subscription?view=graph-rest-1.0.

Are there other implementation differences between the worldwide and national clouds? If so, we have recently established a pattern for documenting workload-level implementation differences in national clouds; we can discuss further.

Lauragra (approved):

PR #19904, opened 2023-01-17
[Doc Improvement][Update List and Publish teamsApp][2804432]

v-ypalikila:
Updated documentation for both for beta and v1, to be aligned with the schema mentioned [teamsApp resource](https://learn.microsoft.com/en-us/graph/api/resources/teamsapp?view=graph-rest-beta).
 

Removed version property from JSON response from the files to correct the content in the articles as per the schema as suggested in the ADO: [Bug 2804432](https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2804432).


Danielabom:
Reached out to the author to get info for Mike Dodaro.

dodaromike:
@Danielabom I can't see the schema, but assuming the version property is not in the TeamsApp entity, this PR is ready to merge.

api-reference/beta/api/appcatalogs-list-teamsapps.md
```diff
@@ -82,34 +82,37 @@ GET https://graph.microsoft.com/beta/appCatalogs/teamsApps?$filter=distributionM
 ```
 
 # [C#](#tab/csharp)
+
```
angelgolfer-ms, line 85: @v-ypalikila , FYI, there isn't a purpose of inserting blank lines like line 90, 95, and so on. 
The lines like 74, and 84-114, are all inserted by an internal tool that runs weekly to add or update these code snippets where necessary. 
Please take a look at this topic in the onboarding docs (CDK): https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211752/Generate-code-snippets

api-reference/beta/api/appcatalogs-list-teamsapps.md
```diff
@@ -206,11 +214,12 @@ Content-Type: application/json
       "id": "b1c5353a-7aca-41b3-830f-27d5218fe0e5",
       "externalId": "f31b1263-ba99-435a-a679-911d24850d7c",
       "displayName": "Test App",
-      "distributionMethod": "organization"
+      "distributionMethod": "Organization"
```
angelgolfer-ms, line 217: Line 217:
@v-ypalikila , what is the purpose of this change?
If you look at the beta metadata, the distribution property is of type teamsAppDistributionMethod, which is an enum appropriately defined with values using camel case. 
As a matter of fact, all API terms in Microsoft Graph, including names of resources, methods, properties, enumerations, and enum values, should all use camel case.

So this change is inappropriate and may confuse users, even if beneath the covers, Microsoft Graph can handle this type of inconsistencies. On the other hand, docs should model the correct use.

angelgolfer-ms (changes requested):
Hi @v-ypalikila , can you please take a look at my comments?
Seems to me the updates to teamsapp-publish.md are necessary, but not those for the list operation. 
Please clarify if there was something else you meant to fix.
Thanks.

api-reference/beta/api/appcatalogs-list-teamsapps.md
v-ypalikila: Removed the capital case for the distribution property. I will ensure to check and not submit such updates.

dodaromike:
It looks like @v-ypalikila has addressed changes requested by @angelgolfer-ms .  I will approve this PR.

dodaromike:
@v-ypalikila @angelgolfer-ms I will try to address the warnings before approving this PR.  The warnings are coming from main.

dodaromike (approved):

api-reference/beta/api/teamsapp-publish.md
```diff
@@ -87,6 +87,10 @@ Content-type: application/zip
 [Zip file containing a Teams app package]
 ```
 
+# [CSharp](#tab/csharp)
```
Lauragra, line 90: Why are you adding this tab and snippets? The SDK snippets are generated via the weekly snippet automation; we don't add the snippets to the topics manually. Please remove the manual snippets.

Lauragra (approved):

PR #19926, opened 2023-01-18
Added coorganizer meeting role to v1.0

frankpeng7:


api-reference/v1.0/resources/meetingparticipantinfo.md
```diff
@@ -19,7 +19,19 @@ Information about a participant in a meeting.
 | :------- | :---------------------------- | :---------------------------------------------------------------------------------- |
 | identity | [identitySet](identityset.md) | Identity information of the participant.                                            |
 | role     | onlineMeetingRole             | Specifies the participant's role in the meeting.  Possible values are `attendee`, `presenter`, `producer`, and `unknownFutureValue`.|
-| upn      | String                        | User principal name of the participant.                                             |
+| role     | [onlineMeetingRole](#onlinemeetingrole-values)     | Specifies the participant's role in the meeting.|
```
Lauragra, line 22: We have the role property listed twice in this table. I assume we should remove the first instance and keep the second one with the link to the values? Was the deletion of the upn property intentional, or did you mean to delete the role row? JFYI, we should list the properties in the Properties tables in alphabetical order.

frankpeng7: Nice catch! I've corrected this

Lauragra (approved):

PR #19988, opened 2023-01-24
Fix docs out of sync with schema - organization entity

FaithOmbongi:
Added the tenantType property to both v1.0 and beta


mmcadamsms (approved):
Seems appropriate to me.

PR #20084, opened 2023-02-01
Adding resources and attributes documentation to Graph 1.0

saylee-dharne-msft:


api-reference/v1.0/api/accesspackage-delete-resourcerolescopes.md
```diff
@@ -0,0 +1,74 @@
+---
+title: "Delete accessPackageResourceRoleScope"
```
saylee-dharne-msft, line 2: Is this operation allowed?

PR #20089, opened 2023-02-01
Create include file - access reviews filter by scope

FaithOmbongi:


api-reference/includes/accessreviews-definition-filter-scope.md
```diff
@@ -0,0 +1,26 @@
...
+
+The value of `{object}` can be one of the following:
+
+|Value|Description|
+|:---     |:---       |
+|`/groups`  |List every accessReviewScheduleDefinition on individual groups (excludes definitions scoped to all Microsoft 365 groups with guest users).|
+|`/groups/{group id}`  |List every accessReviewScheduleDefinition on a specific group (excludes definitions scoped to all Microsoft 365 groups with guest users).|
+|`./members`  |List every accessReviewScheduleDefinition scoped to all Microsoft 365 groups with guest users.|
+|`accessPackageAssignments`  |List every accessReviewScheduleDefinition on an access package.|
+|`roleAssignmentScheduleInstances`  |List every accessReviewScheduleDefinition for service principals assigned to a privileged role.|
```
isabelleatmsft, line 24: I think this description is incorrect. From levi's message to the customer, 'roleAssignmentScheduleInstances' returns every review for a privileged role, not just those scoped to service principals. @leherpel could you confirm whether this is true?

api-reference/includes/accessreviews-definition-filter-scope.md
```diff
@@ -0,0 +1,26 @@
...
+```http
+/identityGovernance/accessReviews/definitions?$filter=contains(scope/microsoft.graph.accessReviewQueryScope/query, '{object}')
+```
+
+The value of `{object}` can be one of the following:
+
+|Value|Description|
+|:---     |:---       |
+|`/groups`  |List every accessReviewScheduleDefinition on individual groups (excludes definitions scoped to all Microsoft 365 groups with guest users).|
+|`/groups/{group id}`  |List every accessReviewScheduleDefinition on a specific group (excludes definitions scoped to all Microsoft 365 groups with guest users).|
```
isabelleatmsft, line 21: @leherpel  does this syntax apply to all review types. Ex: what about roleassignmentscheduleinstances/roleID ?

api-reference/includes/accessreviews-definition-filter-scope.md
```diff
@@ -0,0 +1,26 @@
...
+
+The `$filter` query parameter with the `contains` operator is supported on the **scope** property of **accessReviewScheduleDefinition**. Use the following format for the request:
+
+```http
+/identityGovernance/accessReviews/definitions?$filter=contains(scope/microsoft.graph.accessReviewQueryScope/query, '{object}')
+```
+
+The value of `{object}` can be one of the following:
+
+|Value|Description|
```
isabelleatmsft, line 18: @leherpel what about filtering on reviews of applications and Azure roles? The Azure roles one is tricky, since these are created using ARM APIs and not graph so we definitely wouldnt support that filter on create/list definition - but does the review history report allow for filtering by this type?

isabelleatmsft (changes requested):
some input on correctness needed from eng

api-reference/includes/accessreviews-definition-filter-scope.md
leherpel: This currently only is supported in group reviews and access package assignment reviews, i.e: /identityGovernance/accessReviews/definitions?$filter=scope/query eq '/beta/identityGovernance/entitlementManagement/accessPackageAssignments?$filter=(accessPackageId eq ''137e09cc-10a5-492d-8c7b-4b1315027704'' and assignmentPolicyId eq ''810490ac-a438-40b4-9114-ef8eeec628b2'' and catalogId eq ''d0f10d34-05c2-4786-af64-c59795a84631'')'

leherpel: However a customer can create a review of this type using our Graph API: https://learn.microsoft.com/en-us/graph/accessreviews-scope-concept

api-reference/includes/accessreviews-definition-filter-scope.md
leherpel: We don't support a scope for applications yet in reports as that's not supported in our Graph API: https://learn.microsoft.com/en-us/graph/accessreviews-scope-concept

We do support an Azure role scope but not through our Graph API a user would have to call our ARM API to create the Azure role review and review history report. However, the objects and properties are the same in our ARM and Graph API, i.e `accessReviewScheduleDefinition`

isabelleatmsft (approved):

PR #20093, opened 2023-02-01

maheshsinghania:
Updating to reflect the capacity enhancement for 30 connections (from 10 earlier) and capping the overall item limit at tenant level to 50 million.


msharshkum (approved):
Please check if 15 for compliance should also change with @KumarSeetharaman74

PR #20097, opened 2023-02-01
preventing misunderstandings

brozbab:
internal federation is for the domains that verified


FaithOmbongi (approved):
Thank you @brozbab for raising this PR to clarify this concept. Hope it isn't too much trouble, but can you apply this same change to the beta doc as well? Thanks.

Regardless, I've approved the PR and it will be merged at the end of the day.

PR #20107, opened 2023-02-01
Updates to MGDC policies

isvargasmsft:
This is a request from Amit to add information about estimating copy time for MGDC.

### DIFF
--- a/concepts/data-connect-policies.md
+++ b/concepts/data-connect-policies.md
@@ -41,6 +41,19 @@ There is no charge for extraction of objects from the following datasets:
 - BasicDataSet_v0.GroupMembers
 - BasicDataSet_v0.GroupOwners
 
+## Estimating copy speed
+Extracting all the data from your Microsoft 365 tenant to Azure can take time. The total time a copy takes depends on the number of records to be extracted. This number can vary between tenants of the same data range. If the number of records is known, it is possible to estimate the time it will take to extract most of them. The total time for one extraction job can still be higher as even if one user is not extracted, the extraction will continue until all data is extracted or the deadline of 24 hours is met. 
+
+Based on some assumptions, the following example represents one possible calculation.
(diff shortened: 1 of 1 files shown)


concepts/data-connect-policies.md
```diff
@@ -41,6 +41,19 @@ There is no charge for extraction of objects from the following datasets:
 - BasicDataSet_v0.GroupMembers
 - BasicDataSet_v0.GroupOwners
 
+## Estimating copy speed
+Extracting all the data from your Microsoft 365 tenant to Azure can take time. The total time a copy takes depends on the number of records to be extracted. This number can vary between tenants of the same data range. If the number of records is known, it is possible to estimate the time it will take to extract most of them. The total time for one extraction job can still be higher as even if one user is not extracted, the extraction will continue until all data is extracted or the deadline of 24 hours is met.
```
jasonjoh, line 45:
```suggestion
Extracting data from your Microsoft 365 tenant to Azure can take time. The total time a copy takes depends on the number of records to be extracted. This number can vary between tenants of the same data range. If the number of records is known, it is possible to estimate the time it will take to extract most of them. The total time for one extraction job can still be higher as even if one user is not extracted, the extraction will continue until all data is extracted or the deadline of 24 hours is met. 
```

jasonjoh: A couple of questions to clarify:

- "...vary between tenants of the same data range": what does it mean for tenants to be "of the same data range"?
- "...even if one user is not extracted, the extraction will continue...": what does this mean? Is it because it will retry the user that was not extracted? Also, is the user extracted, or is the user's data?

jasonjoh (changes requested):
I'm not sure this topic is the right place for this information. The time to execute has nothing to do with billing or policies that I can tell.

PR #20108, opened 2023-02-01
list usermemberships v1 permissions table update

psaffaie:
updating permissions table to reflect that for app only permissions user and group permissions are supported and are lesser privileged than Directory permissions


api-reference/v1.0/api/user-list-transitivememberof.md
```diff
@@ -19,9 +19,13 @@ One of the following permissions is required to call this API. To learn more, in
 
 | Permission type                        | Permissions (from least to most privileged)                                  |
 | :------------------------------------- | :--------------------------------------------------------------------------- |
-| Delegated (work or school account)     | User.Read, GroupMember.Read.All, Directory.Read.All, Directory.ReadWrite.All |
+| Delegated (work or school account)     | User.Read.All, GroupMember.Read.All, Group.Read.All, Directory.Read.All, Directory.ReadWrite.All |
```
FaithOmbongi, line 22: Same feedback as in the beta doc...

I think User.Read delegated should be included here; because this API also supports the /me/transitiveMemberOf endpoint whose least privileged permission is User.Read.

Something else to think about, add the GET /me/transitiveMemberOf syntax the HTTP request section (Line 33) below.

FaithOmbongi (approved):
Thank you @psaffaie for your quick work on this PR. Approved.

PR #20115, opened 2023-02-02
Add Change Log For Qna

ZChristine:


changelog/Microsoft.MicrosoftSearch.json
```diff
@@ -592,6 +592,14 @@
           "Description": "Added the **bookmark** member to the **entityType** enumeration.",
           "Target": "entityType"
         },
+        {
```
Lauragra, line 595: Why are we adding this to an existing ChangeList item? Is this to fix an oversight? If not, there should be a doc update that goes along with this, and it should be in a new ChangeList record.

ZChristine: Hi, @Lauragra. Sorry for late response. This is a different one as this is the change log for adding QnA member instead of origin permission. Could you help elaborate more on the existing ChangeList item?

Lauragra:
@ZChristine , have you added the new enum member to the documentation, in addition to adding the changelog entry?

PR #20120, opened 2023-02-02
list groups expand members

psaffaie:
listing groups with advanced capabilities to $expand for members. @FaithOmbongi  would you be able to help me add the sample code request and response


FaithOmbongi:
Hi @psaffaie - I've added the syntax as part of another example so that we can optimize the doc space.

FaithOmbongi (approved):

PR #20167, opened 2023-02-07
revised least privilege

Justinha:


api-reference/v1.0/api/user-assignlicense.md
```diff
@@ -17,9 +17,9 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type      | Permissions (from least to most privileged)              |
 |:--------------------|:---------------------------------------------------------|
-|Delegated (work or school account) | User.ReadWrite.All, Directory.ReadWrite.All    |
+|Delegated (work or school account) | LicenseAssignment.ReadWrite.All    |
```
alvarorahul, line 20: this doens't look correct.... I think it should be

LicenseAssignment.ReadWrite.All, User.ReadWrite.All, Directory.ReadWrite.All


same for application permissions below

alvarorahul (changes requested):

PR #20169, opened 2023-02-08
Add rbac roles to domain api docs v1

franqq:
Update RBAC Roles for the following apis in domains

1. Delete domain
2. Force delete domain
3. Get domain
4. List domainNameReferences
5. List verificationDnsRecords
6. List domains
7. Create domain
8. Update domain
9. Verify domain


FaithOmbongi (commented):
Thank you @franqq for documenting the supported roles for this set of domain APIs. Approved.

FaithOmbongi (approved):

PR #20185, opened 2023-02-09
Fixes Planner Roster typo and toc label

milanholemans:
Noticed a typo in the add Planner Roster member docs.
Also noticed that the `get` page for Planner Roster members was not included in the toc.


milanholemans:
Seems like the pipeline failed on a link which I didn't update

PR #20265, opened 2023-02-16

maheshsinghania:
The connection limit is agnostic of experience. 30 is valid for both search and compliance scenarios. Hence removing the distinction.


maheshsinghania:
@KumarSeetharaman74 - FYI for the update of connection limit for Compliance

PR #20294, opened 2023-02-18
Update author for callRecords and disable Try It feature on snippets

mcm223:
This change for beta and v1.0 updates the author metadata for new PM ownership of the callRecords API. It also changes the ```msgraph-interactive``` widget to ```http```, as Graph Explorer does not support APIs with Application permissions so users trying to use the "try it" feature would end up with 403 errors. This addresses the customer-raised issue here: https://github.com/microsoftgraph/microsoft-graph-docs/issues/14518.


dodaromike:
@mcm223  The warnings appear to be coming from main, not from this PR.  I will try to fix them.

dodaromike (approved):

PR #20319, opened 2023-02-21
add itemidresolver subtype to relationship section

monaray97:


api-reference/beta/resources/externalconnectors-urltoitemresolverbase.md
```diff
@@ -23,7 +23,7 @@ This is an abstract type.
 |priority|Int32|The priority which defines the sequence in which the urlToItemResolverBase instances are evaluated. |
 
 ## Relationships
-None.
+An existing subtype of the urlToItemResolverBase resource type is the [itemIdResolver resource type](externalconnectors-itemidresolver.md).
```
Lauragra, line 26: This is not a relationship. Is this a derived type?

monaray97: it is a derived type. i'd like some way to link the itemidresolver type on this page since this is the only derived type available right now for urltoitemresolvers. what do you recommend for linking to this page?

Lauragra (approved):

PR #20341, opened 2023-02-23
[Doc Improvement][Updated Update teamsAppSettings ]

v-ypalikila:
fixed issue with http code tag in https://learn.microsoft.com/en-us/graph/api/teamsappsettings-update?view=graph-rest-beta&tabs=http.


v-ypalikila:
Issue already resolved on live site. Changing PR to draft state.

PR #20420, opened 2023-03-01
Updated LCW titles

OWinfreyATL:


FaithOmbongi:
Hi @OWinfreyATL -Thank you for leading this.

Suggestion: Let's let the title differentiation for deletedItem and deletedItemContainer remain for now. I'll talk to Angela about this because we do have some exceptions.

FaithOmbongi (approved):
Hi @OWinfreyATL -Thank you for leading this.

Suggestion: Let's let the title differentiation for deletedItem and deletedItemContainer remain for now. I'll talk to Angela about this because we do have some exceptions.

Approved with suggestions.

PR #20436, opened 2023-03-02

yyuank:
@FaithOmbongi we need to remove User.ReadBasic.All app-only permission due to a security bug. Could you please help with this PR and also we need to remove it from All pmerissions and IDs table.


FaithOmbongi (approved):

FaithOmbongi:
Hi @yyuank - As you've confirmed via chat, this permission needs to be removed from the API reference docs as well. Let's engage internally to find an efficient way to do it.

PR #20454, opened 2023-03-03
Jeremyke metered api day one updates

JeremyKelley:
These changes are related to both feedback that came in late on the original publish of the docs as well as adding entries to the TOC.

This also includes contributions to the Teams API pages for better explaining different error conditions for model A and B.


concepts/includes/metered-api-known-limitations.md
```diff
@@ -0,0 +1,12 @@
...
+author: jeremyke
+ms.topic: include
+ms.date: 03/02/2023
+ms.localizationpriority: high
+---
+## Known limitations
+
+The following limitations apply to metered APIs:
+
+- Metered Microsoft 365 APIs and services are not currently available in national cloud deployments, including Microsoft 365 GCC deployments accessed through the worldwide Microsoft Graph endpoint. For details about national clouds, see [National cloud deployments](/graph/deployments.md).
```
angelgolfer-ms, line 11: Line 11:
In general, it might be more complete to first point out where the metered APIs are available before listing the negative (where not available), just in case some customers are concerned if and how they might be affected.

Something like:
"Metered Microsoft 365 APIs and services are currently available only in the Microsoft global environment, and are not available in national cloud deployments, including..."

concepts/includes/metered-api-known-limitations.md
```diff
@@ -0,0 +1,12 @@
...
+author: jeremyke
+ms.topic: include
+ms.date: 03/02/2023
+ms.localizationpriority: high
+---
+## Known limitations
+
+The following limitations apply to metered APIs:
+
+- Metered Microsoft 365 APIs and services are currently available in only the Microsoft global environment and not in national cloud deployments, including Microsoft 365 GCC deployments accessed through the worldwide Microsoft Graph endpoint. For details about national clouds, see [National cloud deployments](/graph/deployments.md).
```
angelgolfer-ms, line 11: Line 11:
This is just FYI, we can close on the following and I'll make the appropriate adjustment in this PR:
To construct parallel phrases, we can do either this:
"... currently available in only {X} and not {Y}, ..." 

or this:
"... currently available only in {X} and not in {Y}, ..."

Right now per your last update, we got:
"... currently available in only {X} and not in {Y}, ..."

I can change this to the second pattern.

concepts/metered-api-overview.md
```diff
@@ -38,12 +38,7 @@ Keep the following considerations in mind when you use metered Microsoft 365 API
...
 
-## Known limitations
-
-The following limitations apply to metered APIs:
-
-- Metered Microsoft 365 APIs and services are not currently available in national cloud deployments, including Microsoft 365 GCC deployments accessed through the worldwide Microsoft Graph endpoint. For details about national clouds, see [National cloud deployments](deployments.md).
-- The target application must be a confidential client application (for example, web application, web API, or daemon/service). Public client applications (desktop and mobile applications) are not supported.
+[!INCLUDE [Metered API Known Limitations](includes/metered-api-known-limitations.md)]
 
 ## See also
```
angelgolfer-ms, line 43: Want to call out that I am adding a link to the Teams-specific topic about their monetized APIs here.
In general, if we consider this topic as the uber overview topic for metered M365 APIs in Microsoft Graph, then there should be links in both directions with other service-specific monetization topics.
So for example, when the files topic eventually goes live, then that topic and this overview topics should have links going both ways with each other.

concepts/toc.yml
```diff
@@ -829,6 +835,8 @@ items:
           - name: Add custom data
             displayName: directory extensions, schema extensions, open extensions, extension attributes
             href: extensibility-overview.md
+          - name: Enable metered APIs
+            href: metered-api-setup.md
           - name: Use Graph Explorer
```
angelgolfer-ms, line 840: Lines 838-839:
No action item for you, just FYI - whenever you update the part of the TOC that's under the "Develop" node, please remember to update the same section in the toc.yml files for v1 and for beta as well. That part of the conceptual TOC is repeated in these other 2 TOC files.
Otherwise if the customer is in the v1 or beta reference and goes to the "Develop" section in the TOC, they wouldn't see the new metered-api-setup.md file there.

concepts/metered-api-overview.md
```diff
@@ -8,27 +8,27 @@ ms.custom: scenarios:getting-started
...
-Whether metered or not, APIs in Microsoft Graph are provided based on the following two principles.
+Whether metered or not, APIs in Microsoft Graph follow these two principles:
 - Customer data ownership: Customer data belongs to the customer. Learn more about how [Microsoft categorizes customer data](https://www.microsoft.com/trust-center/privacy/customer-data-definitions). 
 - Reasonable access: The service provides access to customer content, within [defined limits](throttling-limits.md). 
 
-Metering some APIs helps to ensure the health of the current and future Microsoft Graph ecosystem by balancing platform access and cost. Microsoft will not charge for generally available APIs that are currently free without following the [breaking change policy](versioning-and-support.md).
+Metering some APIs helps to ensure the health of the current and future Microsoft Graph ecosystem by balancing platform access and cost. In the event that a Microsoft Graph API that is included with user subscription licenses becomes metered, that will be a non-backwards compatible change and follow the [versioning, support, and breaking change policies for Microsoft Graph](versioning-and-support.md).
```
angelgolfer-ms, line 18: Line 17: 
No follow-up action necessary, just FYI, I am changing the future tense "will" to "would". 
In general we shouldn't use future tense in docs as everything in the "official" docs should hold as of the current point of time. So even when describing a hypothetical situation like this context here, the logic and plan already exist, so from that perspective, it's reasonable to use not use future tense.

angelgolfer-ms (approved):
Thanks for incorporating the previous feedback in PR 20032, and including other Teams-related updates.
The PR looks good from my POV.

PR #20479, opened 2023-03-06
Adding role requirements to sync API docs

rahuln3223:


FaithOmbongi (approved):
Thank you @rahuln3223 for your swift work on documenting the RBAC roles. Just one missing file https://learn.microsoft.com/en-us/graph/api/directorydefinition-discover.

PR #20502, opened 2023-03-08
Changing from PostAsync to AddAsync

felipeschneider88:
The method PostAsync doesn't exist anymore, we should use AddAsync, this error is widely spread all over the documentation


andrueastman:
Thanks for opening this @felipeschneider88 

The snippet is however correct as it targets v5 of the SDK and snippets for version 4 have been replaced in the documentation. 
We'll therefore close this for now.

PR #20513, opened 2023-03-09
[CloudPC] remove preview tag for CloudPC

Guoan-Tang:


api-reference/beta/toc.yml
```diff
@@ -10564,7 +10564,7 @@ items:
               href: api/intune-tem-telecomexpensemanagementpartner-delete.md
             - name: Update
               href: api/intune-tem-telecomexpensemanagementpartner-update.md
-      - name: Cloud PCs (preview)
+      - name: Cloud PCs
```
Lauragra, line 10567: @Guoan-Tang , is this PR associated with another PR to bring the Cloud PC API to v1.0? If so, can you make this update along with adding the v1.0 docs so we can publish them together?

Guoan-Tang: Hi @Lauragra, as "preview" tag was added during our private beta period, we are supposed to remove the tag after cpc's global announcement. And our product was released in July 2021, therefore the tag need to be removed now.

Lauragra: What is your timeline for publishing the v1.0 docs? Is the API available in v1.0 currently? The preview notation in the beta TOC is there to indicate that an API is only available on the beta endpoint but not v1.0. We remove it when the API and docs are live in v1.0.

PR #20535, opened 2023-03-10
cosmetic changes to support permissions

sseth-msft:


Danielabom (approved):
Thank you for working on this PR @sseth-msft. Based on our chat discussion, these changes do not affect the schema. PR approved.

api-reference/beta/api/security-retentionevent-delete.md
```diff
@@ -21,7 +21,7 @@ One of the following permissions is required to call this API. To learn more, in
 |:---|:---|
 |Delegated (work or school account)|RecordsManagement.ReadWrite.All|
 |Delegated (personal Microsoft account)|Not supported.|
-|Application|Not supported.|
+|Application|RecordsManagement.ReadWrite.All|
```
Lauragra, line 24: Is this correcting an oversight in the docs or are you adding support for application permissions? If the latter, we need an entry in the changelog for this update.

sseth-msft: There is no csdl/schema change here and hence no chqangelogs. App permissions were not corrently working before, we made some fix our our BE to make this work which is way Im updating the docs again.

sseth-msft: @Lauragra  - responded to your comment, please let me know if we can resolve and publish this

Lauragra: We do update the changelog when we enable a permissions scenario, that's why I'm double checking. See example 15 in this topic: https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211742/Changelog

PR #20547, opened 2023-03-10
Fix link for PR 20190, line 8845

dodaromike:
Broken link in beta TOC line 8845 flagged in PR 20190


dodaromike:
Fixed link on line 8845 of beta TOC.  The file flagged by broken link on line 10278 is not in the beta API.

PR #20572, opened 2023-03-13
Clarify metering is for application permissions

erichui-ms:


api-reference/v1.0/api/chatmessage-update.md
```diff
@@ -17,7 +17,8 @@ The update only works for chats where members are Microsoft Teams users. If one
 
 This method does not support federation. Only the user in the tenant who sent the message can perform data loss prevention (DLP) updates on the specified chat message.
 
-[!INCLUDE [teams-model-A-only-disclaimer](../../includes/teams-model-A-only-disclaimer.md)]
+> [!NOTE]
+> When using this API in application permssions, it is metered.  It supports the `model=A` payment model. For details, see [Payment models](/graph/teams-licenses#payment-models). If you don't specify a payment model in your query, the default [evaluation mode](/graph/teams-licenses#evaluation-mode-default-requirements) will be used.
```
Lauragra, line 21: Suggested revision for the first sentence:

When used with application permissions, this API is metered.

api-reference/v1.0/api/chatmessage-update.md
```diff
@@ -17,7 +17,8 @@ The update only works for chats where members are Microsoft Teams users. If one
 
 This method does not support federation. Only the user in the tenant who sent the message can perform data loss prevention (DLP) updates on the specified chat message.
 
-[!INCLUDE [teams-model-A-only-disclaimer](../../includes/teams-model-A-only-disclaimer.md)]
+> [!NOTE]
```
Lauragra, line 20: Are we adding this as text because we need a different variation of the text that we have in the Include file, i.e. to mention application permissions? (As opposed to updating the text of the existing include file.) If we think this will apply to multiple topics, we should create another include file for it. We'll need to add this to the file in the beta folder as well, right?

erichui-ms: I've just searched in GitHub, and this file seems to be the ONLY file that is using that include file.  Thus, let me move this change to the include file instead.

erichui-ms:
Hi, @JeremyKelley.  Could you please review and approve?  Thanks in advance!

Lauragra (approved):

Lauragra:
@erichui-ms , I approved this PR. Not sure if we still need it?

erichui-ms:
Thanks, @Lauragra .  Yes, we still need to publish this.  I am changing the label to "ready to merge" now.

PR #20591, opened 2023-03-14
Fix code snippet list-serviceprincipals-startswith-csharp-snippets.md

ArveSystad:
Encoding error when generated? Should be an actual space, not %20.


andrueastman:
Thanks for your contribution here @ArveSystad 

This has however now been fixed in the snippet generator and the docs should be showing the correct snippets now. Closing this one for now

PR #20648, opened 2023-03-20
Update List People $top description to be more accurate

akapila-msft:
Update List-People $top parameter description to provide a clear description and expectation.


PeterNjorogeMS:
Hi @angelgolfer-ms, We have a one-line change to the documentation and need your help with approving it.

api-reference/v1.0/api/user-list-people.md
```diff
@@ -45,7 +45,7 @@ This method supports the [OData query parameters](/graph/query-parameters) to he
 |$search|string|Search for people by name or alias. Supports Fuzzy matching. Parameter only works for searching the signed-in user's relevant people, not for searching people relevant to other users. Also supports the `topic` keyword to find people based on topics extracted from e-mail conversations with that person. For information and examples, see the *Perform a fuzzy search* section at [Use the People API to get information about the people most relevant to you](/graph/people-insights-overview#perform-a-fuzzy-search). |
 |$select|string|Comma-separated list of properties to include in the response. For optimal performance, only select the subset of properties needed.|
 |$skip|int|Skip the first n results, useful for paging. This is not supported when using *$search*.|
-|$top|int|Number of results to be returned.|
+|$top|int|Maximum number of results to be returned. The actual results may be fewer.|
```
angelgolfer-ms, line 48: @PeterNjorogeMS , I think the original description and your update are both incomplete. AFAIK $top is about the max number of items on a page, and if there are more items that match the query, the result page would include an @odata.nextLink property for the caller to get the next page, and so on, as explained in this [paging](https://learn.microsoft.com/en-us/graph/paging) topic.
$top is to specify the max number of items on a result page. If we say "max number of results...", that would mean the API returns only that number of items and would not return any other items even if they match the query.

For the typical $top behavior, please use this description:
"The maximum number of results to be returned on a results page. For more information, see [top parameter](https://learn.microsoft.com/en-us/graph/query-parameters?tabs=http#top-parameter)."
Because we have the query parameters topic and [paging](https://learn.microsoft.com/en-us/graph/paging) topic, we don't usually explain the details. 

If the list operation implementation is different, then please confirm.

angelgolfer-ms: Please make the same fix to both the beta and v1 topics.

angelgolfer-ms (changes requested):
Hi @akapila-msft , @PeterNjorogeMS ,
Please see my comments and my internal email about items to address.
I added @Danielabom to help follow up.

Thanks.

api-reference/beta/api/user-list-people.md
```diff
@@ -45,7 +45,7 @@ This method supports the following OData query parameters to help customize the
 |$search|string|Search for people by name or alias. Supports Fuzzy matching. Parameter only works for searching the signed-in user's relevant people, not for searching people relevant to other users. Also supports the `topic` keyword to find people based on topics extracted from e-mail conversations with that person. For information and examples, see the *Perform a fuzzy search* section at [Use the People API to get information about the people most relevant to you](/graph/people-insights-overview#perform-a-fuzzy-search).|
 |$select|string|Comma-separated list of properties to include in the response. For optimal performance, only select the subset of properties needed.|
 |$skip|int|Skip the first n results, useful for paging. This is not supported when using *$search*.|
-|$top|int|Number of results to be returned.|
+|$top|int|The maximum number of results to be returned on a results page. For more information, see [top parameter](https://learn.microsoft.com/graph/query-parameters?tabs=http#top-parameter).|
```
angelgolfer-ms, line 48: Line 48;
@akapila-msft , 
FYI, when linking from a reference topic to a conceptual topics, or vice versa, please use a [site-relative link ](https://review.learn.microsoft.com/en-us/help/platform/links-how-to?branch=main#links-to-articles-in-other-repos-or-docsets).

That is, instead of using the HTTP link, just specify /graph/query-parameters?tabs=http#top-parameter.

In the interest of time, I'll fix this for you in my next commit. 
If indeed this is all that you're ready to fix in this PR, then please follow step 8 of the [content workflow](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211773/Content-workflow?anchor=workflow-for-internal-contributors-to-add-new-content-or-update-content) and set the PR to ready to merge.

angelgolfer-ms (approved):
HI @akapila-msft , 
Seems like per your email on 3/24, you'd like to keep this PR to the current clarifications and have the PR published as soon as it's ready, is that right? If so, please take a look at my comment (I have applied the link fixes described in the comment), and set the PR to ready to merge per our content workflow. Then this PR will be published as part of Tuesday's publishing.

PR #20656, opened 2023-03-20
Update callRecord docs with retention details

mcm223:
This change adds details around call record retention and when records are generated, based on common customer questions. It also updates outdated doc author values.


Danielabom (approved):
Thank you for working on this PR @mcm223. PR approved.

PR #20711, opened 2023-03-24
Add ranking hints docs

emzho:
API review work item ID: 15776
CSDL PR ID: 7882008


api-reference/beta/resources/externalconnectors-property.md
```diff
@@ -25,9 +25,10 @@ A [schema](externalconnectors-schema.md) property definition for a Microsoft Sea
...
+| labels        | microsoft.graph.externalConnectors.label collection | Specifies one or more well-known tags added against a property. Labels help Microsoft Search understand the semantics of the data in the connection. Adding appropriate labels would result in an enhanced search experience (e.g. better relevance). Optional.<br><br>The possible values are: `title`, `url`, `createdBy`, `lastModifiedBy`, `authors`, `createdDateTime`, `lastModifiedDateTime`, `fileName`, `fileExtension`, `iconUrl`, `containerName`, `containerUrl`. Note that you must use the `Prefer: include-unknown-enum-members` request header to get the following values in this [evolvable enum](/graph/best-practices-concept#handling-future-members-in-evolvable-enumerations): `iconUrl`, `containerName`, `containerUrl`.|
```
Danielabom, line 28: Quick question, why did you remove "unkownFutureValue"?

api-reference/beta/resources/externalconnectors-property.md
```diff
@@ -25,9 +25,10 @@ A [schema](externalconnectors-schema.md) property definition for a Microsoft Sea
...
 | name          | String            | The name of the property. Maximum 32 characters. Only alphanumeric characters allowed. For example, the property name may not contain control characters, whitespace, or any of the following: `:`, `;`, `,`, `(`, `)`, `[`, `]`, `{`, `}`, `%`, `$`, `+`, `!`, `*`, `=`, `&`, `?`, `@`, `#`, `\`, `~`, `'`, `"`, `<`, `>`, `` ` ``, `^`.  Required.                |
-| type          | microsoft.graph.externalConnectors.propertyType         | The data type of the property. Possible values are: `string`, `int64`, `double`, `dateTime`, `boolean`, `stringCollection`, `int64Collection`, `doubleCollection`, `dateTimeCollection`, `unknownFutureValue`. Required. |
+| rankingHint   | [microsoft.graph.externalConnectors.rankingHint](externalconnectors-rankinghint.md) | Specifies the property ranking hint. Developers can specify which properties are most important, allowing Microsoft Search to determine the search relevance of the content. |
```
Danielabom, line 30: Since you have previously applied the best practice of listing the possible values, we recommend the same in this case , here's the evolvable enums wiki: (https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211749/Evolvable-enumerations)

Danielabom (changes requested):
Hi @emzho, this PR is in great shape. Please review my comments. Thanks!

api-reference/beta/resources/externalconnectors-property.md
emzho: I thought about removing, because it is not a valid value for a partner to pass in, it's only a valid value that our API would return. We've gotten some bad requests from clients calling, trying to set the property value as "unknownFutureValue" which is not correct. I thought removal would be less ambiguous, because the above language seems to imply that unknownFutureValue is a valid value. What are your thoughts around updating this guidance?

api-reference/beta/resources/externalconnectors-property.md
emzho: "rankingHint" isn't an enum, which is why I didn't list out possible values here. I did so in the externalconnectors-rankinghints.md file instead when describing importancesScore

api-reference/beta/resources/externalconnectors-property.md
Danielabom: I understand, and it makes sense. However, since it is an evolvable enum, listing  "unkownFutureValue" as a possible value is a requirement. I'll follow up on Teams.

Danielabom: @angelgolfer-ms

Danielabom (approved):
@emzho, thank you for working on this PR and always clearing my doubts. PR approved.

PR #20766, opened 2023-03-29
Update domain-update.md

abhishekparasurampuram:
PATCH payload shown in documentation could lead to confusion when customers try on federated domains. So adding a note that auth type should be managed to become default domains.

### DIFF
--- a/api-reference/v1.0/api/domain-update.md
+++ b/api-reference/v1.0/api/domain-update.md
@@ -80,6 +80,8 @@ Content-type: application/json
   ]
 }
 ```
+> **Important:**
+> Domain authentication type must be ‘Managed’ to become default domain.
 
 # [C#](#tab/csharp)
 [!INCLUDE [sample-code](../includes/snippets/csharp/update-domain-csharp-snippets.md)]


api-reference/v1.0/api/domain-update.md
```diff
@@ -80,6 +80,8 @@ Content-type: application/json
   ]
 }
 ```
+> **Important:**
+> Domain authentication type must be ‘Managed’ to become default domain.
```
atriagrawal, line 84: Maybe add a link to "DomainAuthenticationType" property.

atriagrawal, line 84: "to be set as default domain"?

abhishekparasurampuram: Updated

abhishekparasurampuram: There is no specific link to the property.

api-reference/v1.0/api/domain-update.md
```diff
@@ -80,6 +80,8 @@ Content-type: application/json
   ]
 }
 ```
+> [!Important]
+> The domain authentication type must be `Managed` to be set as the default.
```
Lauragra, line 84: Please verify this edit and apply the change to the file in the beta folder as well. Thanks.

abhishekparasurampuram: Applied in beta folder as well.

PR #20349, opened 2023-02-24
USQ docs to Beta

BarryShehadeh:
Add USQ docs to BETA.

PR #20815, opened 2023-04-03
USQ - fix toc and a typo

BarryShehadeh:
fix TOC.
update a typo that was introduced here: https://github.com/microsoftgraph/microsoft-graph-docs/pull/20349/commits/3f64b4f0b233169fb249fa1f0e68abbf36da4bac

### DIFF
--- a/api-reference/beta/toc.yml
+++ b/api-reference/beta/toc.yml
@@ -2067,22 +2067,6 @@ items:
             href: api/printtask-get.md
           - name: Update task
             href: api/printtaskdefinition-update-task.md
-        - name: Quota
-          items:
-          - name: Unified storage quota
-            href: resources/unifiedstoragequota.md
-            items:
-            - name: Get
-              href: api/unifiedstoragequota-get.md
-            - name: List
-              href: api/unifiedstoragequota-list-services.md
-          - name: Service storage quota breakdown 
-            href: resources/servicestoragequotabreakdown.md
-            items:
-            - name: Get
-              href: api/servicestoragequotabreakdown-get.md
-            - name: List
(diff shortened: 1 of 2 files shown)


Lauragra (approved):

api-reference/beta/toc.yml
```diff
@@ -15024,6 +15008,22 @@ items:
             href: api/multivaluelegacyextendedproperty-post-multivalueextendedproperties.md
           - name: Get multi-value property
             href: api/multivaluelegacyextendedproperty-get.md
+    - name: Quota
```
Lauragra, line 15011: Where did you mean to put this section in the TOC? It was originally under Cloud printing. It's currently a top-level node; what should it be nested under?

Lauragra (changes requested):
Before we can approve this we need to fix the position in the TOC. It also looks like we don't have the correct WorkloadArea value in the changelog to make this API discoverable.

api-reference/beta/toc.yml
BarryShehadeh: @Lauragra  It was a typo. We meant to have it after "Reports", but I used the wrong reports as a reference where to insert it.
It has nothing to do with Cloud Printing. 
And it has its own workload, a new workload. Which has nothing to do with files. And I think it should have it's own top level. Please let me know if you have better place for it.

Lauragra: Okay gotcha. We minimize adding new top-level nodes if there are logical groupings for APIs, so let us take a look and we can make a recommendation. In the mean time, let's separate the typo fix from the TOC fix.

PR #20821, opened 2023-04-03
Update site-list-permissions.md

micesari:
To call list site's permissions with a delegated work or school account, it must be from a tenant admin account.


api-reference/v1.0/api/site-list-permissions.md
```diff
@@ -17,7 +17,7 @@ One of the following permissions is required to call this API. To learn more, in
 
 |Permission type                        | Permissions (from least to most privileged)
 |:--------------------------------------|:-------------------------------------
-|Delegated (work or school account)     | Sites.FullControl.All
+|Delegated (work or school account)     | Sites.FullControl.All (must be tenant admin)
```
Lauragra, line 20: Please use an asterisk or superscript and add this as a note following the table. Please also update the beta version of this topic, which looks to be out of sync with v1.0.

PR #20833, opened 2023-04-04
Update throttling-limits.md

abheek-das:
Changes:
- Updated the correct value for Outlook ingress throttling limits.
- Added details on batching behavior for Outlook resources.


abheek-das:
Hi @angelgolfer-ms : As per our discussion I added details related to Outlook batching behavior in https://learn.microsoft.com/en-us/graph/throttling-limits#outlook-service-limits. Given this page only contains generic information limited to the batching behavior, I did not make any change to it. It already has a pointer to the throttling & batching page - so do you think we need to add something more here?

dkershaw10 (approved):
LGTM

concepts/throttling-limits.md
```diff
@@ -389,7 +389,7 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
 |------------------------------------------------------------|-----------------|
 | 10,000 API requests in a 10 minute period                  | v1.0 and beta endpoints |
 | 4 concurrent requests                                      | v1.0 and beta endpoints   |
```
angelgolfer-ms, line 391: Line 391:
@abheek-das , with reference to line 407 where you mentioned "the concurrency limits of the Outlook service", I'm deducing what the two lines (391 and 407) are saying is that, only up to 4 requests can be running in parallel regardless of whether it's all those requests on the same mailbox or multiple mailboxes. 
Is that correct?

angelgolfer-ms: If that is correct, then I find the sentence in line 386 misleading:
"If an application exceeds the limit in one mailbox, it does not affect the ability to access another mailbox."

That's not true in the context of concurrent execution - the limit is up to 4 requests at a time, and if all 4 concurrent requests are on one mailbox, calls for another mailbox are going to have to wait, so "does not affect" is false.

If my deduction is correct, then I think we need to pull line 391 out of that table and state separately the limit of 4 concurrent requests can be on the same mailbox or multiple mailboxes.

angelgolfer-ms: Following the assumption of my last comment, I'd suggest under the existing H2 heading "Outlook service limits", add two new H3 sections, to facilitate linking and clarity. Something like this:

```
### Limits per app ID

(State the limit for up to 4 concurrent requests for an app accessing the Outlook service, regardless of whether all requests are for the same mailbox or multiple mailboxes.)

### Limits per app ID and mailbox combination

(Primarily the existing content from lines 386 to 392, excluding line 391.)

File: concepts/throttling-limits.md [hunk 2]
```diff
@@ -403,7 +403,13 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
 | Social and workplace intelligence | <li>[person](/graph/api/resources/person) |
 | To-do tasks API (preview) | <li>[outlookTask](/graph/api/resources/outlooktask) <li> [outlookTaskFolder](/graph/api/resources/outlooktaskfolder) <li>[outlookTaskGroup](/graph/api/resources/outlooktaskgroup) <li> [outlookCategory](/graph/api/resources/outlookcategory) <li> [attachment](/graph/api/resources/attachment)|
 
-
+### Outlook service batching
```
angelgolfer-ms, line 406: Just a comment, based on your Teams message to me on 4/3 - it seems that you have decided not to explicitly call out the change in behavior in this PR that developers will be able to observe - that any batch more than 4 requests used to have only the first 4 requests possibly succeeding and the rest failing with 429.

That's your intention, is that right, not just that you forgot?

BTW, that previous behavior isn't in the known issues topic either, so... nothing in particular to address in that topic now.

angelgolfer-ms, line 406: And, speaking of the known issues topic, the existing [section for JSON batching](https://learn.microsoft.com/en-us/graph/known-issues#json-batching) is just referring to the ongoing behavior for years, is that your perception too, being a partner team that supports batching?

If that's your perception too, then this shouldn't even be in the known issues topic at all. 

In general, per ongoing [guidance](https://dev.azure.com/msazure/One/_wiki/wikis/Microsoft%20Graph%20Partners/211754/Known-issues), the known issues topic is for behaviors that reflect a bug or inadequate design, AND that the related product teams have a plan to fix within a reasonable period of time which we define as 6 months. Otherwise even if the behavior is sub-optimal, the behavior should be documented in the corresponding documentation topics, not in the known issues topic.

concepts/throttling-limits.md
```diff
@@ -403,7 +403,13 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
...
+If a [batch](json-batching.md) request contains multiple requests to Outlook service resources within it, then by default 4 requests will be executed in parallel at any point (irrespective of the target mailbox) to stay within the concurrency limits of the Outlook service. However, the batch execution can be made fully sequential using the [dependsOn](json-batching.md#sequencing-requests-with-the-dependson-property) property.
```
angelgolfer-ms, line 407: @abheek-das ,
FYI for this section, I'm going to make some direct updates in your PR to use language consistent with the existing docs for [batching](https://learn.microsoft.com/en-us/graph/json-batching). For example, in the existing docs we don't use "thread" and instead "individual requests". Even though "thread" is a common computer science concept, in docs in general, we try to use the same terms throughout so to eliminate any ambiguity or guessing. That would help Loc as well.

Please take a look at my upcoming commits.

concepts/throttling-limits.md
```diff
@@ -403,7 +403,13 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
...
+  
+Thus when targetting the same mailbox, applications that have an ability to run multiple batch threads in parallel can either run - 
+1. No more that 4 threads (batches contain dependsOn), or
```
angelgolfer-ms, line 410: Line 410:
@abheek-das , from line 409, I thought you meant a "thread" as in executing an individual request in a batch. So carrying on to line 410,  for a batch request that specifies dependsOn, then every request in the batch should be executed one after another per the specified order, and nothing is concurrent. Why do you say no more than 4 threads? Shouldn't that be one thread per request, up to the limit of 20 requests in a batch?

angelgolfer-ms: Or are you describing the old behavior you explained to me on Teams, that in a batch of up to 20 requests, only 4 requests get excuted, all the other requests in the batch end up in 429?

angelgolfer-ms: Oh, third guess - I think what you're referring to as a "batch thread" in line 409 is "batches executing in parallel". 
So #1 is about each batch using dependsOn where the requests within the batch run in sequential order, so up to 4 such batches can run in parallel.
Is that what you mean?

concepts/throttling-limits.md
```diff
@@ -403,7 +403,13 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
 | Social and workplace intelligence | <li>[person](/graph/api/resources/person) |
 | To-do tasks API (preview) | <li>[outlookTask](/graph/api/resources/outlooktask) <li> [outlookTaskFolder](/graph/api/resources/outlooktaskfolder) <li>[outlookTaskGroup](/graph/api/resources/outlooktaskgroup) <li> [outlookCategory](/graph/api/resources/outlookcategory) <li> [attachment](/graph/api/resources/attachment)|
 
-
+### Outlook service batching
+If a [batch](json-batching.md) request contains multiple requests to Outlook service resources within it, then by default 4 requests will be executed in parallel at any point (irrespective of the target mailbox) to stay within the concurrency limits of the Outlook service. However, the batch execution can be made fully sequential using the [dependsOn](json-batching.md#sequencing-requests-with-the-dependson-property) property.
+  
+Thus when targetting the same mailbox, applications that have an ability to run multiple batch threads in parallel can either run - 
+1. No more that 4 threads (batches contain dependsOn), or 
+2. A single batch request without dependsOn.
```
angelgolfer-ms, line 411: Line 411:
For #2, I think what you're trying to say is, a batch request (each POST to the $batch endpoint) needs to complete first before the next request in the application. So at any one time, if the application specifies multiple batch requests and they don't use dependsOn, then only up to 4 individual requests belonging to the same batch requests can execute at the same time.

concepts/throttling-limits.md
```diff
@@ -403,7 +403,13 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
...
+
```
angelgolfer-ms, line 412: Lines 409-411:
In the interest of time to try save iterations, I'm going to assume the [third guess](https://github.com/microsoftgraph/microsoft-graph-docs/pull/20833/files#r1165015384) is what you're trying to say and will rephrase lines 409-411 for clarity.
Please clarify if that's not what you meant.

concepts/throttling-limits.md
```diff
@@ -403,7 +403,13 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
 | Social and workplace intelligence | <li>[person](/graph/api/resources/person) |
 | To-do tasks API (preview) | <li>[outlookTask](/graph/api/resources/outlooktask) <li> [outlookTaskFolder](/graph/api/resources/outlooktaskfolder) <li>[outlookTaskGroup](/graph/api/resources/outlooktaskgroup) <li> [outlookCategory](/graph/api/resources/outlookcategory) <li> [attachment](/graph/api/resources/attachment)|
 
-
+### Outlook service batching
+If a [batch](json-batching.md) request contains multiple requests to Outlook service resources within it, then by default 4 requests will be executed in parallel at any point (irrespective of the target mailbox) to stay within the concurrency limits of the Outlook service. However, the batch execution can be made fully sequential using the [dependsOn](json-batching.md#sequencing-requests-with-the-dependson-property) property.
+  
+Thus when targetting the same mailbox, applications that have an ability to run multiple batch threads in parallel can either run -
```
angelgolfer-ms, line 409: Line 409:
Why is this limited to targeting the same mailbox?
If you mean "multiple batch threads" as in "multiple batch requests", like multiple POST operations to the $batch endpoint, then what do you mean/ what does it take for "an application to have an ability to run multiple batches in parallel"?

angelgolfer-ms (changes requested):
@abheek-das , please see my comments and my question on Teams. 
There are a few areas that weren't clear to me and seemed to be ambiguous. In the interest of time, I made some best guesses based on existing content, our prior Teams chat, and the content in your PR, and went ahead to rephrase some of the content in the "Outlook service batching" section in the throttling-limits.md topic in your PR. It might turn out to be even further off from what you meant, for that I apologize in advance... at least then we can pre-emptively make more clarifications so that some customers wouldn't end up down my path :-D
Anyway, please verify what I changed in my commit, and clarify where I overlooked or misunderstood. Thanks.

concepts/throttling-limits.md
abheek-das: As we discussed over teams - Outlook limit of 4 concurrent requests is still per app ID AND mailbox. Only during Batch execution, because each time AGS only sends up to 4 requests to Exchange, there can't be more than 4 requests to the same mailbox anyway and so never crossing that Outlook mailbox concurrency limit.

concepts/throttling-limits.md
abheek-das: I am good either way if we want to call out the previous behavior at some place in the docs. The reason I didn't do so till now was because the previous behavior iss no longer of relevance (as they won't see it happen anymore 😄) - but if we want to do that, I am okay with it!

concepts/throttling-limits.md
abheek-das: Yep, your 3rd guess is correct one. Looks like some customers were doing that in the past to get around the issue with our earlier batch execution (i.e. some of the requests within the batch fail if more than 4 land up on a mailbox at the same time).

concepts/throttling-limits.md
abheek-das: Yes. in #2, the batch has 4+ request to Exchange workload to same or different mailboxes. So as AGS is anyways issuing them 4 at a time, and if they are to the same mailbox - there's no point starting another batch thread again having request to the same mailbox in parallel - as the two batch threads executing in parallel will now conflict with each other, resulting in 4+ requests to the same mailbox which can again result in 429s.

Bottom line is, we provide a capability to batch requests - but clients can issue batch requests in multiple ways. We can provide the best guidance on what to do, but they should be ready/able to handle 429s if they deviate!

concepts/throttling-limits.md
abheek-das: Yes, I mean multiple batch requests issued in parallel or the client issues 4 "POST /batch" calls in parallel!

concepts/throttling-limits.md
abheek-das: Line 388-389 needs to be changed, as we discussed the throttling is still at "per app ID AND mailbox" combination & an app's request to one mailbox does not affect its access to other mailboxes. **Only during batch execution** - where AGS is orchestrating requests to workloads based on the contents of the batch call - it issues 4 at a time if it sees multiple request to Exchange workload, without factoring in whether the requests are to same or different mailboxes!

abheek-das: Line 415 - we can probably get rid of "asynchronously". That might create confusion I think - all I mean is app issues 4 "POST /batch" requests in parallel i.e. issuing multiple batch calls concurrently!

concepts/throttling-limits.md
angelgolfer-ms: Yeah, I see this PR is about describing the current behavior, and, because there wasn't a description of the previous behavior where a batch of more than 5 to 20 individual requests to the Outlook service would have those requests fail, this change of behavior now doesn't really have a default home where customers would go look anyway. 

@abheek-das , was that a common complaint before, do you know? I don't seem to find much in that respect on the Internet.

angelgolfer-ms: FYI I looped us in an email.

dkershaw10: @angelgolfer-ms - yes nothing has changed with batching AFAIK.  And I believe the main document describes most of these limitations (maybe except for only supporting sync requests and that there's no transactional capability in batching).  The known-issues topic seems to talk about functionality that we haven't even shipped (like atomicityGroup) too.

I think we can merge a couple of key lines and remove the known issues topic on batching.

abheek-das: Not much on the public forums, but this was a complaint from few partners who were moving from Outlook REST v2.0 --> MS Graph. They faced this issue only with MS Graph endpoint and not the earlier Outlook endpoint.

Jarbas-MSFT:
Hi @Lauragra, this PR is ready for final review.

Lauragra (approved):

concepts/throttling-limits.md
```diff
@@ -402,6 +405,15 @@ Outlook service limits are evaluated for each app ID and mailbox combination. In
 | Social and workplace intelligence | <li>[person](/graph/api/resources/person) |
 | To-do tasks API (preview) | <li>[outlookTask](/graph/api/resources/outlooktask) <li> [outlookTaskFolder](/graph/api/resources/outlooktaskfolder) <li>[outlookTaskGroup](/graph/api/resources/outlooktaskgroup) <li> [outlookCategory](/graph/api/resources/outlookcategory) <li> [attachment](/graph/api/resources/attachment)|
 
+### Outlook service limits for JSON batching
+
+When an app makes a [JSON batch](json-batching.md) request that consists of multiple, _unordered_ individual requests to the Outlook service, by default, Microsoft Graph sends the Outlook service up to 4 individual requests from the batch at a time, regardless of the target mailboxes of those requests. The Outlook service can execute these requests in parallel at any point, also irrespective of the target mailbox. Since Microsoft Graph sends only up to 4 requests to run in parallel, the execution of that batch stays within [Outlook's concurrency limits for the same mailbox](#limits-per-app-id-and-mailbox-combination). 
+
```
angelgolfer-ms, line 411: @Jarbas-MSFT , thank you for stepping up and helping with this PR while I am OOF. :-)
Just wanted to share some guidelines about using numerals vs spelling out numbers as words: I kept "4" as numeral throughout instead of spelling it out as "four", per the [Cloud +AI guidance](https://styleguides.azurewebsites.net/Styleguide/Read?id=2696&topicid=56449):

"Maintain consistency among categories of information. That is, if one number in a category requires a numeral, use numerals for all numbers in that category. 

Our style
Prepay for your 6-month or 12-month plan.  
Save 15 percent on a five-user subscription.

Not our style
Prepay for your six-month or 12-month plan.  
Save 15 percent on a 5-user subscription."

Line 392 refers to 10,000 (API) requests, and other lines in this Outlook section refers to the magic number 4 requests. So I think they can be considered the same "category of information". 

In addition, IMHO not exactly specified in the above guidance, because this topic is full of values representing limits of various types, in general, it might be easier for customers to scan and not overlook, if we specify numerical values as digits as opposed to spelt-out words.

angelgolfer-ms (approved):
Hi @abheek-das , thanks for your patience on this PR while I am OOF. 
I finally checked in the updates per our communications around 4/13 prior to my unexpected OOFage. Please take a look at my updates from the last few hours when you get a chance. If there's anything I still overlooked, please ping me on Teams.
Otherwise, please set the PR to ready to merge so that @Lauragra can  publish the PR.

Thanks.

PR #20858, opened 2023-04-05
Removed reports section as we fixed the only issue

besiler:
The reports section included an issue describing intermittent Azure AD license check failures. This has been fixed for some time. Since it was the only known issue in that section, removed it


besiler:
@FaithOmbongi FYI - updating here to match the Graph docs: https://github.com/microsoftgraph/microsoft-graph-docs/pull/20842

Lauragra (approved):

FaithOmbongi:
@lramosvea - FYI,  a known issue was resolved and we'd removed it from the docs.

PR #20922, opened 2023-04-13
Fix API Doctor errors flagged by new version

millicentachieng:
This PR fixes the following:
- Casing of path segments
- Paths to use key as path segment format
- Remove Objective C code snippets
- Use fully qualified type name for properties not in `microsoft.graph` namespace
- Add sample keys


api-reference/beta/api/attachment-get.md
```diff
@@ -244,7 +244,7 @@ The next example shows how to get an item attachment on a message. The propertie
 }-->
 
 ```msgraph-interactive
-GET https://graph.microsoft.com/beta/me/messages('AAMkADA1M-zAAA=')/attachments('AAMkADA1M-CJKtzmnlcqVgqI=')
+GET https://graph.microsoft.com/beta/me/messages/AAMkADA1M-zAAA=/attachments/AAMkADA1M-CJKtzmnlcqVgqI=
```
FaithOmbongi, line 247: Please verify if this really needs to change. 
I think what was documented was the correct OData syntax that uses alternate key syntax.

api-reference/beta/api/event-post-attachments.md
```diff
@@ -69,7 +69,7 @@ Here is an example of the request.
 }-->
 
 ```http
-POST https://graph.microsoft.com/beta/me/events('AAMkAGI1AAAt9AHjAAA=')/attachments
+POST https://graph.microsoft.com/beta/me/events/AAMkAGI1AAAt9AHjAAA=/attachments
```
FaithOmbongi, line 72: Same comment about alternate key

FaithOmbongi (changes requested):
Hi @millicentachieng - There are some changes that I suspect aren't valid. The syntaxes that were documented are syntaxes for alternate keys and the syntaxes you propose aren't valid for those particular keys. Please validate those particular files.

api-reference/beta/api/attachment-get.md
millicentachieng: For alternate keys, we expect the name of the property to be specified e.g. `/messages(key='AAMkADA1M-zAAA=')`. In this case, the ids are not alternate keys and they should be added as path segments as in v1.0 docs.

millicentachieng: The documented syntax is correct but for consistency, we need all keys to be added as path segments unless they are alternate keys. 

Another example, 
- `https://graph.microsoft.com/v1.0/users('6e7b768e-07e2-4810-8459-485f84f8f204')` 
- `https://graph.microsoft.com/v1.0/users/6e7b768e-07e2-4810-8459-485f84f8f204` 

Both are correct, but the latter is used in the documentation.

PR #20930, opened 2023-04-14
Update credentialuserregistrationcount.md

kvenkit:


kvenkit:
Adding deprecation note

api-reference/beta/resources/credentialuserregistrationcount.md
```diff
@@ -7,7 +7,7 @@ ms.prod: "identity-and-access-reports"
 doc_type: "resourcePageType"
 ---
 
-# credentialUserRegistrationCount resource type
+# credentialUserRegistrationCount resource type (deprecated)
```
Lauragra, line 10: If this resource is deprecated, do we also need to indicate that the associated method is deprecated? Also, please add (deprecated) to the TOC entry for this resource for visibility.

PR #20945, opened 2023-04-16
Update doctype of resource topics with erroneous value in YAML frontmatter

millicentachieng:


FaithOmbongi (approved):
Hi @millicentachieng - Can we automate this check with API Doctor?
I suspect these cases arose because the authors created the files manually from existing docs instead of using autogenerated stubs.

cc @angelgolfer-ms, @Lauragra

FaithOmbongi:
> Some work had been done previously to verify the structure of documents based on their types. If we can check in these changes, it will help us catch any incorrect document type values. Although the work was not deployed initially because API Doctor was being phased out, we should consider if it's still worth pursuing now that we're still using it.
> 
> Please note that this would require us to ensure that all documents follow a specific heading order, depending on their types. It might also entail updating any documents that violate the ordering, which could be quite extensive. Is this something that we want to pursue? Otherwise, PR reviewers will need to validate the doc-type value manually.

Hmm, we'll need to discuss this. 
We currently follow the heading orders; The violators are:
1. Very old docs.
2. Some docs where the author removed some sections - for example, Properties and Relationships sections in resource docs; and Optional query parameters sections in API docs.

PR #20947, opened 2023-04-17
Add comment to instruct API Doctor not to update permissions table

millicentachieng:


FaithOmbongi (approved):
Approving as these files will be revisited after the initial automation.

PR #20957, opened 2023-04-17
Removing useForSignIn property that isn't in the schema

jpettere:
Remove property that is not in schema 'useForSignIn'

cc @Lauragra - A changelog isn't required in this case because property wasn't really deprecated. The docs just went ahead of the schema.


FaithOmbongi (changes requested):
Hi @jpettere - Please also remove the useForSignIn property from the JSON representation section. Thanks

PR #20975, opened 2023-04-18
docs: adds taxonomy-picker.md to document mgt-taxonomy-picker

anoopt:
Added documentation for `mgt-taxonomy-picker`

To do:
- [ ] `TermStore.Read.All` permission to be provided for the tenant used in the story book.
- [ ] Stories need to updated with correct term set id, term id and site id after that.

If this cannot be done, should the documentation be updated with images as per the [spec](https://github.com/microsoftgraph/microsoft-graph-toolkit/blob/next/fluentui/specs/mgt-taxonomy-picker.md)?


musale:
@anoopt I think images will do for now because we have no ETA on when the permissions will be set.

concepts/toolkit/components/taxonomy-picker.md
```diff
@@ -0,0 +1,151 @@
+---
+title: "Taxonomy picker component in the Microsoft Graph Toolkit"
+description: "The taxonomy picker is a component that can query the Microsoft Graph API for Taxonomy and render a dropdown control with terms, allowing selection of a single term based on the specified term set id or a combination of the specified term set id and the specified term id"
+ms.localizationpriority: medium
+author: anoopt
+---
+
+# Taxonomy picker component in the Microsoft Graph Toolkit
+
+The taxonomy picker is a component that can query the [Microsoft Graph API for Taxonomy](/graph/api/resources/termstore-term?view=graph-rest-1.0) and render a dropdown control with terms, allowing selection of **a single** term based on the specified term set `id` or a combination of the specified term set `id` and the specified term `id`.
```
sebastienlevert, line 10:
```suggestion
The taxonomy picker is a component that can query the [Microsoft Graph API for Taxonomy](/graph/api/resources/termstore-term?view=graph-rest-beta) and render a dropdown control with terms, allowing selection of **a single** term based on the specified term set `id` or a combination of the specified term set `id` and the specified term `id`. 

```

concepts/toolkit/components/taxonomy-picker.md
```diff
@@ -0,0 +1,151 @@
...
+
+![mgt-taxonomy-picker-termset](./images/mgt-taxonomy-picker-termset.png)
+
+The following example shows the use of the `mgt-taxonomy-picker` component to select a child term of a term in a term set.
+
+![mgt-taxonomy-picker-term](./images/mgt-taxonomy-picker-term.png)
+
+The following example shows the use of the `mgt-taxonomy-picker` component to select a child term of a term in a term set in a different language.
+
+![mgt-taxonomy-picker-locale-term](./images/mgt-taxonomy-picker-locale-term.png)
```
sebastienlevert, line 29: To avoid going back to docs and edit, I'd use the same approach to iframe mgt.dev here and just showcase the default story.

sebastienlevert (changes requested):

concepts/toolkit/components/taxonomy-picker.md
sebastienlevert: Just realized the scope issues with the Sandbox... Let's leave it like this for now.

sebastienlevert (approved):

musale (approved):

PR #21106, opened 2023-04-28
Update phoneauthenticationmethod.md

jpettere:
Update description to clarify alternate phones only support voice calls.


FaithOmbongi (approved):
Hi @jpettere - Please apply the same updates to the beta doc as well. Thanks.

PR #21149, opened 2023-05-02
Add notice about required license to ThreatIntelligence APIs

joerattazzi-microsoft:
Threat Intelligence APIs require a license for usage. I've followed a pattern I saw on the InTune pages.

There are no changelog changes here because there was no change to the APIs themselves.


api-reference/beta/api/security-article-get.md
```diff
@@ -13,6 +13,8 @@ Namespace: microsoft.graph.security
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
+> **Note:** The Microsoft Graph API for Threat Intelligence requires an [active Defender Threat Intelligence Portal license and API add-on license](https://go.microsoft.com/fwlink/?linkid=2235706) for the tenant.
```
noah-msft-brabec, line 16: Nit on the phrasing, "The Microsoft Graph API for [Microsoft Defender] Threat Intelligence..." sounds more completed. But if thats not how it gets marketed to customers, as in, if this part of the API from graphs context is commonly referred to as just "Threat Intelligence", then this look good to me.

noah-msft-brabec (approved):

api-reference/beta/api/security-article-get.md
Lauragra: I agree with the suggestion; please apply the change throughout. Another thing to consider is to create an includes file that you insert in every topic, so you only have to maintain the content in one place and it's easier to update if needed.

joerattazzi-microsoft: Great suggestion. That will be much easier. I'll refactor with the include file

joerattazzi-microsoft:
Just a note - I'm trying the change with one file. If that works, I'll refactor the rest

api-reference/beta/api/security-article-get.md
joerattazzi-microsoft: @Lauragra , @noah-msft-brabec  - updated and changed to use an include. I chose to make it a top-level include, so we can reuse the same notice when we go GA.

Example of what this looks like:
image

joerattazzi-microsoft:
@Lauragra  - just checking in on this one. It's been a few weeks. Thanks!

api-reference/includes/threatintelligence-api-disclaimer.md
```diff
@@ -0,0 +1,10 @@
+---
+author: joerattazzi-microsoft
+ms.topic: include
+ms.date: 05/09/2023
+---
+
+<!-- markdownlint-disable MD041-->
+
+> [!IMPORTANT]
```
Lauragra, line 9: I wonder if you want to change this from IMPORTANT to NOTE. On the review site, with the two Importants next to each other, they might lose emphasis. The Note will make the box a different color at least for some contrast.

Lauragra:
> @Lauragra - just checking in on this one. It's been a few weeks. Thanks!

@joerattazzi-microsoft , just made one small comment to differentiate the two alerts.

api-reference/includes/threatintelligence-api-disclaimer.md
joerattazzi-microsoft: @Lauragra - We had done some investigation, and there didn't look to be a standard. 

This one uses two Important labels:
https://learn.microsoft.com/en-us/graph/api/resources/searchentity?view=graph-rest-beta

This one uses a mix and match approach: https://learn.microsoft.com/en-us/graph/api/resources/accessreviewinstancedecisionitem?view=graph-rest-beta

I've updated to Note, per your suggestion

joerattazzi-microsoft:
image

joerattazzi-microsoft:
@Lauragra  - sorry to bug again. We're working on GA documentation going. I'd like to include this there as well if this is good.

Thanks!

Lauragra (approved):

PR #21177, opened 2023-05-04
AAD applications are not supported

woter1832:
Make it abundantly clear AAD applications are not supported

### DIFF
--- a/api-reference/v1.0/api/authenticationmethod-resetpassword.md
+++ b/api-reference/v1.0/api/authenticationmethod-resetpassword.md
@@ -24,6 +24,7 @@ One of the following permissions is required to call this API. To learn more, in
 
 > [!IMPORTANT]
 > The operation cannot be performed on a user's own account. Only an administrator with the appropriate permissions can perform this operation.
+> AAD applications are **not** supported.
 
 |Permission type      | Permissions (from least to most privileged)              |
 |:---------------------------------------|:-------------------------|


FaithOmbongi:
Hi @woter1832 - Thank you for your contribution.

Can you provide more context as to this change? For example, when you say Azure AD applications aren't supported, do you mean that a user can't call this API from a Microsoft-owned application? Please help me understand the comment.

woter1832:
@FaithOmbongi,

Authenticating with an AAD application is not supported. It must be a user.

AAD app = SPN or App registration. (client_id / client_secret).

I wasted a whole day messing around with Graph permissions trying to get this endpoint to work with an AAD app. I want to prevent others going through the same pain by explicitly calling it out.

FaithOmbongi:
> @FaithOmbongi,
> 
> Authenticating with an AAD application is not supported. It must be a user.
> 
> AAD app = SPN or App registration. (client_id / client_secret).
> 
> I wasted a whole day messing around with Graph permissions trying to get this endpoint to work with an AAD app. I want to prevent others going through the same pain by explicitly calling it out.

Hi @woter1832 - Ah, now I've understood.

Yes, that's correct, app-only scenarios aren't supported.
But we already communicate this in the "Application" row of the permissions table which shows "Not supported."
Let me know if this helps.

FaithOmbongi:
Hi @woter1832 - I hope my response cleared up how we document unsupported access scenarios per API. I'll close this PR.

Feel free to file a docs issue and leave suggestions on how we can improve.

PR #21221, opened 2023-05-09

maheshsinghania:
Added external group membership limits per users for search experience. Query will fail the queries from users with higher than 10000 groups memberships


maheshsinghania:
@Lauragra - request your review for this PR.

PR #21237, opened 2023-05-09
Update Key Credentials patch info

franqq:
Provide more info on patching to prevent partners from getting blocked while patching without correct type and usage combination

Incident: https://portal.microsofticm.com/imp/v3/incidents/details/384705352/home


FaithOmbongi (approved):
Hi @franqq - I've left one suggestion for minor formatting updates. Otherwise, approved with suggestions.

ngash (approved):

PR #21255, opened 2023-05-11
Update accesspackageassignmentpolicy-update.md

saylee-dharne-msft:
Updated the documentation. We support PUT but not PATCH for the assignment policy, so the linked text snippet (I included body below) does not apply to it. 

"In the request body, supply only the values for properties that should be updated. Existing properties that are not included in the request body will maintain their previous values or be recalculated based on changes to other property values."


markwahl-msft (approved):

FaithOmbongi (approved):
Hi @saylee-dharne-msft - Please also update the beta doc. Thank you.

PR #21354, opened 2023-05-19
Update the max size value limitation.

yiwenwang-ms:


Danielabom (approved):
Based on my chat with @yiwenwang-ms, there are no schema changes. PR approved.

PR #21368, opened 2023-05-19
EM assignment policy update: add table description for request body

markwahl-msft:
Last weeks' change #21255 removed the incorrect include from accesspackageassignmentpolicy-update.  The beta article has a sentence on what the structure of the request body is, before the table, but the 1.0 article just has the table.  This adds that sentence to the 1.0 article, in place of the no longer applicable include, that specifies what is the payload structure of the request body.  (This question came from a support request).


FaithOmbongi (approved):
Hi @markwahl-msft - I've added another text...approved with comments.

PR #21383, opened 2023-05-23
Fix typo in audioCodec enum

mcm223:
Fixing a typo in the audioCodec enum to match the exact casing returned from the API.

Resolves Issue #21372


Danielabom (approved):
PR approved.

PR #21482, opened 2023-05-30
Fix doc bug #17962

keylimesoda:
Sample incorrectly suggests loading dataKey property, when it is the data property that should be included at this stage.

As it was written, you'd just be decrypting the encryption key again instead of decrypting the payload itself.

https://github.com/microsoftgraph/microsoft-graph-docs/issues/17962


Danielabom (approved):
PR approved.

Issue #20602, opened 2023-03-15
"Unrecognized query argument specified: ''." when using filter with  https://graph.microsoft.com/beta/groups/delta


gfree76:
When I try to use a filter against the beta/groups/delta endpoint, I get an error regardless of what I try passing in. 

here's an example: 
```
curl -X GET "https://graph.microsoft.com/beta/users/delta/?$filter=startsWith%28displayName%2C%20%27Geoff%27%29" \
    -H "Authorization: Bearer $TOKEN" | jq '.' 
```
returns
```
{
  "error": {
    "code": "Request_BadRequest",
    "message": "Unrecognized query argument specified: ''.",
    "innerError": {
      "date": "2023-03-15T16:47:28",
      "request-id": "c64282b0-4d76-48d3-9d81-c0b213874b7e",
      "client-request-id": "c64282b0-4d76-48d3-9d81-c0b213874b7e"
    }
  }
}
```

I get the same thing when I try the format from https://learn.microsoft.com/en-us/graph/api/group-list?view=graph-rest-1.0&tabs=http#request-1
```
curl -X GET "https://graph.microsoft.com/beta/groups/delta/?$filter=hasMembersWithLicenseErrors+eq+true" \
    -H "Authorization: Bearer $TOKEN" | jq '.' 
```

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: bdaf58ca-8b6c-cddd-f583-f8617b0450a6
* Version Independent ID: 08b26245-df55-d5d5-a6d2-edc69eda3fe3
* Content: [Get incremental changes for groups - Microsoft Graph](https://learn.microsoft.com/en-us/graph/delta-query-groups?tabs=http)
* Content Source: [concepts/delta-query-groups.md](https://github.com/microsoftgraph/microsoft-graph-docs/blob/main/concepts/delta-query-groups.md)
* Product: **non-product-specific**
* Technology: **microsoft-graph**
* GitHub Login: @FaithOmbongi
* Microsoft Alias: **ombongifaith**

FaithOmbongi:
Thank you @gfree76 for raising this issue.

[...]

PR #21484, opened 2023-05-30
Clarifying that DQ can only filter by Object ID

keylimesoda:
Fix for this bug here:  https://github.com/microsoftgraph/microsoft-graph-docs/issues/20602

The initial bug was opened against an example, but we've updated the overview section to emphasize this clarification, rather than repeating it in each example.


Danielabom (approved):
PR approved.

PR #21485, opened 2023-05-31
Breadcrumb investigation

kcpitt:


Lauragra:
@kcpitt , what is the expected result you want to see with this PR?

kcpitt:
> @kcpitt , what is the expected result you want to see with this PR?

Breadcrumb only shows "Learn" which is the platform fallback. Should be "Learn | Microsoft Graph"

this was working a few months ago and then it stopped. Working to troubleshoot.

Lauragra:
@mimisasouvanh , do you know if we need to update the TOC file as well? See this PR: https://github.com/microsoftgraph/microsoft-graph-docs/pull/21485/files

PR #21509, opened 2023-06-02
Update authorizationpolicy-update.md with roles

myra-ramdenbourg:


api-reference/v1.0/api/authorizationpolicy-update.md
```diff
@@ -23,6 +23,9 @@ One of the following permissions is required to call this API. To learn more, in
 | Delegated (personal Microsoft account) | Not supported. |
 | Application                            | Policy.ReadWrite.Authorization|
 
+When calling on behalf of a user, the user needs to belong to the following [Azure AD roles](/azure/active-directory/roles/permissions-reference):
++ Privileged Role Administrator
```
FaithOmbongi, line 27: It's only one role (well would be two but we're avoiding the Global Admin role) so no need for a list.
```suggestion
When calling on behalf of a user, the user needs to have the *Privileged Role Administrator* [Azure AD role](/azure/active-directory/roles/permissions-reference).
```

FaithOmbongi (approved):
Hi @myra-ramdenbourg - I've left a suggestion inline. Please address it and apply the same change to the beta file before merging this PR.

Otherwise, approved with comments.

FaithOmbongi (approved):
Accidentally set the previous approval. Check the previous approved with suggestions comments.

PR #21539, opened 2023-06-06
Update identity-governance-pim-rules-overview.md

chkoide:
Changing the setting from Azure portal, rule name in Graph API seems to be different. I would be appreciated if anyone check the appropriate rule name of No.2 and No.8.


FaithOmbongi (approved):
Thank you @chkoide for catching this error. I've pushed a commit before approving your changes. This PR will be merged on Thursday.

FaithOmbongi (approved):

PR #21562, opened 2023-06-07
Best practices: remove mention of x-ms-ags-diagnostic

syprieur:
Remove mention of `x-ms-ags-diagnostic`:
 - It was never intended for customers to log.
 - It is not required nor helpful to diagnose customers MSGraph requests.


dkershaw10 (approved):
LGTM

FaithOmbongi (approved):

PR #21580, opened 2023-06-08
Embed video for change notifications using Event Hubs

FaithOmbongi:


concepts/change-notifications-delivery-event-hubs.md
```diff
@@ -16,6 +16,8 @@ Good examples of high throughput scenarios include applications subscribing to a
 
 The article guides you through the process of managing your Microsoft Graph subscription and how to receive change notifications through Azure Event Hubs.
 
+> [!VIDEO https://www.youtube-nocookie.com/embed/d_T_DYwmN3w]
```
Lauragra, line 19: @FaithOmbongi , FYI, embedded YouTube videos are discouraged on the Learn platform (even though we do have this in our docs as a legacy) - see the guidance here: https://review.learn.microsoft.com/en-us/help/platform/contribute-video-publish?branch=main#youtube-and-other-hosting-platforms

I'm going to hold off on merging this until we decide whether we want to continue this practice as it is not in line with the guidance.

FaithOmbongi: Ah, thanks for making me aware @Lauragra. 

Q: Are we allowed to link to the YouTube video in the "See also" section which means customers will be sent out of the Learn platform?

PR #21588, opened 2023-06-08
Rich notifications doc refresh (Batch 1)

keylimesoda:
First part of re-work rich notification document for clarity


FaithOmbongi (approved):
LGTM!

concepts/includes/change-notifications-rich-notifications-supported-resources.md
```diff
@@ -0,0 +1,50 @@
...
+| Teams [channel][] | Changes to channels in all teams: `/teams/getAllChannels`<br/><br/>Changes to channel in a specific team: `/teams/{id}/channels` | - |
+| Teams [conversationMember][] | Changes to membership in a specific team: `/teams/{id}/members`<br/><br/><!-- Changes to membership in all channels under a specific team:  `/teams/{id}/channels/getAllMembers` --><br/><br/>Changes to membership in a specific chat: `/chats/{id}/members`<!-- <br/><br/>Changes to membership for all Teams chats: `/chats/getAllMembers`<br/><br/>Changes to membership for all channels across the entire tenant: `/teams/getAllChannels/getAllMembers` * --> | - |
```
FaithOmbongi, line 24: Hi @keylimesoda - I'm unable to trace these endpoints to the Graph schema or any API reference documentation:

- /teams/{id}/channels/getAllMembers
- /teams/{id}/channels/getAllMembers
- /chats/getAllMembers

Ideally, we shouldn't document them (because there's no related schema) but I wonder if there are customers making use of these API endpoints and it's just a case of the schema being behind? Can you help to verify?

I've commented them out in the meantime.

FaithOmbongi (approved):
Everything LGTM.

 I've pushed a commit to convert the supported resources into a table, please review. I've also pointed out a case of three endpoints that are schemaless and might need some internal investigation.

concepts/includes/change-notifications-rich-notifications-supported-resources.md
keylimesoda: I believe these are the references for the APIs (you only mentioned two endpoints in your notes, first one was repeated?)

teams/{id}/channels/getAllMembers is here:  https://learn.microsoft.com/en-us/graph/teams-changenotifications-teammembership

chats/getallmembers appears in the API documentation here: https://learn.microsoft.com/en-us/graph/teams-changenotifications-chatmembership

FaithOmbongi: _Fixed to show the three endpoints._ 

Notice both references you're linking to are in the change notifications conceptual docs. Before a resource/endpoint can support change notifications, it should be valid and obviously documented in ⁠[the API reference docs](https://learn.microsoft.com/en-us/graph/api/overview?view=graph-rest-beta) which is where devs would go for info on licensing, permissions, payload structure, etc. But there's no API reference doc for the three endpoints and the `getAllMembers` function isn't ⁠[in the schema](https://graph.microsoft.com/beta/$metadata) either.

PR #21607, opened 2023-06-09
Small tweak to set expectations for when call records will be available

keylimesoda:
Based on customer incident https://portal.microsofticm.com/imp/v3/incidents/details/395238914/home

Helping set customer expectations correctly to avoid confusion.


FaithOmbongi (approved):
LGTM!

PR #21621, opened 2023-06-12
Update identity set for v1.0

arifsaikat-microsoft:
NICE would like to have a clear documentation with examples on different identity types, so that they can filter out the participant reporting on their side. Without filtering applications/bots in the report, it causes confusion and compliance issues. 

They pointed out this doc, which does not have clear description of different identity types and the example JSON has additional types (for example encrypted) that is not listed in the types.


Danielabom (approved):
@arifsaikat-microsoft, there is a build issue. Please review the logs:  /resources/permission.md/grantedToIdentities/member:
       Missing properties: response was missing these required properties: applicationInstance, conversation, conversationIdentityType, encrypted, onPremises, guest, phone

PR #21641, opened 2023-06-14
Update authentication-update.md

DanielBradley1:
I have removed the option for 'UnknownFutureValue' as it is reporting as an unsupported authentication type.


jpettere (approved):

Lauragra:
@jpettere , the metadata is still showing unknownFutureValue as a member. Are you sure we should merge this?

PR #21644, opened 2023-06-14
[Azure AD] Custom security attribute assignments promote to v1.0

rolyon:


FaithOmbongi (approved):
LGTM! Thanks @rolyon!

PR #21676, opened 2023-06-16
TEST - Add new doc stubs with permission stubs generated

wandeg:


api-reference/beta/includes/permissions/user-delete-testjamesnoperms-permissions.md
```diff
@@ -0,0 +1,6 @@
+|Permission type|Least privileged permission|Higher privileged permissions|
```
Lauragra, line 1: Global: Please make sure that the include files include the required metadata, as defined in the spec.

api-reference/beta/includes/permissions/user-delete-testjamesnoperms-permissions.md
```diff
@@ -0,0 +1,6 @@
...
+|:---|:---|:---|
+|Delegated (work or school account)|Not supported.|Not supported.|
```
Lauragra, line 3: Global: Many of the permission tables don't appear to actually list any permissions. Is that by design? Can we try a test where we pull permissions into tables in the include files?

FaithOmbongi (commented):
Hi @wandeg - I've left a few comments inline + agree with Laura's suggestion on the metadata. Please review.

PR #21739, opened 2023-06-20
Update intune-androidforwork-devicemanagement-get.md

SteveWilkins1123:


Lauragra:
@SteveWilkins1123 , how are you making sure that these updates to the Intune files won't get overwritten in the next drop?

SteveWilkins1123:
The scripts will have to be run each time the docs get over written to inject the meta data once again

PR #21760, opened 2023-06-22
Change app role assignment least privilege permissions for user and groups

baprzyst:


baprzyst:
@microsoft-github-policy-service agree company="Microsoft"

FaithOmbongi:
Hi @psignoret - Please help me validate this change + whether it should also apply to delegated permissions. Thanks.

psignoret:
@FaithOmbongi, Application.ReadWrite.OwnedBy requires that the calling app's service principal is owner of the resource service principal (identified by resourceId) in the app role assignment. I think it's important that this requirement be included, because otherwise some developers will be confused why only having Application.ReadWrite.OwnedBy and User.Read.All isn't sufficient.

All the app role assignment articles would need to be updated accordingly, for both beta and v1.0: POST users/{id}/appRoleAssignments, POST groups/{id}/appRoleAssignments, POST servicePrincipals/{id}/appRoleAssignments, and POST servicePrincipals/{id}/appRoleAssignedTo.

PR #21765, opened 2023-06-22
Update application-post-onlinemeetings.md

mervynmanilall:
Hello,

I would like to submit this PR to contribute to the application-post-onlinemeetings.md
which is the documentation for configuring creating teams meeting link using Microsoft Graph. I noticed that we could add the link for the documentation that shows how to get the token for the authorization. This will greatly help developers to achieve this faster.

In this PR, I have added a line that will provide the url to get the token for the authorization
I hope this addition is useful and informative for your users. Please let me know if there are any changes you would like me to make or if you have any feedback.

Thank you for your time and consideration.

### DIFF
--- a/api-reference/v1.0/api/application-post-onlinemeetings.md
+++ b/api-reference/v1.0/api/application-post-onlinemeetings.md
@@ -49,6 +49,9 @@ POST /users/{userId}/onlineMeetings
 
 If the request contains an `Accept-Language` HTTP header, the `content` of `joinInformation` will be in the language and locale variant specified in the `Accept-Language` header. The default content will be in English.
 
+
+To get the `token` for the `Authorization`, follow [Get access on behalf of a user](/graph/auth-v2-user?tabs=http#3-request-an-access-token).
+
 ## Request body
 In the request body, supply a JSON representation of an [onlineMeeting](../resources/onlinemeeting.md) object.
 


mervynmanilall:
@Lauragra / @ailae , can you check my changes please

api-reference/beta/api/application-post-onlinemeetings.md
```diff
@@ -51,6 +51,8 @@ POST /users/{userId}/onlineMeetings
 
 If the request contains an `Accept-Language` HTTP header, the `content` of `joinInformation` will be in the language and locale variant specified in the `Accept-Language` header. The default content will be in English.
 
+To get the `token` for the `Authorization`, follow [Get access on behalf of a user](/graph/auth-v2-user?tabs=http#3-request-an-access-token).
```
Lauragra, line 54: Suggested revision for this text:
For details about how to get the `token` for the `Authorization` header, see [Get access on behalf of a user](/graph/auth-v2-user?tabs=http#3-request-an-access-token).

Lauragra, line 54: This header is common to all methods; is there a reason to provide this cross-reference in just these topics, or is the suggestion to add this more globally?

mervynmanilall: Hello,

This documentation does not explicitly mention where and how to get the 'Authorization', despite mentioning that it is required. To get it, more search is required, which may lead to other sources and more time spent.
This documentation is already very good and providing all the information required here itself will make it the single source of truth, which is better.
Adding this reference globally to all the places requiring it, is also a very good suggestion. I have implemented only this one so far; that is why I did the suggestion here.

Regards.

mervynmanilall:
@FaithOmbongi , can you review my changes please

mervynmanilall:
Dear @Lauragra , any update please?

mervynmanilall:
@FaithOmbongi , can you review my changes please?

FaithOmbongi:
Hi @mervynmanilall - First, thank you for suggesting these changes to the docs. We value developer feedback and appreciate that this kind of feedback that will help most developers, especially those new to Microsoft Graph, to be successful in working with the APIs.

I understand the intent behind the change, but I'm not sure that adding this statement is the most impactful way to go for the following reasons:

| Feedback | Suggestions                                                         |
|---------------|-----------------------------------------------------------------|
|The statement links to only the [delegated access guidance](https://learn.microsoft.com/en-us/graph/auth-v2-user?tabs=http) yet the API supports both delegated and [app-only access](https://learn.microsoft.com/en-us/graph/auth-v2-service?tabs=http) which are done differently.        | Linking to both guidance is important. |
| The authorization requirement applies to _all_ of Microsoft Graph and therefore all APIs, not just the two APIs.          | Therefore, this cross-linking to the access tokens guidance would be useful for all API docs. For this reason, I'd recommend that our team implements a single consistent cross-linking in _all_ API docs. |
| The "Request headers" table in all API docs mentions that an app needs a bearer token to successfully call the API.        | Clearly, we need more information and that's why you proposed the update to cross-link to the guidance. But maybe we can cross-link from the table row, for all API docs, instead of having to add a statement in individual docs. This is in the spirit of my second statement about consistency and I think applying this change will address your feedback for all Microsoft Graph docs. |

You've shared the following useful feedback that we can run with to help us start the internal process.

> I noticed that we could add the link for the documentation that shows how to get the token for the authorization. This will greatly help developers to achieve this faster.

And also your answer to @Lauragra's comment:

> Hello,
> 
> This documentation does not explicitly mention where and how to get the 'Authorization', despite mentioning that it is required. To get it, more search is required, which may lead to other sources and more time spent. This documentation is already very good and providing all the information required here itself will make it the single source of truth, which is better. Adding this reference globally to all the places requiring it, is also a very good suggestion. I have implemented only this one so far; that is why I did the suggestion here.
> 
> Regards.

Please allow us to discuss your feedback internally and identify a way to have all docs consistently point to the guidance which I think will make your suggestion have the most impact.

PR #21769, opened 2023-06-22
Usage examples for authentication behaviors API

medhir:
Adds API examples for listing, setting, and un-setting authentication behaviors. 

Originally planned to have this as just a resource type per discussion with @FaithOmbongi, but want to ensure that customers who are going through [this recently published MSRC advisory](https://msrc.microsoft.com/blog/2023/06/potential-risk-of-privilege-escalation-in-azure-ad-applications/) have adequate support in addressing security fixes.


api-reference/beta/api/authenticationbehaviors-get.md
```diff
@@ -0,0 +1,83 @@
...
+GET /applications(appId='{appId}')/authenticationBehaviors/removeUnverifiedEmailClaim
+```
+
+### Response
+
+The following is an example of the response.
+
+<!-- {
+  "blockType": "response",
+  "@odata.type": "boolean"
```
medhir, line 72: Do you know what the correct type to place here is @FaithOmbongi? This keeps failing during API doctor validation

FaithOmbongi: The @odata.type would be microsoft.graph.application behause authentication behaviors is a property of the application entity.

FaithOmbongi (changes requested):
Hi @medhir -

I understand the need to increase visibility for authenticationBehaviors, but we should do it in a way that customers and downstream tools will undertand the docs.

The API reference docs must always be mappable to the schema. Even the file names follow the file naming guidelines and they are understood by upstream and downstream tooling because they align with the guidance. So we can’t create new authenticationBehaviors API docs because you only have API docs for entities, and authenticationBehaviors isn’t one. Even if we published the docs, upstream tooling like permissions automation model won’t understand them, and downstream tooling like all SDKs won’t understand them either, because neither expects these docs.

I pushed a commit to start off a discussion and shared the reasoning internally. See the details of the cmmit here https://github.com/microsoftgraph/microsoft-graph-docs/pull/21769/commits/3e5965247752d06fbeeedcd217bc4704678bcad8

If this makes sense, please feel free to make any edits to the new conceptual article.

medhir:
Closing in favor of https://github.com/microsoftgraph/microsoft-graph-docs/pull/21787

PR #21842, opened 2023-06-27
Update addActivities urls with external segment

sacampbe-msft:
The "external" url segment is missing on these pages. Updating to avoid future churn and confusion.


emzho (approved):

Danielabom (approved):
PR approved.

PR #21938, opened 2023-07-05
07 05 22 jupetter authmethods

jpettere:
Remove add/edit on self for phone and email methods

### DIFF
--- a/api-reference/beta/api/authentication-post-emailmethods.md
+++ b/api-reference/beta/api/authentication-post-emailmethods.md
@@ -1,7 +1,7 @@
 ---
 title: "Create emailAuthenticationMethod"
 description: "Create a new emailAuthenticationMethod object."
-author: "mmcla"
+author: "jpettere"
 ms.localizationpriority: medium
 ms.prod: "identity-and-sign-in"
 doc_type: "apiPageType"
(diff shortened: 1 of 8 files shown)


api-reference/beta/api/authentication-post-emailmethods.md
```diff
@@ -35,6 +25,8 @@ One of the following permissions is required to call this API. To learn more, in
 
 [!INCLUDE [rbac-authentication-methods-apis-write-others](../includes/rbac-for-apis/rbac-authentication-methods-apis-write-others.md)]
 
+Users without one of the supported Azure AD roles cannot add their own email authentication method.
```
FaithOmbongi, line 28: Hmm @jpettere - Please confirm...Is this the logic we're going by and the authZ behavior? I don't think it really encourages least-privilege decisions.

- Global Admin with UserAuthenticationMethod.ReadWrite _which is least privileged permission scoped to self_ **cannot** update their auth method.
- Global Admin needs _higher privileged permission UserAuthenticationMethod.ReadWrite.All which is scoped to all users_ to update **their own** auth method.

jpettere: @FaithOmbongi some context:

I learned we don't support self operations (except for in certain 1st party apps) so needed to remove anything about self in the update / create APIs for phone and email methods. I saw that the beta doc for updating phone methods already didn't have anything about self operations except for this one liner about certain roles so I copied it. I just testing it out though and don't think it works for these roles either. Could you possibly test to confirm? Then I'll remove the sentence entirely.

jpettere: Updated the descriptions and removed the /me syntax. The limitation is only for create/update operations which is only supported for phone and email today. I think the RBAC note is used in some other docs so for now I've kept it separate.

I see I'm getting a build error due to the example in the beta patch phone methods doc, but its the same example used in all the other ones so I'm not entirely sure what's wrong 😕.

FaithOmbongi (approved):
Hi @jpettere - I've pushed a commit to resolve the validation error and left one comment to address.

PR #21953, opened 2023-07-06
Update MGT providers.md to support CustomHosts

techienickb:
Update MGT providers.md to support customHosts


sebastienlevert:
@gavinbarron please have a look at this PR as it's not out-of-sync with v3.0.1! I don't have all the context to be able to review. Thanks!

PR #21968, opened 2023-07-07
Update permission-grant.md

programaths:
Added the encoding algorithm.

That would greatly help users!


programaths:
@microsoft-github-policy-service agree

PR #21990, opened 2023-07-11
Update TOC to include SLA

jamesw-msft:
Update TOC to include SLA


msewaweru (changes requested):
@jamesw-msft - See my comment below for your action.

msewaweru (approved):
@jamesw-msft -LGTM! Thanks.

api-reference/beta/toc.yml
```diff
@@ -15348,6 +15348,16 @@ items:
               href: api/reportroot-list-applicationsignindetailedsummary.md
             - name: Get sign-in details
               href: api/applicationsignindetailedsummary-get.md
+        - name: Azure AD SLA attainment
+          items:
+          - name: Azure AD Authentication
+            href: resources/azureadauthentication.md
+          - name: Service Level Agreement
+            href: resources/servicelevelagreementroot.md
+          - name: Service Level Agreement attainment
```
Lauragra, line 15357: Per our guidelines, we don't surface complex types in the TOC. I think here the TOC structure can simply be:

Azure AD authentication
    Get

Also note that per our guidelines, we use sentence case in the TOC, so only the first letter is capitalized except for proper nouns.

jamesw-msft: Removed the complex type and fixed the casing per our chat offline.

PR #21993, opened 2023-07-11
promote teams app settings to v1.0

sweta-thapliyal:


api-reference/v1.0/api/teamsappsettings-update.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+#### Response
+
+<!-- {
+  "blockType": "response"
+} -->
+
+```http
+HTTP/1.1 204 No Content
```
AkJo, line 86: I do not think that this is surfaced in V1.0 yet. We should remove it.

PR #22014, opened 2023-07-11
Update customextensionclientconfiguration.md

soneff:
Add documentation for maximumRetries in clientConfiguration


api-reference/beta/resources/customextensionclientconfiguration.md
```diff
@@ -13,12 +13,13 @@ Namespace: microsoft.graph
 
 [!INCLUDE [beta-disclaimer](../../includes/beta-disclaimer.md)]
 
-Connection settings that define how long Azure AD can wait for a a response from an external app before it shuts down the connection when trying to trigger the external app. Only **timeoutInMilliseconds** is supported.
+Connection settings that define how long Azure AD can wait for a a response from an external app before it shuts down the connection when trying to trigger the external app. Only **timeoutInMilliseconds** and **maximumRetries** are supported.
```
FaithOmbongi, line 16: The second statement is redundant because the two are all the properties in this resource.
```suggestion
Connection settings that define how long Azure AD can wait for a a response from an external app before it shuts down the connection when trying to trigger the external app.
```

api-reference/beta/resources/customextensionclientconfiguration.md
```diff
@@ -13,12 +13,13 @@ Namespace: microsoft.graph
...
 
 ## Properties
 |Property|Type|Description|
 |:---|:---|:---|
-|timeoutInMilliseconds|Int32|The max duration in milliseconds that Azure AD will wait for a response from the external app before it shuts down the connection. The valid range is between `200` and `2000` milliseconds. Default duration is `1000`.|
+|timeoutInMilliseconds|Int32|The max duration in milliseconds that Azure AD will wait for a response from the external app before it shuts down the connection. The valid range is between `200` and `2000` milliseconds. If null, the service default will apply.|
+|maximumRetries|Int32|The max number of retries that Azure AD will make to the external API. Values of 0 or 1 are supported. If null, the service default will apply.|
```
FaithOmbongi, line 22: What are the "service default" values?
```suggestion
|timeoutInMilliseconds|Int32|The max duration in milliseconds that Azure AD will wait for a response from the external app before it shuts down the connection. The valid range is between `200` and `2000` milliseconds. If `null`, the service default will apply.|
|maximumRetries|Int32|The max number of retries that Azure AD will make to the external API. Values of 0 or 1 are supported. If `null`, the service default will apply.|
```

FaithOmbongi (approved):
Thanks @soneff for making this fix. Left two minor suggestions and approved with comments.

api-reference/beta/resources/customextensionclientconfiguration.md
soneff: For the custom auth extensions, the default values are 2000ms and 1 retry. But from eng it sounds like other services could set different default values. Let me see if we could convey that more clearly...

PR #22041, opened 2023-07-12
v1 labels and fpd - resource and api

sseth-msft:


api-reference/v1.0/resources/security-labelsroot.md
```diff
@@ -0,0 +1,67 @@
+---
+title: "labelsRoot resource type"
+description: "Represents the entry point for the retentionLabels resource type."
+author: "sseth"
+ms.localizationpriority: medium
+ms.prod: "security"
+doc_type: resourcePageType
+---
+
```
angelgolfer-ms, line 9: Please update the security-api-overview.md and security-recordsmanagement-overview.md as I explained in my email (8/2 afternoon).

api-reference/v1.0/toc.yml
```diff
@@ -7912,32 +7912,116 @@ items:
...
-                    href: api/security-retentioneventtype-post.md
-                  - name: Get
-                    href: api/security-retentioneventtype-get.md
-                  - name: Update
-                    href: api/security-retentioneventtype-update.md
-                  - name: Delete
-                    href: api/security-retentioneventtype-delete.md
+            - name: Overview
+              href: resources/security-recordsmanagement-overview.md
+            - name: Label
```
angelgolfer-ms, line 7917: FYI to be consistent with how and what we expose in the TOC for other sets of API in Graph, I'm going to remove the "Label" topic from the TOC. This is a "root" kind of entity, it's sufficient to document the entity but not expose it in the TOC.

angelgolfer-ms: Also, in the beta TOC, you should have removed the "(preview)"  label that's been appended to the "Retention label" TOC node. In general, when you GA an API set, that set is no longer in preview and in the beta reference, we should not mark it with "(preview)" anymore. 
More info about versioning and "preview" here: https://learn.microsoft.com/en-us/graph/versioning-and-support

api-reference/v1.0/api/security-authoritytemplate-get.md
```diff
@@ -0,0 +1,92 @@
...
+<!-- {
+  "blockType": "response",
+  "truncated": true,
+  "@odata.type": "microsoft.graph.security.authorityTemplate"
+}
+-->
+``` http
+HTTP/1.1 200 OK
+Content-Type: application/json
+
```
angelgolfer-ms, line 79: GLOBAL:
@sseth-msft , I just noticed, all the beta and v1 API topics documented for the various types of file plan descriptor templates have example requests and responses that contain placeholder (id values in the GET and delete operation topics) and type information in the response body (list, GET, create operation topics)!
In an example request that involves an id, you need to specify a real-looking ID value.
In an example response that involves a response body, you need to provide real-looking value for each of the properties in the response body.

Examples are useless if they don't show real-looking values.

Can you please take this work item?

api-reference/v1.0/resources/security-authoritytemplate.md
```diff
@@ -0,0 +1,58 @@
...
+|Method|Return type|Description|
+|:---|:---|:---|
+|[List authorities](../api/security-labelsroot-list-authorities.md)|[microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md) collection|Get a list of the [microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md) objects and their properties.|
+|[Create authorities](../api/security-labelsroot-post-authorities.md)|[microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md)|Create a new [microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md) object.|
+|[Get authorities](../api/security-authoritytemplate-get.md)|[microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md)|Read the properties and relationships of a [microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md) object.|
+|[Delete authorities](../api/security-labelsroot-delete-authorities.md)|None|Delete a [microsoft.graph.security.authorityTemplate](../resources/security-authoritytemplate.md) object.|
+
+## Properties
+|Property|Type|Description|
+|:---|:---|:---|
```
angelgolfer-ms, line 28: GLOBAL:
@sseth-msft , for both the beta and v1 versions of each of the 6 types derived from filePlanDescriptorTemplate, you have the description for the createdBy, createdDateTime, displayName properties incorrect.

Instead of saying:
Inherited from [microsoft.graph.security.filePlanDescriptor](../resources/security-fileplandescriptor.md).

you should say:
Inherited from [microsoft.graph.security.filePlanDescriptorTemplate](../resources/security-fileplandescriptortemplate.md).

Can you please fix them?  (Total 3x6x2=36 descriptions.)

api-reference/v1.0/resources/security-fileplandescriptor.md
```diff
@@ -0,0 +1,73 @@
...
+ms.localizationpriority: medium
+ms.prod: "security"
+doc_type: resourcePageType
+---
+
+# filePlanDescriptor resource type
+
+Namespace: microsoft.graph.security
+
+Represents a group of columns to improve the manageability and organization of the content you need to label.
```
angelgolfer-ms, line 14: Line 14, in both v1 and beta:
The use of the term "column" here, in this context in the developer docs, is not helpful, bordering confusing. 
It seems like other than this topic and security-fileplandescriptortemplate.md (which currently share the same description), there's no other mention of what you mean by a column or where such "columns" appear.

Perhaps this came from the admin docs section title "[File plan descriptor columns](https://learn.microsoft.com/en-us/purview/file-plan-manager?view=o365-worldwide#file-plan-descriptors-columns)"?
In that topic, because they explain and show a screenshot of a file plan in the Microsoft Purview compliance center, it might be more understandable what a "column" means.

But in the developer docs, we don't usually show as much admin UI-side content, so let's just not explain descriptors by introducing another term "column".

angelgolfer-ms: Applicable to both v1 and beta:
My next comment is about how this resource is named ("filePlanDescriptor"), what properties it contains, and, how we describe this resource and each of the 6 actual descriptors. 

(a) According to the [admin doc explanation](https://learn.microsoft.com/en-us/purview/file-plan-manager?view=o365-worldwide#file-plan-descriptors-columns), even though this resource is named "filePlanDescriptor", this is actually a _file plan_ that contains optional, actual descriptors. 

(b) Each of what we call in the developer docs - appliedCategory, authority, citation, department, filePlanDescriptorReference, and subCategory - corresponds to a file plan descriptor in the 2 screenshots of the compliance center in that [section](https://learn.microsoft.com/en-us/purview/file-plan-manager?view=o365-worldwide#file-plan-descriptors-columns).

So, you can see it's illogical to name this resource as filePlanDescriptor (singular descriptor), and define it with 5 properties each of which is an optional descriptor, as described in the admin docs. (Being aligned conceptually with the admin docs is important because often the similar admin audience uses these features either in the admin center or programmatically.)

To make it less ambiguous, FYI I'm changing the description for filePlanDescriptor as follows:
```
Represents a set of optional descriptors to supplement a [retention label](security-retentionlabel.md) and improve the manageability and organization of content in Microsoft 365 for an organization.
```

api-reference/v1.0/resources/security-fileplandescriptorbase.md
```diff
@@ -0,0 +1,37 @@
...
+ms.localizationpriority: medium
+ms.prod: "security"
+doc_type: resourcePageType
+---
+
+# filePlanDescriptorBase resource type
+
+Namespace: microsoft.graph.security
+
+Represents the name for each [file plan descriptor](../resources/security-fileplandescriptor.md) column associated with a particular label.
```
angelgolfer-ms, line 14: Applicable to both v1 and beta:
Line 14:
Per my comment in security-fileplandescriptor.md, the use of "column" is not helpful. FYI I'm going to remove that term.

angelgolfer-ms: Also, the link to fileplandescriptor.md is inappropriate. As mentioned in my comments for that topic, that resource is not an individual descriptor, it represents a set of descriptors.
So instead of linking to that resource, we should actually mention this resource is the base type for each of the actual 6 descriptor resources, and link to them.

api-reference/v1.0/resources/security-fileplanreference.md
```diff
@@ -0,0 +1,40 @@
...
+ms.localizationpriority: medium
+ms.prod: "security"
+doc_type: resourcePageType
+---
+
+# filePlanReference resource type
+
+Namespace: microsoft.graph.security
+
+Represents the file plan descriptor of type filePlanReference applied to a particular retention label.
```
angelgolfer-ms, line 14: For both v1 and beta:
Line 14:
FYI the description of each of the 6 descriptors actually doesn't explain the purpose of the descriptor or what it is meant to be used for, because the description simply repeats the same term as the name of the resource.
While some of the descriptor names can be considered intuitive, I find filePlanReference more ambiguous, for example, is this about a reference made to the file plan, or a reference this file plan makes?

After checking the screenshot in the admin [topic](https://learn.microsoft.com/en-us/purview/file-plan-manager?view=o365-worldwide#file-plan-descriptors-columns) about the Microsoft Purview compliance portal, then I realize this is meant to be a reference ID.

FYI I am going to change this description to make it more informative.

In general, please remember that in any description in the docs, do not use only the same words as the name of that API element, be it a resource, property, method, enumeration member :-)

api-reference/v1.0/resources/security-departmenttemplate.md
```diff
@@ -0,0 +1,59 @@
...
+|[Create departments](../api/security-labelsroot-post-departments.md)|[microsoft.graph.security.departmentTemplate](../resources/security-departmenttemplate.md)|Create a new [microsoft.graph.security.departmentTemplate](../resources/security-departmenttemplate.md) object.|
+|[Get departments](../api/security-departmenttemplate-get.md)|[microsoft.graph.security.departmentTemplate](../resources/security-departmenttemplate.md)|Read the properties and relationships of a [microsoft.graph.security.departmentTemplate](../resources/security-departmenttemplate.md) object.|
+|[Delete departments](../api/security-labelsroot-delete-departments.md)|None|Delete a [microsoft.graph.security.departmentTemplate](../resources/security-departmenttemplate.md) object.|
+
+## Properties
+|Property|Type|Description|
+|:---|:---|:---|
+|createdBy|[microsoft.graph.identitySet](/graph/api/resources/identityset)|Represents the user who created the department. Inherited from [microsoft.graph.security.filePlanDescriptor](../resources/security-fileplandescriptor.md).|
```
angelgolfer-ms, line 30: For both v1 and beta:
I'm using lines 30 and 31 here as examples, but I think my comment is applicable to the 2 properties of each of the 6 resources that are templates.
I think createdBy and createdDateTime actually refer to the adding of a value to that corresponding descriptor, is that correct?
Just like the createdBy and createdDateTime properties of a retentionLabel, those properties refer to adding the retention label, and since these descriptors supplement what's being tracked by a retention label, it makes sense that the API design is to track the similar metadata for each descriptor, if that descriptor is used? 

After all, it doesn't seem to make sense to say, "date and time the department was created", and be tracking the date/time of the creation of an entire department in an organization, that is related to a piece of data in Microsoft 365?

angelgolfer-ms: @sseth-msft , 
Upon your confirmation for my last comment - that createdBy and createdDateTime are metadata for including a file descriptor to supplement a retention label - then please update the descriptions for these 2 properties for each of the 6 descriptor templates.

api-reference/v1.0/api/security-labelsroot-post-citations.md
```diff
@@ -0,0 +1,107 @@
...
+## Request headers
+|Name|Description|
+|:---|:---|
+|Authorization|Bearer {token}. Required.|
+|Content-Type|application/json. Required.|
+
+## Request body
+In the request body, supply a JSON representation of the [microsoft.graph.security.citationTemplate](../resources/security-citationtemplate.md) object.
+
+You can specify the following properties when creating a **citationTemplate**.
```
angelgolfer-ms, line 43: For both v1 and beta:
GLOBAL:
I'm using this example - the create operation for citationTemplate - but it looks like all the create operation for the 6 template resources have the similar issue.
Issue:
In lines 43-49, you only list 3 properties. What about setting the createdBy property, which you show in the example request in line 73?
Action for you:
(a) Please add the createdBy property to the table in the Request body section.
(b) Please clarify in the resource topic for citationTemplate, in the Properties table, description for createdDateTime, add to the end of the existing description "Read-only." (That would be after the sentence that in another comment I requested you to fix, "Inherited from microsoft.graph.security.filePlanDescriptorTemplate.").
(c) And just to make it consistent for the properties in that Properties table, add the same "Read-only." sentence to the end of the description of the id property as well. Just so we are clear, in general, the id property of all the resources of type entity are read-only in Microsoft Graph, assigned on creation. Since we mark one of the properties in the same table as "read-only" in this case, I don't want to create ambiguity that all the other properties in the same table are not read-only if they are  not marked as such. :-)

Please do (a) to (c) for all the 6 template resources, for each of v1 and beta.

api-reference/v1.0/resources/security-retentionlabel.md
```diff
@@ -0,0 +1,92 @@
...
+|[Delete retentionLabel](../api/security-retentionlabel-delete.md)|None|Delete a [retentionLabel](../resources/security-retentionlabel.md) object.|
+|[List retentionEventType](../api/security-retentioneventtype-list.md)|[microsoft.graph.security.retentionEventType](../resources/security-retentioneventtype.md) collection|Get the retentionEventType resources from the exapnd eventType navigation property.|
+|[Add retentionEventType](../api/security-retentioneventtype-post.md)|[microsoft.graph.security.retentionEventType](../resources/security-retentioneventtype.md)|Add eventType by adding the relevant odata property when creating a label.|
+
+
+## Properties
+|Property|Type|Description|
+|:---|:---|:---|
+|actionAfterRetentionPeriod|microsoft.graph.security.actionAfterRetentionPeriod| Specifies the action to take on a document with this label applied during the retention period. The possible values are: `none`, `delete`, `startDispositionReview`, `unknownFutureValue`.|
```
angelgolfer-ms, line 31: For both v1 and beta:
Line 31:
I think this description is incorrect. The property name says "after", after the retention period. 
This should read:
"Specifies what action to take on the document with this label after the value specified by the retentionDuration property expires. The possible values are: `none`, `delete`, `startDispositionReview`, `unknownFutureValue`."

FYI I'm going to make this fix for both v1 and beta. Calling this out for you.

PR #22078, opened 2023-07-15
Updated links in the docs to point to the new known issues site

lramosvea:


FaithOmbongi (approved):
Hi @lramosvea - I pushed a commit to address the pending issues that were in your Excel sheet. Only two issues are unresolved now.

PR #22113, opened 2023-07-18
[Doc Improvement][Create channel note added][3203414]

v-sdhakshina:


Prasad-MSFT (commented):
Better to change the text as below:
"If you want to add a guest account to the channel, use `guest` in `roles` property."

Prasad-MSFT (approved):
changes look fine to me

Wajeed-msft (approved):

PR #22157, opened 2023-07-19
Update Microsoft.Security.Ediscovery.json

Danielabom:
Fixing changelog.


angelgolfer-ms:
Thanks for following up, @Danielabom , the fix looks good.

PR #21988, opened 2023-07-10
Windows Updates catalog docs changes

angiechen22:
I am a product owner of the Windows Update namespace in Microsoft Graph. The team has recently made several additions to the catalog endpoint that have been [approved by the Graph board](https://microsoftgraph.visualstudio.com/onboarding/_git/onboarding/pullrequest/14164?path=/reviews/17447-windows-updates-quality-update-catalog-entry-metadata-additions/api.md).

Changed /resources/ files

- windowsupdatesfeatureupdatecatalogentry
- windowsupdatesqualityupdatecatalogentry
- *new* windowsupdatesknowledgebasearticle
- *new* windowsupdatescveinformation
- *new* windowsupdatesproductrevision

Changed /concepts/ docs:

- windowsupdates-deploy-update
- windowsupdates-deploy-expedited-update

PR #22188, opened 2023-07-21
Edits in PR 21988

JarbasHorst:
Fixed issues in PR #21988.
Fixed member's order as per schema file.


angiechen22 (approved):
Thanks for the edits (and the changelog)!

PR #22199, opened 2023-07-21
adds user container type for Planner

wghilliard:
This PR adds a new container type for Planner Plans. Please refer to the internal graph pull request for more details!

### DIFF
--- a/api-reference/beta/resources/enums.md
+++ b/api-reference/beta/resources/enums.md
@@ -3000,6 +3000,7 @@ Possible values for user account types (group membership), per Windows definitio
 |roster|
 |project|
 |driveItem|
+|user|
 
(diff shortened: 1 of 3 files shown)


api-reference/beta/resources/enums.md
```diff
@@ -3000,6 +3000,7 @@ Possible values for user account types (group membership), per Windows definitio
 |roster|
 |project|
 |driveItem|
+|user|
```
TarkanSevilmis, line 3003: This is also listed in the places that use it, and it has also samples and other information like the URL that needs to be edited.

wghilliard: added content to the overview page, please advise if there is another specific place where content should be added

PR #22245, opened 2023-07-26
Update user-sendmail.md

MSFT-AN:
Additional note for 202 Accepted HTTPS response is a requested update via a customer case.


MSFT-AN (commented):
Thanks for the review

PR #22338, opened 2023-08-03
[Doc Issue][Update appcatalogs list teamsapps]

v-ypalikila:
removed azureADAppId from List teamsApp responses.


Meghana-MSFT (approved):
ok

PR #22359, opened 2023-08-04
Moved limitations note to bottom of table for better readability.

keylimesoda:
This section is often embedded in other documents.  Needed to move the resource limits note from the top to the bottom of the table for readability.


FaithOmbongi (approved):
LGTM!

PR #22361, opened 2023-08-04
Reorganized for better readability

keylimesoda:
Grouped together all the tables related to various properties of subscribable resources.


FaithOmbongi (approved):
LGTM!

PR #22364, opened 2023-08-04
Significant re-write to focus on user-centric view

keylimesoda:
Existing documentation did not sufficiently guide users in creating endpoints that can reliably receive notifications.

Refactoring, reorganized, and rewrote significant portions of this document to help users succeed in setting up notifications.

### DIFF
--- a/concepts/change-notifications-delivery-webhooks.md
+++ b/concepts/change-notifications-delivery-webhooks.md
@@ -12,24 +12,68 @@ ms.date: 03/23/2023
 
 # Receive change notifications through webhooks
 
-A webhook is a HTTP-based user-defined callback API that you can set up in your infrastructure to receive change notifications and events from a service, such as Microsoft Graph. You must configure the webhook using a well-known and accessible HTTPS-secured endpoint.
+A webhook is a HTTP-based user-defined callback API that you can set up in your infrastructure to receive change notifications and events from a service, such as Microsoft Graph. To use webhooks you will need to standup a publicly accessible HTTPS-secured endpoint that will receive the notifications.
 
-To receive change notifications through webhooks, you need to create a subscription to the resource for which you want to be notified of changes. While the subscription is valid, Microsoft Graph sends a notification to your app whenever a change is detected on the resource.
+You can create a subscription to the resource for which you want to be notified of changes. While the subscription is valid, Microsoft Graph sends a notification to your endpoint whenever a change is detected on the resource.
 
-The article guides you through the process of managing your Microsoft Graph subscription and how to receive change notifications through webhooks.
+The article guides you through the process of implementing your webhook endpoint, subscribing to and managing Microsoft Graph subscriptions, and how to receive change notifications through webhooks.
+
+## Implementing a webhook endpoint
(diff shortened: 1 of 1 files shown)


concepts/change-notifications-delivery-webhooks.md
```diff
@@ -12,24 +12,68 @@ ms.date: 03/23/2023
...
+      - An endpoint will exit the "slow" state once less than 10% of responses take longer than 3 seconds in a 10-minute window.
+
+2. An endpoint will be marked "drop" once more than 15% of responses take longer than 3 seconds in a 10-minute window.
+      - Once an endpoint has been marked "drop", any new notifications will be dropped, for up to 10 minutes
+      - An endpoint will exit the "drop" state once less than 15% of responses take longer than 3 seconds in a 10-minute window.
+
+If your endpoint is unable to meet these performance characteristics, consider using [Event Hub](/graph/change-notifications-delivery-event-hubs) or [Event Grid](/azure/event-grid/subscribe-to-graph-api-events?context=graph/context) as a target for receiving notifications.
+
+### Authentication
+When you create your subscription, an access token will be sent to your endpoint. This access token will generally expire within 1 hour.
```
FaithOmbongi, line 49: Should also mention that before it expires, Microsoft Graph will send lifecycle notifications of the impending expiry to the lifecycle notification endpoint. The expiry doesn't happen without notice.

concepts/change-notifications-delivery-webhooks.md
```diff
@@ -12,24 +12,68 @@ ms.date: 03/23/2023
...
+
+If an access token expires, notifications will not be delivered. But Microsoft Graph will continue to retry sending each notification for up to 4 hours. So if the access token is refreshed within 4 hours of expiration, unsent notifications will be delivered.
```
FaithOmbongi, line 51: > If an access token expires, notifications will not be delivered. But Microsoft Graph will continue to retry sending each notification for up to 4 hours.

Q: Are we saying that Microsoft Graph will try to resend the notifications in a 4-hour span even though it knows the endpoint can't receive them? Doesn't that then guarantee a throttling scenario?

> So if the access token is refreshed within 4 hours of expiration, unsent notifications will be delivered.

1. In the lifecycle notifications doc, we say that after the access token expires, you can no longer "refresh" it but you must create a new subscription thereby getting a new access token.
2. Your statement also contradicts the info we have on the lifecycle notifications doc where we say that "any notifications about resource changes that happen when the change notification delivery pauses and the time when the app successfully creates the subscription again _would be lost._".

One of the two references is therefore misleading.

FaithOmbongi (changes requested):
Hi @keylimesoda - Thank you for these well-thought changes!

I've left two comments in the Authentication section. Please review and address them.

concepts/change-notifications-delivery-webhooks.md
keylimesoda: Authentication token and subscription have different lifecycles with different behavior.

An authentication token is used to continue to ensure that your endpoint is secure and valid.  This does not in any way impact throttling.

If your subscription expires you have to create a new one.  If your authentication token expires, you can reauthorize.

I've updated this section to try to make it more clear.  Please re-review.

PR #22381, opened 2023-08-07
add cpc  getRawConnectionReports api

Gutentag0004:


RuiHou105 (changes requested):
@Gutentag0004 
1.Please also update changelog to summarize your change.
2.Please create roadmap work item and attach the work item number in comment. Here is an [example](https://microsoftgraph.visualstudio.com/Roadmap/_workitems/edit/18470/).

